PR #571, opened 2015-08-30
Fix gmock_doctor.py for Python3

srouquette:


BillyDonahue:
needs to keep working with Python 2. Will it?

srouquette:
yes. one of the difference between 3 and 2 is that print is now a function, so you need to call it with parenthesis.
it was partially done in the script (look just above my first change), I just "fixed" what was preventing the script from running for me.
I only updated the print syntax.

PR #580, opened 2015-09-07
Make tests pass on MacOS with -std=c++11.

dmircevski:
The MacOS version of clang++ doesn't allow tuples of references to
rvalues when -std=c++11 is used.

https://llvm.org/bugs/show_bug.cgi?id=20855

### DIFF
--- a/googlemock/test/gmock-more-actions_test.cc
+++ b/googlemock/test/gmock-more-actions_test.cc
@@ -325,7 +325,7 @@ TEST(InvokeTest, FunctionThatTakes10Arguments) {
 
 // Tests using Invoke() with functions with parameters declared as Unused.
 TEST(InvokeTest, FunctionWithUnusedParameters) {
-  Action<int(int, int, double, const string&)> a1 =
+  Action<int(int, int, double, const string)> a1 =
       Invoke(SumOfFirst2);
   EXPECT_EQ(12, a1.Perform(make_tuple(10, 2, 5.6, string("hi"))));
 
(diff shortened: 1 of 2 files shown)


googlemock/test/gmock-more-actions_test.cc
```diff
@@ -325,7 +325,7 @@ TEST(InvokeTest, FunctionThatTakes10Arguments) {
 
 // Tests using Invoke() with functions with parameters declared as Unused.
 TEST(InvokeTest, FunctionWithUnusedParameters) {
-  Action<int(int, int, double, const string&)> a1 =
+  Action<int(int, int, double, const string)> a1 =
```
BillyDonahue: The 'const' in a function argument type is meaningless.

Can you change the .travisci.yml and travis.sh scripts to prove out this fix?
It would good to have automatic checking of the osx c++11 build that you say is broken.

dmircevski: Removed "const" from value parameters.

I can modify Travis scripts in a separate change that can be merged prior to this one.  It's possible, depending on which clang version Travis uses, that it may not experience this failure.  Regardless, this fixes undefined behaviour -- see the bug link up top.

BillyDonahue:
I can't tell what's going on in that llvm bug, exactly.
I hesitate to strip the references from these actions, though. I believe there's a big difference in gtest between a by-reference Action parameter and a by-value Action parameter.
It would be nice to see exactly which build problems we see in c++11.

dmircevski:
With respect to dropping "std::" in "std::string":

If there's some subtle reason why "string" is OK without "std::" on, say, line 211 but must have "std::" on line 754, then it's probably a bad reason.  My change merely makes the code consistent.

I'm OK with changing every occurrence of "string" to "std::string" in another PR.

BillyDonahue:
@dmircevski dmircevski, Can you give me a quick review of https://github.com/google/googletest/pull/583 ?

BillyDonahue:
I haven't reviewed these sites closely to answer whether the change is ok
or not. I agree that it looks like just sloppiness. I just don't want to
change the meaning of the existing tests here. I suspect we'll want some
std::string tests and some string tests just to check the interop of gtest
between those two potentially different classes. But the tests you're
changing don't seem to be focused on string behavior at all, so it really
does look like a fixable oversight. I just don't want to open that can of
worms in a PR that's about C++11 interop.

dmircevski:
@BillyDonahue: done.  Sorry, I'm on the East Coast and was asleep when #583 was created.

PR #581, opened 2015-09-07
Add support for Bazel build targets and fix some tests.

jraisher:
Not all unit tests are included in the new build file since the
tests are not all simple C++ tests.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -10,12 +10,19 @@ licenses(["notice"])
 
 cc_library(
     name = "googletest",
+    srcs = ["googletest/src/gtest_main.cc"],
+    deps = [
+        ":googletest_no_main",
+    ],
+)
+
+cc_library(
+    name = "googletest_no_main",
     srcs = glob([
     	"googletest/src/gtest.cc",
     	"googletest/src/gtest-death-test.cc",
     	"googletest/src/gtest-filepath.cc",
     	"googletest/src/gtest-internal-inl.h",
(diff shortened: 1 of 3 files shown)


jraisher:
Thanks for the comments, I've incorporated them and updated the CR.

BUILD
```diff
@@ -0,0 +1,614 @@
...
+    ]),
+    includes = [
+        "googletest",
+        "googletest/include",
+    ],
+    copts = [
+        "-g",
+        "-Wall",
+        "-Wextra",
+<<<<<<< HEAD
```
balasanjay: Bad merge, here and below.

BUILD
```diff
@@ -0,0 +1,614 @@
...
+licenses(["notice"])
+
+cc_library(
+    name = "googletest",
+    srcs = glob([
+    	"googletest/src/gtest.cc",
+    	"googletest/src/gtest-death-test.cc",
+    	"googletest/src/gtest-filepath.cc",
+    	"googletest/src/gtest-internal-inl.h",
+    	"googletest/src/gtest_main.cc",
```
balasanjay: Can you split gtest_main.cc into its own separate target?

We have a library that needs to include gtest to get the FRIEND_TEST macro, but we don't want a main in that situation.

gennadiycivil:
I see failed Travis build https://travis-ci.org/google/googletest/builds/80834800.
Please review if this is still relevant.

PR #583, opened 2015-09-08
gmock unit tests: avoid make_tuple for lvalueref tuples.

BillyDonahue:
Googlemock has some tuples containing lvalue refs in its unit tests. These tuples are created with make_tuple, which is given temporaries. The make_tuple is in a function argument list.

A possibly overzealous static_assert in libc++'s std::tuple ctor
is firing in our 'Perform(make_tuple("hi"))' calls, so
we can't use its make_tuple here. Instead we will use
explicitly-constructed tuples constructed from non-temporary strings.

Workaround for llvm bug:
    https://llvm.org/bugs/show_bug.cgi?id=20855

An alternative to https://github.com/google/googletest/pull/580 .


BillyDonahue:
@dmircevski dmircevski, Can you give me a quick review?

dmircevski:
LGTM

PR #589, opened 2015-09-17
Fix the googlemock autotools build.

### LINKED ISSUE #588: googlemock autotools are broken
I'm trying to switch our CI tools from https://github.com/google/googlemock to the new github version, but the build is broken when using autotools.

googletest works correctly, but googlemock fails with 

    aclocal: error: configure.ac:1: file 'gtest/m4/acx_pthread.m4' does not exist

If I fix that, there are more failures caused by the gtest directory not existing. Symlinking googlemock/gtest to googletest allows the build to succeed.

I'm happy to send patches, I just need to know what the desired layout is.

nomis52:
Closed #588

### DIFF
--- a/googlemock/configure.ac
+++ b/googlemock/configure.ac
@@ -1,4 +1,4 @@
-m4_include(gtest/m4/acx_pthread.m4)
+m4_include(../googletest/m4/acx_pthread.m4)
 
 AC_INIT([Google C++ Mocking Framework],
         [1.7.0],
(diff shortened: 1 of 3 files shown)


nomis52:
The travis failure is occurring even without this patch.

BillyDonahue:
Yeah, we're currently broken. It's something to do with missing traits in a C++11 stdlib. Not you.

PR #593, opened 2015-09-20
Adding missing header files to the distribution.

martinmaly:
When trying to use the googletest in another project I found that there are some files missing in pkginclude. As a result, "make check" inside unpacked distribution fails. With this fix, it passes.


BillyDonahue:
brilliant, thanks!

PR #596, opened 2015-09-23
Allow explicit bool operators to work with ASSERT_TRUE variants

adamlesinski:
Inserting an explicit cast to bool allows the explicit bool operator
to function just as it would in an if statement.


BillyDonahue:
I'm confused by this PR's motivation.
This is supposed to already work, using the template constructor for AssertionResult.
Does that NOT work with contextually convertible T ?

https://github.com/google/googletest/blob/master/googletest/include/gtest/gtest.h#L274

adamlesinski:
Ah, apologies, looks like my patch was based on a really old copy of the codebase.

PR #602, opened 2015-10-02
Implement GetThreadCount for AIX.

mtbrandy:


googletest/src/gtest-port.cc
```diff
@@ -58,6 +58,10 @@
 # include <sys/procfs.h>
 #endif  // GTEST_OS_QNX
 
+#if GTEST_OS_AIX
+# include <procinfo.h>
```
BillyDonahue: also #include <sys/types.h>, according to https://www-01.ibm.com/support/knowledgecenter/ssw_aix_71/com.ibm.aix.basetrf1/getprocs.htm

googletest/src/gtest-port.cc
```diff
@@ -146,6 +150,19 @@ size_t GetThreadCount() {
   }
 }
 
+#elif GTEST_OS_AIX
+
+size_t GetThreadCount() {
+  struct procentry64 entry;
```
BillyDonahue: AIX has getprocs and getprocs64, depending on whether you're building for a 32-bit or 64-bit target.
How do you know you're building specifically for 64-bit AIX?

mtbrandy: getprocs64 can be used for either execution mode.  It simply provides a size-invariant procentry64 structure.

googletest/src/gtest-port.cc
mtbrandy: sys/types.h provides the definition for pid_t.  This is implicitly included in any of the std.* headers.  I will explicitly add it if you want.

googletest/src/gtest-port.cc
BillyDonahue: Ah yes. You're right: "Alternatively, the struct procsinfo64 and sizeof (struct procsinfo64) can be used by 32-bit getprocs to return full 64-bit process information"

googletest/src/gtest-port.cc
BillyDonahue: Please add it explicitly. I don't want to rely on the presence of other includes. Those could disappear.

mtbrandy: Done.

mtbrandy:
Anything else you need from me to close this out?

mtbrandy:
Any update here?  The failing check appears to be unrelated.

BillyDonahue:
Looks good to me

Issue #451, opened 2015-07-28
Support building as shared library with -fvisibility=hidden on Linux


GoogleCodeExporter:
```
On non-Windows systems GTEST_API_ is defined as empty macros which assumes 
default visibility.

However if embedder's build is configured with -fvisibility=hidden GTEST_API_ 
should be defined as __attribute__ ((visibility ("default"))) otherwise 
embedder will not be able to link.

Example of such embedder is Chromium.
 
```

Original issue reported on code.google.com by `vegorov@google.com` on 19 Jan 2014 at 7:28

d-meiser:
@BillyDonahue I think this can be closed now.

PR #605, opened 2015-10-05
Fix building of gtest with -fvisibility=hidden [#451]

d-meiser:
This fixes [#451]. I've signed the CLA.

### DIFF
--- a/googletest/include/gtest/internal/gtest-port.h
+++ b/googletest/include/gtest/internal/gtest-port.h
@@ -920,14 +920,14 @@ using ::std::tuple_size;
 #endif  // GTEST_HAS_SEH
 
 #ifdef _MSC_VER
-
 # if GTEST_LINKED_AS_SHARED_LIBRARY
 #  define GTEST_API_ __declspec(dllimport)
 # elif GTEST_CREATE_SHARED_LIBRARY
 #  define GTEST_API_ __declspec(dllexport)
 # endif
-
-#endif  // _MSC_VER
+#elif __GNUC__ >= 4 || defined(__clang__)
+# define GTEST_API_ __attribute__((visibility ("default")))
+#endif // _MSC_VER
 
 #ifndef GTEST_API_
 # define GTEST_API_
--- a/googletest/CMakeLists.txt
+++ b/googletest/CMakeLists.txt
@@ -22,6 +22,11 @@ option(gtest_build_samples "Build gtest's sample programs." OFF)
 
 option(gtest_disable_pthreads "Disable uses of pthreads in gtest." OFF)
 
+option(
+  gtest_hide_internal_symbols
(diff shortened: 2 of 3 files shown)


d-meiser:
Is there any interest in these changes? Do I need to clean something up?

d-meiser:
Is this the right place to submit pull requests? readme suggests it is but perhaps that's out of date?

BillyDonahue:
Yes this is the correct place. The queue is a little neglected right now.
Sorry.

d-meiser:
No problem. Let me know if I can do anything to make it easier to review or merge this PR.

googletest/include/gtest/internal/gtest-port.h
```diff
@@ -920,18 +920,22 @@ using ::std::tuple_size;
...
-# define GTEST_API_
-#endif
+#elif __GNUC__ >= 4 || defined(__clang__)
+# if GTEST_CREATE_SHARED_LIBRARY
+#  define GTEST_API_ __attribute__((visibility ("default")))
+# else
+#  define GTEST_API_
+# endif
+#else
+# ifndef GTEST_API_
```
BillyDonahue: The ifelse chain is hard to follow. How about moving this fallback out of the entire staircase, leaving it as an unconditional cleanup after the last #endif ?

googletest/include/gtest/internal/gtest-port.h
```diff
@@ -920,18 +920,22 @@ using ::std::tuple_size;
...
 #  define GTEST_API_ __declspec(dllexport)
 # endif
-
-#endif  // _MSC_VER
-
-#ifndef GTEST_API_
-# define GTEST_API_
-#endif
+#elif __GNUC__ >= 4 || defined(__clang__)
+# if GTEST_CREATE_SHARED_LIBRARY
```
BillyDonahue: Maybe this __attribute would be useless if the visibility preset wasn't going to be "hidden".

googletest/include/gtest/internal/gtest-port.h
d-meiser: Great suggestion.

googletest/include/gtest/internal/gtest-port.h
d-meiser: I'm not sure I fully understand. Are you proposing to get rid of the `#if GTEST_CREATE_SHARED_LIBRARY` because the `__attribute__((visibility ("default")))` doesn't have an effect if we're building a static library? I think that would be a valid simplification. I'll try that.

d-meiser:
Thanks for the feedback. I'll update the pull request, hopefully by tonight.

d-meiser:
@BillyDonahue I've cleaned up the definition of GTEST_API_. Let me know if this is better. Thanks.

PR #614, opened 2015-10-18
Fix C++11 clang builds in Travis CI

DukeXar:
Existing Travis configuration installs clang 3.0, which seems not being able to locate gcc 4.9.
Using white-listed clang 3.7 package instead.


mark-chromium:
Looks like this or something like it would be a good one to take. All of the pending pull requests except for this one show a red X because of the failing clang C++11 build.

PR #619, opened 2015-10-21
Add an environment variable to control the location of temporary files

bsilver8192:
Bazel says in their test environment documentation
(http://bazel.io/docs/test-encyclopedia.html) that tests should not
write to /tmp. However, googletest unconditionally writes files there.
This fixes that by having googletest use the GTEST_TMP_DIR
environment variable instead, which Bazel will soon set automatically
when running tests.

I've previously submitted this [on the mailing list](https://groups.google.com/forum/#!topic/googletestframework/_yvYL1KoFbk), but it seems like the more common path for submissions is pull requests.

### DIFF
--- a/googletest/src/gtest-port.cc
+++ b/googletest/src/gtest-port.cc
@@ -953,7 +953,15 @@ class CapturedStream {
     // use /sdcard.
     char name_template[] = "/sdcard/gtest_captured_stream.XXXXXX";
 #  else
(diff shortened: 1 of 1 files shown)


gennadiycivil:
TravisCI build failed. Doing some cleanup and closing outstanding PRs that 
1) don't have all checks OK

PR #620, opened 2015-11-02
Allow specifying file/line for SCOPED_TRACE

caitp:
Example use:

```c++
void custom_assert(const Foo& actual, int expected_1, expected_2, expected_3) {
  ASSERT_EQ(expected_1, actual.thing_1());
  // ...
}

#define CUSTOM_ASSERT(actual, expected_1, expected_2, expected_3, message) \
  do { \
    SCOPED_TRACE_(message, __FILE__, __LINE__); \
    custom_assert(actual, expected_1, expected_2, expected_3); \
  } while(0)
```

Grouping re-used assertions into a subroutine, while still keeping the stack trace clean

### DIFF
--- a/googletest/include/gtest/gtest.h
+++ b/googletest/include/gtest/gtest.h
@@ -2116,9 +2116,14 @@ GTEST_API_ AssertionResult DoubleLE(const char* expr1, const char* expr2,
 // of the dummy variable name, thus allowing multiple SCOPED_TRACE()s
 // to appear in the same block - as long as they are on different
 // lines.
-#define SCOPED_TRACE(message) \
+#define SCOPED_TRACE(...) GTEST_VA_SELECT_(SCOPED_TRACE, __VA_ARGS__) 
+#define SCOPED_TRACE_1(message) \
   ::testing::internal::ScopedTrace GTEST_CONCAT_TOKEN_(gtest_trace_, __LINE__)(\
     __FILE__, __LINE__, ::testing::Message() << (message))
+#define SCOPED_TRACE_3(file, line, message) \
+  ::testing::internal::ScopedTrace GTEST_CONCAT_TOKEN_(gtest_trace_, line)(\
+    file, line, ::testing::Message() << (message))
+
 
 // Compile-time assertion for type equality.
 // StaticAssertTypeEq<type1, type2>() compiles iff type1 and type2 are
--- a/googletest/include/gtest/internal/gtest-internal.h
+++ b/googletest/include/gtest/internal/gtest-internal.h
@@ -76,6 +76,10 @@
 #define GTEST_CONCAT_TOKEN_(foo, bar) GTEST_CONCAT_TOKEN_IMPL_(foo, bar)
 #define GTEST_CONCAT_TOKEN_IMPL_(foo, bar) foo ## bar
 
+#define GTEST_VA_SIZE_(...) GTEST_VA_SIZE_IMPL_(__VA_ARGS__, 9, 8, 7, 6, 5, 4, 3, 2, 1)
+#define GTEST_VA_SIZE_IMPL_(_1, _2, _3, _4, _5, _6, _7, _8, _9, SIZE, ...) SIZE
+#define GTEST_VA_SELECT_(name, ...) GTEST_CONCAT_TOKEN_(name ## _, GTEST_VA_SIZE_(__VA_ARGS__))(__VA_ARGS__)
+
 class ProtocolMessage;
 namespace proto2 { class Message; }
 


googletest/include/gtest/gtest.h
```diff
@@ -2116,9 +2116,11 @@ GTEST_API_ AssertionResult DoubleLE(const char* expr1, const char* expr2,
 // of the dummy variable name, thus allowing multiple SCOPED_TRACE()s
 // to appear in the same block - as long as they are on different
 // lines.
-#define SCOPED_TRACE(message) \
-  ::testing::internal::ScopedTrace GTEST_CONCAT_TOKEN_(gtest_trace_, __LINE__)(\
-    __FILE__, __LINE__, ::testing::Message() << (message))
+#define SCOPED_TRACE_(message, file, line) \
```
caitp: `SCOPED_TRACE2`? `SCOPED_TRACE_EX`? I'm calling it `MY_SCOPED_TRACE`, but it feels bad rewriting macros that use internal APIs.

BillyDonahue: I think it should be (file, line, message) rather than (message, file, line). This will allow for variadic extension of message.... Can we do something clever with __VA_ARGS__ to implement both SCOPED_TRACE variants under the same name?

googletest/include/gtest/gtest.h
```diff
@@ -2116,9 +2116,11 @@ GTEST_API_ AssertionResult DoubleLE(const char* expr1, const char* expr2,
 // of the dummy variable name, thus allowing multiple SCOPED_TRACE()s
 // to appear in the same block - as long as they are on different
 // lines.
-#define SCOPED_TRACE(message) \
-  ::testing::internal::ScopedTrace GTEST_CONCAT_TOKEN_(gtest_trace_, __LINE__)(\
-    __FILE__, __LINE__, ::testing::Message() << (message))
+#define SCOPED_TRACE_(file, line, message) \
```
caitp: updated to `file, line, message`

googletest/include/gtest/gtest.h
caitp: as far as I'm aware, the only way to do this would still require defining multiple macro names (eg selecting the macro impl based on number of arguments). I don't know if that's any better

BillyDonahue: http://stackoverflow.com/questions/16683146/can-macros-be-overloaded-by-number-of-arguments

googletest/include/gtest/gtest.h
```diff
@@ -2116,9 +2116,14 @@ GTEST_API_ AssertionResult DoubleLE(const char* expr1, const char* expr2,
 // of the dummy variable name, thus allowing multiple SCOPED_TRACE()s
 // to appear in the same block - as long as they are on different
 // lines.
-#define SCOPED_TRACE(message) \
+#define SCOPED_TRACE(...) GTEST_VA_SELECT_(SCOPED_TRACE, __VA_ARGS__) 
+#define SCOPED_TRACE_1(message) \
```
caitp: Updated to use single name

BillyDonahue: can we do the same trick without any variadic macro sorcery?
Have the macro forward all of its arguments to a C++ function and use ordinary overload resolution to figure out the number and meaning of the fields?

caitp: The macro is declaring a variable in the calling scope, using a name derived from the line of the call. So I'm not positive that can fully work

BillyDonahue: It can still do that.
Does the __LINE__ that forms the variable name have to be the same as the
'line' that's printed in the SCOPED_TRACE?

caitp: Okay, yes, the declaration name doesn't matter a whole lot, so it could be `__LINE__` always --- but the ScopedTrace() constructor, if called with one parameter, can't derive the correct `__LINE__`, so the message output would be weird

BillyDonahue: On Tue, Nov 24, 2015 at 2:02 PM, ⭐caitp⭐ <notifications@github.com> wrote:

> In googletest/include/gtest/gtest.h
> <https://github.com/google/googletest/pull/620#discussion_r45778673>:
>
> > @@ -2116,9 +2116,14 @@ GTEST_API_ AssertionResult DoubleLE(const char* expr1, const char* expr2,
> >  // of the dummy variable name, thus allowing multiple SCOPED_TRACE()s
> >  // to appear in the same block - as long as they are on different
> >  // lines.
> > -#define SCOPED_TRACE(message) \
> > +#define SCOPED_TRACE(...) GTEST_VA_SELECT_(SCOPED_TRACE, __VA_ARGS__)
> > +#define SCOPED_TRACE_1(message) \
>
> Okay, yes, the declaration name doesn't matter a whole lot, so it could be
> __LINE__ always --- but the ScopedTrace() constructor, if called with one
> parameter, can't derive the correct __LINE__, so the message output would
> be weird
>
I was suggesting going through a dispatching function that calls the
ScopedTrace constructor.
The only purpose of the dispatching function would be to sort out the
argument list and produce a ScopedTrace constructor call. I haven't tried
it.

caitp: in order to avoid a regression in the current behaviour, it would be something like

```c++
#define SCOPED_TRACE(...) \
    ScopedTrace derived_name = ScopedTrace::factory(__FILE__, __LINE__, __VA_ARGS__);
```

```diff
  // The d'tor pops the info pushed by the c'tor.
  //
  // Note that the d'tor is not virtual in order to be efficient.
  // Don't inherit from ScopedTrace!
  ~ScopedTrace();

+ static ScopedTrace factory(const char* caller_file, int caller_line, const Message& message) {
+     return ScopedTrace(caller_file, caller_line, message);
+ }
+
+ static ScopedTrace factory(const char* caller_file, int caller_line, const char* file, int line,
+                                            const Message& message) {
+     return ScopedTrace(file, line, message);
+ }
- private:
-  GTEST_DISALLOW_COPY_AND_ASSIGN_(ScopedTrace);
```

which would probably work, but the is less strict WRT copy/assign. Allowing the move constructor might be a good workaround for that

BillyDonahue:
I'm going back to the CUSTOM_ASSERT usage example that kicked off this thread and I don't understand it. Wouldn't an ordinary SCOPED_TRACE have __FILE__ and __LINE__ of the place where CUSTOM_ASSERT is used? Is that not desired? What's "clean" mean w.r.t. this stack trace.

caitp:
It has a extra entry in the s trace pointing back to the caller of the group of assertions, making it easier to find a point of failure

BillyDonahue:
Sorry I still don't get it.
Can you give a before and after example?

caitp:
Sure, something like this:

```c++
#define MY_SCOPED_TRACE(message, file, line) \
  ::testing::internal::ScopedTrace GTEST_CONCAT_TOKEN_(gtest_trace_, line)(\
    __FILE__, __LINE__, ::testing::Message() << (message))

#define CHECK_TOKEN(message, actual, type, begin, end, line, column) \
  do {                                                               \
    MY_SCOPED_TRACE(message, __FILE__, __LINE__);                    \
    check_token(actual, type, begin, end, line, column);             \
  } while(0)

#define CHECK_SCAN(name, type, begin, end, line, col) \
  CHECK_TOKEN("Scanning " name, s.next(), type, begin, end, line, col)

TEST(Scanner, Tokens) {
  TestScanner s("//abc\n(){}[].++--**+-*/%\n"
                "/*foo-bar-baz\n*\n\n*/\n"
                "!!===&&||&|&=|=~~=^^==+=-=*=/=%=:;<<=<=<<<>>=>=>>>\n"
                "0.4 0xABCDEF 0b1010101 0123456789 0o1234567");

  CHECK_SCAN("short comment", Token::COMMENT, 0, 6, 1, 1);

  CHECK_SCAN("Token::LPAREN", Token::LPAREN, 6, 7, 2, 1);
  CHECK_SCAN("Token::RPAREN", Token::RPAREN, 7, 8, 2, 2);
  CHECK_SCAN("Token::LBRACE", Token::LBRACE, 8, 9, 2, 3);
  // ...
```

In the case of an error, the output will indicate both the line of the specific inner assertion which failed, and a line with more context.

If I modify the first assertion set to expect an incorrect token, I see 2 lines reported for the failure. 1, the actual assertion which failed, and 2, the line belonging to the assertion group in use.

```bash
[==========] Running 1 test from 1 test case.
[----------] 1 test from Scanner
[ RUN      ] Scanner.Tokens
test/frontend/frontend_test.h:34: Failure
Value of: actual.type()
  Actual: 2
Expected: expected_type
Which is: 3
Google Test trace:
test/frontend/test-scanner.cc:13: Scanning Token::LPAREN
```

Without the extra MY_SCOPED_TRACE, I'm missing the line providing more context, and only have this:

```bash
[==========] Running 1 test from 1 test case.
[----------] 1 test from Scanner
[ RUN      ] Scanner.Tokens
test/frontend/frontend_test.h:34: Failure
Value of: actual.type()
  Actual: 2
Expected: expected_type
Which is: 3
```

The other way around this is to define groups of assertions in macros, but this has some limitations (not revealing the line of the specific failed assertion, for instance).

The solution isn't perfect, but the extra context provided is helpful while debugging test failures

BillyDonahue:
`/* (admittedly, this could just as easly be SCOPED_TRACE(message), but still )*/ `

^^^ I guess that's what I'm asking: but still what?
Isn't that all you need?

caitp:
it works in this case, not necessarily all cases --- but you know what, I'll leave it until it's needed

Issue #606, opened 2015-10-07
Can't compile with MinGW (Win8 x64)


WesleyCeraso:
Hi, I'm having trouble building GTest with MinGW, I used bisect to find the commit that broke the build:

a6340420b9cee27f77c5b91bea807121914a5831 is the first bad commit
commit a6340420b9cee27f77c5b91bea807121914a5831
Author: kosak kosak@google.com
Date:   Mon Mar 24 21:58:25 2014 +0000

```
Implement threading support for gtest on Windows.
Also, stop using localtime(). Instead, use localtime_r() on most systems, localtime_s() on Windows.
```

The build step output:

$ cmake --build .
[ 25%] Building CXX object CMakeFiles/gtest.dir/src/gtest-all.cc.obj
In file included from C:/Users/wprudencio/gtest-1.7.0-hunter-11/include/gtest/internal/gtest-internal.h:40:0,
from C:/Users/wprudencio/gtest-1.7.0-hunter-11/include/gtest/gtest.h:58,
from C:\Users\wprudencio\gtest-1.7.0-hunter-11\src\gtest-all.cc:39:
C:/Users/wprudencio/gtest-1.7.0-hunter-11/include/gtest/internal/gtest-port.h:1704:3: error: 'AutoHandle' does not name a type
AutoHandle thread_;
^
In file included from C:\Users\wprudencio\gtest-1.7.0-hunter-11\src\gtest-all.cc:43:0:
C:/Users/wprudencio/gtest-1.7.0-hunter-11/src/gtest-death-test.cc:626:3: error: 'AutoHandle' does not name a type
AutoHandle write_handle_;
^
C:/Users/wprudencio/gtest-1.7.0-hunter-11/src/gtest-death-test.cc:628:3: error: 'AutoHandle' does not name a type
AutoHandle child_handle_;
^
C:/Users/wprudencio/gtest-1.7.0-hunter-11/src/gtest-death-test.cc:633:3: error: 'AutoHandle' does not name a type
AutoHandle event_handle_;
^
C:/Users/wprudencio/gtest-1.7.0-hunter-11/src/gtest-death-test.cc: In member function 'virtual int testing::internal::WindowsDeathTest::Wait()':
C:/Users/wprudencio/gtest-1.7.0-hunter-11/src/gtest-death-test.cc:645:36: error: 'child_handle_' was not declared in this scope
const HANDLE wait_handles[2] = { child_handle_.Get(), event_handle_.Get() };
^
[...]

PR #608, opened 2015-10-09
Properly disable phtreads when compiling with MinGW

andoks:
fix #606

PR #621, opened 2015-11-03
Fix compilation on MinGW with pthread

orgads:


BillyDonahue:
Can you describe the problem this is solving?
What is it about MinGW's pthread implementation that's causing trouble?

orgads:
Done in the commit message.

DanielSWolf:
:+1: I'm having the same build problem. This pull request fixes it for me.

fredZen:
It fixes issue #606 

I hope the following excerpt from gtest-port.h clarifies the situation

    # if GTEST_HAS_NOTIFICATION_
    ...
    # elif GTEST_HAS_PTHREAD // line 1461
    // This branch is taken on MinGW, which prevents the next elif from getting taken
    ...
    # elif GTEST_OS_WINDOWS && !GTEST_OS_WINDOWS_PHONE && !GTEST_OS_WINDOWS_RT
    // On MinGW, this branch gets skipped, so AutoHandle doesn't get defined
    ...
    class GTEST_API_ AutoHandle {
    ...
    };
    ...
    # endif  // GTEST_HAS_NOTIFICATION_
    
    # if GTEST_HAS_MUTEX_AND_THREAD_LOCAL_
    ...
    # elif GTEST_OS_WINDOWS && !GTEST_OS_WINDOWS_PHONE && !GTEST_OS_WINDOWS_RT
    // This branch is taken even on MinGW
    ...
    class GTEST_API_ ThreadWithParamBase {
    ...
     private:
      AutoHandle thread_; // line 1782, breaks because AutoHandle does not name a type
    };

fredZen:
See also #608

BillyDonahue:
If MinGW has a broken Pthread, let's have GTEST_HAS_PTHREAD be unset for that platform, instead of having a compound conditional whenever we check for Pthread support. Would that work?

orgads:
@BillyDonahue: Better now?

BillyDonahue:
Don't undef and redefine PTHREAD macro. Just fold MINGW into the os test when we set it originally.

orgads:
And what if the user chooses to define PTHREAD?

BillyDonahue:
i mean GTEST_HAS_PTHREAD

orgads:
Me too :)

Should it be used if the user explicitly defines it or not?

If not, how do you suggest to do it?

orgads:
Let's start with being consistent within this header. Improvements can come later.

orgads:
ping?

googletest/include/gtest/internal/gtest-port.h
```diff
@@ -604,6 +604,14 @@ struct _RTL_CRITICAL_SECTION;
     || GTEST_OS_QNX || GTEST_OS_FREEBSD || GTEST_OS_NACL)
 #endif  // GTEST_HAS_PTHREAD
 
+// On MinGW, we can have both GTEST_OS_WINDOWS and GTEST_HAS_PTHREAD
+// defined, but we don't want to use MinGW's pthreads implementation, which
+// has conformance problems with some versions of the POSIX standard.
+#if GTEST_HAS_PTHREAD && GTEST_OS_WINDOWS_MINGW
```
BillyDonahue: I don't want to undef something that was possibly set explicitly.
We should just deduce it correctly the first time.

```
#if !defined(GTEST_HAS_PTHREAD) && GTEST_OS_WINDOWS_MINGW
# define GTEST_HAS_PTHREAD 0
#endif
```

KindDragon:
Duplicate of PR  #721

PR #627, opened 2015-11-11
Fix warnings encountered in MSVC build of gtest/gmock tests

mark-chromium:
This fixes these warnings:

…\gtest\googletest\test\gtest-port_test.cc(78) : error C2220: warning treated as error - no 'object' file generated
…\gtest\googletest\test\gtest-port_test.cc(78) : warning C4309: 'static_cast' : truncation of constant value
…\gtest\googletest\test\gtest-port_test.cc(79) : warning C4309: 'static_cast' : truncation of constant value

…\gtest\googlemock\test\gmock-matchers_test.cc(2712) : error C2220: warning treated as error - no 'object' file generated
        …\gtest\googlemock\test\gmock-matchers_test.cc(2706) : while compiling class template member function 'testing::gmock_matchers_test::FloatingPointTest<float>::FloatingPointTest(void)'
        …\gtest\googlemock\test\gmock-matchers_test.cc(2896) : see reference to function template instantiation 'testing::gmock_matchers_test::FloatingPointTest<float>::FloatingPointTest(void)' being compiled
        …\gtest\googlemock\test\gmock-matchers_test.cc(2896) : see reference to class template instantiation 'testing::gmock_matchers_test::FloatingPointTest<float>' being compiled
…\gtest\googlemock\test\gmock-matchers_test.cc(2712) : warning C4267: 'argument' : conversion from 'size_t' to 'const unsigned int', possible loss of data
…\gtest\googlemock\test\gmock-matchers_test.cc(2714) : warning C4267: 'argument' : conversion from 'size_t' to 'const unsigned int', possible loss of data
…\gtest\googlemock\test\gmock-matchers_test.cc(2716) : warning C4267: 'argument' : conversion from 'size_t' to 'const unsigned int', possible loss of data
…\gtest\googlemock\test\gmock-matchers_test.cc(2717) : warning C4267: 'argument' : conversion from 'size_t' to 'const unsigned int', possible loss of data
…\gtest\googlemock\test\gmock-matchers_test.cc(2718) : warning C4267: 'argument' : conversion from 'size_t' to 'const unsigned int', possible loss of data
…\gtest\googlemock\test\gmock-matchers_test.cc(2721) : warning C4267: 'argument' : conversion from 'size_t' to 'const unsigned int', possible loss of data
…\gtest\googlemock\test\gmock-matchers_test.cc(2723) : warning C4267: 'argument' : conversion from 'size_t' to 'const unsigned int', possible loss of data

### DIFF
--- a/googlemock/test/gmock-matchers_test.cc
+++ b/googlemock/test/gmock-matchers_test.cc
@@ -2708,21 +2708,16 @@ class FloatingPointTest : public testing::Test {
         zero_bits_(Floating(0).bits()),
         one_bits_(Floating(1).bits()),
         infinity_bits_(Floating(Floating::Infinity()).bits()),
-        close_to_positive_zero_(Floating::ReinterpretBits(
-            static_cast<Bits>(zero_bits_ + max_ulps_/2))),
-        close_to_negative_zero_(-Floating::ReinterpretBits(
-            static_cast<Bits>(zero_bits_ + max_ulps_ - max_ulps_/2))),
-        further_from_negative_zero_(-Floating::ReinterpretBits(
-            static_cast<Bits>(zero_bits_ + max_ulps_ + 1 - max_ulps_/2))),
-        close_to_one_(Floating::ReinterpretBits(
-            static_cast<Bits>(one_bits_ + max_ulps_))),
-        further_from_one_(Floating::ReinterpretBits(
-            static_cast<Bits>(one_bits_ + max_ulps_ + 1))),
+        close_to_positive_zero_(ReinterpretBits(zero_bits_ + max_ulps_/2)),
+        close_to_negative_zero_(ReinterpretBits(
+            zero_bits_ + max_ulps_ - max_ulps_/2)),
+        further_from_negative_zero_(-ReinterpretBits(
+            zero_bits_ + max_ulps_ + 1 - max_ulps_/2)),
+        close_to_one_(ReinterpretBits(one_bits_ + max_ulps_)),
+        further_from_one_(ReinterpretBits(one_bits_ + max_ulps_ + 1)),
         infinity_(Floating::Infinity()),
-        close_to_infinity_(Floating::ReinterpretBits(
-            static_cast<Bits>(infinity_bits_ - max_ulps_))),
-        further_from_infinity_(Floating::ReinterpretBits(
-            static_cast<Bits>(infinity_bits_ - max_ulps_ - 1))),
+        close_to_infinity_(ReinterpretBits(infinity_bits_ - max_ulps_)),
+        further_from_infinity_(ReinterpretBits(infinity_bits_ - max_ulps_ - 1)),
         max_(Floating::Max()),
         nan1_(Floating::ReinterpretBits(Floating::kExponentBitMask | 1)),
         nan2_(Floating::ReinterpretBits(Floating::kExponentBitMask | 200)) {
@@ -2806,6 +2801,12 @@ class FloatingPointTest : public testing::Test {
   // Some NaNs.
   const RawType nan1_;
   const RawType nan2_;
+
+ private:
+  template <typename T>
+  static RawType ReinterpretBits(T value) {
+    return Floating::ReinterpretBits(static_cast<Bits>(value));
+  }
 };
 
 // Tests floating-point matchers with fixed epsilons.
--- a/googlemock/test/gmock-matchers_test.cc
+++ b/googlemock/test/gmock-matchers_test.cc
@@ -2708,19 +2708,21 @@ class FloatingPointTest : public testing::Test {
         zero_bits_(Floating(0).bits()),
         one_bits_(Floating(1).bits()),
         infinity_bits_(Floating(Floating::Infinity()).bits()),
-        close_to_positive_zero_(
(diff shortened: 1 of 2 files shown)


googlemock/test/gmock-matchers_test.cc
```diff
@@ -2778,7 +2778,7 @@ class FloatingPointTest : public testing::Test {
 
   // Pre-calculated numbers to be used by the tests.
 
-  const size_t max_ulps_;
+  const unsigned int max_ulps_;
```
BillyDonahue: max_ulps_ is initialized from a size_t in Floating<FltType>.
We should just be mirroring that type here.

googletest/test/gtest-port_test.cc
```diff
@@ -75,8 +75,8 @@ TEST(IsXDigitTest, WorksForNarrowAscii) {
 }
 
 TEST(IsXDigitTest, ReturnsFalseForNarrowNonAscii) {
-  EXPECT_FALSE(IsXDigit(static_cast<char>(0x80)));
-  EXPECT_FALSE(IsXDigit(static_cast<char>('0' | 0x80)));
+  EXPECT_FALSE(IsXDigit(static_cast<char>(0x80u)));
+  EXPECT_FALSE(IsXDigit(static_cast<char>('0' | 0x80u)));
```
BillyDonahue: I don't see the improvement here.
If 'char' is signed, you've got an impl-defined out of range unsigned-to-signed conversion.
What would work reliably?

static_cast<char>(static_cast<signed char>(-1)) ?

Is there a more concise way to say it?

googlemock/test/gmock-matchers_test.cc
mark-chromium: Meaning that you want to change Floating::kMaxUlps to unsigned int too, or meaning that you want to leave both of them as size_t and find another way (a bunch of casts, I guess) to silence the warnings about narrowing from size_t to unsigned int?

BillyDonahue: The latter. Leave it as a size_t and figure the warnings out as they occur. Why is there so much assignment of size_t to unsigned? Should those unsigned really be size_t?

mark-chromium: I showed the warnings here: https://github.com/mark-chromium/googletest/commit/8b085f0d2168e6e54712d295dcf26feac92bb9fc

They’re all of the form

        close_to_positive_zero_(
            Floating::ReinterpretBits(zero_bits_ + max_ulps_/2)),

With RawType = float, Bits (the type of zero_bits_ and ReinterpretBits()’ argument type) is unsigned int. Because max_ulps_ is size_t, with a 64-bit size_t, the expression for the argument is computed as size_t, and then the compiler issues a warning about narrowing to unsigned int to pass to ReinterpretBits().

googletest/test/gtest-port_test.cc
mark-chromium: This is not more concise but perhaps it expresses the intent better:

  static_cast<char>(std::numeric_limits<signed char>::min())

This is the most concise:

  '\x80'

although for the one where '0' is ORed in, the static_cast<char>() is still needed around the entire expression.

mark-chromium: (or '\200' if you prefer)

mark-chromium:
I updated gtest-port_test IsXDigitTest.ReturnsFalseForNarrowNonAscii to use char constants instead of casting signed/unsigned int constants to char.

I also updated gmock-matchers_test FloatTest.* to leave decltype(max_ulps_) alone and instead cast at the points of use. This is every point of use, so don’t really think that this is an improvement, but it makes no difference to correctness if you prefer this.

googlemock/test/gmock-matchers_test.cc
```diff
@@ -2708,19 +2708,21 @@ class FloatingPointTest : public testing::Test {
...
+            static_cast<Bits>(one_bits_ + max_ulps_ + 1))),
         infinity_(Floating::Infinity()),
-        close_to_infinity_(
-            Floating::ReinterpretBits(infinity_bits_ - max_ulps_)),
-        further_from_infinity_(
-            Floating::ReinterpretBits(infinity_bits_ - max_ulps_ - 1)),
+        close_to_infinity_(Floating::ReinterpretBits(
+            static_cast<Bits>(infinity_bits_ - max_ulps_))),
+        further_from_infinity_(Floating::ReinterpretBits(
+            static_cast<Bits>(infinity_bits_ - max_ulps_ - 1))),
```
BillyDonahue: can you write the ReinterpretBits(static_cast<Bits>(...)) as a single short private function to reduce the duplication of ugliness?

mark-chromium: Yeah, but the function’s argument needs to be templated because it’s not possible to know what it’s going to be. FloatingPointTest is instantiated with RawType = float and RawType = double, so Bits will be an unsigned 32-bit type or an unsigned 64-bit type. Meanwhile, size_t will be either 32 bits or 64 bits depending on the system. I don’t know how to express decltype(who_wins_the_promotion_race) without C++11.

I updated the pull request.

mark-chromium:
Ping?

PR #628, opened 2015-11-11
Adding CMakeLists.txt to the root dir

fjunqueira:
Adding functionality to build googletest from the root directory

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,16 @@
+cmake_minimum_required(VERSION 2.6.2)
+
+project( googletest-distribution )
+
+enable_testing()
+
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -0,0 +1,16 @@
+cmake_minimum_required(VERSION 2.6.2)
+
+project( googletest-distribution )
+
+enable_testing()
+
+option(BUILD_GTEST "Builds the googletest subproject" OFF)
```
BillyDonahue: why OFF?

fjunqueira: Googlemock already builds googletest. I did this so googletest is not built twice and googlemock is used by default.

PR #632, opened 2015-11-23
Add CMake install rules for gtest libraries and headers

nocnokneo:
This is useful when GoogleTest is pulled in as an `ExternalProject` in a CMake superbuild.

### DIFF
--- a/googletest/CMakeLists.txt
+++ b/googletest/CMakeLists.txt
@@ -93,6 +93,14 @@ target_link_libraries(gtest_main gtest)
 
 ########################################################################
 #
+# Install rules
+install(TARGETS gtest gtest_main
+  DESTINATION lib)
+install(DIRECTORY ${gtest_SOURCE_DIR}/include/gtest
+  DESTINATION include)
+
+########################################################################
+#
 # Samples on how to link user tests with gtest or gtest_main.
 #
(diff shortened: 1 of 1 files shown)


googletest/CMakeLists.txt
```diff
@@ -93,6 +93,14 @@ target_link_libraries(gtest_main gtest)
 
 ########################################################################
 #
+# Install rules
```
BillyDonahue: I can't read this syntax. What's it doing?

nocnokneo: gtest and gtest_main libraries are installed to `<CMAKE_INSTALL_PREFIX>/lib`

The `include/gtest` directory is installed to `<CMAKE_INSTALL_PREFIX>/include`

fjunqueira:
I made a pull request to build googletest and googlemock from the project root dir. [628](https://github.com/google/googletest/pull/628)

This allows the project to be built when used as an ExternalProject as well. I believe [628](https://github.com/google/googletest/pull/628) and [632](https://github.com/google/googletest/pull/632) solve the same "problem".

nocnokneo:
PR #628 is nice to have, but it's not related to this PR. This PR adds install rules so that the `INSTALL` step of of `ExternalProject_Add` works.

PR #633, opened 2015-11-24
Fix missing-field-initializers GCC/Clang warning (issue 433).

plopresti:


seirl:
Why not just `{}`?

plopresti:
Do you mean {} for the entire initializer? That would not necessarily work
because PTHREAD_MUTEX_INITIALIZER is not guaranteed to be the same as
zero-initialization.

Or do you mean {} instead of pthread_t()? The would not necessarily work
either because pthread_t is not guaranteed to be a structure; POSIX allows
it to be an arithmetic type (and even required it to be arithmetic prior to
Issue 6). The code as currently written complies rigidly with the C++03 and
POSIX specs regardless of the underlying types involved.

seirl:
Oh, I didn't know that this was supposed to be compatible with C++03. Nevermind then :)

PR #646, opened 2015-12-01
Add table of contents to docs

petersohn:
The TOCs are created with the following script: https://github.com/petersohn/rcfiles/blob/master/home/bin/gen_toc

This is the improved version of this script: https://github.com/sabel83/metashell/blob/master/tools/gen_toc


KindDragon:
:+1:

googlemock/docs/CheatSheet.md
```diff
@@ -559,4 +597,4 @@ See this [recipe](http://code.google.com/p/googlemock/wiki/CookBook#Using_Check_
 
 | `--gmock_catch_leaked_mocks=0` | Don't report leaked mock objects as failures. |
 |:-------------------------------|:----------------------------------------------|
-| `--gmock_verbose=LEVEL`        | Sets the default verbosity level (`info`, `warning`, or `error`) of Google Mock messages. |
\ No newline at end of file
+| `--gmock_verbose=LEVEL`        | Sets the default verbosity level (`info`, `warning`, or `error`) of Google Mock messages. |
```
BillyDonahue: what changed here? I don't see it.

BillyDonahue:
I worry about keeping these ToC up-to-date.
And about relying on somebody's script.
Can GitHub generate ToC? Some Markdown interpreters must have this as an extension.

googlemock/docs/CheatSheet.md
petersohn: Nothing. A newline maybe?

KindDragon:
GitHub can't do that :cry:

KindDragon:
It should be script that update all TOC that everyone can run

petersohn:
Maybe this script should be included in this repo, then it can be used every time a document is updated.

Farwaykorse:
This implementation relies on the specific way GitHub uses to make headers linkable.

To make it portable an anchor could be added to the actual headers:

     ## Exception Assertions ## <a id="h_exception_assertions" />

This will work in any Markdown implementation, including GitHub. Without loss of plain-text readability.

Adding a prefix might be a good idea. (e.g. `h` for header)  This would help detect and prevent accidental use of the GitHub reference. Which would silently break use outside of GitHub.com.

gennadiycivil:
This is sadly very out of date. There is also general question of TOC usefullness and its maintenance.

Issue #652, opened 2015-12-03
Global mock - double free or corruption


ajneu:
Hi, 

When using a global (static) mock,
and setting `::testing::GMOCK_FLAG(catch_leaked_mocks) = false;` 
(--> the global mock could **outlive** 
~~RUN_ALL_TESTS()~~ `g_mock_object_registry` (in whose destructor leakage is checked - [link](https://github.com/google/googletest/blob/1f8fe13618a5b23c937140a7c9ad498cd93392e1/googlemock/src/gmock-spec-builders.cc#L550)), 
so I switch off gmock's leakage detection...)

But I then get  
**double free or corruption**
right at the end of the program.
.
.
.

Here's some sample code https://github.com/ajneu/global_mock (see directory 1)
Please use the included cmake-file.
.
.
.
.

PS: 
Why use a global mock ??
The above repo is just a dummy. 
But in a real project, I'm using a mock for a static global-function that I need to use as a callback. 
I need to test that I can hook in such static functions (as opposed to member functions), so I'm using a global mock class, so that I can set appropriate expectations.

ajneu:
The problem is related to livetimes, 
regarding the _order of destruction_ between __two globals__:

On the one hand we have my global mock:
[`MockRenderer mock_renderer;`](https://github.com/ajneu/global_mock/blob/e5a7cb2dd005abc58a7e06e3bee313f13d30864d/test_printer.cpp#L9)

On the other hand we have the global g_mock_object_registry:
[`MockObjectRegistry g_mock_object_registry;`](https://github.com/google/googletest/blob/1f8fe13618a5b23c937140a7c9ad498cd93392e1/googlemock/src/gmock-spec-builders.cc#L599)

.
.

What happens in the example (with the corruption)
>Here's some sample code https://github.com/ajneu/global_mock

is:
1. Destructor of `g_mock_object_registry`: [`~MockObjectRegistry()`](https://github.com/google/googletest/blob/1f8fe13618a5b23c937140a7c9ad498cd93392e1/googlemock/src/gmock-spec-builders.cc#L550)
2. Destructor of mock_renderer: [`~MockRenderer()`](https://github.com/ajneu/global_mock/blob/e5a7cb2dd005abc58a7e06e3bee313f13d30864d/mockrenderer.h#L9)

That results in 
[...]

PR #653, opened 2015-12-03
global mocks outliving registry, don't access destructed registry

ajneu:
Global mocks may outlive the registry. 
Currently such mocks - in their destructors - still access the destructed registry!
(This commit should prevent that, and hence prevent mem-corruption.)

Related to #652

### DIFF
--- a/googlemock/include/gmock/gmock-spec-builders.h
+++ b/googlemock/include/gmock/gmock-spec-builders.h
@@ -215,6 +215,10 @@ class GTEST_API_ UntypedFunctionMockerBase {
       const void* untyped_args)
           GTEST_LOCK_EXCLUDED_(g_gmock_mutex);
 
+  void set_registry_destructed() {
+    registry_destructed = true;
+  }
+
  protected:
   typedef std::vector<const void*> UntypedOnCallSpecs;
 
@@ -239,6 +243,8 @@ class GTEST_API_ UntypedFunctionMockerBase {
 
   // All expectations for this function mocker.
   UntypedExpectations untyped_expectations_;
+
+  bool registry_destructed;
 };  // class UntypedFunctionMockerBase
 
 // Untyped base class for OnCallSpec<F>.
@@ -1468,7 +1474,8 @@ class FunctionMockerBase : public UntypedFunctionMockerBase {
         GTEST_LOCK_EXCLUDED_(g_gmock_mutex) {
     MutexLock l(&g_gmock_mutex);
     VerifyAndClearExpectationsLocked();
-    Mock::UnregisterLocked(this);
+    if (!registry_destructed)
+      Mock::UnregisterLocked(this);
     ClearDefaultActionsLocked();
   }
 
--- a/googlemock/include/gmock/gmock-spec-builders.h
+++ b/googlemock/include/gmock/gmock-spec-builders.h
@@ -215,8 +215,8 @@ class GTEST_API_ UntypedFunctionMockerBase {
       const void* untyped_args)
           GTEST_LOCK_EXCLUDED_(g_gmock_mutex);
 
-  void set_registry_destructed() {
-    registry_destructed = true;
+  void SetRegistryDestructed() {
+    registry_destructed_ = true;
   }
 
  protected:
@@ -244,7 +244,7 @@ class GTEST_API_ UntypedFunctionMockerBase {
   // All expectations for this function mocker.
   UntypedExpectations untyped_expectations_;
 
-  bool registry_destructed;
+  bool registry_destructed_;
 };  // class UntypedFunctionMockerBase
 
 // Untyped base class for OnCallSpec<F>.
@@ -383,12 +383,12 @@ class GTEST_API_ Mock {
   static void AllowLeak(const void* mock_obj)
       GTEST_LOCK_EXCLUDED_(internal::g_gmock_mutex);
 
-  // Tells Google Mock about global static mocks, so that they are
-  // not falsely detected as leaks, if they outlive the
-  // global g_mock_object_registry
-  // (this is if a global static mock's destructor is called AFTER
+  // Tells Google Mock about mocks with static storage duration,
+  // so that they are not falsely detected as leaks, if they
+  // outlive the static g_mock_object_registry
+  // (this is if a static mock's destructor is called AFTER
   //  g_mock_object_registry's destructor)
-  inline static void MarkGlobalStatic(const void* mock_obj) {
+  static void MarkStatic(const void* mock_obj) {
      AllowLeak(mock_obj);
   }
 
@@ -1483,7 +1483,7 @@ class FunctionMockerBase : public UntypedFunctionMockerBase {
         GTEST_LOCK_EXCLUDED_(g_gmock_mutex) {
     MutexLock l(&g_gmock_mutex);
     VerifyAndClearExpectationsLocked();
-    if (!registry_destructed)
+    if (!registry_destructed_)
       Mock::UnregisterLocked(this);
     ClearDefaultActionsLocked();
   }
--- a/googlemock/include/gmock/gmock-spec-builders.h
+++ b/googlemock/include/gmock/gmock-spec-builders.h
@@ -79,8 +79,6 @@
 
 namespace testing {
 
-class MockObjectRegistry; // forward declaration
-
 // An abstract handle of an expectation.
 class Expectation;
 
@@ -242,7 +240,7 @@ class GTEST_API_ UntypedFunctionMockerBase {
   // All expectations for this function mocker.
   UntypedExpectations untyped_expectations_;
 
-  MockObjectRegistry *&g_mock_object_registry_p;
+  bool &g_mock_object_registry_valid_;
 };  // class UntypedFunctionMockerBase
 
 // Untyped base class for OnCallSpec<F>.
@@ -1472,7 +1470,7 @@ class FunctionMockerBase : public UntypedFunctionMockerBase {
         GTEST_LOCK_EXCLUDED_(g_gmock_mutex) {
     MutexLock l(&g_gmock_mutex);
     VerifyAndClearExpectationsLocked();
-    if (g_mock_object_registry_p)
+    if (g_mock_object_registry_valid_)
       Mock::UnregisterLocked(this);
     ClearDefaultActionsLocked();
   }
--- a/googlemock/include/gmock/gmock-spec-builders.h
+++ b/googlemock/include/gmock/gmock-spec-builders.h
@@ -383,6 +383,15 @@ class GTEST_API_ Mock {
   static void AllowLeak(const void* mock_obj)
       GTEST_LOCK_EXCLUDED_(internal::g_gmock_mutex);
 
+  // Tells Google Mock about global static mocks, so that they are
+  // not falsely detected as leaks, if they outlive the
+  // global g_mock_object_registry
+  // (this is if a global static mock's destructor is called AFTER
+  //  g_mock_object_registry's destructor)
+  inline static void MarkGlobalStatic(const void* mock_obj) {
+     AllowLeak(mock_obj);
+  }
+
   // Verifies and clears all expectations on the given mock object.
   // If the expectations aren't satisfied, generates one or more
   // Google Test non-fatal failures and returns false.
--- a/googlemock/src/gmock-spec-builders.cc
+++ b/googlemock/src/gmock-spec-builders.cc
@@ -487,8 +487,31 @@ bool UntypedFunctionMockerBase::VerifyAndClearExpectationsLocked()
       // takes care of it.
       untyped_expectation->MaybeDescribeExtraMatcherTo(&ss);
       untyped_expectation->DescribeCallCountTo(&ss);
-      Expect(false, untyped_expectation->file(),
-             untyped_expectation->line(), ss.str());
+      if (! registry_destructed) {
+         Expect(false, untyped_expectation->file(),
+                untyped_expectation->line(), ss.str());
+      } else {
+         /* We are at the end of the program.
+            There is a high chance that the call ...
+
+            Expect(false, untyped_expectation->file(),
+                untyped_expectation->line(), ss.str());
+
+            ... will fail (segfault), because
+
+            Expect()                                                 calls
+            GetFailureReporter()->ReportFailure                which calls
+            AssertHelper::operator=(const Message& message)    which calls
+            UnitTest::GetInstance()                            which returns a pointer to
+            static UnitTest instance;
+
+            And that static instance is probably already destructed. (Same as the registry).
+            Therefore write out the message here.
+         */
+         std::cout << "Warning: Mocks (typically global static mocks), should not be verified on destruction (at program ending),\n"
+                      "but explicitly by calling Mock::VerifyAndClearExpectations(&your_global_mock_obj)\n"
+                   << untyped_expectation->file() << ':' << untyped_expectation->line() << ": Failure\n" << ss.str() << std::endl;
+      }
     }
   }
 
@@ -578,7 +601,9 @@ class MockObjectRegistry {
              << state.first_used_test << ")";
       }
       std::cout << " should be deleted but never is. Its address is @"
-           << it->first << ".";
+           << it->first << ".\n"
+              "   If the mock happens to be a global static, it should be marked with ::testing::Mock::MarkGlobalStatic(&my_global_mock)\n"
+              "   and it's expectations explicitly verified, e.g. with Mock::VerifyAndClearExpectations(&my_global_mock);";
       leaked_count++;
     }
     if (leaked_count > 0) {
@@ -665,7 +690,7 @@ void Mock::AllowLeak(const void* mock_obj)
   internal::MutexLock l(&internal::g_gmock_mutex);
   g_mock_object_registry.states()[mock_obj].leakable = true;
 }
-
+   
 // Verifies and clears all expectations on the given mock object.  If
 // the expectations aren't satisfied, generates one or more Google
 // Test non-fatal failures and returns false.
--- a/googlemock/src/gmock-spec-builders.cc
+++ b/googlemock/src/gmock-spec-builders.cc
@@ -508,8 +508,8 @@ bool UntypedFunctionMockerBase::VerifyAndClearExpectationsLocked()
             And that static instance is probably already destructed. (Same as the registry).
             Therefore write out the message here.
          */
-         std::cout << "Warning: Mocks (typically global static mocks), should not be verified on destruction (at program ending),\n"
-                      "but explicitly by calling Mock::VerifyAndClearExpectations(&your_global_mock_obj)\n"
+         std::cout << "   Warning: Mocks (typically global static mocks), should not be verified on destruction (if this is at program ending!),\n"
+                      "   but explicitly by calling Mock::VerifyAndClearExpectations(&your_global_mock_obj)\n"
                    << untyped_expectation->file() << ':' << untyped_expectation->line() << ": Failure\n" << ss.str() << std::endl;
       }
     }
--- a/googlemock/src/gmock-spec-builders.cc
+++ b/googlemock/src/gmock-spec-builders.cc
@@ -50,7 +50,7 @@
 
 namespace testing {
 
-static MockObjectRegistry *g_mock_object_registry_ptr;
+static bool g_mock_object_registry_valid;
 
 namespace internal {
 
@@ -271,7 +271,7 @@ void ReportUninterestingCall(CallReaction reaction, const string& msg) {
 }
 
 UntypedFunctionMockerBase::UntypedFunctionMockerBase()
-    : mock_obj_(NULL), name_(""), g_mock_object_registry_p(g_mock_object_registry_ptr) {}
+    : mock_obj_(NULL), name_(""), g_mock_object_registry_valid_(g_mock_object_registry_valid) {}
 
 UntypedFunctionMockerBase::~UntypedFunctionMockerBase() {}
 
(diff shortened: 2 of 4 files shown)


ajneu:
Usage:

When using global mocks, one should now mark them as leakable.

The behaviour when `g_mock_object_registry` is destructed is then:
it checks if a still-registered mock is marked as leakable ([link](https://github.com/google/googletest/pull/653/files#diff-369617aab0757c0c98dec11d4fcd3e14R560)).
If so, then it accesses that mock and sets it's member `registry_destructed` to `true` ([link](https://github.com/google/googletest/pull/653/files#diff-369617aab0757c0c98dec11d4fcd3e14R565) and [link](https://github.com/google/googletest/pull/653/files#diff-294a2f78240ff7f44557b6f0490f9312R218))

* If that still-registered mock is on the free store (heap), then it was really leaked.
* If that still-registered mock is a global, then it will not really leak: It's destructor will still be called. But because we've now marked `registry_destructed == true`, that global mock's destructor will now conditionally (in the if) **not** call `Mock::UnregisterLocked` (if it would, we'd end up accessing the destructed registry and get corruption): [Link](https://github.com/google/googletest/pull/653/files#diff-294a2f78240ff7f44557b6f0490f9312R1477)

ajneu:
Regarding the usage comment above:

> `::testing::Mock::AllowLeak(&my_global_mock);` 

might be confusing, since globals cannot leak.

.
.
Perhaps one should additionally have a new function
```
::testing::Mock::MarkGlobalStatic(&my_global_mock);
```

That has a **sensible name**.

But (under the hood) the function `::testing::Mock::MarkGlobalStatic` would do exactly the same as [`testing::Mock::AllowLeak`](https://github.com/google/googletest/blob/1f8fe13618a5b23c937140a7c9ad498cd93392e1/googlemock/src/gmock-spec-builders.cc#L657), i.e. set `leakable = true`.

ajneu:
To see the code-improvements black-on-white, you can run the following tests:

**Current codebase and mem-corruption**
* clone  https://github.com/ajneu/global_mock.git
* then build each of the 4 examples (e.g. `cd 1; mkdir build; cd build; cmake ..; ./mytest`)
* run with: `valgrind ./mytest` and see it fail!!

**This Pull-request with no mem-corruption**
For the examples above, edit the lines of the 4 CMakeLists.txt 
* [1/CMakeLists.txt](https://github.com/ajneu/global_mock/blob/8fe0414197bbe0d6fda015d1fdf0759ba0096823/1/CMakeLists.txt#L65)
* [2/CMakeLists.txt](https://github.com/ajneu/global_mock/blob/8fe0414197bbe0d6fda015d1fdf0759ba0096823/2/CMakeLists.txt#L65)
* [3/CMakeLists.txt](https://github.com/ajneu/global_mock/blob/8fe0414197bbe0d6fda015d1fdf0759ba0096823/3/CMakeLists.txt#L65)
* [4/CMakeLists.txt](https://github.com/ajneu/global_mock/blob/8fe0414197bbe0d6fda015d1fdf0759ba0096823/4/CMakeLists.txt#L65)

changing it from the official repo
```cmake
GIT_REPOSITORY https://github.com/google/googletest.git # https://github.com/ajneu/googletest.git
```
to my pull-request-repo
```cmake
GIT_REPOSITORY https://github.com/ajneu/googletest.git
```

The rerun make and then each of the four examples with `valgrind ./mytest` and behold zero errors ;)

BillyDonahue:
Well, if we're being pedantic, `MarkGlobalStatic` isn't 100% right either unless the object has external linkage in the global namespace.

googlemock/include/gmock/gmock-spec-builders.h
```diff
@@ -215,6 +215,10 @@ class GTEST_API_ UntypedFunctionMockerBase {
       const void* untyped_args)
           GTEST_LOCK_EXCLUDED_(g_gmock_mutex);
 
+  void set_registry_destructed() {
```
BillyDonahue: Should follow the naming convention of SetOwnerAndName above.

googlemock/include/gmock/gmock-spec-builders.h
```diff
@@ -239,6 +243,8 @@ class GTEST_API_ UntypedFunctionMockerBase {
 
   // All expectations for this function mocker.
   UntypedExpectations untyped_expectations_;
+
+  bool registry_destructed;
```
BillyDonahue: private. use trailing underscore.

googlemock/include/gmock/gmock-spec-builders.h
```diff
@@ -377,6 +383,15 @@ class GTEST_API_ Mock {
   static void AllowLeak(const void* mock_obj)
       GTEST_LOCK_EXCLUDED_(internal::g_gmock_mutex);
 
+  // Tells Google Mock about global static mocks, so that they are
+  // not falsely detected as leaks, if they outlive the
+  // global g_mock_object_registry
+  // (this is if a global static mock's destructor is called AFTER
+  //  g_mock_object_registry's destructor)
+  inline static void MarkGlobalStatic(const void* mock_obj) {
```
BillyDonahue: already inline because it's in a class.

googlemock/src/gmock-spec-builders.cc
```diff
@@ -487,8 +487,31 @@ bool UntypedFunctionMockerBase::VerifyAndClearExpectationsLocked()
       // takes care of it.
       untyped_expectation->MaybeDescribeExtraMatcherTo(&ss);
       untyped_expectation->DescribeCallCountTo(&ss);
-      Expect(false, untyped_expectation->file(),
-             untyped_expectation->line(), ss.str());
+      if (! registry_destructed) {
+         Expect(false, untyped_expectation->file(),
+                untyped_expectation->line(), ss.str());
+      } else {
+         /* We are at the end of the program.
```
BillyDonahue: too much commentary.

googlemock/src/gmock-spec-builders.cc
```diff
@@ -487,8 +487,31 @@ bool UntypedFunctionMockerBase::VerifyAndClearExpectationsLocked()
       // takes care of it.
       untyped_expectation->MaybeDescribeExtraMatcherTo(&ss);
       untyped_expectation->DescribeCallCountTo(&ss);
-      Expect(false, untyped_expectation->file(),
-             untyped_expectation->line(), ss.str());
+      if (! registry_destructed) {
```
BillyDonahue: if (registry_destructed_) {
  ... do stuff...
  return;
}
... rest of function unchanged, not even indentation changes.

googlemock/src/gmock-spec-builders.cc
```diff
@@ -572,7 +601,9 @@ class MockObjectRegistry {
              << state.first_used_test << ")";
       }
       std::cout << " should be deleted but never is. Its address is @"
-           << it->first << ".";
+           << it->first << ".\n"
+              "   If the mock happens to be a global static, it should be marked with ::testing::Mock::MarkGlobalStatic(&my_global_mock)\n"
+              "   and it's expectations explicitly verified, e.g. with Mock::VerifyAndClearExpectations(&my_global_mock);";
```
BillyDonahue: s/it's/its/

googlemock/src/gmock-spec-builders.cc
```diff
@@ -572,7 +601,9 @@ class MockObjectRegistry {
              << state.first_used_test << ")";
       }
       std::cout << " should be deleted but never is. Its address is @"
-           << it->first << ".";
+           << it->first << ".\n"
+              "   If the mock happens to be a global static, it should be marked with ::testing::Mock::MarkGlobalStatic(&my_global_mock)\n"
```
BillyDonahue: 'global' is being used loosely here. I'd rather avoid the term.

googlemock/src/gmock-spec-builders.cc
```diff
@@ -659,7 +690,7 @@ void Mock::AllowLeak(const void* mock_obj)
   internal::MutexLock l(&internal::g_gmock_mutex);
   g_mock_object_registry.states()[mock_obj].leakable = true;
 }
-
+
```
BillyDonahue: revert this whitespace?

googlemock/src/gmock-spec-builders.cc
ajneu: >if (registry_destructed_) {
... do stuff...
return;
}
... rest of function unchanged, not even indentation changes.

That `if` is actually in a for-loop. One probably should not cut short the iterations by an immediate return (return is also bool, not void). 

My thinking was: 
Use a conditional, to choose between `Expect` or `std::cout`.
Note: Both branches (if and else) rely on common code (`::std::stringstream ss;` ...)

ajneu:
> Well, if we're being pedantic, MarkGlobalStatic isn't 100% right either unless the object has external linkage in the global namespace.

Ok right. I've renamed it to `MarkStatic`. In the comments I now use "static storage location" instead of the loose "global".

Thanks for looking at the code. I've incorporated your suggestions. (In 2 locations I've also shortened the lines.)

googlemock/include/gmock/gmock-spec-builders.h
```diff
@@ -377,6 +383,15 @@ class GTEST_API_ Mock {
   static void AllowLeak(const void* mock_obj)
       GTEST_LOCK_EXCLUDED_(internal::g_gmock_mutex);
 
+  // Tells Google Mock about mocks with static storage duration,
+  // so that they are not falsely detected as leaks, if they
+  // outlive the static g_mock_object_registry
+  // (this is if a static mock's destructor is called AFTER
+  //  g_mock_object_registry's destructor)
+  static void MarkStatic(const void* mock_obj) {
```
BillyDonahue: If this is nothing but a synonym for the same AllowLeak function, maybe we shouldn't have it.
It's expanding the API. Maybe someone wouldn't know whether to use AllowLeak or MarkStatic, and they don't need to choose.

googlemock/src/gmock-spec-builders.cc
```diff
@@ -487,8 +487,24 @@ bool UntypedFunctionMockerBase::VerifyAndClearExpectationsLocked()
       // takes care of it.
       untyped_expectation->MaybeDescribeExtraMatcherTo(&ss);
       untyped_expectation->DescribeCallCountTo(&ss);
-      Expect(false, untyped_expectation->file(),
-             untyped_expectation->line(), ss.str());
+      if (! registry_destructed_) {
```
BillyDonahue: The if (registry_destructed_) case can go first, with a return.
Then you don't change the Expect(false,... ); statement.

googlemock/src/gmock-spec-builders.cc
```diff
@@ -487,8 +487,24 @@ bool UntypedFunctionMockerBase::VerifyAndClearExpectationsLocked()
...
+        Expect(false, untyped_expectation->file(),
+               untyped_expectation->line(), ss.str());
+      } else {
+        // Program nearing end: static registry already destructed.
+        // Other statics (such as obj pointed to by return of
+        // UnitTest::GetInstance(); which is called by Expect)
+        // probably also already destructed...
+        // -> Write out message immediately, instead of calling Expect.
+        std::cout << "   Warning: Mocks (typically mocks with static storage duration), "
```
BillyDonahue: The phrasing is weird. Eliminate the parens. The parens are optional information.
If you take them out, you have a false statement.
Mocks can normally be verified on destruction.

googlemock/src/gmock-spec-builders.cc
```diff
@@ -551,19 +567,25 @@ class MockObjectRegistry {
...
-
     int leaked_count = 0;
     for (StateMap::const_iterator it = states_.begin(); it != states_.end();
          ++it) {
-      if (it->second.leakable)  // The user said it's fine to leak this object.
+      const MockObjectState& state = it->second;
+
+      if (!GMOCK_FLAG(catch_leaked_mocks)
+          || state.leakable)              // The user said it's fine to leak this object.
+      {
```
BillyDonahue: join brace to previous line.

googlemock/src/gmock-spec-builders.cc
```diff
@@ -551,19 +567,25 @@ class MockObjectRegistry {
...
+        for (FunctionMockers::iterator mocker_it = state.function_mockers.begin();
+             mocker_it != state.function_mockers.end();
+             ++mocker_it) {
+           (*mocker_it)->SetRegistryDestructed();
```
BillyDonahue: Would life be easier if we just never allowed the global registry to die?

googlemock/include/gmock/gmock-spec-builders.h
```diff
@@ -239,6 +243,8 @@ class GTEST_API_ UntypedFunctionMockerBase {
 
   // All expectations for this function mocker.
   UntypedExpectations untyped_expectations_;
+
+  bool registry_destructed_;
```
BillyDonahue: Maybe we should just keep a Registry* instead of a bool.
That's the registry we'd have to unregister with when we die, and that registry can reset this pointer to nullptr when IT dies.

googlemock/src/gmock-spec-builders.cc
ajneu: ???

(You probably missed the previous replay about that: [ref](https://github.com/google/googletest/pull/653#discussion_r46755716).
I'll repeat it below, since your suggestion is not clear to me.).

>if (registry_destructed_) case can go first

That if is actually in a for-loop. One probably should not cut short the iterations by an immediate return (return is also bool, not void).

My thinking was:
Use a conditional, to choose between Expect or std::cout.
Note also: Both branches (if and else) rely on common code (::std::stringstream ss; ...)

(Am I missing something?)

googlemock/src/gmock-spec-builders.cc
ajneu: Hmmm...

Either:

1. Tell people using static-storage-lifetime-mocks, to compile googlemock as a shared lib. I've noticed that when setting up with `cmake -DBUILD_SHARED_LIBS=ON ..`, the global registry in the shared lib `gmock.so`, seems to outlive other statics. (But I'm not sure about the legalese, i.e. exact standards guarantees, etc. (Seems platform specific, so I'm not sure there can be any guarantees))  
OR
2. Put the global registry on the free store (heap) and just don't delete it. (Not sure that's a nice way to go? But see below: stuff like this is already in gmock in certain ifdefs...)
3. Possibly use a shared pointer that's held in mocks, to make lifetime guarantees (nice in C++11, but I suppose googlemock wants to be compatible with C98??) 

But note:

>Would life be easier if we just never allowed the global registry to die?

It's not just the registry that is involved. There are **other statics** that cause the same trouble. An example is the object pointed to by the pointer returned from `UnitTest::GetInstance()`: it is a static ([ref](https://github.com/google/googletest/blob/35fb11efbe1a2761ce923f49a9df1a430e5d16be/googletest/src/gtest.cc#L3986)), that (I think) dies too early in my [example 3](https://github.com/ajneu/global_mock/tree/master/3): Try it! -- the segfault is in the call-tree of `Expect` . 
(Funnily enough one finds... an ifdef Version that reads [as follows](https://github.com/google/googletest/blob/35fb11efbe1a2761ce923f49a9df1a430e5d16be/googletest/src/gtest.cc#L3983) - basically never allowing it to die.)

I'm not sure how many static objects in total, are involved in such lifetime issues...

ajneu:
Regarding [your comment](https://github.com/google/googletest/pull/653#discussion_r46763185)
>Maybe we should just keep a Registry* instead of a bool.
That's the registry we'd have to unregister with when we die, and that registry can reset this pointer to nullptr when IT dies.

Hmmm... yes maby...

Here are some tests (C++98 ... quite ugly to initialize vectors):

Consider this mem-corruption example:
.
.
**corrupt printing of vector contents** (just an example)
```cpp
#include <iostream>
#include <vector>

class Registry {
public:
   Registry() : vec(arr, arr + sizeof(arr) / sizeof(*arr))
   {}
   
   ~Registry()
   {
      std::cout << "Destructor of Registry" << std::endl;
   }
   void print() {
      std::cout << "Print Registry: " << std::endl;
      for (std::vector<int>::const_iterator it = vec.begin(); it != vec.end(); ++it) {
         std::cout << '\t' << *it << std::endl;
      }
   }
private:
   std::vector<int> vec;
   static const int arr[3];
};
const int Registry::arr[] = {1, 2, 3};



extern Registry registry;

void print() {
   registry.print();
}

class MyClass {
public:
   ~MyClass() {
      print();
   }
};

MyClass myclass;
Registry registry;

int main()
{
   return 0;
}
```
.
.
**Possibility a) - registry will not die (put it on the heap an leak it)**
```cpp
#include <iostream>
#include <vector>

class Registry {
public:
   Registry() : vec(arr, arr + sizeof(arr) / sizeof(*arr))
   {}
   
   ~Registry()
   {
      std::cout << "Destructor of Registry" << std::endl;
   }
   void print() {
      std::cout << "Print Registry: " << std::endl;
      for (std::vector<int>::const_iterator it = vec.begin(); it != vec.end(); ++it) {
         std::cout << '\t' << *it << std::endl;
      }
   }
private:
   std::vector<int> vec;
   static const int arr[3];
};
const int Registry::arr[] = {1, 2, 3};



extern Registry *registry;

void print() {
   registry->print();
}

class MyClass {
public:
   ~MyClass() {
      print();
   }
};

MyClass myclass;
Registry *registry;

int main()
{
   registry = new Registry;
   return 0;
}
```
.
.
**Possibility b) - your suggestion**
>Maybe we should just keep a Registry* instead of a bool.
That's the registry we'd have to unregister with when we die, and that registry can reset this pointer to nullptr when IT dies.

```cpp
#include <iostream>
#include <vector>

class Registry; // forward declaration
Registry *registry_ptr = 0;

class Registry {
public:
   Registry() : vec(arr, arr + sizeof(arr) / sizeof(*arr))
   {
      registry_ptr = this;
   }
   
   ~Registry()
   {
      std::cout << "Destructor of Registry" << std::endl;
      registry_ptr = 0;
   }
   void print() {
      std::cout << "Print Registry: " << std::endl;
      for (std::vector<int>::const_iterator it = vec.begin(); it != vec.end(); ++it) {
         std::cout << '\t' << *it << std::endl;
      }
   }
private:
   std::vector<int> vec;
   static const int arr[3];
};
const int Registry::arr[] = {1, 2, 3};




void print() {
   if (registry_ptr != 0)
      registry_ptr->print();
}

class MyClass {
public:
   ~MyClass() {
      print();
   }
};

MyClass myclass;
Registry registry;

int main()
{
   return 0;
}
```

Somehow I quite like **Possibility b) - your suggestion** above. I'm not sure how well that example translates to gmock/gtest, but it seems to avoid the destruction problems of statics classes quite nicely, by using a static pointer. Such a static pointer is a single simple POD, which (I've read up) has a trivial destructor (!) and is therefore safe.

_A trivial destructor is a destructor that performs no action. Objects with trivial destructors don't require a delete-expression and may be disposed of by simply deallocating their storage. All data types compatible with the C language (POD types) are trivially destructible._ (from [here](http://en.cppreference.com/w/cpp/language/destructor#Trivial_destructor))

ajneu:
>Such a static pointer is a single simple POD, which (I've read up) has a trivial destructor (!) and is therefore (probably???) safe.

According to this example, it's safe:
```cpp
#include <iostream>
#include <cassert>

class X {
public:
   ~X() {
      std::cout << "~X" << std::endl;
   }
};

extern int b;

class Y {
public:
   ~Y() {
      std::cout << "~Y\tb: " << b << std::endl;
      assert(b == 123);
   }
};


X x;         // destructs last
Y y;         // destructs 2nd (can even access an already destructed object b. Reason: b is a POD and trivially destructed)
int b = 123; // destructs 1st (but nothing happens, since it is a POD that's trivially destructed)

int main()
{
   
   return 0;
}
```

ajneu:
Ok, I've got some code that uses static pointers pod...
(for registry: [here](https://github.com/ajneu/googletest/blob/78b60e876517be6d91734d046f11ba6be3c9b144/googlemock/src/gmock-spec-builders.cc#L53) and for UnitTest instance [decl](https://github.com/ajneu/googletest/blob/78b60e876517be6d91734d046f11ba6be3c9b144/googletest/include/gtest/gtest.h#L1328), [defin](https://github.com/ajneu/googletest/blob/78b60e876517be6d91734d046f11ba6be3c9b144/googletest/src/gtest.cc#L3975))

...which are set to a static class's `this` at the classes construction
(registry: [link](https://github.com/ajneu/googletest/blob/78b60e876517be6d91734d046f11ba6be3c9b144/googlemock/src/gmock-spec-builders.cc#L547) and UnitTest: [link](https://github.com/ajneu/googletest/blob/78b60e876517be6d91734d046f11ba6be3c9b144/googletest/src/gtest.cc#L4313))

...and set to `nullptr` at destruction.
(registry: [link](https://github.com/ajneu/googletest/blob/78b60e876517be6d91734d046f11ba6be3c9b144/googlemock/src/gmock-spec-builders.cc#L557) and UnitTest: [link](https://github.com/ajneu/googletest/blob/78b60e876517be6d91734d046f11ba6be3c9b144/googletest/src/gtest.cc#L4318))

See https://github.com/ajneu/googletest/tree/static_pointer

ajneu:
Hi Billy,

how do you like this commit? 
(Explanations are in the comment [above](https://github.com/google/googletest/pull/653#issuecomment-162663910))

ajneu:
> Damn... some tests fail. Seems the original commits were better than this method...

Fixed now. All tests pass.

ajneu:
Hi, 

this code is ready to be pulled from. 
(It's become quiet around this PR, but I'd like to see it, make it into googlemock. Comments?)

Thanks.

gennadiycivil:
This is sadly of of date, I see last activity on  Jan 8, 2016. There are a lot of outdated changes.

I believe that the best way forward is to close this PR and re-start the work with up to date source base.

PR #654, opened 2015-12-04
Fix lots of broken links in documentation

cmorse:
Also make some links relative so they work in forked repositories


BillyDonahue:
incredible. thanks!

Issue #656, opened 2015-12-06
Minimum CMake version should be 2.6.4, not 2.6.2


audiofanatic:
I'm just doing some testing of a change I'm working on for issue 655 and noticed that the CMakeLists.txt files list 2.6.2 as the minimum CMake version. My testing shows, however, that these do not build with 2.6.2 (chokes on the `include(cmake/internal_utils.cmake)` line in googletest/CMakeLists.txt), they only seem to build successfully with 2.6.4 or later. I'm using a pretty standard OS X 10.10.5 machine for testing. I think the `cmake_minimum_required()` lines need to be updated to 2.6.4, which is what the developer guidelines list as the minimum anyway.

BillyDonahue:
Sounds right

thedrow:
Should be closed no?

audiofanatic:
The CMakeLists.txt files have not actually been updated yet, they still specify 2.6.2 as the minimum version. I'll try to put together a pull request later today to address this.

BillyDonahue:
Ok thanks

Issue #655, opened 2015-12-06
Add gtest/gmock header dirs to targets if CMake version supports it


audiofanatic:
From version 2.8.11, CMake provides the `target_include_directories()` command. This can be used to attach header file directories to the target so that anything linking against that target also gets those header paths included automatically. This makes it much easier for client code to use libraries like gtest/gmock because simply linking against them is enough.

When gtest or gmock are being added to an existing project via `add_subdirectory()`, the `gtest` and `gmock` CMake targets are directly available. It would be great if `target_include_directories()` commands were added to both of these targets so that client code didn't have to explicitly add the header paths themselves. Such a change would obviously have to be protected with a conditional to only try to do so if the CMake version is 2.8.11 or later.

I'm about to make and test the required changes so I can create a pull request for this. Just creating this issue so other devs are aware and also because the dev guidelines asked me to. ;)

audiofanatic:
The linked pull request was merged, so closing this issue. #701 is a related but separate issue from the initial request here.

PR #657, opened 2015-12-06
Add include dirs to targets if CMake version supports it

audiofanatic:
I've verified on OS X 10.10.5 that with this change, googletest and googlemock builds with the following CMake versions:

- 2.6.4
- 2.8.10.2
- 2.8.11.1
- 3.4.1

I've also verified that the include directories are successfully attached to the targets when using CMake 2.8.11.1 and 3.4.1, and are not attached when using CMake 2.6.4 or 2.8.10.2, hence confirming the expected behaviour.

### DIFF
--- a/googlemock/CMakeLists.txt
+++ b/googlemock/CMakeLists.txt
@@ -92,6 +92,14 @@ cxx_library(gmock_main
             src/gmock-all.cc
             src/gmock_main.cc)
 
+# If the CMake version supports it, attach header directory information
+# to the targets for when we are part of a parent build (ie being pulled
+# in via add_subdirectory() rather than being a standalone build).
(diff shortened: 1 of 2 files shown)


audiofanatic:
Addresses issue #655

audiofanatic:
As mentioned in #656, the minimum CMake version should probably be 2.6.4, not 2.6.2 as currently required by the `cmake_minimum_required()` lines. If this is indeed the case, then the test for whether CMAKE_VERSION is defined here in this change set can be removed, leaving just the test for the CMAKE_VERSION value (CMAKE_VERSION is defined from 2.6.3 onwards).

BillyDonahue:
thanks

PR #658, opened 2015-12-07
Added docs for ExternalProject download during CMake step

audiofanatic:
Added content to the gtest and gmock build instructions with links to an
article explaining how to download googletest sources at CMake time
and then use `add_subdirectory()` to incorporate it directly into an
existing build. The preference for using CMake was made clearer in
the gmock readme and the advantage of building gtest and gmock
directly as part of an existing CMake build has been highlighted.

### DIFF
--- a/googletest/README.md
+++ b/googletest/README.md
@@ -59,7 +59,13 @@ cross-platform.). If you don't have CMake installed already, you can
 download it for free from <http://www.cmake.org/>.
 
 CMake works by generating native makefiles or build projects that can
-be used in the compiler environment of your choice.  The typical
+be used in the compiler environment of your choice.  You can either
+build Google Test as a standalone project or it can be incorporated
+into an existing CMake build for another project.
+
+#### Standalone CMake Project ####
+
+When building Google Test as a standalone project, the typical
 workflow starts with:
 
     mkdir mybuild       # Create a directory to hold the build output.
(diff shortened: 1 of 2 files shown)


googletest/README.md
```diff
@@ -80,13 +86,37 @@ using Visual Studio.
...
+
+To achieve this, the project needs to know where to find the gtest
+sources. One approach is to embed the sources directly into your
+project's source tree, but this is not recommended. Using something
+like git submodules may be a suitable alternative for some projects,
+but a more flexible and less invasive approach is to get the CMake
+project to download/copy the gtest sources as part of the configure
+step, then pull that downloaded gtest directory into the build using
+`add_subdirectory()`. This technique is relatively straightforward
+and is fully explained in
```
BillyDonahue: I'd like a quick example here so we're not completely dependent on the crascit article.

audiofanatic:
Hopefully that's about the level of example code you wanted to see. It should be enough for people to see how it's done without needing to read the article. There's a little more detail in the article explaining what's happening, but the essentials are there in the googletest readme now.

audiofanatic:
Sorry about the last commit, had to force push a correction to the commit message. I've updated the docs so that it is clearer about the other ways CMake can be used and to give less prominence to the Crascit article. I think it is more in keeping with the rest of the document's style now. The merging of #657 also meant I was able to update the example to be a bit cleaner.

audiofanatic:
Coming back to this after no activity/response for nearly a year. I've rebased and squashed the original set of commits to make it easier to see what has been changed (i.e. just the two .md files). I also slightly updated the instructions to include error checking on the two `execute_process()` calls. Is there anything more needed in order to get this merged?

PR #660, opened 2015-12-08
fixed link in googlemock documentation

waynix:
Hi I did repair the links in the documentation.
I tried to use relative links wherever possible so that it will work on other servers.


BillyDonahue:
thanks

PR #663, opened 2015-12-10
The googlemock CMakeLists.txt does not have proper install targets

sztomi:
Currently, `make install` does not install the gmock headers and libs.


BillyDonahue:
Duplicate of PR #650

PR #665, opened 2015-12-13
Removing gmock/gtest root directory from the list of include directories

bucienator:
Currently, if a project pulls in the gmock-all.cc or gtest-all.cc directly, the other .cc files included from the *-all.cc files are not found, unless the root directory of the gmock/gtest projects are added as include folders.
In my opinion it is cleaner to have only the "include" directory of gmock/gtest added as an include directory. This is what this fix does.

Was there a specific reason for setting it up in the current way?


gennadiycivil:
This is quite a large change and old one (Dec 13, 2015) . I am closing it, please feel free to re-open if you feel that this is important still

PR #670, opened 2015-12-23
Update GTEST_PROJECT_URL_ in internal/gtest-port.h

dawikur:
As repository moved to github. All urls need to be changed. There is
still plenty to do.


BillyDonahue:
thx!

PR #674, opened 2016-01-01
2.6.4 is the minimum CMake version, so enforce it

audiofanatic:
Addresses #656


audiofanatic:
@BillyDonahue This one seems to have sat idle for about a year now. It's small and should be pretty uncontroversial, so any reason it hasn't been merged?

BillyDonahue:
No reason. Looks good. Thanks

PR #677, opened 2016-01-06
Fixed GTestDir 

kerys:
GMock will not build successfully in VS 2010 unless GTestDir is fixed.


KindDragon:
Should be fixed in PR #723 with other errors

gennadiycivil:
Dup of [723](https://github.com/google/googletest/pull/723)

Issue #667, opened 2015-12-15
Support "noexcept" function specifiers


goofacz:
I'm looking forward to see "noexcept" function specifier support in Google Mock. At the moment I deal with number of virtual methods annotated with "noexcept" specifier and I cannot mock them using Google Mock.

pepe82sh:
@gennadiycivil Is this closed because there is a solution to work with? I followed the "trail" to #1532, but that one is still open.

gennadiycivil:
I have been cleaning up older and inactive GitHub googletest issues. You may see an issue "closed" if it appears to be inactive/abandoned 
Thank you

Issue #546, opened 2015-08-25
need to mock methods with usual access modifiers (volatile and const volatile) and exception specs (throw ...)


BillyDonahue:
_From @GoogleCodeExporter on August 24, 2015 22:39_

```
Currently gmock don't have macros for mocking volatile and const volatile 
methods.  There are several options to solve this:

1. provide

  MOCK_VOLATILE_METHOD0
  MOCK_VOLATILE_METHOD0_T
  MOCK_VOLATILE_METHOD0_WITH_CALLTYPE
  MOCK_VOLATILE_METHOD0_T_WITH_CALLTYPE

and friends as needed.

2. expose a variation of GMOCK_METHOD0_ to let the user write

  GMOCK_METHOD0(volatile, , Foo, int());
  GMOCK_METHOD0_T(const volatile, , Bar, int());

and so on.

#1 means adding 40 more macros and doesn't scale well.  I like #2 better.  Its 
syntax isn't as nice as #1, but good enough for occasional use.

If we go with #2, we should consider supporting exception specs at the same 
time.
```

Original issue reported on code.google.com by `w...@google.com` on 8 Sep 2010 at 8:55

_Copied from original issue: google/googlemock#122_

BillyDonahue:
_From @GoogleCodeExporter on August 24, 2015 22:39_

```
Correction: #1 means 80 more macros when we consider 'const volatile', and 160 
when we add exception spec into the picture.  Definitely not the way to go.
```

Original comment by `w...@google.com` on 8 Sep 2010 at 8:56

BillyDonahue:
_From @GoogleCodeExporter on August 24, 2015 22:39_

```
I've took a look into updating my partial implementation of #1 into something 
more like #2.

One issue is generating the identifier for the FunctionMocker member, which 
needs to include qualifiers.
Currently I supply volatile as another parameter to GMOCK_METHOD0_. Adding 
exception specs we end up with 6 parameters. This seems a bit much:
  GMOCK_METHOD0(const, volatile, exception-spec, call-type, name, signature);

The first 3 parameters are only separate arguments so that the FunctionMocker 
member identifier can be generated. With an alternate way to generate the 
identifier I think we could merge those 3 into a single param:
  GMOCK_METHOD0(qualifiers, call-type, name, signature);

Some options for generating the identifier:
- Rely on the __LINE__ or __COUNTER__ (non standard, in gcc-4.3 and msvc) 
macros.
- Might be possible to use some other macro hackery.

[...]

PR #681, opened 2016-01-11
Add support for "noexcept" specifier in MOCK_METHOD... macros.

goofacz:
I implemented "noexcept" function specifier in MOCK_METHOD... macros. Solution for issue #667.

### DIFF
--- a/googlemock/include/gmock/gmock-generated-function-mockers.h
+++ b/googlemock/include/gmock/gmock-generated-function-mockers.h
@@ -696,6 +696,55 @@ using internal::FunctionMocker;
 #define MOCK_CONST_METHOD9(m, ...) GMOCK_METHOD9_(, const, , m, __VA_ARGS__)
 #define MOCK_CONST_METHOD10(m, ...) GMOCK_METHOD10_(, const, , m, __VA_ARGS__)
 
+#ifdef GTEST_LANG_CXX11
+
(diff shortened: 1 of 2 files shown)


BillyDonahue:
We're looking for a solution that won't have a combinatoric problem if
other annotations need to be added to the method, like override or
ref-qualifiers (& or &&).

goofacz:
You're right. The idea behind my changes was to provide consequent interface (we have MOCK_METHODn, so we should have MOCK_NOEXCEPT_METHODn etc.). I was aware it opens doors for combinatoric problem.

What about giving users freedom of setting  function specifiers? You can introduce 
MOCK_METHODn_WITH_SPECIFIERS(method, specifiers, ...)? Then MOCK_METHODn and MOCK_CONST_METHODn can be reimplemented on top of MOCK_METHODn_WITH_SPECIFIERS without breaking API. 

Embedding specifier in macros' names closes possibility of extending them in future (maybe we will see new specifier in e.g. C++21?).

Bagira80:
A more general solution, which supports several combinations of function-qualifiers, was originally provided in issue #546.
The original patches do no longer apply cleanly, but I sorted the problems out and fixed them.
These changes are provided in pull-request #794.

However, for supporting functions with rvalue-references (`&&`) as parameters some additional internals of GoogleMock need to be adjusted, I assume.

Just so that you know:
We have been using these patches in our company for several years now and they worked well.

gennadiycivil:
No CLA, closing

PR #682, opened 2016-01-12
update relative path to google test directory

tsekityam:
after moving mock files to googlemock/ sub directory,  the relative path to google test directory is changed.


eldruin:
This should already be solved after PR #773 was merged. Please close if so.

PR #688, opened 2016-01-19
Python 3 support

### LINKED ISSUE #434: Errors running Make with Python 3
```
What steps will reproduce the problem?
1. Install Python 3.
2. Attempt to run 'make', which throws syntax errors in ./scripts/*.py.

What is the expected output? What do you see instead?
Expected it to use compile cleanly using Python 2, not the newer system default.

What version of Google Test are you using? On what operating system?
gtest-1.6.0, Arch Linux
```

Original issue reported on code.google.com by `egt...@gmail.com` on 19 May 2013 at 5:49

tamland:
Adds support for running googletests own tests and the 'fuse' script with python 3. Fixes #434

### DIFF
--- a/googletest/scripts/fuse_gtest_files.py
+++ b/googletest/scripts/fuse_gtest_files.py
@@ -60,7 +60,10 @@ __author__ = 'wan@google.com (Zhanyong Wan)'
 
 import os
 import re
-import sets
+try:
+  from sets import Set as set  # For Python 2.3 compatibility
+except ImportError:
+  pass
 import sys
 
 # We assume that this file is in the scripts/ directory in the Google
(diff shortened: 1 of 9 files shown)


googletest/scripts/fuse_gtest_files.py
```diff
@@ -90,10 +93,10 @@ def VerifyFileExists(directory, relative_path):
   """
 
   if not os.path.isfile(os.path.join(directory, relative_path)):
-    print 'ERROR: Cannot find %s in directory %s.' % (relative_path,
-                                                      directory)
-    print ('Please either specify a valid project root directory '
-           'or omit it on the command line.')
+    print('ERROR: Cannot find %s in directory %s.' % (relative_path,
```
BillyDonahue: I'm not a python expert, but doesn't this print function syntax require an import from the future?

tamland: Nope, that's only for *enforcing* new style syntax. Haven't done that as I've tried to keep python2.3 support (the print_function future is only available in 2.6+) ;)

tamland:
@BillyDonahue Will this be accepted?

BillyDonahue:
So the parens syntax is ok in 2.3?

tamland:
Yes, it will work in python 2.3. The parentheses is a widely used 'hack' for writing python 2/3 compatible code (in 2.x it will be interpreted the same, a keyword + an expression, only enclosed in parentheses. Nothing special)

PR #691, opened 2016-01-22
Add HasParam() for value-parametrized tests

rakhimov:
There is no restriction for value-parametrized tests
to be called only within TEST_P.
Upon calling within TEST_F,
SetUp, TearDown, and other member functions of a test fixture
can be made aware of the call without a parameter.

[Review on Reviewable](https://reviewable.io/reviews/google/googletest/691)

### DIFF
--- a/googletest/include/gtest/gtest.h
+++ b/googletest/include/gtest/gtest.h
@@ -1747,6 +1747,11 @@ class WithParamInterface {
   typedef T ParamType;
   virtual ~WithParamInterface() {}
 
+  // Returns true if the test is called with a parameter,
+  // i.e. within TEST_P.
+  // This function is helpful for test fixture member functions.
+  bool HasParam() const { return parameter_; }
+
   // The current parameter value. Is also available in the test fixture's
   // constructor. This member function is non-static, even though it only
   // references static data, to reduce the opportunity for incorrect uses


googletest/include/gtest/gtest.h
```diff
@@ -1751,6 +1751,11 @@ class WithParamInterface {
   typedef T ParamType;
   virtual ~WithParamInterface() {}
 
+  // Returns true if the test is called with a parameter,
+  // i.e. within TEST_P.
+  // This function is helpful for test fixture member functions.
+  bool HasParam() const { return parameter_; }
```
BillyDonahue: I'm a little confused about this. When wouldn't the caller know the kind of test fixture they're using?

rakhimov:
There is no way of knowing inside of test fixture member functions (SetUp, TearDown, constructor, destructor, etc).
For example, the following use case is valid.
TEST_F(ParamFixture, TEST_FOO) {}
The test may run with the default parameter or no parametrization.

BillyDonahue:
On Fri, Jan 22, 2016 at 1:18 PM, Olzhas Rakhimov <notifications@github.com>
wrote:

> There is no way of knowing inside of test fixture member functions (SetUp,
> TearDown, constructor, destructor, etc).
> For example, the following use case is valid.
> TEST_F(ParamFixture, TEST_FOO) {}
> The test may run with the default parameter or no parametrization.
>

Sorry, i think i need a little more context to evaluate that claim. What's
ParamFixture and what's TEST_FOO?

rakhimov:
```cpp

class FooTest : public ::testing::TestWithParam<const char*>
```

BillyDonahue:
Unfortunately, this answers neither of my questions.

    class FooTest : public ::testing::TestWithParam<const char*>

I need to know about TEST_FOO and ParamFixture. Thanks.

rakhimov:
Sorry, I accidentally submitted unfinished version. Take a look at the updated one.

BillyDonahue:
it will crash if called from a TEST_F anyway.
How would HashParam help?

rakhimov:
It will compile without a problem,
and there is no restriction on not using ParamFixture with TEST_F.
An example can be:
```cpp
virtual void SetUp { 
  if (!HasParam()) {
    // Don't call GetParam().
    // we are not inside TEST_P.
    // Do some default setup.
  } else {
   // Use the GetParam() instead of the defaults.
  }
}
```

eldruin:
I think a class that derives from testing::TestWithParam should always have a parameter. It would be unintuitive otherwise as _the parameter_ is what they add to the testing::Test class. The fact that it also derives from testing::Test and thus technically you can use it for a TEST_F (provided that you do not call GetParam() ) sounds to me like an opportunistic reuse (no offense intended).

Maybe an alternative for you would be to create a normal derivative class of testing::Test for your test, where you do whatever you would do without the parameter, and in a derivative of this one and testing::WithParamInterface<T>, you add anything parameter-related.

rakhimov:
I wouldn't consider the use without parameter an opportunistic case,
it is rather expected to behave well under TEST_F
because of the public inheritance and required "is a" substitutability.
I am exactly trying to avoid the extra class with "WithParamInterface".
I think having TEST_P instead of TEST_F is enough to deduce
that tests are parametrized.
Additionally renaming the test case with "CaseWithParam" is redundant and extra code.
The real thing I am missing is a default parameter for the parametrized test cases.
I am using HasParam() to set the default.
If there's a way to set a default parameter,
I wouldn't need to check.

rakhimov:
I signed it!

gennadiycivil:
AppVeyor and Travis CI are not OK Doing some cleanup and closing outstanding PRs that 
1) don't have all checks OK

PR #701, opened 2016-02-01
target_include_directories for gmock, should also include gtest headers

ajneu:
Regarding #655
As one can see, the target `gmock` or `gmock_main` ([ref](https://github.com/google/googletest/blob/a9b73f8139a92315f37e48a3ff3d43e3929055bc/googlemock/CMakeLists.txt#L84)), has the same functionality as as `gtest` or `gtest_main` ([ref](https://github.com/google/googletest/blob/a9b73f8139a92315f37e48a3ff3d43e3929055bc/googletest/CMakeLists.txt#L90)), therefore when linking against target `gmock` or `gmock_main`, one should get not only the gmock-includes, but also the gtest-includes.

### DIFF
--- a/googlemock/CMakeLists.txt
+++ b/googlemock/CMakeLists.txt
@@ -96,8 +96,8 @@ cxx_library(gmock_main
 # to the targets for when we are part of a parent build (ie being pulled
 # in via add_subdirectory() rather than being a standalone build).
 if (DEFINED CMAKE_VERSION AND NOT "${CMAKE_VERSION}" VERSION_LESS "2.8.11")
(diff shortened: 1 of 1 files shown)


audiofanatic:
This one has sit for some time now. Anything preventing it from being merged?

bolry:
Should not the gtest directories go before the gmock directories?

audiofanatic:
Actually, I think the more correct change would be to make gmock link to gtest (as PUBLIC). The gtest include directory would then be picked up as an interface include directory automatically.

audiofanatic:
@bolry I don't think the order should really matter anyway.

gennadiycivil:
No CLA, closing

Issue #533, opened 2015-08-25
Use override keyword with new C++11 compilers


BillyDonahue:
_From @GoogleCodeExporter on August 24, 2015 22:40_

```
The C++11 standard introduces new "override" and "final" keywords in virtual 
function declarations. The "override" keyword is especially useful in context 
of creating mock functions. Macros MOCK_METHOD* expand to virtual methods that 
have to be present in base class in order to be useful as mocks. And, since we 
have to define mocks and class interface separately, this creates a possible 
gap for bugs when changing method prototypes.

For example, changing a type of argument in base class method, without updating 
the associated MOCK_METHOD* macro, will define the mock method as overloading 
and compilers will silently allow such buggy code to compile. However, using 
the new C++11 feature, mismatched function declarations will result in compile 
error, thus allowing the programmer to clearly state his purpose.

In my projects, I patched GMock macros to optionally include the override 
keyword - see the attached diff. I think such improvement merits for inclusion 
in upstream, so I decided to create this issue.
```

Original issue reported on code.google.com by `piot...@gmail.com` on 15 Dec 2012 at 9:23

Attachments:
- [GMock-use-override.patch](https://storage.googleapis.com/google-code-attachments/googlemock/issue-157/comment-0/GMock-use-override.patch)

_Copied from original issue: google/googlemock#157_

BillyDonahue:
_From @GoogleCodeExporter on August 24, 2015 22:40_



Original comment by `w...@google.com` on 8 Mar 2013 at 5:42
* Changed state: **Accepted**

BillyDonahue:
_From @GoogleCodeExporter on August 24, 2015 22:40_

```
has this been fixed? just spend a few hours searching for a bug because there 
was no override keyword present :(
```

Original comment by `alex.lan...@googlemail.com` on 12 Nov 2013 at 1:37

BillyDonahue:
_From @GoogleCodeExporter on August 24, 2015 22:40_

```
Beware that this might not be a good thing to do, because the original 
statement:

"Macros MOCK_METHOD* expand to virtual methods that have to be present in base 
class in order to be useful as mocks"

is flawed, because there are situations where a mock method is not present in 
the mocked base class and still can be useful.

In the fact some of techniques described in the GoogleMock cookbook rely on 
creating mock methods which are not present in the original interface (base 
class), for example:
https://code.google.com/p/googlemock/wiki/CookBook#Simplifying_the_Interface_wit
[...]

PR #703, opened 2016-02-03
Use override keyword (reference: google/googletest#533)

AriehSchneier:
reference: google/googletest#533

This enables the override keyword (you need to turn on the GMOCK_USE_OVERRIDE macro to enable this feature, which allows people to migrate at will).


AriehSchneier:
I wasn't sure about adding tests for this new feature as its a compile time feature, not a runtime feature... Possibly you would need to change the CI process to build both with and without GMOCK_USE_OVERRIDE set.

AriehSchneier:
Cancelled this version as I accidentally stomped some changes (new to git/github)

PR #704, opened 2016-02-03
Use override keyword

AriehSchneier:
This enables the override keyword (you need to turn on the GMOCK_USE_OVERRIDE macro to enable this feature, which allows people to migrate at will).

reference: google/googletest#533

### DIFF
--- a/googlemock/include/gmock/gmock-generated-function-mockers.h
+++ b/googlemock/include/gmock/gmock-generated-function-mockers.h
@@ -353,9 +353,9 @@ using internal::FunctionMocker;
     GTEST_CONCAT_TOKEN_(gmock##constness##arity##_##Method##_, __LINE__)
 
 // INTERNAL IMPLEMENTATION - DON'T USE IN USER CODE!!!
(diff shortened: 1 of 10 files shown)


AriehSchneier:
I wasn't sure about adding tests for this new feature as its a compile time feature, not a runtime feature... Possibly you would need to change the CI process to build both with and without GMOCK_USE_OVERRIDE set.

BillyDonahue:
It's "NONOVERRIDE" if anything, but this is a controversial feature.
We want to go with a solution that scales without a combinatoric explosion in the macro count. Can that be done?

AriehSchneier:
The difference for override is we would like to change the default to include override and need a way to specify removal. So it is different to the other cases which need to add modifiers.

paulsapps:
Can't varadic templates be used instead of all these macros since override is C++11 only?

gennadiycivil:
This issue is fairly old and there hasn't been much activity on it. Closing, but please re-open if it still occurs.

PR #706, opened 2016-02-04
Allow macros inside of parametrized test names.

sheepmaster:
This allows doing things like TEST_P(TestFixture, MAYBE(TestName)) for nicer conditional test disabling.

### DIFF
--- a/googletest/include/gtest/gtest-param-test.h
+++ b/googletest/include/gtest/gtest-param-test.h
@@ -1375,7 +1375,10 @@ internal::CartesianProductHolder10<Generator1, Generator2, Generator3,
 }
 # endif  // GTEST_HAS_COMBINE
 
-
+// Use a macro to stringify the test (case) name, because direct stringification
+// does not work if one of the arguments is itself a macro
+// (https://gcc.gnu.org/onlinedocs/cpp/Stringification.html).
+# define GTEST_STRINGIFY_(name) #name
 
 # define TEST_P(test_case_name, test_name) \
   class GTEST_TEST_CLASS_NAME_(test_case_name, test_name) \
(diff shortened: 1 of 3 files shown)


sheepmaster:
@BillyDonahue, friendly ping? 😃

BillyDonahue:
i'm not sure about this one.

the pre-stringified form might be desired.

sheepmaster:
I don't think the unexpanded form would even be legal as a test name – after all, it would have to contain parentheses. I haven't seen any instances of test names that contain non-identifier characters.

Also, the current behavior is inconsistent with TEST_F, which does macro expansion in test names. (Another reason why I don't think anyone uses unexpanded macros in test names, as it would not work with TEST_F).

BillyDonahue:
On Tue, Feb 9, 2016 at 4:34 PM, Bernhard Bauer <notifications@github.com>
wrote:

> I don't think the unexpanded form would even be legal as a test name –
> after all, it would have to contain parentheses.
>
Wouldn't a macro that isn't function-like also be expanded?
#define X abc123

TEST_F(Case, X)   <== Would this 'X' turn into abc123

sheepmaster:
True, a non-function macro would also be expanded. In your specific example however, the "X" would already get expanded today, because TEST_F does that (and Chromium relies on this heavily to conditionally disable tests). It's just TEST_P that behaves differently for no apparent reason, and this has caused breakages in Chromium in the past.

BillyDonahue:
ok. well can we have a test that does the TEST_F expansion near the test
for TEST_P expansion just to emphasize that they should behaviorally match?

sheepmaster:
Sure! Done.

gennadiycivil:
No CLA, closing

sheepmaster:
Umm, where did you get that information? I'm a Chromium committer working for Google and created the commit from my Chromium account, so I should be covered under the Google / Chromium CLA. A search for my Github username on SignCLA also shows that I've signed the CLA on Dec 17, 2014 (https://goto.google.com/lsvwz).

Feel free to ping me directly (bauerb@) if you have any questions. Thanks!

gennadiycivil:
The PR showed that CLA was still "pending" please feel free to re-open, (I have been cleaning up for some time now)

gennadiycivil:
I just tried to re-open and it still shows
cla/google — Waiting for status to be reported. I am going to close again, please re-open and lets take a look

sheepmaster:
I can't seem to reopen this request, but I created a new one, which also seems to pass the CLA check.

PR #711, opened 2016-02-13
pull

wlsgur0726:


KindDragon:
Please use cmake generated solution for VS2013/2015 or just update VS2010 if you need one

gennadiycivil:
Out of date, does not build, no proper description. Cleaning up

PR #713, opened 2016-02-16
Change error message of EXPECT_EQ to treat lhs and rhs equivalently.

DarthGandalf:


pjaytycy:
Was there any discussion regarding why this commit was a good idea?

BillyDonahue:
Sorry, not yet. I think it's nice to treat both sides equivalently though.

apbarrero:
I like the current _expected vs actual_ convention; failure messages are more precise.

BillyDonahue:
I would call `expected` vs `actual` less precise, not more precise.
EXPECT_EQ doesn't assume anything about lhs or rhs other than that they are the left and right sides of an == that we expect to evaluate to a value contextually convertible to bool. Higher-level semantics are up to the caller.  This makes EXPECT_EQ and EXPECT_NE semanticaly neutral, aligned better with their cousins EXPECT_LE, EXPECT_GE, EXPECT_LT, EXPECT_GT. That's the thought, anyway.

Bu11etmagnet:
`expected` vs `actual` is useful, but the current version (expected on the lhs, actual on the rhs) makes it awkward and counter-intuitive, especially if switching between EXPECT_EQ and EXPECT_GT, for example. When you write EXPECT_GT(foo, bar), usually `foo` is the actual value and `bar` is the expected one (although EXPECT_GT has no concept of expected and actual), and so it's exactly the opposite to EXPECT_EQ/NE.

If the code is EXPECT_EQ(0, val) and the tested code changes so `val` is now greater than zero, the smallest change to make the test passing again is ... EXPECT_LT(0, val) !

"Semantically neutral" sounds nice, but it actually loses information, because the test writer knows which value is the expected one and which value is the actual one. It is up to the test writer to correctly order the arguments to EXPECT_EQ. The arguments of EXPECT_EQ are rarely on equal footing: almost always one of them is a tested value and the other is the expected value. Having this information in the error message is very useful:

```
expected: Chico
actual: Karl

```
carries much more information than

```
Chico
does not equal
Karl
```

when I'm left wondering whether I got Chico and it should have been Karl, or I got Karl and it should have been Chico.

Bu11etmagnet:
So, my preference would be for all EXPECT_ macros to have the concept of expected/actual, but have the actual always on the left and expected on the right. Failing that, maybe we could have two sets,  EXPECT_xx (with expected and actual) and COMPARE_xx (where there's only lhs and rhs).

BillyDonahue:
> When you write EXPECT_GT(foo, bar), usually foo is the actual value and bar is the expected one

I don't agree with that.

I was going to write that my hope is that if there's an expected and actual, the author would write EXPECT_THAT(actual, matcher); Where matcher is something like Eq or Lt.  But yeah, these matchers will put the `actual` on the lhs of a relational operator anyway.

The stronger argument though is that the original Yoda ordering `EXPECT_EQ(expected,actual)` has hundreds of thousands of calls. Too many of those calls get the ordering wrong, so the diagnostics are clearer for the incorrect cases when we don't apply a sense of expected or actual to those arguments.

DarthGandalf:
There is `EXPECT_THAT(actual, Eq(expected))` (or even without `Eq()`),
which has concept of expected/actual, though it's in gmock.

2016-03-18 14:13 GMT+00:00 Ráduly, Csaba <notifications@github.com>:

Bu11etmagnet:
Looks like EXPECT_THAT is a usable replacement for what EXPECT_EQ did, although the information is a bit less detailed.

PR #719, opened 2016-02-21
Fix formatting in AdvancedGuide.md

paul-wilkinson:
Put occurrences of "#include" in a code span so they are not interpreted
as headers.

Other documents were not broken because the #include was not at the
start of the line, but put them in code spans anyway just in case the
text gets refilled in the future.


BillyDonahue:
thanks!

PR #724, opened 2016-02-24
Read Bazel's $XML_OUTPUT_FILE environment variable

duggelz:
If $XML_OUTPUT_FILE is set, and $GTEST_OUTPUT and --gtest_output are not
specified, produce output as if GTEST_OUTPUT=xml:$XML_OUTPUT_FILE had
been set.

### DIFF
--- a/googletest/include/gtest/internal/gtest-port.h
+++ b/googletest/include/gtest/internal/gtest-port.h
@@ -2546,8 +2546,7 @@ bool ParseInt32(const Message& src_text, const char* str, Int32* value);
 // corresponding to the given Google Test flag.
 bool BoolFromGTestEnv(const char* flag, bool default_val);
 GTEST_API_ Int32 Int32FromGTestEnv(const char* flag, Int32 default_val);
-const char* StringFromGTestEnv(const char* flag, const char* default_val);
-std::string OutputFromGTestEnv(const char* default_val);
+std::string StringFromGTestEnv(const char* flag, const char* default_val);
 
 }  // namespace internal
 }  // namespace testing
--- a/googletest/include/gtest/internal/gtest-port.h
+++ b/googletest/include/gtest/internal/gtest-port.h
@@ -2547,7 +2547,7 @@ bool ParseInt32(const Message& src_text, const char* str, Int32* value);
 bool BoolFromGTestEnv(const char* flag, bool default_val);
 GTEST_API_ Int32 Int32FromGTestEnv(const char* flag, Int32 default_val);
 const char* StringFromGTestEnv(const char* flag, const char* default_val);
-std::string OutputFromGTestEnv(const char * default_val);
+std::string OutputFromGTestEnv(const char* default_val);
 
 }  // namespace internal
 }  // namespace testing
--- a/googletest/include/gtest/internal/gtest-port.h
+++ b/googletest/include/gtest/internal/gtest-port.h
@@ -2547,9 +2547,9 @@ bool ParseInt32(const Message& src_text, const char* str, Int32* value);
 bool BoolFromGTestEnv(const char* flag, bool default_val);
 GTEST_API_ Int32 Int32FromGTestEnv(const char* flag, Int32 default_val);
 const char* StringFromGTestEnv(const char* flag, const char* default_val);
+std::string OutputFromGTestEnv(const char * default_val);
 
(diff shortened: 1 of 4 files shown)


duggelz:
The naming, placement, etc, of this function was a bit arbitrary.  Let me know if you'd like it somewhere else.  I ran "make test", and tested with Bazel manually.

b/26945103

googletest/include/gtest/internal/gtest-port.h
```diff
@@ -2547,9 +2547,9 @@ bool ParseInt32(const Message& src_text, const char* str, Int32* value);
 bool BoolFromGTestEnv(const char* flag, bool default_val);
 GTEST_API_ Int32 Int32FromGTestEnv(const char* flag, Int32 default_val);
 const char* StringFromGTestEnv(const char* flag, const char* default_val);
+std::string OutputFromGTestEnv(const char * default_val);
```
BillyDonahue: use consistent * placement.

googletest/src/gtest-port.cc
```diff
@@ -1235,5 +1235,24 @@ const char* StringFromGTestEnv(const char* flag, const char* default_value) {
...
+// a filename without the "xml:" prefix of GTEST_OUTPUT.
+std::string OutputFromGTestEnv(const char* default_value) {
+#if defined(GTEST_GET_STRING_FROM_ENV_)
+  return GTEST_GET_STRING_FROM_ENV_("output", default_value);
+#endif  // defined(GTEST_GET_STRING_FROM_ENV_)
+  const char* value = StringFromGTestEnv("output", NULL);
+  if (value) {
+    return value;
+  }
+  value = posix::GetEnv("XML_OUTPUT_FILE");
```
BillyDonahue: Should this be a call to StringFromGTestEnv("XML_OUTPUT_FILE", NULL) ?
I think the idea of StringFromGTestEnv is to add an abstraction for posix::GetEnv.

googletest/include/gtest/internal/gtest-port.h
duggelz: Fixed

googletest/src/gtest-port.cc
duggelz: I agree it's clumsy.  The problem is that StringFromGTestEnv would append a GTEST_ to the front, and look for GTEST_XML_OUTPUT_FILE, which wouldn't be right.

There are a couple similar calls elsewhere, e.g.:

    ./src/gtest-port.cc:1062:  const char* temp_dir = posix::GetEnv("TEMP");
    ./src/gtest.cc:2926:    const char* const term = posix::GetEnv("TERM");
    ./src/gtest.cc:4212:      NULL : internal::posix::GetEnv("TEST_PREMATURE_EXIT_FILE"));

googletest/src/gtest-port.cc
```diff
@@ -1235,5 +1235,24 @@ const char* StringFromGTestEnv(const char* flag, const char* default_value) {
   return value == NULL ? default_value : value;
 }
 
+// Reads and returns GTEST_OUTPUT and/or XML_OUTPUT_FILE environment
```
BillyDonahue: Please describe the relative priority of these variables.
We use GTEST_OUTPUT first if present, then fall back to XML_OUTPUT_FILE.

duggelz: Ok, restructured this to use StringFromGTestEnv as you mentioned above, and added comments about the relative priority.

googletest/test/gtest_env_var_test.py
```diff
@@ -98,6 +100,25 @@ def testEnvVarAffectsFlag(self):
       TestFlag('death_test_use_fork', '1', '0')
       TestFlag('stack_trace_depth', '0', '100')
 
+  def testXmlOutputFile(self):
+    """Test that $XML_OUTPUT_FILE affects the output flag."""
+
+    try:
+      # $XML_OUTPUT_FILE is overridden by $GTEST_OUTPUT
```
BillyDonahue: Can you restructure this test so that each of the scenarios is independent?
Right now each is building on state from the prior scenario.
Maybe refactoring a little with a subfunction would be clearer.

duggelz: Done

PR #736, opened 2016-03-08
Fixing containers' content matching

jhindin:
The pull request supposedly fixes the issue https://github.com/google/googlemock/issues/191

### DIFF
--- a/googlemock/include/gmock/gmock-matchers.h
+++ b/googlemock/include/gmock/gmock-matchers.h
@@ -2593,8 +2593,8 @@ class ContainerEqMatcher {
       }
       return !printed_header2 && !printed_header;
     }
-    return false;
 
+    return false;
   }
 
  private:
--- a/googlemock/include/gmock/gmock-matchers.h
+++ b/googlemock/include/gmock/gmock-matchers.h
@@ -2591,9 +2591,10 @@ class ContainerEqMatcher {
           UniversalPrint(*it, os);
         }
       }
+      return !printed_header2 && !printed_header;
     }
-
     return false;
+
   }
 
  private:


googlemock/include/gmock/gmock-matchers.h
```diff
@@ -2591,9 +2591,10 @@ class ContainerEqMatcher {
           UniversalPrint(*it, os);
         }
       }
+      return !printed_header2 && !printed_header;
     }
-
```
BillyDonahue: please revert this whitespace churn at the end here.

googlemock/include/gmock/gmock-matchers.h
```diff
@@ -2591,9 +2591,10 @@ class ContainerEqMatcher {
           UniversalPrint(*it, os);
         }
       }
+      return !printed_header2 && !printed_header;
```
BillyDonahue: I don't understand. Why are these printed_header bools determining the truth value of the matcher?

jhindin: When containers comparison fails at line 2554, the matcher attempts to compare containers by explicit traversal.
First, the matcher goes over value container and tests the presence of all its' elements in expected containe - lines 2561-2575.  Then the matcher goes over expected container and tests the presence of all its' elements in a value container, in lines 2579-2592. If container don't match, the error is sent to 'os' and correspondent flags are raised. 
If upon these two passes, both print flags are clear, both containers held the same entries's set and matcher should return 'true'. 

Perhaps, better logic would be run these two passes regardless of 'os' stream availability.

jhindin:
Hi

  After looking at code during the weekend, I came to realization that I've misunderstood the purpose of  the code in lines 2558-2593. According to my newly acquired understanding, the code in these lines serves to create concise explanation in case containers' comparison operation at line 2555 fails.
The explanation doesn't cover all cases, but it is not expected to double-guess the containers' comparison. Please, disregard this patch, I will create the new one, with more concise output.

PR #738, opened 2016-03-11
Create gtest-extend.h

turdusmerula:
I needed more verbosity from googletest during my dev process for exception assertions checks, mostly to show the content of the exception to help identify why unit test failed.

I fastly prototyped a solution that fits my needs but I know it's far from perfect. I propose it here to see if it can result in a new extension to googletest:

Two new macros:
- ASSERT_THROW_V: if the exception is any kind of std::exception then prints

`[ THROWED  ] exception content in green`
- ASSERT_NO_THROW_V: if an exception of any kind of std::exception throwed then prints

`[ THROWED  ] exception content in red`

It needs to be improved as it does not currently use the message coloring system of googletest (don't know where it stands), but the idea is there.

### DIFF
--- a/googletest/include/gtest/gtest-extend.h
+++ b/googletest/include/gtest/gtest-extend.h
@@ -1,0 +1,61 @@
+#ifndef GTEST_INCLUDE_GTEST_EXTEND_GTEST_H_
+#define GTEST_INCLUDE_GTEST_EXTEND_GTEST_H_
+
+#include <iostream>
+
+#define NONE "\033[0m"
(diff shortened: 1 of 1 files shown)


gennadiycivil:
Thank you for your submission. It is not  clear how and where this proposed mechanism would plug in with googletest. 
You are providing a new header file but it is not clear where this would be included, how the mechanism would work, there are no tests to cover this.
I suggest a discussion or providing more complete solution for consideration. In this form it is not acceptable.

PR #739, opened 2016-03-13
More concise output when containers are not equal

jhindin:
The output of ContainerEq matcher made more concise in cases where the containers contains the same set of entries, but container's comparison operator return 'false'; also, size difference is reported.


gennadiycivil:
Travis CI build failed. Doing some cleanup and closing outstanding PRs that 
1) don't have all checks OK

PR #741, opened 2016-03-16
Fix whitespace in paragraph

mbinna:


dcow:
Respectfully, I think this change should be reverted. It is common practice to use double space between sentences in fixed-width fonts. The markdown will render proper html, the double spaces exist for the benefit of someone reading the fixed-width source.

dcow:
Further, this is now inconsistent with respect to the rest of the markdown files in the project.

dcow:
That extra line in there can still be removed. I'd just change the spacing between sentences back to two spaces.

BillyDonahue:
Personally, I'm a single-spacer. The only thing I cared about when evaluating this PR was consistency within the file. I scrolled down and saw another single-space and accepted the merge.

But now that you mention it, if I'd scrolled up, I'd see more double-spaced sentences. So the file was already inconsistent and this PR neither made things better nor worse.

I'd like to convert it all to single-space just to end the controversy.

mbinna:
As @BillyDonahue correctly pointed out, the intent of this pull request is to merge the two paragraphs into one. I did not know that it is "common practice to use double space between sentences in fixed-width fonts". Is this an unwritten rule or is it documented somewhere? 

There are arguments for both approaches (single space, double space), but consistency is key. This should be handled in a separate pull request.

dcow:
In general, you should leave the formatting to whatever is displaying your text, anyway. But, we inevitably operate in the realm between the two worlds. For the record, I'm a single-spacer too for [everything save fixed-width fonts](http://graphicdesign.stackexchange.com/questions/2160/why-all-the-hate-two-spaces-after-the-period). I just find it easier to pick out sentences when reading long terminals full of text if there are two spaces. Indeed the reason the second space was introduced is because of type writers. Fixed width environments are the same thing. I didn't mean to kick up any dust. I just saw this change and wondered *why*.

I like this op-ed: *[Space Invaders](http://www.slate.com/articles/technology/technology/2011/01/space_invaders.html)* (= It's what convinced me to drop my double space habits elsewhere.

PR #721, opened 2016-02-22
Fix compilation of googletest with MinGW using Win32 threads

ilmagico:
Fixes #708.

Allows to compile googletest with MinGW when pthreads support is disabled.
Also, the cmake build automatically disables pthreads supportwhen MinGW is used.

Tested on Win7, with MinGW 4.8.1, cmake 3.5.0. All tests pass, except for gmock-matchers_test that fails because of a limitation of MinGW not being able to handle "too many sections", most likely unrelated to the issue being addressed here (see below).

Unfortunately, I am not able to (legally) test with MSVC compiler, anybody please help test.

Also tested on Linux, although I expect no differences.

cmake command line used for testing:
`cmake -G "MinGW Makefiles" -Dgtest_build_samples=ON -Dgmock_build_samples=ON -Dgtest_build_tests=ON -Dgmock_build_tests=ON ..`

MinGW "too many section" issue:
```
c:/mingw-2/bin/../lib/gcc/mingw32/4.8.1/../../../../mingw32/bin/as.exe: CMakeFiles\gmock-matchers_test.dir\test\gmock-matchers_test.cc.obj: too many sections (43319)
C:\Users\ilmagico\AppData\Local\Temp\cc4PXqrl.s: Assembler messages:
C:\Users\ilmagico\AppData\Local\Temp\cc4PXqrl.s: Fatal error: can't write CMakeFiles\gmock-matchers_test.dir\test\gmock-matchers_test.cc.obj: File too big
c:/mingw-2/bin/../lib/gcc/mingw32/4.8.1/../../../../mingw32/bin/as.exe: CMakeFiles\gmock-matchers_test.dir\test\gmock-matchers_test.cc.obj: too many sections (43319)
C:\Users\ilmagico\AppData\Local\Temp\cc4PXqrl.s: Fatal error: can't close CMakeFiles\gmock-matchers_test.dir\test\gmock-matchers_test.cc.obj: File too big
```
and (after using `make --keep-going`)
```
98% tests passed, 1 tests failed out of 59

Total Test time (real) =  26.45 sec

The following tests FAILED:
          9 - gmock-matchers_test (Not Run)
```
Apparently this limitation is lifted in recent versions of MinGW64 (when compiling in 64-bit mode), if I get a chance I'll try that too.

PR #750, opened 2016-03-27
Fix for Mingw32

M-griffin:
_MSC_VER is not caught in Mingw32, added _WIN32 to catch properly in
GCC.


KindDragon:
It should be fixed in PR#721

M-griffin:
Indeed it does, thanks for the heads up as I missed that.  I'll close this and wait for PR #721

Issue #692, opened 2016-01-23
Memory leak in ThreadLocalRegistryImpl::GetThreadLocalsMapLocked() on Windows


klugier:
Hello,

I am trying to port gtest to U++ bazar project (www.ultimatepp.org) , but it's memory dectors shows that there are memory leaks on MS Windows. I fount that one of this leaks comes from static ThreadLocalRegistryImpl method "ThreadIdToThreadLocals\* GetThreadLocalsMapLocked()" - gtest-port.cpp - line 532. Below is the original implementation:

``` cpp
  // Returns map of thread local instances.
  static ThreadIdToThreadLocals* GetThreadLocalsMapLocked() {
    mutex_.AssertHeld();
    static ThreadIdToThreadLocals* map = new ThreadIdToThreadLocals;
    return map;
  }
```

As you can see the problem here is that this static variable is never deleted in the gtest code. The solution here is simply, instead of allocating on stack let's allocate this on heap:

``` cpp
  // Returns map of thread local instances.
  static ThreadIdToThreadLocals* GetThreadLocalsMapLocked() {
    mutex_.AssertHeld();
    static ThreadIdToThreadLocals map;
    return &map;
  }
```

In this solution destructor of ThreadIdToThreadLocals  will be called when all static variable should be cleaned.

Sincerely,
Zbigniew Rębacz

nachokruss:
That would make things worse as local variables are destroyed after GetThreadLocalsMapLocked returns.

dawikur:
@nachokruss local yes, but static are not.

nachokruss:
Aha. You are right didn't see it was static, maybe put in a pull request then?

klugier:
Thanks for discussion guys. I will create pull request in near future.

coreykosak:
This is not a leak per se (where a leak is defined as "allocated memory
that has become inaccessible from any pointer"), but rather a failure to
destroy all allocated objects at program shutdown.

Put another way, that object is always reachable, so it's not a leak.

So far as I am aware (maybe the gtest owners want to comment more
authoritatively), I don't think we care about carefully destroying every
last object at shutdown.

Is there an argument (other than hygeine / warnings from the leak detector
you are using) why this design choice hurts your project?

paulsapps:
Its annoying to find real leaks with lots of on purpose leaked objects :)

gennadiycivil:
Should be addressed by #1142

PR #757, opened 2016-04-13
Fix memory leak (reachable) in ThreadIdToThreadLocals (Platform MS Windows)

klugier:
Hello,

I am trying to port gtest to U++ bazar project (www.ultimatepp.org) , but it's memory dectors shows that there are memory leaks on MS Windows. I fount that one of this leaks comes from static ThreadLocalRegistryImpl method "ThreadIdToThreadLocals\* GetThreadLocalsMapLocked()" - gtest-port.cpp - line 532. Below is the original implementation:

``` cpp
  // Returns map of thread local instances.
    mutex_.AssertHeld();
    static ThreadIdToThreadLocals* map = new ThreadIdToThreadLocals;
    return map;
  }
```

As you can see the problem here is that this static variable is never deleted in the gtest code. The solution here is simply, instead of allocating on stack let's allocate this on heap:

``` cpp
  // Returns map of thread local instances.
    mutex_.AssertHeld();
    static ThreadIdToThreadLocals map;
    return &map;
  }
```

In this solution destructor of ThreadIdToThreadLocals will be called when all static variable should be cleaned.

Sincerely,
Zbigniew Rębacz

---

Discussion in https://github.com/google/googletest/issues/692

### DIFF
--- a/googletest/src/gtest-port.cc
+++ b/googletest/src/gtest-port.cc
@@ -531,8 +531,8 @@ class ThreadLocalRegistryImpl {
   // Returns map of thread local instances.
   static ThreadIdToThreadLocals* GetThreadLocalsMapLocked() {
     mutex_.AssertHeld();
-    static ThreadIdToThreadLocals* map = new ThreadIdToThreadLocals;
-    return map;
+    static ThreadIdToThreadLocals map;
+    return &map;
   }
 
   // Protects access to GetThreadLocalsMapLocked() and its return value.


klugier:
I signed it!

klugier:
The above solution leads to crash. So, I am glad nobody response to it for 1.5 year. However, I found that some folks reports the other solution. I hope the problem will be solved in the future.

In U++ GTest bazaar project we deal with it by ignoring this leak and everything working as expected without warnings.

PR #759, opened 2016-04-22
Added explicit calling convention when needed

irwand:
It's needed in 2 different cases:
- when we define main function
- when we accept function pointers for standard runtime functions

This is only needed for Win32, since it supports different calling conventions. This change allows users to change MSVC default calling convention (using /Gz or /Gr) and still compiles.

### DIFF
--- a/googlemock/src/gmock_main.cc
+++ b/googlemock/src/gmock_main.cc
@@ -43,7 +43,7 @@
 
 GTEST_API_ int _tmain(int argc, TCHAR** argv) {
 #else
-GTEST_API_ int main(int argc, char** argv) {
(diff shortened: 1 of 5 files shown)


irwand:
To maintainer: I've signed CLA agreement under my gmail address, which is different from the email address I've used here. I've changed it so that github knows about that gmail address, as well as putting my github registered email address as alternate email in my google account. I don't know how to re-trigger the cla/google check or how to fix it at this point. Please advise, or take my word and we can get past that issue? Thx!

gennadiycivil:
Doing some cleanup and closing outstanding PRs that 
1) don't have all checks OK

Issue #761, opened 2016-04-24
WillOnce and WillRepeatedly should accept a list of actions


gsasha:
Then in most cases, we don't need to use DoAll.
This is both cleaner, as in, less code to write, and simpler - switching from one action to two and back doesn't involve plumbing changes.

PR #762, opened 2016-04-24
WillOnce and WillRepeatedly can accept s list of actions, so there is…

gsasha:
… no longer need to add DoAll
#761


gsasha:
I signed it!

gennadiycivil:
Apologies for dropping the ball on this, it is very old now and I see no other way but close it. Please re-submit if you think this is still relevant

PR #768, opened 2016-05-05
Generate import configs for other CMake projects to include

andreyvps:
for both Debug and Release builds by appending 'd' to the binary names
of the debug binaries.

### DIFF
--- a/googlemock/CMakeLists.txt
+++ b/googlemock/CMakeLists.txt
@@ -92,10 +92,21 @@ cxx_library(gmock_main
             src/gmock-all.cc
             src/gmock_main.cc)
 
+
 # If the CMake version supports it, attach header directory information
 # to the targets for when we are part of a parent build (ie being pulled
 # in via add_subdirectory() rather than being a standalone build).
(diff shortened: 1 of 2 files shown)


andreyvps:
I signed it!

Knitschi:
I did something similar in the hunter package fork of google test here: https://github.com/hunter-packages/googletest

I used the functions `write_basic_package_version_file()`  and `configure_package_config_file()` from the `CMakePackageConfigHelpers` module instead of the 

```
install(EXPORT GtestTargets
  FILE GmockConfig.cmake
  NAMESPACE Gtest::
  DESTINATION cmake
)
```
function that you use. Do you know what the difference is? The [documentation of the CMakePackageConfigHelpers](https://cmake.org/cmake/help/latest/module /CMakePackageConfigHelpers.html) states that this approach yields relocateable packages because the Config.cmake file does not contain hardcoded pathes. Do you know if the package that is generated with your implementation is relocateable?

gennadiycivil:
Conflicting files
googlemock/CMakeLists.txt
googletest/CMakeLists.txt

gennadiycivil:
This PR is fairly old and with merge conflicts. At the same time looks potentially useful.
About to close, ping me otherwise

Issue #683, opened 2016-01-14
sign mismatch warnings inconsistencies between MSVC and GCC/Clang


gpakosz:
I noticed the following gtest behavior.

On Windows, when compiling with MSVC, if you write:

```
uint32_t a = 1;
int32_t b = -1;
EXPECT_EQ(a, b);
```

You don't get any warning while GCC/Clang emits `-Wsign-compare`.

This is caused by commit 4b83461e9772cce62e84310060fe84172e9bf4ba by @zhanyong.wan that disables MSVC warning `4389` in `CmpHelperEQ`.

```
commit 4b83461e9772cce62e84310060fe84172e9bf4ba
Author: zhanyong.wan <zhanyong.wan@861a406c-534a-0410-8894-cb66d6ee9925>
Date:   Thu Jan 29 06:49:00 2009 +0000

    Fixes some warnings when compiled with MSVC at warning level 4.
```

Now, what should the proper behavior be?
1. should we see warnings when comparing signed/unsigned values with `ASSERT|EXPECT_EQ`?
   - in which case, `-Wsign-compare` should be silenced when compiling with GCC/Clang
2. should all `ASSERT|EXPECT_` macros emit warnings
   - in which case warning `4389` shouldn't be silenced in `CmpHelperEQ`
3. should sign/unsigned mismatch warnings be silenced for all `ASSERT|EXPECT_` macros? 
   - in which case both `CmpHelperEQ` and `GTEST_IMPL_CMP_HELPER_` definitions should silence warnings

I have a PR for each case mentioned above but I don't know which behavior to choose
- I think I would favor option 1.
- option 2. could be seen as a regression. 
- option 3. is dubious in the following example:
  
  ```
  uint32_t a = 1;
  int32_t b = -1;
  EXPECT_GT(a, b);
  ```

What do you think?

gpakosz:
Anyone?

BillyDonahue:
I don't know what to say here. It's hard to change googletest behavior in either direction. People are very picky in either direction when warnings change. If you enable a warning, some will demand that it be turned back off. If you disable a warning, some will demand that it be turned back on. 

I think these macros should emit exactly what the equivalent direct comparison would emit.

    uint32_t a = 1;
    int32_t b = -1;
    EXPECT_GT(a, b);

I don't care if this is dubious. I know it fails because 'b' is converted to uint32_t and then compared to a. But that's what the usual arithmetic conversions dictate and who is googletest to get in the way?

[...]

PR #778, opened 2016-05-18
Re-enable MSVC++ C4389 warning in CmdHelperEq()

gpakosz:
This PR re-enables C4389 as discussed in https://github.com/google/googletest/issues/683#issuecomment-196694742

C4389 was inhibited in commit 4b83461 making behavior inconsistent with other compilers.

### DIFF
--- a/googletest/include/gtest/gtest.h
+++ b/googletest/include/gtest/gtest.h
@@ -1386,11 +1386,9 @@ AssertionResult CmpHelperEQ(const char* lhs_expression,
                             const char* rhs_expression,
                             const T1& lhs,
                             const T2& rhs) {
-GTEST_DISABLE_MSC_WARNINGS_PUSH_(4389 /* signed/unsigned mismatch */)
   if (lhs == rhs) {
     return AssertionSuccess();
   }
-GTEST_DISABLE_MSC_WARNINGS_POP_()
 
   return CmpHelperEQFailure(lhs_expression, rhs_expression, lhs, rhs);
 }


gpakosz:
cc @BillyDonahue :)

BillyDonahue:
This PR will break some users. I'm reluctant to do it at least for now.
We need to make a v1.8.0 release summarizing the current state of googletest. Then we can start changing stuff like this once the wheel is turning.

gpakosz:
You're right it's going to change things for users.

I would argue it's for the good. I disagree with qualifying this PR as "breaking things for some users". The way googletest currently hides warnings is bad:

```
uint32_t a = 1;
int32_t b = -1;
EXPECT_GT(a, b);
```

BillyDonahue:
It will cause them to do work. They will be in a better position to do that
when they're upgrading to a new release.

gennadiycivil:
is this still relevant? @gpakosz

gpakosz:
What do you mean by "out of date"? The PR merges just fine and I believe it's still relevant.

I believe `GTEST_DISABLE_MSC_WARNINGS_PUSH_(4389 /* signed/unsigned mismatch */)` should be removed from `googletest/include/gtest/gtest.h` and `googlemock/scripts/gmock/gtest/gtest.h`

gennadiycivil:
@gpakosz
Ok so lets see that we could prove this.

This PR  restores equivalent behavior between MSVC and GCC/Clang - correct?

The reason I am hesitant is that yes this changes behavior. I don't see many complaints from users and there appears not to be a widespread ask for the changes.  I understand completely that this is not 100% logical however I am not yet convinced that these changes belong to the master as opposed to just your fork as an individual preference.

We can talk more I am open to hear arguments.

Thank you
G

gpakosz:
@gennadiycivil 

This PR indeed restores equivalent bevahior between MSCV and GCC/Clang.

And well I don't really see how I could pitch more for it. By the principle of least surprise, I believe it's wrong googletest silences warnings. The current behavior is also covering bugs and I'm sorry if people out there don't realize this. Which I wouldn't call a personal preference :)

gennadiycivil:
@gpakosz coming back to this PR, could you please merge the new changes to get it even with the rest of the code. I would like to take a look again, thank you very much!
G

gpakosz:
done

gennadiycivil:
@gpakosz could you please provide before/after output to clearly demonstrate how this change addresses the issue. It would be helpful

gpakosz:
Sure. As I mentioned at [the beginning of this discussion](https://github.com/google/googletest/pull/778#issuecomment-228146440), the current `master` behavior is inconsistent between gcc/clang and msvc:

If you compile the following `main.cpp` program,
```
#include <cstdint>
#include <gtest/gtest.h>

TEST(warnings, C4389)
{
  uint32_t a = 1;
  int32_t b = -1;
  EXPECT_GT(a, b);
}

int main(int argc, char* argv[])
{
  testing::InitGoogleTest(&argc, argv);

  int status = RUN_ALL_TESTS();

  return status;
}
```

You get to control the level of warning you want and it's not uncommon for projects to compile with `-Wsign-compare` (turned on by `-Wall` when compiling C++ code with g++) which produces:
```
→ /tmp g++ -Wsign-compare -I. main.cpp gtest/gtest-all.cc
In file included from main.cpp:2:
./gtest/gtest.h:20199:28: warning: comparison of integers of different signs: 'const unsigned int' and 'const int' [-Wsign-compare]
GTEST_IMPL_CMP_HELPER_(GT, >);
~~~~~~~~~~~~~~~~~~~~~~~~~~~^~
./gtest/gtest.h:20179:12: note: expanded from macro 'GTEST_IMPL_CMP_HELPER_'
  if (val1 op val2) {\
      ~~~~ ^  ~~~~
main.cpp:8:3: note: in instantiation of function template specialization 'testing::internal::CmpHelperGT<unsigned int, int>' requested here
  EXPECT_GT(a, b);
  ^
./gtest/gtest.h:20959:44: note: expanded from macro 'EXPECT_GT'
  EXPECT_PRED_FORMAT2(::testing::internal::CmpHelperGT, val1, val2)
                                           ^
1 warning generated.
```

The problem is that googletest silently inhibits the equivalent [`C4389`](https://msdn.microsoft.com/en-us/library/bx7x617k.aspx?f=255&MSPPError=-2147217396) warning and there's nothing you can do about it beside patching googletest itself.

I believe it's better to let the users which user of warning they want. Also I believe it's beneficial to let people realize comparing signed and unsigned integers is a bad idea. Isn't `1` greater than `-1` after all?

gennadiycivil:
@gpakosz once again, could you please merge the new changes to get it even with the rest of the code. 
I am testing internally in the meantime. 
Thank you

gpakosz:
I rebased this PR branch onto `origin/master`

PR #783, opened 2016-05-22
Performance fixes reported by cppcheck

tobbi:
Non-primitive iterators should use prefix increments instead of postfix increments.


tobbi:
I signed it!

tobbi:
I am also getting stuff like Variable  is reassigned a value before the old one has been used.
Do you care about those issues?

PR #792, opened 2016-06-03
Fix location capture in value parametrized tests

taka-no-me:
The location has to be associated with instance of internal ParameterizedTestCaseInfo::TestInfo structure instead of parent  ParameterizedTestCaseInfo object. Otherwise all cases sharing the same fixture inherit location information from the first test case met by compiler.


taka-no-me:
> Once you've signed, please reply here (e.g. I signed it!) and we'll verify. Thanks.

Done

gennadiycivil:

PR #793, opened 2016-06-03
Fix Google Mock's python tests

zonr:
They have been stopped working for a while because of a bug in CMake. Enabling the tests caught another bugs related to the googletest path. After these 2 fixes, `make test` for Google Mock should include 2 more tests `gmock_leak_test` and `gmock_output_test` and runs normally.


gennadiycivil:
continuous-integration/travis-ci/pr — The Travis CI build could not complete due to an error
Cleaning up PRs that
1) Don't have all checks OK
2) Have been around for a while with no activity

PR #794, opened 2016-06-07
[Fix 546] Added general macros MOCK_QUALIFIED_METHOD*.

Bagira80:
This fixes issue #546.

These new macros `MOCK_QUALIFIED_METHODx` are a generalized version of the
former macros `MOCK_METHODx` and `MOCK_CONST_METHODx`. They allow any
combination of function qualifiers (const, volatile, noexcept).

Versions for different call-types do also exist
(`MOCK_QUALIFIED_METHODx_WITH_CALLTYPE`) as well as templated versions
(`MOCK_QUALIFIED_METHODx_T` and `MOCK_QUALIFIED_METHODx_T_WITH_CALLTYPE`).

Example:

```
struct Example {
  virtual void func() = 0;
  virtual int constFunc(double) const = 0;
  virtual int constNoexceptFunc() const noexcept = 0;
};

struct ExampleMock : public Example {
  MOCK_QUALIFIED_METHOD0(func, , void());
  MOCK_QUALIFIED_METHOD1(constFunc, const, int(double));
  MOCK_QUALIFIED_METHOD0(constNoexceptFunc, const noexcept, int());
};
```

**Note**
This mainly applies the patches (from Edgar Ellis) referenced from issue #546 and fixes the errors which occurred while doing so (because the code evolved since the time the original patches were created).

### DIFF
--- a/googlemock/include/gmock/gmock-generated-function-mockers.h
+++ b/googlemock/include/gmock/gmock-generated-function-mockers.h
@@ -320,6 +320,15 @@ class FunctionMocker<R(A1, A2, A3, A4, A5, A6, A7, A8, A9, A10)> : public
   }
 };
 
+
+template<typename T> T& RemoveVolatileCast(T & v) {
+  return v;
+}
+template <typename T>
+T& RemoveVolatileCast(T volatile& v) {
+  return const_cast<T&>(v);
+}
+
 }  // namespace internal
 
 // The style guide prohibits "using" statements in a namespace scope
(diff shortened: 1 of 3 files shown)


evgen48:
Hi @KindDragon , do you know status of that patch? Can we kindly ask that this one to be applied in nearest future, since that is the only possibility for me to mock noexcept methods?

KindDragon:
I don't know

Bagira80:
Hi @BillyDonahue ,
could you please tell me what else is needed to get these macros into GoogleMock?

We have been using them successfully at our company for more than 3 years now.
Everything compiles cleanly without any errors.
The build-errors with Visual Studio 2015 (v140) that occurred on Appveyor are unrelated to these patches.

If you have another solution or do not want these patches, then please tell me. In that case I could close the ticket.

googlemock/include/gmock/gmock-generated-function-mockers.h
```diff
@@ -320,6 +320,15 @@ class FunctionMocker<R(A1, A2, A3, A4, A5, A6, A7, A8, A9, A10)> : public
   }
 };
 
+
+template<typename T> T& RemoveVolatileCast(T & v) {
```
BillyDonahue: Consistency:
  `(T& v)`, not `(T & v)`
  `template <typename T>`, not `template<typename T>`.

googlemock/include/gmock/gmock-generated-function-mockers.h.pump
```diff
@@ -99,6 +99,15 @@ $if i >= 1 [[
 
 
 ]]
+
+template<typename T> T& RemoveVolatileCast(T & v) {
```
BillyDonahue: Consistency with whitespace: `T& v`,  `template <typename T>`.

googlemock/include/gmock/gmock-generated-function-mockers.h.pump
```diff
@@ -140,24 +148,34 @@ $var as = [[$for j, [[gmock_a$j]]]]
...
   } \
   ::testing::MockSpec<__VA_ARGS__>& \
-      gmock_##Method($matcher_as) constness { \
-    GMOCK_MOCKER_($i, constness, Method).RegisterOwner(this); \
-    return GMOCK_MOCKER_($i, constness, Method).With($as); \
+      gmock_##Method($matcher_as) qualifiers { \
+    ::testing::internal::RemoveVolatileCast(*this) \
+        .GMOCK_MOCKER_($i, Method) \
+        .RegisterOwner(& ::testing::internal::RemoveVolatileCast(*this); \
+    return ::testing::internal::RemoveVolatileCast(*this) \
```
BillyDonahue: I don't understand what this is doing well enough to tell whether it's safe to cast the volatility away.
Why is it safe? Maybe document it.

Bagira80: To be honest, as I am neither the original author of these patches (I just applied them, fixed some obvious patch-errors and provided the pull-request; see my note in the first message and the referenced issue ticket) nor have I originally written the macros that get patched, I cannot tell for sure.

What I understand is that the macros `GMOCK_METHOD$i` are the macros that get called by the public macros `MOCK_METHOD$i` (and the new `MOCK_QUALIFIED_METHOD$i` macros).
`GMOCK_METHOD$i` then calls `GMOCK_MOCKER_($i, Method)` which just creates the name of the member-object which is the mocker for the particular function.
And to be able to call the required non-volatile member-function (`SetOwnerAndName`, `RegisterOwner`, `Invoke` etc.) the volatile qualifier must be casted away.

If the `this` object was not created as volatile in the first place, the volatile qualifier just got added by calling the (mocked) volatile function and can be safely casted away.
However, you are right, at this point we cannot tell where the volatile qualifier was added to `this`. Therefore we might trigger undefined behavior.

But is it possible (or useful at all) to create a volatile mock-object? (I cannot think of any useful use-case.)
If not, we could probably safely assume that volatile just got added by calling the (mocked) function and can therefore by casted away without triggering undefined behavior.

irwand:
National Instruments, the company I work for, has also been using this successfully since Sept 2016. We'd appreciate it if this patch makes it into the master branch :-)
Thank you much!

gennadiycivil:
continuous-integration/appveyor/pr — AppVeyor build failed
Doing cleanup, closing PRs that 
1) Dont have all checks OK
2) Have been around for a while with no activity.

irwand:
Please don't close this. This patch is very useful.
I don't actually see appveyor build fail record in here.

zhenechish:
As i see appveyor errors are :
C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\include\hash_map(17): error C2338: <hash_map> is deprecated and will be REMOVED. Please use <unordered_map>. You can define _SILENCE_STDEXT_HASH_DEPRECATION_WARNINGS to acknowledge that you have received this warning. [C:\projects\googletest\_build\googlemock\gtest\gtest-printers_test.vcxproj]
C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\include\hash_set(17): error C2338: <hash_set> is deprecated and will be REMOVED. Please use <unordered_set>. You can define _SILENCE_STDEXT_HASH_DEPRECATION_WARNINGS to acknowledge that you have 
Not related to the patch, 

I guess this warning already fixed in master, so just need to rebase patch maybe.

Bagira80:
I created a second [pull-request](https://github.com/google/googletest/issues/1168) (#1168) with the same changes, just this time rebased to the current master.
I hope that helps.

**Note:** For code using C++14 and above I nowadays recommend using [*Trompeloeil*](https://github.com/rollbear/trompeloeil) instead of *GoogleMock*.
It's just so much easier to use.

PR #795, opened 2016-06-08
Fix compilation error when both gtest and <tr1/tuple> are included (GTEST_LANG_CXX11=1)

oliora:
Fix compilation error when both gtest and `<tr1/tuple>` are included and `GTEST_LANG_CXX11` is defined.


oliora:
I signed it!

Issue #807, opened 2016-06-24
Add gyp file


midned:
What about adding a `.gyp` file, making easier to include `googletest` in projects using it?

PR #809, opened 2016-06-26
Added .gyp file for generating projects

alv-c:
Add a very simple `.gyp` file, making a lot easier to include `googletest`/`googlemock` to a project already using GYP. I created an issue for this #807 

Improvisations to this `.gyp` file could be made, I started using GYP not so long time ago so I'm still a beginner at using it; I just tested it, and it works properly

### DIFF
--- a/googletest.gyp
+++ b/googletest.gyp
@@ -1,0 +1,52 @@
+{
+    'targets': [
+        {
+            'target_name': 'googletest',
+            'type': 'static_library',
+            'sources': [
+                './googletest/src/gtest-all.cc'
+            ],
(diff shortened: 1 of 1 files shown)


alv-c:
I signed it!

gennadiycivil:
This is quite old, not sure how much interest there is in using GYP. Please feel free to re-open with more information if you still think this is still needed

alv-c:
Right now I'm not doing any C/C++ based development. But I still think this would be nice. This way I could simply clone this repo and include it in my project with ease.

PR #723, opened 2016-02-23
Run MSVC tests on AppVeyor

KindDragon:
Run tests on appveyor https://ci.appveyor.com/project/KindDragon/googletest

PR #811, opened 2016-06-27
VS2015/2017 build fix and enable C++11 features 

KindDragon:
Depend on PR #723 

Commit  1695708 allow use C++11 features available in VS2015,  fix failed tests with GMock under VC++2015 compiler

### DIFF
--- a/googletest/include/gtest/internal/gtest-port.h
+++ b/googletest/include/gtest/internal/gtest-port.h
@@ -323,7 +323,7 @@
 // -std={c,gnu}++{0x,11} is passed.  The C++11 standard specifies a
 // value for __cplusplus, and recent versions of clang, gcc, and
 // probably other compilers set that too in C++11 mode.
-# if __GXX_EXPERIMENTAL_CXX0X__ || __cplusplus >= 201103L
+# if __GXX_EXPERIMENTAL_CXX0X__ || __cplusplus >= 201103L || _MSC_VER >= 1900
 // Compiling in at least C++11 mode.
 #  define GTEST_LANG_CXX11 1
 # else
@@ -355,12 +355,16 @@
 #if GTEST_STDLIB_CXX11
 # define GTEST_HAS_STD_BEGIN_AND_END_ 1
 # define GTEST_HAS_STD_FORWARD_LIST_ 1
-# define GTEST_HAS_STD_FUNCTION_ 1
+# if !defined(_MSC_VER) || (_MSC_FULL_VER >= 190023824) // works only with VS2015U2 and better
+#   define GTEST_HAS_STD_FUNCTION_ 1
+# endif
 # define GTEST_HAS_STD_INITIALIZER_LIST_ 1
 # define GTEST_HAS_STD_MOVE_ 1
 # define GTEST_HAS_STD_SHARED_PTR_ 1
 # define GTEST_HAS_STD_TYPE_TRAITS_ 1
 # define GTEST_HAS_STD_UNIQUE_PTR_ 1
+# define GTEST_HAS_UNORDERED_MAP_ 1
+# define GTEST_HAS_UNORDERED_SET_ 1
 #endif
 
 // C++11 specifies that <tuple> provides std::tuple.
--- a/googletest/test/gtest_catch_exceptions_test_.cc
+++ b/googletest/test/gtest_catch_exceptions_test_.cc
@@ -138,7 +138,7 @@ TEST_F(CxxExceptionInConstructorTest, ThrowsExceptionInConstructor) {
 }
 
 // Exceptions in destructors are not supported in C++11.
-#if !defined(__GXX_EXPERIMENTAL_CXX0X__) &&  __cplusplus < 201103L && _MSC_VER < 1900
+#if !GTEST_LANG_CXX11
 class CxxExceptionInDestructorTest : public Test {
  public:
   static void TearDownTestCase() {
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,15 @@
 # Ignore CI build directory
 build/
 xcuserdata
+
+# Visual Studio files
+*.sdf
+*.opensdf
+*.VC.opendb
+*.suo
+*.user
+_ReSharper.Caches/
+Win32-Debug/
+Win32-Release/
+x64-Debug/
+x64-Release/
--- a/googlemock/test/gmock-generated-matchers_test.cc
+++ b/googlemock/test/gmock-generated-matchers_test.cc
@@ -120,7 +120,7 @@ TEST(ArgsTest, AcceptsOneTemplateArg) {
 }
 
 TEST(ArgsTest, AcceptsTwoTemplateArgs) {
-  const tuple<short, int, long> t(4, 5, 6L);  // NOLINT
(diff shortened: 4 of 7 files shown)


googletest/test/gtest_list_tests_unittest.py
```diff
@@ -71,7 +71,7 @@
 TypedTest/0\.  # TypeParam = (VeryLo{245}|class VeryLo{239})\.\.\.
   TestA
   TestB
-TypedTest/1\.  # TypeParam = int\s*\*
+TypedTest/1\.  # TypeParam = int\s*\*( __ptr64)?
```
BillyDonahue: There seems to be a lot going on here that isn't technically necessary.
I would like to have some idea of what's being changed but I can't review all of this at once.
Can you write a little in the "conversation" tab about what exactly is being changed and why?

KindDragon: This commit disappear after you merge PR #723. So ignore this commit here.
Specifically, these changes correct VC++ 64-bit tests

coreykosak:
I'm curious about something. We used to rely on CMake to auto-generate our Visual Studio solution and project files. Are we no longer doing that?

googletest/include/gtest/internal/gtest-port.h
```diff
@@ -355,7 +355,9 @@
 #if GTEST_STDLIB_CXX11
 # define GTEST_HAS_STD_BEGIN_AND_END_ 1
 # define GTEST_HAS_STD_FORWARD_LIST_ 1
-# define GTEST_HAS_STD_FUNCTION_ 1
+# if !defined(_MSC_VER) || (_MSC_FULL_VER >= 190023824) // works only with VS2015U2 and better
+#   define GTEST_HAS_STD_FUNCTION_ 1
```
KindDragon: VS2015 without updates doesn't support all required std::function features

googletest/include/gtest/internal/gtest-port.h
```diff
@@ -616,12 +618,21 @@ struct _RTL_CRITICAL_SECTION;
 // Determines if hash_map/hash_set are available.
 // Only used for testing against those containers.
 #if !defined(GTEST_HAS_HASH_MAP_)
-# if _MSC_VER
+# if defined(_MSC_VER) && (_MSC_VER < 1900)
 #  define GTEST_HAS_HASH_MAP_ 1  // Indicates that hash_map is available.
 #  define GTEST_HAS_HASH_SET_ 1  // Indicates that hash_set is available.
```
KindDragon: `hash_map` and `hash_set` not available in VS2015, but it support `unordered_map`

googletest/test/gtest_catch_exceptions_test_.cc
```diff
@@ -138,7 +138,7 @@ TEST_F(CxxExceptionInConstructorTest, ThrowsExceptionInConstructor) {
 }
 
 // Exceptions in destructors are not supported in C++11.
-#if !defined(__GXX_EXPERIMENTAL_CXX0X__) &&  __cplusplus < 201103L
+#if !GTEST_LANG_CXX11
```
KindDragon: We have warning here on VS2015

    ~CxxExceptionInDestructorTest() {
       GTEST_SUPPRESS_UNREACHABLE_CODE_WARNING_BELOW_(
           throw std::runtime_error("Standard C++ exception"));
    }

KindDragon:
@coreykosak I only want support old projects for old VS (2010 and less), because some users depend on them. Old VS projects cannot run tests. New users should use CMake generated projects.

But now we have VS2015 projects #773 :disappointed:

coreykosak:
I guess I don't understand the full story. CMake should be able to generate projects for any Visual Studio version, and it can support conditionals to enable or disable tests on some versions.

This is my assumption anyway. I haven't actually tried it.

KindDragon:
Yes you right, but some people prefer don't use CMake and they use old projects.
We can remove old projects at some point.

coreykosak:
I agree. Sorry, I didn't say my whole thought, which is.

Even if *we* publish the solution and project files for all VS versions (and I think we should), those files should be generated by CMake, not hand-edited. Put another way, our job is to get CMakeLists.txt right, run it for all the versions, and then publish the result.

KindDragon:
@BillyDonahue anything else is not clear in this PR?

KindDragon:
@BillyDonahue any news? I promise that when we will fix VS 2015 build I stop posting messages there :smile:

googlemock/include/gmock/gmock-matchers.h
```diff
@@ -515,7 +515,7 @@ template <typename T, typename M>
 class MatcherCastImpl {
  public:
   static Matcher<T> Cast(const M& polymorphic_matcher_or_value) {
-    // M can be a polymorhic matcher, in which case we want to use
+    // M can be a polymorphic matcher, in which case we want to use
```
BillyDonahue: This is not an MSVC build fix.
Please narrow your changes down for review.

If you want to fix spelling errors, do it separately.

KindDragon: ok :+1:

KindDragon: fixed

KindDragon:
Rebased to remove changes which were in PR #821

KindDragon:
@BillyDonahue Do I need remove anything else from PR?

googletest/include/gtest/internal/gtest-port.h
```diff
@@ -616,12 +618,21 @@ struct _RTL_CRITICAL_SECTION;
...
 # endif  // _MSC_VER
 #endif  // !defined(GTEST_HAS_HASH_MAP_)
 
+// Determines if unordered_map/unordered_set are available.
+// Only used for testing against those containers.
+#if !defined(GTEST_HAS_UNORDERED_MAP_)
+# if defined(_MSC_VER) && (_MSC_VER >= 1900)
```
eldruin: std::unordered_set and std::unordered_map tests should also be enabled on other platforms, I would say this check should be #if GTEST_LANG_CXX11

KindDragon:
@BillyDonahue PR now has only one commit. VS2010 solution changes moved to separate PR.

marco-m:
I agree with @coreykosak: cmake should be used to detect what the compiler supports, instead of adding conditional macro logic in gtest include files.
 
If somebody is using _old_version of VS and doesn't want to run cmake, why don't they simply use an old version of gtest? :-)

KindDragon:
I don't add it. I just update currently used `#ifdef``s

If you know how to that, patches welcome.

KindDragon:
> I agree with @coreykosak: cmake should be used to detect what the compiler supports, instead of adding conditional macro logic in gtest include files.

Why do we have to do it, if condional macro works pretty fine for all use cases?

googletest/test/gtest_catch_exceptions_test_.cc
paulsapps: So GTEST_SUPPRESS_UNREACHABLE_CODE_WARNING_BELOW_ also needs to disable "it throws even though it says it shouldn't throw" warning?

gennadiycivil:
This looks somewhat messy, a lot of conversation, merge conflicts now.
I am closing it and looking forward to clean PR that can be discussed anew and actually merged in.

Issue #775, opened 2016-05-16
Error C2440 'return': cannot convert from 'const testing::internal::EqMatcher<testing::gmock_matchers_test::ConvertibleFromAny>' to 'testing::Matcher<testing::gmock_matchers_test::ConvertibleFromAny>'	gmock_test


mazong1123:
I've got following error message when building gmock_test in visual studio 2015:

>D:\Workspace\github\googletest\googlemock\include\gmock/gmock-matchers.h(555): error C2440: 'return': cannot convert from 'const testing::internal::EqMatcher<testing::gmock_matchers_test::ConvertibleFromAny>' to 'testing::Matcher<testing::gmock_matchers_test::ConvertibleFromAny>'
2>  D:\Workspace\github\googletest\googlemock\include\gmock/gmock-matchers.h(555): note: No constructor could take the source type, or constructor overload resolution was ambiguous
2>  D:\Workspace\github\googletest\googlemock\include\gmock/gmock-matchers.h(546): note: while compiling class template member function 'testing::Matcher<testing::gmock_matchers_test::ConvertibleFromAny> testing::internal::MatcherCastImpl<T,M>::CastImpl(const M &,testing::internal::BooleanConstant<true>)'
2>          with
2>          [
2>              T=testing::gmock_matchers_test::ConvertibleFromAny,
2>              M=testing::internal::EqMatcher<testing::gmock_matchers_test::ConvertibleFromAny>
2>          ]
2>  D:\Workspace\github\googletest\googlemock\include\gmock/gmock-matchers.h(531): note: see reference to function template instantiation 'testing::Matcher<testing::gmock_matchers_test::ConvertibleFromAny> testing::internal::MatcherCastImpl<T,M>::CastImpl(const M &,testing::internal::BooleanConstant<true>)' being compiled
2>          with
2>          [
2>              T=testing::gmock_matchers_test::ConvertibleFromAny,
2>              M=testing::internal::EqMatcher<testing::gmock_matchers_test::ConvertibleFromAny>
2>          ]
2>  D:\Workspace\github\googletest\googlemock\include\gmock/gmock-matchers.h(611): note: see reference to class template instantiation 'testing::internal::MatcherCastImpl<T,M>' being compiled
2>          with
2>          [
2>              T=testing::gmock_matchers_test::ConvertibleFromAny,
2>              M=testing::internal::EqMatcher<testing::gmock_matchers_test::ConvertibleFromAny>
2>          ]
[...]

PR #820, opened 2016-07-14
Fixed issue #775. Mark the ConvertibleFromAny constructor as explicit, and fix operator overload issue for Unprintable

mazong1123:
I've seen the fix somewhere a couple of weeks ago, and I failed to find the source again. I'm actually not the author of this PR, someone else has already provided the exactly same solution somewhere.

Anyway, this PR fixes the issue #775 somehow.


mazong1123:
I accidentally use wrong account to commit the change. I'll close this PR and send another one with right account.

PR #821, opened 2016-07-14
Fixed issue #775. Mark the ConvertibleFromAny constructor as explicit, and fix operator overload issue for Unprintable

mazong1123:
I've seen the fix somewhere a couple of weeks ago, and I failed to find the source again. I'm actually not the author of this PR, someone else has already provided the exactly same solution somewhere.

Anyway, this PR fixes the issue #775 somehow.

### DIFF
--- a/googlemock/test/gmock-matchers_test.cc
+++ b/googlemock/test/gmock-matchers_test.cc
@@ -613,7 +613,7 @@ TEST(MatcherCastTest, FromSameType) {
 struct ConvertibleFromAny {
   ConvertibleFromAny(int a_value) : value(a_value) {}
   template <typename T>
-  ConvertibleFromAny(const T& /*a_value*/) : value(-1) {
+  explicit ConvertibleFromAny(const T& /*a_value*/) : value(-1) {
     ADD_FAILURE() << "Conversion constructor called";
   }
   int value;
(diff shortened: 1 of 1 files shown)


googlemock/test/gmock-matchers_test.cc
```diff
@@ -867,11 +867,12 @@ class Unprintable {
  public:
   Unprintable() : c_('a') {}
 
-  bool operator==(const Unprintable& /* rhs */) { return true; }
  private:
   char c_;
 };
 
+inline bool operator==(const Unprintable&, /* lhs */ const Unprintable& /* rhs */) { return true; }
```
BillyDonahue:
```
inline bool operator==(const Unprintable& /* lhs */,
                       const Unprintable& /* rhs */) {
  return true;
}
```

mazong1123: Should I reformat the code style as your note and submit a new PR again?

mazong1123:
@BillyDonahue just formatted the operator== method coding style.

PS: The appveryor build for MSVC 14 always failed. But that should be caused by another issue. Please check if this patch fix #775  :)

BillyDonahue:
Looks good, thanks!

PR #836, opened 2016-07-21
Add flag to disable installation of gtest.

bacek:
Whet gtest used as subproject via add_directory() it will add itself
into installation targets. Sometime it's not feasible. Adding new option
will allow parent project to manage gtest installation using

SET(gtest_install OFF CACHE BOOL "Don't install gtest" FORCE)

before invoking add_project.


gennadiycivil:
AppVeyor build failed
Cleaning up PRs that

Don't have all checks OK
Have been around for a while with no activity

PR #843, opened 2016-08-04
Add x64 configuration

fojogit:


KindDragon:
I have already this changes in PR #811

fojogit:
@KindDragon Ok great! I see the x64 configs added in the msvc\2010 folder, is there another change that applies those to the msvc\2015 files?

KindDragon:
@fojogit I think we should remove it because in general that project unsupported. Users should use CMake generated projects

gennadiycivil:
AppVeyor build failed
Cleaning up PRs that
1) Don't have all checks OK
2) Have been around for a while with no activity

PR #846, opened 2016-08-06
fix python tests when CMAKE_BUILD_TYPE is specified

jmacfarlane:
Currently, the Python tests are looking in the wrong place for binaries when CMAKE_BUILD_TYPE is specified.  This has not previously been caught because the Travis CI script does not currently set the build type.

This merge request updates the Travis CI script to catch this failure and fixes the failure itself.

There are still other Travis CI variables that don't seem to be used like SHARED_LIB, STATIC_LIB, and CMAKE_PKG.  Perhaps these should be reviewed as well.

### DIFF
--- a/googletest/cmake/internal_utils.cmake
+++ b/googletest/cmake/internal_utils.cmake
@@ -235,20 +235,29 @@ function(py_test name)
   # We are not supporting Python tests on Linux yet as they consider
   # all Linux environments to be google3 and try to use google3 features.
   if (PYTHONINTERP_FOUND)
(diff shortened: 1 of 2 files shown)


jmacfarlane:
fixed for multi-config generators

jmacfarlane:
Unfortunately CMAKE_CFG_INTDIR could not be used because of the way environment variable expansion works in `add_test`.  Using a conditional instead.

jmacfarlane:
There may be a better way to do this, but supports multi-config generators at least as well as before.

gennadiycivil:

PR #847, opened 2016-08-06
fix broken syntax for CMake 2.6.2

jmacfarlane:
Fixes some CMake syntax that breaks on CMake version 2.6.2.


gennadiycivil:
Please provide before/after test results

jmacfarlane:
It looks like the issue was fixed by commit c0059a79f82d0ecc3e9bfdc12146a5b84befc70a by bumping the minimum CMake version to 2.6.4.

gennadiycivil:
CMake v2.6.4 or newer is a requirement these days.

PR #856, opened 2016-08-24
Fix Mingw-w64 build

KindDragon:
@BillyDonahue PR #721 break build with Mingw-w64, this PR should fix this

PR #867, opened 2016-09-02
Clarifying documentation language

marzojr:
The old language might mislead someone into thinking that the access level on the base class itself was changed. This happened to a co-worker of mine, a (relative) novice in C++.


marzojr:
The CLA is signed.

marzojr:
AppVeyor build is likely due to the issue fixed by #856

PR #871, opened 2016-09-05
Update Primer.md

tommyleo2:
fix bug in string comparison assertions table where _str2_ is mistakenly written as _str_2


tommyleo2:
Signed it.

Issue #873, opened 2016-09-12
Google Mock README.md broken relative links


dandezille:
The README.md for googlemock has a number of relative links, for example
Google Mock for Dummies -> docs/ForDummies.md

These do not seem to be working, the example above points to:
https://github.com/google/googletest/tree/master/docs/ForDummies.md

but the actual document is found at:
https://github.com/google/googletest/blob/master/googlemock/docs/ForDummies.md

It appears that the links generated are to locations in the source tree, rather than the blob. I created a fork and attempted to fix, but I could not get anywhere. 

Happy to make the changes myself but would need some input from someone with more Github Markdown experience. An obvious solution would be going to a "repo/blob/master/..." format link without using the relative path, but not sure if this is acceptable?

BillyDonahue:
I don't understand the difference between the 'blob' and the 'source tree'.
But it doesn't work now and it does work with this patch, so good enough.

PR #874, opened 2016-09-12
Fixing broken relative links in Google Mock documentation

sejr:
This pull request offers a solution to issue #873, and ensures that every relative link in [googletest/googlemock/README.md](https://github.com/google/googletest/blob/master/googlemock/README.md) is working properly. It is likely that there are similar relative link issues in other parts of the documentation, which I will review.


sejr:
I signed it!

Issue #624, opened 2015-11-05
`ThreadIdToThreadLocals` objects leaked.


FrankStain:
Hi there.
https://github.com/google/googletest/blob/master/googletest/src/gtest-port.cc#L516
Here some object is created. But it's never deleted. Memory leaks detection pointing me on this allocation and on `void Mutex::ThreadSafeLazyInit()` as well.

Can you fix this leaks so leaks detection tool will stop pointing me on it?

Leaks of `ThreadIdToThreadLocals` can be fixed here:
https://github.com/google/googletest/blob/master/googletest/src/gtest-port.cc#L408
Currently i fix it via counting of total elements that remains inside of `thread_local_values` and deleting it when `thread_local_values` becomes empty. But i don't satisfied with this fix as it maybe break some feature and i actually don't clearly understand this code.

FrankStain:
UPD:

Hre is detailed leaks report.
```
d:\th\minkernel\crts\ucrt\src\appcrt\heap\debug_heap.cpp (450): App!heap_alloc_dbg + 0x18 bytes
d:\th\minkernel\crts\ucrt\src\appcrt\heap\debug_heap.cpp (492): App!_malloc_dbg
d:\th\minkernel\crts\ucrt\src\appcrt\heap\malloc.cpp (22): App!malloc
f:\dd\vctools\crt\vcstartup\src\heap\new_scalar.cpp (19): App!operator new + 0xA bytes

d:\dev\googletest\googletest\src\gtest-port.cc (516): App!testing::internal::ThreadLocalRegistryImpl::GetThreadLocalsMapLocked + 0x44 bytes
d:\dev\googletest\googletest\src\gtest-port.cc (416): App!testing::internal::ThreadLocalRegistryImpl::OnThreadLocalDestroyed + 0x5 bytes
d:\dev\googletest\googletest\src\gtest-port.cc (538): App!testing::internal::ThreadLocalRegistry::OnThreadLocalDestroyed
d:\dev\googletest\googletest\include\gtest\internal\gtest-port.h (1851): App!testing::internal::ThreadLocal<std::vector<testing::internal::TraceInfo,std::allocator<testing::internal::TraceInfo> > >::~ThreadLocal<std::vector<testing::internal::TraceInfo,std::allocator<testing::internal::TraceInfo> > > + 0x2B bytes
d:\dev\googletest\googletest\src\gtest.cc (4363): App!testing::internal::UnitTestImpl::~UnitTestImpl + 0x13 bytes
[...]

PR #892, opened 2016-10-05
Silence false positive memory leaks reported by MSVC debug CRT

scottslacksmith:
There are few Issues (e.g. #624 , #692 and #757) requesting a "memory leak" fix for MSVC debug builds. In fact these are not true memory leaks but false positives due to limitations of MSVC's lightweight memory leak detection mechanism (see below). But I think this is missing the real point that #624 , #692 and #757 are trying to solve. It's highly desirable that all false positive leaks be either **fixed** or **silenced** otherwise it becomes impractical to identify real leaks using MSVC's leak report. I normally endeavour to fix all false positive leaks (e.g. as suggested by #757) but in this instance silencing is the best solution as it's impossible to test that the destructors aren't called too early for all possible build configuration that may exist in the wild. 

As background, MSVC debug CRT can only detect if a memory allocation is missing a matching deallocation. Note MSVC debug CRT is always tracking memory allocations however the memory leak report is disabled by default. As you can’t avoid paying for its cost, you may as well use it. The memory leak report can be enabled by calling the following function
`_CrtSetDbgFlag(_CrtSetDbgFlag(_CRTDBG_REPORT_FLAG) | _CRTDBG_LEAK_CHECK_DF);`
anywhere before exiting main. 
For example, the following are the false positive leaks reported by gtest_prod_test.exe before this change;

```
Detected memory leaks!
Dumping objects ->
{750} normal block at 0x015DF938, 8 bytes long.
Data: <  ]     > 00 F9 5D 01 00 00 00 00
{749} normal block at 0x015DEE60, 32 bytes long.
Data: <` ] ` ] ` ]     > 60 EE 5D 01 60 EE 5D 01 60 EE 5D 01 01 01 CD CD
{748} normal block at 0x015DF900, 12 bytes long.
Data: <8 ] ` ]     > 38 F9 5D 01 60 EE 5D 01 00 00 00 00
{747} normal block at 0x015DA0F8, 24 bytes long.
Data: <                > FF FF FF FF FF FF FF FF 00 00 00 00 00 00 00 00
Object dump complete.

```

As you can see from above it’s not easy to identify which leaks are false
positives or not. Consequently, if false positive leaks are not fixed or
silenced, then it becomes impractical to identify real memory leaks.

### DIFF
--- a/googletest/src/gtest-port.cc
+++ b/googletest/src/gtest-port.cc
@@ -293,16 +293,16 @@ class MemoryIsNotDeallocated
 public:
   MemoryIsNotDeallocated() : old_crtdbg_flag_(0) {
 #ifdef _MSC_VER
-    old_crtdbg_flag_ = ::_CrtSetDbgFlag(_CRTDBG_REPORT_FLAG);
+    old_crtdbg_flag_ = _CrtSetDbgFlag(_CRTDBG_REPORT_FLAG);
     // Set heap allocation block type to _IGNORE_BLOCK
-    ::_CrtSetDbgFlag(old_crtdbg_flag_ & ~_CRTDBG_ALLOC_MEM_DF);
+    _CrtSetDbgFlag(old_crtdbg_flag_ & ~_CRTDBG_ALLOC_MEM_DF);
 #endif // _MSC_VER
   }
 
   ~MemoryIsNotDeallocated() {
 #ifdef _MSC_VER
     // Restore the original _CRTDBG_ALLOC_MEM_DF flag
-    ::_CrtSetDbgFlag(old_crtdbg_flag_);
+    _CrtSetDbgFlag(old_crtdbg_flag_);
 #endif // _MSC_VER
   }
 


scottslacksmith:
A possibly more elegant solution is to defined an immortal pointer. For example
```
template<typename T>
struct immortal_delete {
  void operator()(T*) const noexcept {}
};

// Is an unique_ptr where the held object can never be destroyed.
// The motivation is to self-document objects that are intentionally 
// never destroy/deallocated because they are reachable for the entire
// lifetime of an app.
template<typename T>
using immortal_unique_ptr = std::unique_ptr<T, immortal_delete<T>>;

template<class T, class... Args> 
immortal_unique_ptr<T> make_immortal_unique(Args&&... args)
{
  MemoryIsNotDeallocated memoryIsNotDeallocated;
  return (immortal_unique_ptr<T>(new T(std::forward<Args>(args)...)));
}

```
In which case
```
  // Returns map of thread local instances.
    mutex_.AssertHeld();
    static ThreadIdToThreadLocals* map = new ThreadIdToThreadLocals;
    return map;
  }

```
becomes
```
  // Returns map of thread local instances.
    mutex_.AssertHeld();
    static immortal_unique_ptr<ThreadIdToThreadLocals> map = immortal_unique_ptr<ThreadIdToThreadLocals>();
    return map.get();
  }

```
The advantage is that we clearly document which object constructions are intentionally never destroyed. For example, it could handy in a different context where an app is constantly loading and unloading  .so/.dll plugins then I believe a function static like 
`static ThreadIdToThreadLocals* map = new ThreadIdToThreadLocals;`
could be classified as a true memory leak.

gennadiycivil:
Apologies for not getting to this sooner. As a preparation for 1.8.x release I am cleaning up older PR that are in CI failed status and/or have merge conflicts.
Please submit a new PR if you think this is still important/relevant.

PR #895, opened 2016-10-06
[893] Added a minimum Windows2000 (0x0500) to enable OpenThread

AlexanderPaniutin:
To fix a compiling error "OpenThread is not defined".

By default MinGW does not define WINNT other than **0x0400**. But OpenThread is only after **0x0500** Win2000. See winbase.h

```
#if (_WIN32_WINNT >= 0x0500) || (_WIN32_WINDOWS >= 0x0490)
WINBASEAPI HANDLE WINAPI OpenThread(DWORD,BOOL,DWORD);
#endif
```

As per instructions from windef.h

```
#ifndef WINVER
#define WINVER 0x0400
/*
 * If you need Win32 API features newer the Win95 and WinNT then you must
 * define WINVER before including windows.h or any other method of including
 * the windef.h header.
 */
```


AlexanderPaniutin:
I signed it!

AlexanderPaniutin:
w32api.h not found on appveyor CI?
OK. I can remove it and deving 0x0500 instead of Windows2000.

AlexanderPaniutin:
I signed it!

googletest/src/gtest.cc
```diff
@@ -105,10 +105,9 @@
 
 // cpplint thinks that the header is already included, so we want to
 // silence it.
-// Add require of at least Win 2000 to enable OpenThread
-# include <w32api.h> // windows names: Windows2000
+// Add require of at least Win 2000 0x0500 to enable OpenThread
```
KindDragon: Maybe add a comment that is required only for MinGW?

AlexanderPaniutin: Do you think its for MinGW only?

KindDragon: Yes, it builds fine on Windows and with MINGW64

KindDragon: I think enough. Sorry, I don't have MinGW-w64 now

AlexanderPaniutin: updated

gennadiycivil:
No CLA

PR #896, opened 2016-10-06
PUMP line breaking fix

stryku:
Hello (:

I was playing with PUMP in my own project, and in one file I had line which has 80 spaces indentation and a '\' at the end. Apparently PUMP had problem with it because it was consuming whole RAM etc.
After quick research I found that the WrapCode method support maximum only <80 characters long line and it goes crazy in some cases, so I did two things:
-Introduced 'max_reasonable_indent' as 60 characters. If line has longer indentation it's already weird and we'll only help it, stripping that indentation away.

Was plying some more and found couple more problematic cases, so
-Introduced FindSplitPos function, which cover corner cases that I found.

If I just broke some 'line breaking rules' and this should be done in different way, please let me know.

### DIFF
--- a/googletest/scripts/pump.py
+++ b/googletest/scripts/pump.py
@@ -739,14 +739,14 @@ def WrapComment(line, output):
 
 
 def FindSplitPos(seg, max_len, splitter):
-  space_at = seg.find(splitter)
-  if space_at == -1:
+  splitter_at = seg.find(splitter)
+  if splitter_at == -1:
     return len(seg)
-  else:
-    if space_at > max_len:
-      return space_at
-    else:
-      return seg.rfind(splitter, 0, max_len)
+
+  if splitter_at > max_len:
+    return splitter_at
+
+  return seg.rfind(splitter, 0, max_len)
 
 
 def WrapCode(line, line_concat, output):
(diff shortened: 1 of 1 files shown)


stryku:
I signed it!

googletest/scripts/pump.py
```diff
@@ -751,7 +768,8 @@ def WrapCode(line, line_concat, output):
     # If the line is still too long, wrap at a space.
     while cur_line == '' and len(seg.strip()) > max_len:
       seg = seg.lstrip()
-      split_at = seg.rfind(' ', 0, max_len)
+      split_at = FindSplitPos(seg, max_len, ' ')
+
```
BillyDonahue: remove blank line

googletest/scripts/pump.py
```diff
@@ -738,10 +738,27 @@ def WrapComment(line, output):
...
+  max_line_length = 80
+  max_reasonable_indent = 60
+
   indent = len(line) - len(line.lstrip())
+  if indent >= max_reasonable_indent:
+    indent = 0
+
   prefix = indent*' '  # Prefix of the current line
-  max_len = 80 - indent - len(line_concat)  # Maximum length of the current line
+  max_len = max_line_length - indent - len(line_concat)  # Maximum length of the current line
```
BillyDonahue: this line of python is now too long

googletest/scripts/pump.py
```diff
@@ -738,10 +738,27 @@ def WrapComment(line, output):
     output.append(prefix + cur_line.strip())
 
 
+def FindSplitPos(seg, max_len, splitter):
+  space_at = seg.find(splitter)
+  if space_at == -1:
+    return len(seg)
+  else:
```
BillyDonahue: don't need else after a returning 'if'.

stryku:
Thanks, good points.
I have a question. Will googletest regenerate *.pump files in the repo, in pre build step, or I should do this manually, and push changes?

gennadiycivil:
This issue is fairly old and there hasn't been much activity on it. Closing, but please re-open if it still occurs.

PR #910, opened 2016-10-18
Disable VS2015 build before it is fixed

KindDragon:
So other PR's is not ended in failure due to VS2015


KindDragon:
@BillyDonahue please merge at least that PR  to quick-fix build for other PR's

PR #911, opened 2016-10-19
Make GTEST_TEST_NO_THROW_ output the exception message

BrukerJWD:
When I am using *_NO_THROW statements and they fail, I want to know what was the cause for the exception. The

```
Actual: it throws.
```

message doesn't help me a lot, so I extended the macro to save the messages from std::exceptions into a string variable, which is then printed as well. Throwing reasonable exceptions I thereby can fix most of the problems without using breakpoints.

### DIFF
--- a/googletest/include/gtest/internal/gtest-internal.h
+++ b/googletest/include/gtest/internal/gtest-internal.h
@@ -1148,18 +1148,24 @@ class NativeArray {
       fail(gtest_msg.value)
 
 #define GTEST_TEST_NO_THROW_(statement, fail) \
+  std::string GTEST_CONCAT_TOKEN_(gtest_exmsg_testnothrow_,__LINE__); \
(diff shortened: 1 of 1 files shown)


BrukerJWD:
I'm still waiting for the CLA.

As the integration tests show: This PR removes the possibility to write code like
`if (someting) EXPECT_NO_THROW(start());`
Instead using brackets are necessary:
`if (someting) { EXPECT_NO_THROW(start()); }`

BrukerJWD:
@travnick: This looks like a very good solution!
Unfortunately my company fails to allow me to sign a CLA so perhaps you could create a PR on your own?

gennadiycivil:
No CLA

travnick:
`Google Individual CLA | Mikołaj Milej | May 15, 2018 03:12 PDT`? It's not enough?

PR #919, opened 2016-11-02
Fix the link to the float comparison article

delaitre:
As explained on the previous link, the article has been rewritten and moved.


delaitre:
I signed it!

PR #924, opened 2016-11-05
Fix #923 - support CMAKE_CROSSCOMPILING_EMULATOR for tests

### LINKED ISSUE #923: Testcases fail to run in cross compilation, due to legacy syntax of cmake add_test()
I'm porting googletest to VxWorks using cmake cross-compilation. Since cmake-3.3 , the CMAKE_CROSSCOMPILING_EMULATOR feature allows even running testcases on a remote system -- it automatically logs in and runs the test. This is an extremely powerful and helpful feature!

But this fails with googletest, since in googletest/cmake/internal_utils.cmake line 219, legacy syntax is used for the add_test() comand: This command can only find executables on the local host, and doesn't consider the CMAKE_CROSSCOMPILING_EMULATOR.

To resolve this problem, please consider this simple change in internal_utils.cmake line 219 -- simply replace this legacy syntax:
    add_test(${name} ${name})
with this newer syntax:
    add_test(NAME ${name} COMMAND ${name})

moberhuber:
Replaced legacy syntax of cmake add_test() with more modern syntax.
This allows running gtests's own tests on remote (cross) systems
using CMAKE_CROSSCOMPILING_EMULATOR with cmake-3.3 or newer.

### DIFF
--- a/googletest/cmake/internal_utils.cmake
+++ b/googletest/cmake/internal_utils.cmake
@@ -216,7 +216,7 @@ find_package(PythonInterp)
 # from the given source files with the given compiler flags.
 function(cxx_test_with_flags name cxx_flags libs)
   cxx_executable_with_flags(${name} "${cxx_flags}" "${libs}" ${ARGN})
-  add_test(${name} ${name})
+  add_test(NAME ${name} COMMAND ${name})
 endfunction()
 
 # cxx_test(name libs srcs...)


moberhuber:
I just signed the CLA as an individual contributor.

moberhuber:
Ping, could this pull request be considered ?

The change is trivial and significantly improves support for cross-compiling / cross-testing with cmake.

gennadiycivil:
@moberhuber Sorry for the long delay on this. Could you please explain how this change relates to CMAKE_CROSSCOMPILING_EMULATOR usage?

moberhuber:
@gennadiycivil IIRC cmake just doesn't work with the old syntax when cross-compiling. See also https://stackoverflow.com/a/34203427 which also suggests you just have to use the "NAME ... COMMAND" syntax.

PR #925, opened 2016-11-05
update README.md with tiny-dnn

edgarriba:
tiny-dnn in its recent version updated the testing framework to GTest
https://github.com/tiny-dnn/tiny-dnn/pull/361

We would like to be mentioned as consumers of the project.


edgarriba:
I signed it!

PR #955, opened 2016-12-07
Fixing compilation on SunOS 5.10

dsalmin:
Fixing compilation on SunOS 5.10 (Generic_150400-11 Sparc) SunPro compiler 5.12.0 (SolarisStudio 12.3). Added string initialization cuz of SunOS cannot cast char to std string without calling constructor. Added template function to gtest printers because of instatniating from non-template code error. Also changed function for converting char array to string vector because of unsupporting by old cruel SunOS


dsalmin:
I signed it!

dsalmin:
Hi. will you commit it in future or not? and when if yes? because i did not found any errors as result of my code. only C2338 is presented in compile output. correct me if wrong

gennadiycivil:
I see broken AppVeyor builds potentially indicating windows compilation problem. https://ci.appveyor.com/project/BillyDonahue/googletest/build/130

It may or may not be related to the change. Please check

dsalmin:
Not related. Deprecated function warning treated as error.

2017-08-03 5:11 GMT+03:00 Gennadiy Civil <notifications@github.com>:

audiofanatic:
Anything stopping this fix from being merged now? It seems to have passed all required tests and contribution requirements from the discussion trail above.

gennadiycivil:
Aplogies for the long wait and thank you for your contribution. This makes general changes not guarded by if SunOS. Is there any reason we are not guarding the changes?

PR #957, opened 2016-12-08
Fix a few documentation nits in the mock dummies guide

sglass68:
Add a hyphen to 'ad hoc' and change 'distance' to plural.


sglass68:
I am a googler (sjg@) so should not need to sign the agreement.

PR #962, opened 2016-12-13
Fix a typo

klimkin:
Help says "FILE_PATH defaults to test_details.xml", but the real path is test_detail.xml.


klimkin:
I signed it!

-Лёша

PR #965, opened 2016-12-22
Annotate ColoredPrintf with the format attribute and fix bugs.

davidben:
I'm not sure if the `__MINGW_PRINTF_FORMAT` mess is necessary. We have it in BoringSSL's version of this macro to appease some Android build warning. I've included it here since I see some evidence GTest supports MinGW, but it may be more general than is necessary.

googletest doesn't currently build with clang's very aggressive `-Wformat-nonliteral` warning. It requires that all non-literal format strings come from the argument of a function annotated with a compatible format attribute.

Fixing that reports that `ColoredPrintf`'s callers weren't passing the normal `-Wformat` warning. Some messages were passed directly into the format string rather than via `"%s"`.

### DIFF
--- a/googletest/include/gtest/internal/gtest-port.h
+++ b/googletest/include/gtest/internal/gtest-port.h
@@ -874,6 +874,23 @@ using ::std::tuple_size;
 # define GTEST_ATTRIBUTE_UNUSED_
 #endif
 
+// Use this annotation before a function that takes a printf format string.
+#if defined(__GNUC__) && !defined(COMPILER_ICC)
+# if defined(__MINGW_PRINTF_FORMAT)
+// MinGW has two different printf implementations. Ensure the format macro
+// matches the selected implementation. See
+// https://sourceforge.net/p/mingw-w64/wiki2/gnu%20printf/.
+#  define GTEST_ATTRIBUTE_PRINTF_(string_index, first_to_check) \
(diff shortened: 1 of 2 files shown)


googletest/include/gtest/internal/gtest-port.h
```diff
@@ -874,6 +874,23 @@ using ::std::tuple_size;
 # define GTEST_ATTRIBUTE_UNUSED_
 #endif
 
+// Use this annotation before a function that takes a printf format string.
```
BillyDonahue: I'd rather not introduce this complex definition, and we will just eyeball the printf formats, which are all internal.

davidben: The trouble is clang's `-Wformat-nonliteral` basically requires you do it. One could argue whether this warning is useful, but considering it managed to find a bug in `ShouldShard`, this seems valuable to me.

BoringSSL, being security-sensitive code, errs on the side of enabling warnings. For googletest to be useful for us, we would need it too to pass those warnings.

davidben: I can probably cut down the MinGW stuff if you don't care about that. I erred on the side of generality since it looked like you all supported a lot of things? MinGW has some weirdness due to it having two different `printf`s which depend on whether `__USE_MINGW_ANSI_STDIO` is set. It's a mess.

PR #966, opened 2016-12-22
Honour symbol visibility for all target types.

reuk:
This is mostly useful for hiding symbols when the libraries are built
statically.
Solves warnings about 'direct access to global weak symbol' when linking against
other libraries built with hidden symbols.


reuk:
I signed the CLA.

PR #971, opened 2016-12-26
Make VerifyAndClear work with Nice, Naggy and Strict mock wrappers.

### LINKED ISSUE #969: Calling Mock::VerifyAndClearExpectations on StrictMock doesn't seem to clear expectations.
I expect all three tests to fail. But `Test.TestPasses` does not.
```cpp
class C {
 public:
  MOCK_METHOD0(f, void());
};

TEST(Test, TestPasses) {
  testing::StrictMock<C> callback;
  EXPECT_CALL(callback, f());
  // Should fail here because nobody called f().
  testing::Mock::VerifyAndClearExpectations(&callback);
  // Should also fail here, because:
  // * expectations are cleared;
  // * it's an unexpected call;
  // * it's a StrictMock.
  callback.f();
}

TEST(Test, TestFails1) {
  C callback;
  EXPECT_CALL(callback, f());
  // Correctly fails here.
  testing::Mock::VerifyAndClearExpectations(&callback);
  // Correctly warns of unexpected call.
  callback.f();
}

TEST(Test, TestFails2) {
  testing::StrictMock<C> callback;
  EXPECT_CALL(callback, f());
  testing::Mock::VerifyAndClearExpectations(&callback);
  EXPECT_CALL(callback, Call()).Times(0);
  // It fails here, because f() called more times than expected.
  callback.f();
  // And it fails here because nobody called f().
}
```

Is this a bug or am I missing something?

projedi:
This fixes #969.

Adds the casts from ctors and dtors of `NiceMock`, `NaggyMock`, `StrictMock` to `VerifyAndClearExpectations` and `VerifyAndClear`.

I've only added a test reproducing #969. Can add more if you agree with the change.

### DIFF
--- a/googlemock/test/gmock-nice-strict_test.cc
+++ b/googlemock/test/gmock-nice-strict_test.cc
@@ -38,6 +38,9 @@
 
 // This must not be defined inside the ::testing namespace, or it will
 // clash with ::testing::Mock.
+// Some tests below expect this not having virtual functions table so
+// that address of StrictMock<Mock> and address of its underlying Mock are
+// different.
 class Mock {
  public:
   Mock() {}
@@ -505,6 +508,16 @@ TEST(StrictMockTest, AcceptsClassNamedMock) {
   strict.DoThis();
 }
 #endif  // !GTEST_OS_SYMBIAN && !GTEST_OS_WINDOWS_MOBILE
+
+// Tests that Mock::VerifyAndClearExpectations called on StrictMock correctly
+// verifies and clears set expectations.
+TEST(StrictMockTest, VerifyAndClearExpectations) {
+  StrictMock< ::Mock> strict;
+  EXPECT_CALL(strict, DoThis());
+  EXPECT_NONFATAL_FAILURE(Mock::VerifyAndClearExpectations(&strict),
+      "never called");
+  EXPECT_NONFATAL_FAILURE(strict.DoThis(), "Uninteresting mock function call");
+}
 
 }  // namespace gmock_nice_strict_test
 }  // namespace testing


alex-ac:
@BillyDonahue, Take a look, please.

googlemock/include/gmock/gmock-spec-builders.h
```diff
@@ -367,6 +367,15 @@ enum CallReaction {
 
 }  // namespace internal
 
+template <typename M>
```
BillyDonahue: Can we use a #include instead of these forward decls?

googlemock/include/gmock/gmock-spec-builders.h
```diff
@@ -380,14 +389,56 @@ class GTEST_API_ Mock {
...
+    return VerifyAndClearExpectationsImpl(internal::ImplicitCast_<M*>(mock_obj));
+  }
 
   // Verifies all expectations on the given mock object and clears its
   // default actions and expectations.  Returns true iff the
   // verification was successful.
-  static bool VerifyAndClear(void* mock_obj)
-      GTEST_LOCK_EXCLUDED_(internal::g_gmock_mutex);
+  template <class M>
+  static bool VerifyAndClear(M* mock_obj)
```
BillyDonahue: Just constrain VerifyAndClear to only accept a Matcher*.
I'd try an `enable_if` for that.

googlemock/include/gmock/gmock-spec-builders.h
projedi: Circular dependency: these are defined in `gmock-generated-generated-nice-strict.h` which itself #includes `gmock-spec-builders.h`.

googlemock/include/gmock/gmock-spec-builders.h
projedi: Sorry, didn't get that. `VerifyAndClear` wants a mock, not a matcher. Am I missing something?

BillyDonahue: Ok. Sorry. Not a Matcher. We could use a trait to detect Mock wrappers then, and avoid the circular dependency.

googlemock/include/gmock/gmock-spec-builders.h
BillyDonahue: Then IMO that's pointing to a structural flaw, and forward decls aren't the fix.
Let's use a trait and have the mock wrappers specialize the trait.

projedi:
Added `MockTraits` with `AddressOf` method. Is this what you had in mind?

googlemock/include/gmock/gmock-generated-nice-strict.h.pump
```diff
@@ -94,38 +94,49 @@ class $clazz : public MockClass {
...
 
  private:
   GTEST_DISALLOW_COPY_AND_ASSIGN_($clazz);
 };
 
+namespace internal {
+
+template <typename MockClass>
+struct MockTraits<$clazz<MockClass> > {
+  static void* AddressOf($clazz<MockClass>* mock_obj) {
```
BillyDonahue: We shouldn't need `void*`.
I was thinking of something less intrusive.

    template <typename M>
    struct UnwrappedMockType { using type = M; };

    // Each of these specializations would follow their class template definitions.
    template <typename M>
    struct UnwrappedMockType<StrictMock<M>> { using type = M; };
    template <typename M>
    struct UnwrappedMockType<NiceMock<M>> { using type = M; };
    template <typename M>
    struct UnwrappedMockType<NaggyMock<M>> { using type = M; };

You'd use an ImpliciitCast to `UnwrappedMockType<M>::type*` where needed.

projedi: Ok, done. Externally used `Mock` methods (except `Register` and `RegisterUseByOnCallOrExpectCall`) now take `M*` and cast it to `UnwrappedMockType<M>::type*`.

projedi:
@BillyDonahue could you, please, take another look?

projedi:
@BillyDonahue please, take a look.

projedi:
@BillyDonahue could you, please, take another look?

projedi:
@BillyDonahue can you please take another look? All the tests seem to pass now.

projedi:
I've also looked in chromium for relevant problems and found https://cs.chromium.org/chromium/src/components/subresource_filter/content/browser/content_ruleset_service_unittest.cc?l=167 it uses a combination of `StrictMock<MockFunction<void()>>` and `Mock::VerifyAndClearExpectations` which triggers the issue.

projedi:
@gennadiycivil could you, please, take a look?

gennadiycivil:
Aplogies for not getting to this sooner and thank you for your contribution. As part of release 1.8.x preparation I am closing all older PRs that have conflicts and/or failed CI. Please resolve conflicts.

gennadiycivil:
Also no CLA

projedi:
Apparently, the bug got fixed already. I've left a test reproducing the original issue, and it passes locally (on macOS with clang). Let's see if the entire CI matrix passes.

projedi:
The test has passed everywhere. I think the test alone is still useful (to help avoid regressions), but I'm fine with just declining the PR. #969 is fixed as far as I'm concerned.

gennadiycivil:
@projedi I think it is better not to merge it as the original purpose of it is no longer relevant.
Thank you again , please consider contributing again

PR #975, opened 2016-12-30
CMake: use threads if allowed and found, not just if found.

LebedevRI:
If the user's cmakelists.txt first look for threads using
find_package(Threads), then set(gtest_disable_pthreads ON),
and then include googletest. GoogleTest will not look for
threads. But since they have already been found before in
user's cmakelists, it will use them regardless.

This helped me fix build issue in https://github.com/darktable-org/rawspeed/commit/e079559f9e05597183f681cd94cff42b9c51a9d7
on windows/MSYS2, even though there are threads, and they
are usable, googletest build was failing with issues
about AutoHandle. I was first looking for threads, and only
then including googletest, so no matter the value of
gtest_disable_pthreads, it failed.

The other obvious solution is for user to first include
googletest, and only then look for threads by himself.

-----

Thoughts? Changes?
Needs a reproducer?
Should `NOT gtest_disable_pthreads AND NOT MINGW AND CMAKE_USE_PTHREADS_INIT` be stored as one variable (`HAS_THREADS`?) and that variable be used? May prevent breakage in the future.

### DIFF
--- a/googletest/cmake/internal_utils.cmake
+++ b/googletest/cmake/internal_utils.cmake
@@ -48,10 +48,14 @@ endmacro()
 macro(config_compiler_and_linker)
   # Note: pthreads on MinGW is not supported, even if available
   # instead, we use windows threading primitives
+  unset(GTEST_HAS_PTHREAD)
   if (NOT gtest_disable_pthreads AND NOT MINGW)
     # Defines CMAKE_USE_PTHREADS_INIT and CMAKE_THREAD_LIBS_INIT.
     set(THREADS_PREFER_PTHREAD_FLAG ON)
     find_package(Threads)
+    if (CMAKE_USE_PTHREADS_INIT)
+      set(GTEST_HAS_PTHREAD ON)
+    endif()
   endif()
 
   fix_default_compiler_settings_()
@@ -126,7 +130,8 @@ macro(config_compiler_and_linker)
     set(cxx_no_rtti_flags "")
   endif()
 
-  if (CMAKE_USE_PTHREADS_INIT)  # The pthreads library is available and allowed.
+  # The pthreads library is available and allowed?
+  if (DEFINED GTEST_HAS_PTHREAD)
     set(GTEST_HAS_PTHREAD_MACRO "-DGTEST_HAS_PTHREAD=1")
   else()
     set(GTEST_HAS_PTHREAD_MACRO "-DGTEST_HAS_PTHREAD=0")
@@ -159,7 +164,7 @@ function(cxx_library_with_type name type cxx_flags)
       PROPERTIES
       COMPILE_DEFINITIONS "GTEST_CREATE_SHARED_LIBRARY=1")
   endif()
-  if (CMAKE_USE_PTHREADS_INIT)
+  if (DEFINED GTEST_HAS_PTHREAD)
     target_link_libraries(${name} ${CMAKE_THREAD_LIBS_INIT})
   endif()
 endfunction()
--- a/googlemock/CMakeLists.txt
+++ b/googlemock/CMakeLists.txt
@@ -164,7 +164,7 @@ if (gmock_build_tests)
   cxx_test(gmock_link_test gmock_main test/gmock_link2_test.cc)
   cxx_test(gmock_test gmock_main)
 
-  if (CMAKE_USE_PTHREADS_INIT)
+  if (DEFINED GTEST_HAS_PTHREAD)
     cxx_test(gmock_stress_test gmock)
   endif()
 


LebedevRI:
Is the approach wrong? Should i change something?

LebedevRI:
Dear developers.
This PR fixes a bug. The bug is small, but it is real. The fix is small.
Is there anything wrong with the fix? Could this be merged, please?
Roman.

LebedevRI:
Still, ping.

gennadiycivil:
Hi Roman thank you for your contribution and sorry for the delay.
Could you please include set of test outputs before/after that shows the original problem and how the problem has been fixed.

LebedevRI:
Hi. Thank you for looking into this!

Hmm, not sure this will suffice, but this is what i have right now:

* My cmake build system was searching for threads before `add_subdirectory(googletest)`
* The output before:
https://ci.appveyor.com/project/LebedevRI/rawspeed/build/150/job/fjlp28h6c73opalq#L460
* The [workaround](https://github.com/LebedevRI/rawspeed/commit/845991cbcf3fc7e5782a66d3ed84afd5ec41122b) was applied to my code
* The output with workaround:
https://ci.appveyor.com/project/LebedevRI/rawspeed/build/151/job/0iowaxts2ttrklhn

As you can see, if i don't search for threads before `add_subdirectory(googletest)`, the googletest builds fine. Thus, if googletest is to always respect it's own cmake option `gtest_disable_pthreads`, the result should be identical.

Is that sufficient information?

LebedevRI:
@gennadiycivil 

Slightly different explanation:
1. you look for threads only if some conditions are not met
https://github.com/google/googletest/blob/8604c4adac40573f806cfadae44e22f8dfaf212a/googletest/cmake/internal_utils.cmake#L51-L55
2. But then everywhere else you link to threads simply if threads are marked as found:
https://github.com/google/googletest/blob/8604c4adac40573f806cfadae44e22f8dfaf212a/googletest/cmake/internal_utils.cmake#L129-L133
https://github.com/google/googletest/blob/8604c4adac40573f806cfadae44e22f8dfaf212a/googletest/cmake/internal_utils.cmake#L162-L164
https://github.com/google/googletest/blob/8604c4adac40573f806cfadae44e22f8dfaf212a/googlemock/CMakeLists.txt#L167-L169

Now, it would be logical that if you did not look for threads, then `CMAKE_USE_PTHREADS_INIT` should not be defined. But unfortunately, that is not a safe assumption to make, since [README.md#incorporating-into-an-existing-cmake-project](https://github.com/google/googletest/blob/master/googletest/README.md#incorporating-into-an-existing-cmake-project) recommends for googletest to be added directly into parent cmake build. And that parent build may have already looked for (and found) the threadsm and that `CMAKE_USE_PTHREADS_INIT` will be there regardless of whether https://github.com/google/googletest/blob/8604c4adac40573f806cfadae44e22f8dfaf212a/googletest/cmake/internal_utils.cmake#L51-L55 did search for it or not.

I think the problem is painfully obvious...

Do you need more info?

LebedevRI:
Hmm. I'm honestly not sure how to test for such a problem.
Are there any tests for such cmake issues already?

gennadiycivil:
Looks to me that this is somewhat related to #1239 by @hethi ... could you please collaborate and see if unified solution can be found? I also think that we will need a test after such solution presents itself

LebedevRI:
@gennadiycivil 
1. that pr would not help/work here, since that define would still be erroneously defined:
https://github.com/google/googletest/blob/8604c4adac40573f806cfadae44e22f8dfaf212a/googletest/cmake/internal_utils.cmake#L129-L133
2. yes, i guess #1239 also makes sense, but they are unrelated.

This pr still needs to land. As for tests, what do you want that test to do? Maybe that will give me an idea...

hethi:
I agree to @LebedevRI - this PR is unrelated to #1239.
@gennadiycivil - as this is a configuration issue, it's not possible to add a test IMHO.

googletest/cmake/internal_utils.cmake
```diff
@@ -159,7 +160,7 @@ function(cxx_library_with_type name type cxx_flags)
       PROPERTIES
       COMPILE_DEFINITIONS "GTEST_CREATE_SHARED_LIBRARY=1")
   endif()
-  if (CMAKE_USE_PTHREADS_INIT)
+  if (NOT gtest_disable_pthreads AND NOT MINGW AND CMAKE_USE_PTHREADS_INIT)
```
hethi: As this term is repeated identically to line 167 in `CMakeLists.txt`, I personally would prefer to set one CMake variable with this term, so future changes are in one place only. As you suggested already in the issue's discussion.

hethi: But I'm not a maintainer here :-)

LebedevRI: Sure, that i can do easily :)

gennadiycivil: +1 on @hethi 's suggestion

LebedevRI: Already done.

LebedevRI:
(So, anything left for me to do?)

LebedevRI:
Thank you very much :)

PR #976, opened 2016-12-30
Fix #819: Using FormatFileLocation to insert a TestPartResult

### LINKED ISSUE #819: Test Event: OnTestPartResult() gives C++ exception upon not finding file_name
```
void ConfigurableEventListener::OnTestPartResult(const TestPartResult& result)
{
    LOG_DEBUG("START");
    LOG_DEBUG("result.file_name: %p", result.file_name());
    if(result.file_name())
        LOG_DEBUG("result.file_name: %s", result.file_name());
    eventListener->OnTestPartResult(result);
    LOG_DEBUG("END");
}
```

I have derived Event Listener for my GTEST App with name of ConfigureEventListener.
When tests get exited with SUCCESS()/ADD_FAILURE() an API OnTestPartResult is called which internally calls default eventListener's OnTestPartResult. This gives an C++ Exception -
unknown file: Failure
C++ exception with description "basic_string::_S_construct null not valid" thrown in the test fixture's constructor.

While investigating, I found that because result.file_name() returned NULL, this exception occurred. 

Gtest is not able to find the File name of the Source from which the event has occurred. 

Need to confirm the implementations of TestPartResults class in Gtest.

```
void OnTestPartResult(const TestPartResult& test_part_result) {
    const char* file_name = test_part_result.file_name();
    if (file_name == NULL)
      file_name = "";
    SendLn("event=TestPartResult&file=" + UrlEncode(file_name) +
           "&line=" + StreamableToString(test_part_result.line_number()) +
           "&message=" + UrlEncode(test_part_result.message()));
  }
```

thejcannon:
This will fix issue #819, since the result.file_name() may return NULL/the result.line_number() may be <0.
The current code attempts to insert NULL into the ostream causing an exception on some platforms.

### DIFF
--- a/googletest/src/gtest-test-part.cc
+++ b/googletest/src/gtest-test-part.cc
@@ -32,6 +32,7 @@
 // The Google C++ Testing Framework (Google Test)
 
 #include "gtest/gtest-test-part.h"
+#include "gtest/internal/gtest-port.h"
 
 // Indicates that this translation unit is part of Google Test's
 // implementation.  It must come before gtest-internal-inl.h is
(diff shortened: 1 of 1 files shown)


MrElusive:
@gennadiycivil Are there any concerns about this change?

gennadiycivil:
Aplogies for not getting to this sooner and thank you for your contribution. As part of release 1.8.x preparation I am closing all older PRs that have conflicts and/or failed CI. Please resolve conflicts.

PR #977, opened 2017-01-03
fix building on VS 2015

DoDoENT:
VS 2015 fully supports C++14 (and some C++17 features), but still defines __cplusplus as 199711L. However, VS 2015 does not support std::tr1::tuple anymore and this causes failure to compile GTest.

This change adds msvc flag which turns on all latest c++ features and also in `gtest-port.h` performs compiler version check while determining if C++11 is supported.


DoDoENT:
I signed it!

KindDragon:
Hi. We have already PR #811 about that

KindDragon:
Are you sure that the new commits should be in the PR?

PR #991, opened 2017-01-20
Pass MSVC's C4826 warning.

davidben:
MSVC has an optional warning which flags when 32-bit pointers get cast into a 64-bit value. This is a little overaggressive I think, but to ease compiling in projects with aggressive warnings, fix this by defining `internal::UIntPtr` (analog of `uintptr_t`) and use that instead.


googletest/include/gtest/internal/gtest-port.h
```diff
@@ -2505,6 +2506,7 @@ typedef TypeWithSize<4>::Int Int32;
 typedef TypeWithSize<4>::UInt UInt32;
 typedef TypeWithSize<8>::Int Int64;
 typedef TypeWithSize<8>::UInt UInt64;
+typedef TypeWithSize<sizeof(void*)>::UInt UIntPtr;
```
BillyDonahue: Why not uintptr_t ?

davidben: Are you all okay using `uintptr_t`? If so, that works for me. My assumption was that, given the rest of these` typedef`s existed, you wanted to support platforms which were missing some of these.

BillyDonahue: You're right. I should look more closely before commenting.
`uintptr_t` is an "optional" definition in <cstdint>.
I guess we should stick to what you've defined it as.
Should you also define a signed `IntPtr`?

We really need to start killing off obsolete compiler support!

davidben: `IntPtr` wouldn't be used yet, but it'd be easy enough to add it for completeness. Do you prefer to include it?

googletest/include/gtest/gtest-printers.h
```diff
@@ -423,9 +423,9 @@ void DefaultPrintTo(IsNotContainer /* dummy */,
       // void*.  However, we cannot cast it to const void* directly,
       // even using reinterpret_cast, as earlier versions of gcc
       // (e.g. 3.4.5) cannot compile the cast when p is a function
```
BillyDonahue: Alternative idea to fix your warning.
Let's just get rid of the gcc 3.4.5 support here, and use `*os << reinterpret_cast<const void*>(p);`.

davidben: No tears shed on loss of GCC 3.4.5 here!

Let me check real quick if MSVC complains about that. C is kinda unsure about the whole von Neumann architecture thing and doesn't promise data and function pointers are the same size. My experience is GCC and Clang don't care but MSVC, of all compilers, sometimes gets grumpy about this. :-(

davidben: Looks like it's just fine in all our compilers! I'll update the PR.

davidben: Done.

gennadiycivil:
Looks outdated now , could you please see if you can resolve merge conflicts

davidben:
@gennadiycivil To be fair, it probably wasn't outdated when I uploaded it way back. ;-) Rebased.

Was this the wrong place to send patches? Should I be pinging somewhere internally? I have two others that also have stalled for quite a while.

googletest/include/gtest/gtest-printers.h
```diff
@@ -426,13 +426,8 @@ void DefaultPrintTo(WrapPrinterType<kPrintFunctionPointer> /* dummy */,
...
     // T is a function type, so '*os << p' doesn't do what we want
-    // (it just prints p as bool).  We want to print p as a const
-    // void*.  However, we cannot cast it to const void* directly,
-    // even using reinterpret_cast, as earlier versions of gcc
-    // (e.g. 3.4.5) cannot compile the cast when p is a function
-    // pointer.  Casting to UInt64 first solves the problem.
-    *os << reinterpret_cast<const void*>(
-        reinterpret_cast<internal::UInt64>(p));
+    // (it just prints p as bool).  Cast p to const void* to print it.
+    *os << reinterpret_cast<const void*>(p);
```
gennadiycivil: Looking at the comment it seems to say that "no you can not cast directly to const void* as this would break older versions of gcc.  
However this seems to be exactly what this PR suggests. Could you please comment?

davidben: See [this thread](https://github.com/google/googletest/pull/991#discussion_r115274698) with @BillyDonahue earlier on this PR and the commit message. I originally added a `internal::UIntPtr` (mimicking the real `uintptr_t`).

gennadiycivil:
Closing for now, will wait for new internal updates as discussed

davidben:
Please reopen this. This change is still a requirement.

PR #1006, opened 2017-02-10
Fix links to the google test docs

sdutoit:
Previously all of these links were providing 404s.


gennadiycivil:
Looks out of date, redundant now.

PR #1007, opened 2017-02-11
Pass the -Wmissing-declarations warning.

davidben:
This makes it easier to use GTest in projects that build with the
-Wmissing-declarations warning.

### DIFF
--- a/googlemock/include/gmock/gmock-matchers.h
+++ b/googlemock/include/gmock/gmock-matchers.h
@@ -3615,6 +3615,10 @@ BoundSecondMatcher<Tuple2Matcher, Second> MatcherBindSecond(
   return BoundSecondMatcher<Tuple2Matcher, Second>(tm, second);
 }
 
+// Joins a vector of strings as if they are fields of a tuple; returns
+// the joined string.  This function is exported for testing.
+GTEST_API_ string JoinAsTuple(const Strings& fields);
+
 // Returns the description for a matcher defined using the MATCHER*()
 // macro where the user-supplied description string is "", if
 // 'negation' is false; otherwise returns the description of the
--- a/googletest/include/gtest/gtest-param-test.h
+++ b/googletest/include/gtest/gtest-param-test.h
@@ -1407,21 +1407,21 @@ internal::CartesianProductHolder10<Generator1, Generator2, Generator3,
 // type testing::TestParamInfo<class ParamType>, and return std::string.
 //
 // testing::PrintToStringParamName is a builtin test suffix generator that
-// returns the value of testing::PrintToString(GetParam()). It does not work
-// for std::string or C strings.
+// returns the value of testing::PrintToString(GetParam()).
 //
 // Note: test names must be non-empty, unique, and may only contain ASCII
-// alphanumeric characters or underscore.
+// alphanumeric characters or underscore. Because PrintToString adds quotes
+// to std::string and C strings, it won't work for these types.
 
 # define INSTANTIATE_TEST_CASE_P(prefix, test_case_name, generator, ...) \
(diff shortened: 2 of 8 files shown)


BillyDonahue:
I don't know that it's useful to make this flag happy.

BillyDonahue:
Sorry didn't mean to close the PR. But this flag is awful picky about definitions in what's essentially a main()-style test file.

davidben:
We need GTest---at minimum the headers, but it's easier if the whole thing passes---to pass these warnings for it to be useful to BoringSSL. We get decent mileage out of this warning:
* If we forget to mark a `static` function as such, it catches us.
* If we forget to `#include` the header for a function, it catches us. This is important because, if you don't do that, you lose all type-checking between the function declaration and the definition.

(Plus, if I recall, one of our consumers set this warning. For the same reason it's valuable for GTest to pass a strict set of warnings to appease your downstreams, it's valuable for BoringSSL to pass and test against a strict set of warnings to appease ours.)

gennadiycivil:
This branch has conflicts that must be resolved
Use the web editor or the command line to resolve conflicts.

googletest/include/gtest/gtest-param-test.h
```diff
@@ -1419,14 +1419,14 @@ internal::CartesianProductHolder10<Generator1, Generator2, Generator3,
 // alphanumeric characters or underscore.
 
 # define INSTANTIATE_TEST_CASE_P(prefix, test_case_name, generator, ...) \
-  ::testing::internal::ParamGenerator<test_case_name::ParamType> \
+  static ::testing::internal::ParamGenerator<test_case_name::ParamType> \
```
gennadiycivil: @davidben please dont edit this file firectly. Rather, change the corresponding pump file and run the pump script.

googlemock/include/gmock/gmock-matchers.h
```diff
@@ -3614,6 +3614,10 @@ BoundSecondMatcher<Tuple2Matcher, Second> MatcherBindSecond(
   return BoundSecondMatcher<Tuple2Matcher, Second>(tm, second);
 }
 
+// Joins a vector of strings as if they are fields of a tuple; returns
+// the joined string.
+GTEST_API_ string JoinAsTuple(const Strings& fields);
```
gennadiycivil: @davidben I am missing where this is used?

googletest/include/gtest/gtest-param-test.h
davidben: Done. Interestingly, it doesn't look like anyone has run that script for a while. Some comments have been out of sync since 794ef030ebad671b699abdfd8e0474f93045be91. :-)

googlemock/include/gmock/gmock-matchers.h
davidben: It's used in gmock-matchers_test.cc by replicating the declaration. My bad, I forgot to remove the duplicate one.

davidben:
Rebased and fixed up a few more that have since snuck in.

googlemock/include/gmock/gmock-matchers.h
gennadiycivil: @davidben sorry, I still see it here, I thought we were removing it?

gennadiycivil:
I think we still need to remove [duplicate declaration](https://github.com/google/googletest/pull/1007/files#diff-59ca716c9f8866771fae4c61dd13962cR3620)

davidben:
Hrm? It's declared in gmock-matchers.h, defined in gmock-matchers.cc, and used in gmock-matchers.cc and gmock-matchers_test.cc, both of which include gmock-matchers.h. Am I missing something?

(There was a duplicate in gmock-matchers_test.cc, but I've removed that now.)

PR #1016, opened 2017-02-14
Fix temp dir on Android.

yabinc:
/sdcard is no longer available for processes running in app
context starting from Android O. Instead, first try /data/local/tmp,
which is usually used as the temporary directory, then try the
current directory, which is usually accessible in app context.


yabinc:
The build failures seem all unrelated.

gennadiycivil:
This is, sadly out of date and has conflicting files. Could you please re-visit and re-submit if still relevant...
Thank you

yabinc:
Yes, it has conflicts now. I created a new pull request at https://github.com/google/googletest/pull/1187.

PR #1023, opened 2017-02-22
Fix broken links to docs sub-pages

klimkin:
The links were using blob/tree/master, which should be blob/master to work properly.


gennadiycivil:
Looks like https://github.com/google/googletest/blob/master/googletest/docs/Primer.md is fine now.?

gennadiycivil:
This is no longer relevant, looks like it was also done here , https://github.com/google/googletest/commit/ba638689e40fc60c2e5194fc777bf30541deeff8#diff-a8bd9c0a0a6f1596f1c9cdcc5a7fd294, closing

Issue #1028, opened 2017-02-25
Links on googletest/googlemock/ReadMe.md are dead


pike4:
"basics", "Google mock for dummies", "Cheat Sheet", and "Cookbook" links under the "Getting Started" section in the GoogleMock ReadMe are dead. 

They point to: googletest/blob/tree/master/googlemock/*
They should point to: googletest/blob/master/googlemock/*

BillyDonahue:
Thanks for reporting.

PR #1029, opened 2017-02-26
Remove /tree/ from Readme.md links.

BillyDonahue:
Address dead links reported in Issue #1028


hethi:
@BillyDonahue - Do you still need branch BillyDonahue-patch-3? When forking the project, this branch will also be created in the fork, which is confusing a little bit; and it seems to me, the branch is not used anymore, as it was merged in this pull request?

PR #1034, opened 2017-03-02
Minimal changes to fix build failures on Microsoft Visual Studio 2015

dankegel:
Should fix #1033 

Fixes the following errors:

```
hash_map(17): error C2338: <hash_map> is deprecated and will be REMOVED. Please use <unordered_map>. You can define _SILENCE_STDEXT_HASH_DEPRECATION_WARNINGS to acknowledge that you have received this warning. [googlemock\gtest\gtest-printers_test.vcxproj]
hash_set(17): error C2338: <hash_set> is deprecated and will be REMOVED. Please use <unordered_set>. You can define _SILENCE_STDEXT_HASH_DEPRECATION_WARNINGS to acknowledge that you have received this warning. [googlemock\gtest\gtest-printers_test.vcxproj]
googletest\test\gtest_catch_exceptions_test_.cc(152): error C2220: warning treated as error - no 'object' file generated
googletest\test\gtest_catch_exceptions_test_.cc(152): warning C4297: 'CxxExceptionInDestructorTest::~CxxExceptionInDestructorTest': function assumed not to throw an exception but does
  googletest\test\gtest_catch_exceptions_test_.cc(152): note: destructor or deallocator has a (possibly implicit) non-throwing exception specification
```

There were already some checks for _MSC_VER in the code, so this commit
continues in that vein.

PR #1046, opened 2017-03-17
MSVC project work

tplunket:
This PR started just as an addition of VS2015 project files for gtest. The old project files were for VS2003 so they were moved to a subdirectory and the new project files were added to another subdirectory, mirroring what had been done in gmock. 64 bit targets were also added. This required output directories to be modified a little bit; x86/Win32 and x64 directories get all build outputs for their respective targets. I also tweaked the way that debug outputs get the 'd' suffixed on their name to allow more property sharing between projects and to not stomp things like the debug database when building multiple targets at once.

A bunch of common properties were pulled out of both projects' project files, into their respective projects' props files. Finally, scripts were added to automate building all targets and running all test executables.


tplunket:
I signed it!

tplunket:
The appveyor failed tests appear to be addressed by PR #1034.

KindDragon:
PR #811 should fix this too, but no one merges them for more than 8 months 😞

gennadiycivil:
Not sure why AppVeyor build failed, please take a look, and resubmit with working build for this to be considered
https://ci.appveyor.com/project/BillyDonahue/googletest/build/209

PR #1058, opened 2017-03-31
Fix out-of-source build in example

martijnotto:
The example for incorporating googletest into an existing cmake project was missing a CMAKE_BINARY_DIR which breaks out-of-source builds


martijnotto:
Signed the CLA.

martijnotto:
I don't use my google account with github (I only have it to be able to use Android) nor do I wish to link it. If that means you cannot merge this pull request, that's your business. I have agreed to the CLA.

martijnotto:
This isn't going anywhere so I'm closing the pull request.

Issue #930, opened 2016-11-08
Double free or corruption (fasttop) with google mock


agirault:
I get the following problem when using google mock for a CMake project on Linux.

`gmock` and `gtest` are built with CMake. Even without using any google mock features, the error below will occur as long as google mock is linked against my test executable. No problem if linked against google test only.

This occurs after the tests completion. My tests all run good but this segfault occurs right after:

```
*** Error in `/home/virtualfls/Projects/IMSTK/debug/bin/imstk_geometry_test_driver': double free or corruption (fasttop): 0x00000000007086c0 ***
======= Backtrace: =========
/lib64/libc.so.6(+0x7277f)[0x7fffe9eca77f]
/lib64/libc.so.6(+0x78026)[0x7fffe9ed0026]
/lib64/libc.so.6(+0x78d53)[0x7fffe9ed0d53]
/usr/lib64/libstdc++.so.6(_ZNSsD2Ev+0x40)[0x7fffea7ebcb0]
/home/virtualfls/Projects/IMSTK/debug/lib/libgtest.so(_ZN7testing8internal20StringStreamToStringEPSt18basic_stringstreamIcSt11char_traitsIcESaIcEE+0xe0)[0x7ffff7b898a6]
/home/virtualfls/Projects/IMSTK/debug/lib/libgtest.so(_ZNK7testing7Message9GetStringEv+0x2e)[0x7ffff7b85ed2]
/home/virtualfls/Projects/IMSTK/debug/lib/libgtest.so(_ZN7testing8internal18StreamableToStringIiEESsRKT_+0x42)[0x7ffff7ba238d]
/home/virtualfls/Projects/IMSTK/debug/lib/libgtest.so(_ZN7testing8internal18FormatFileLocationEPKci+0x8c)[0x7ffff7b9a0df]
/home/virtualfls/Projects/IMSTK/debug/lib/libgtest.so(_ZN7testing8internal8GTestLogC2ENS0_16GTestLogSeverityEPKci+0x6e)[0x7ffff7b9a378]
/home/virtualfls/Projects/IMSTK/debug/lib/libgmock.so(_ZN7testing8internal11ThreadLocalIPNS_8SequenceEED2Ev+0x54)[0x7ffff76f0b1e]
/lib64/libc.so.6(__cxa_finalize+0x8f)[0x7fffe9e8febf]
/home/virtualfls/Projects/IMSTK/debug/lib/libgmock_main.so(+0x84653)[0x7ffff73b4653]
```
or
[...]

PR #1059, opened 2017-03-31
Fix ODR violation of gtest global variables in gmock

igorsugak:
When both gtest and gmock are DSOs configured by cmake, and are dynamically
linked into a program, the program crashes with a double free error upon exit,
if the test program is invoked with at least one gtest flag specified,
or if there is a death test in the program. This is due to ODR violation
of global `std::string` variables (holding command line arguments) defined in
gtest but duplicated in gmock, because gtest is linked into gmock.
This updates cmake configuration for gmock to remove gtest from gmock
libraries. Notice that the gmock configure script also doesn't link gtest into gmock.
This also removes gmock from gmock_main to be consistent with gtest and
gtest_main.

Tested by running googletest unit tests as well by running tests in a
larger codebase that uses googletest (both DSO and .a) as a testing framework 
on a linux system with gcc-4.9 and gcc-5.4.

Potentially also fixes some bugs mentioned in https://github.com/google/googletest/issues/930

### DIFF
--- a/googlemock/CMakeLists.txt
+++ b/googlemock/CMakeLists.txt
@@ -81,16 +81,11 @@ endif()
 # Google Mock libraries.  We build them using more strict warnings than what
 # are used for other targets, to ensure that Google Mock can be compiled by
 # a user aggressive about warnings.
-cxx_library(gmock
-            "${cxx_strict}"
-            "${gtest_dir}/src/gtest-all.cc"
-            src/gmock-all.cc)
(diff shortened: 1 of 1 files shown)


igorsugak:
Looks like this broke googletest on windows and build on linux when cmake configures from googlemock (not root) directory. I'll update my patch to handle this.

KiNgMaR:
I have run into this error on Linux and can confirm that the patch fixes it.

gennadiycivil:
Both Travis and AppVeyor failed. Closing

PR #1077, opened 2017-05-02
Changes to make TempDir() public

### LINKED ISSUE #1076: Make TempDir() available publicly
The TempDir() functionality in googletest needs to be available publicly. Currently it only exists in internal:: , need to make it available in testing::

gennadiycivil:
Fixes #1076.

### DIFF
--- a/googletest/src/gtest.cc
+++ b/googletest/src/gtest.cc
@@ -5385,4 +5385,23 @@ void InitGoogleTest(int* argc, wchar_t** argv) {
 #endif  // defined(GTEST_CUSTOM_INIT_GOOGLE_TEST_FUNCTION_)
 }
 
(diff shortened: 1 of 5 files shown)


googletest/src/gtest.cc
```diff
@@ -5385,4 +5385,23 @@ void InitGoogleTest(int* argc, wchar_t** argv) {
 #endif  // defined(GTEST_CUSTOM_INIT_GOOGLE_TEST_FUNCTION_)
 }
 
+std::string TempDir() {
```
BillyDonahue: The implementation of TempDir is very related to portability so I think it should remain there as-is.
This change might just expose a forwarding function called testing::TempDir() that calls it?

PR #1078, opened 2017-05-04
Pick up GTEST_API_ definition in gtest/internal/custom/gtest-port.h.

pwnall:
This makes it possible for a port to define the specifier used for exported symbols without having to change Google Test.

Background: I'm trying to roll Google Test from 1.7.0 to 1.8.0 (or above) into Chromium and I'm running into a crasher in the ThreadLocal destructor for a static variable defined in Google Mock with the GTEST_API_ specifier. Some details are at https://bugs.chromium.org/p/chromium/issues/detail?id=630705#c28 and I can supply more, if it helps.

I realize this PR is not addressing the root cause of the crashes, but being able to replace the GTEST_API_ definition lets me make forward progress, and I can imagine that being able to define GTEST_API_ would come in handy for other projects as well. Will you please consider merging this?

### DIFF
--- a/googletest/include/gtest/internal/gtest-port.h
+++ b/googletest/include/gtest/internal/gtest-port.h
@@ -930,6 +930,11 @@ using ::std::tuple_size;
 
 #endif  // GTEST_HAS_SEH
 
+// GTEST_API_ qualifies all symbols that must be exported. The definitions below
+// are guarded by #ifndef to give embedders a chance to define GTEST_API_ in
+// gtest/internal/custom/gtest-port.h
+#ifndef GTEST_API_
+
 #ifdef _MSC_VER
 # if GTEST_LINKED_AS_SHARED_LIBRARY
 #  define GTEST_API_ __declspec(dllimport)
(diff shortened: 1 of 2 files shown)


nico:
It sounds like #451 was for chromium, but it's actively harmful for chromium -- so maybe we can just revert #605 instead?

googletest/include/gtest/internal/gtest-port.h
```diff
@@ -938,11 +940,11 @@ using ::std::tuple_size;
 # endif
 #elif __GNUC__ >= 4 || defined(__clang__)
 # define GTEST_API_ __attribute__((visibility ("default")))
```
nico: Or only do this if GTEST_LINKED_AS_SHARED_LIBRARY is defined, that's probably the best fix.

pwnall:
I think that reverting #605 is an orthogonal concern. GTEST_API_ really seems like an embedder concern to me, and I think that trying to get all the cases in right in Google Test is a tough battle. I claim that having an escape hatch there is desirable, no matter what the built-in definition ends up being.

pwnall:
Thank you very much!

PR #1082, opened 2017-05-06
Add missing initializers to enable -Weff-c++ warning

pavelkryukov:
Hi

I use -Weff-c++ warning in my project but the rules cannot be applied to Googletest
Actually, there are only few changes required to enable this warning in Googletest builds.

Thanks,
Pavel


pavelkryukov:
I signed it!

PR #1083, opened 2017-05-06
Add missing initializers as -Weff-c++ option advises

pavelkryukov:
Hi

I use -Weff-c++ warning in my project but I found that the rules cannot be applied to Googletest.
Actually, there are only few changes required to enable this option in Googletest builds.

There are some tricky cases still left:
 * `class MutexBase` has no default initialization for all fields. If explicit constructor is defined, build for C++03 will be broken.
 * -Weff-c++ requires `return *this` for every `operator=`. Nevertheless, `void AssertHelper::operator=(const Message& message) const` cannot have it.

Thanks,
Pavel


googletest/include/gtest/internal/gtest-internal.h
```diff
@@ -307,7 +307,7 @@ class FloatingPoint {
   // around may change its bits, although the new value is guaranteed
   // to be also a NAN.  Therefore, don't expect this constructor to
   // preserve the bits in x when x is a NAN.
-  explicit FloatingPoint(const RawType& x) { u_.value_ = x; }
+  explicit FloatingPoint(const RawType& x) : u_() { u_.value_ = x; }
```
BillyDonahue: What does it mean to default-construct a union of RawType and Bits?
It's going to create a default-valued RawType, then overwrite it with 'x'.
I am not sure of the usefulness, portability, or legality of 'u_()'.
The eff-c++ flag might be too picky here? What does it say?

googletest/include/gtest/internal/gtest-linked_ptr.h
```diff
@@ -149,12 +149,12 @@ class linked_ptr {
...
   // Take over ownership of a raw pointer.  This should happen as soon as
   // possible after the object is created.
-  explicit linked_ptr(T* ptr = NULL) { capture(ptr); }
+  explicit linked_ptr(T* ptr = NULL) : value_(), link_() { capture(ptr); }
   ~linked_ptr() { depart(); }
 
   // Copy an existing linked_ptr<>, adding ourselves to the list of references.
-  template <typename U> linked_ptr(linked_ptr<U> const& ptr) { copy(&ptr); }
-  linked_ptr(linked_ptr const& ptr) {  // NOLINT
+  template <typename U> linked_ptr(linked_ptr<U> const& ptr) : value_(), link_() { copy(&ptr); }
```
BillyDonahue: copy(&ptr) is about to unconditionally overwrite these data members. There's no reason to initialize them before that.

pavelkryukov: I see. The target is to prevent us from having unitialized values from `copy(linked_ptr const&)` behavior changes.

googletest/include/gtest/internal/gtest-internal.h
pavelkryukov: > What does it mean to default-construct a union of RawType and Bits?
> I am not sure of the .. portability, or legality of 'u_()'.

There is a good explanation: http://stackoverflow.com/questions/321351/initializing-a-union-with-a-non-trivial-constructor

From C++03, 12.1 Constructors, pg 190

The implicitly-defined default constructor performs the set of initializations of the class that would be performed by a user-written default constructor for that class with an empty mem-initializer-list (12.6.2) and an empty function body.

From C++03, 8.5 Initializers, pg 145

To default-initialize an object of type T means:

if T is a non-POD class type (clause 9), the default constructor for T is called (and the initialization is ill-formed if T has no accessible default constructor);
if T is an array type, each element is default-initialized;
otherwise, the object is zero-initialized.
To zero-initialize an object of type T means:

if T is a scalar type (3.9), the object is set to the value of 0 (zero) converted to T;
if T is a non-union class type, each non static data member and each base-class subobject is zero-initialized;
**if T is a union type, the object’s first named data member is zero-initialized;**
if T is an array type, each element is zero-initialized;
if T is a reference type, no initialization is performed.

> I am not sure of the usefulness .. of 'u_()'.

OK, I may add `explicit FloatingPointUnion(const RawType& v)` constructor to keep only one assignment

googletest/include/gtest/internal/gtest-linked_ptr.h
BillyDonahue: We should cover that with unit tests, though. They don't have a runtime cost! :)

BillyDonahue:
This is a big change and it is in many cases causing data members to be initialized twice. Once with a default value and then immediately overwritten. This doesn't feel right to me.

pavelkryukov:
> Once with a default value and then immediately overwritten.

It is equal to the original behavior, as object members were implicitly initialized by default constructors and then get overwritten. Now the initialization by default constructor becomes explicit, that is the point.

BillyDonahue:
The linked_ptr fields are POD so they weren't implicitly initialized.
The explicitness of initializing a vector `: v()` instead of just omitting it feels like a downgrade to me. I don't agree about the benefits of writing it out.

PR #1084, opened 2017-05-08
rename "test case" to "test suite"

hethi:
Unfortunately, googletest is using the term "test case" for a collection of
several tests, which is usually named "test suite" - while the usual use of
"test case" is to describe the single test.

So when googletest says "Running x tests from y test cases", several other
test environments would say "Running x test cases from y test suites".

To avoid confusion, I just changed "test case" to "test suite" in the startup
message.


hethi:
I signed it!

hethi:
I signed it!

hethi:
I will propose a more complete change related to #1085 at a later time, closing this pull request in between.

PR #1091, opened 2017-05-15
Fix -Wmicrosoft-cast warnings when using gtest with clang on Windows.

nico:
This upstreams a Google-internal change.

Original CL description:
The C++ standard says that function pointers are not implicitly
convertible to object pointers. Visual Studio disregards that and allows
implicit conversion between function pointers and object points, and
enough code relies on this that clang follows suit in
Microsoft-compatibility mode.

However, clang emits a -Wmicrosoft-cast warning when such a conversion
is done:

E:\b\c\b\win_clang\src\sandbox\win\src\sync_dispatcher.cc(42,7):
    warning: implicit conversion between pointer-to-function and
             pointer-to-object is a Microsoft extension [-Wmicrosoft-cast]

This change fixes this warning in gtest, while hopefully not changing
any behavior. The change does two things:

1. It replaces the if in DefaultPrintTo with SFINAE
2. In C++11 mode, it uses enable_if<is_function<>> instead of
   ImplicitlyConvertible<T*, const void*> to check if the
   explicit cast is needed.

With this change, functions will use the branch with the reintpret_casts
with Visual Studio and clang/win, and clang no longer needs to warn
that it implicitly converts a function pointer to a void pointer.

### DIFF
--- a/googletest/include/gtest/gtest-printers.h
+++ b/googletest/include/gtest/gtest-printers.h
@@ -364,11 +364,18 @@ class UniversalPrinter;
 template <typename T>
 void UniversalPrint(const T& value, ::std::ostream* os);
 
+enum DefaultPrinterType {
+  kPrintContainer,
+  kPrintPointer,
+  kPrintFunctionPointer,
+  kPrintOther,
+};
+template <DefaultPrinterType type> struct WrapPrinterType {};
+
 // Used to print an STL-style container when the user doesn't define
 // a PrintTo() for it.
 template <typename C>
-void DefaultPrintTo(IsContainer /* dummy */,
-                    false_type /* is not a pointer */,
+void DefaultPrintTo(WrapPrinterType<kPrintContainer> /* dummy */,
                     const C& container, ::std::ostream* os) {
   const size_t kMaxCount = 32;  // The maximum number of elements to print.
   *os << '{';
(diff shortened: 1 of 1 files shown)


googletest/include/gtest/gtest-printers.h
```diff
@@ -364,11 +364,18 @@ class UniversalPrinter;
 template <typename T>
 void UniversalPrint(const T& value, ::std::ostream* os);
 
+enum DefaultPrinterType {
```
BillyDonahue: This technique is fine, but I'd prefer a few tag types here instead of an enum and WrapPrinterType.
```
struct ContainerPrintTag{};
struct PointerPrintTag{};
struct FunctionPointerPrintTag{};
struct OtherPrintTag{};
```

googletest/include/gtest/gtest-printers.h
```diff
@@ -468,13 +470,24 @@ void PrintTo(const T& value, ::std::ostream* os) {
...
-  //   PrintTo(T* x, ...);
-  DefaultPrintTo(IsContainerTest<T>(0), is_pointer<T>(), value, os);
+  // Note that MSVC and clang-cl do allow an implicit conversion from
+  // pointer-to-function to pointer-to-object, but clang-cl warns on it.
+  // So don't use ImplicitlyConvertible if it can be helped since it will
+  // cause this warning, and use a separate overload of DefaultPrintTo for
+  // function pointers so that the `*os << p` in the object pointer overload
+  // doesn't cause that warning either.
+  DefaultPrintTo(
+      WrapPrinterType<sizeof(IsContainerTest<T>(0)) == sizeof(IsContainer)
```
BillyDonahue: Let's factor this meta-logic out of the mainline of the code with a trait.
```
DefaultPrintTo(typename PrintToTag<T>::type(), value, os);
```
And then have the traits of T analyzed off to the side.

nico:
This merges a google-interal change already reviewed by a googltest owner. The motivation is to reduce the diff between google-internal googletest and open-source googletest. I'd prefer to land this without changes.

Issue #1097, opened 2017-05-18
Abort on nested exception from dtor of Mock object


rgolovanov:
There is a problem I met using GMock with 3rdParty UT-engine based on exceptions handling.

If I enables throwing exceptions from GMock: `::testing::GTEST_FLAG(throw_on_failure) = true`
Then in test case if I created a **local** Mock object with invalid expectation call then I will get exception in exception:
1. Because of invalid arguments passed to the method.
2. During stack unwinding we will destroy Mock object and will call `VerifyAndClearExpectationsLocked();` from`~FunctionMockerBase()`

These behavior obviously leads to process termination.

The suggestion is to check whether we already in unhandled exception or not before Verifing expectations in destructor:
```
    if (!std::uncaught_exception())
    {
       VerifyAndClearExpectationsLocked();
    }
```

gennadiycivil:
I have been cleaning up older and inactive GitHub googletest issues. You may see an issue "closed" if it appears to be inactive/abandoned 
Thank you

PR #1098, opened 2017-05-18
Don't throw nested exception from dtor of Mock object

rgolovanov:
fix of #1097 
There is a problem happened when GMock is used with 3rdParty UT-engine based on exceptions handling.

If I enables throwing exceptions from `GMock: ::testing::GTEST_FLAG(throw_on_failure) = true`
and if I created a local Mock object with invalid expectation call then I will get exception in exception:

1. Because of invalid arguments passed to the method.
2. During stack unwinding we will destroy Mock object and will call `VerifyAndClearExpectationsLocked(); `from `~FunctionMockerBase()`
These behavior obviously leads to process termination.

The suggestion is to check whether we already in unhandled exception or not before Verifing expectations in destructor.


rgolovanov:
@BillyDonahue could you please review these few lines.

lor-anonymous-spirit:
@rgolovanov, he merges google's internal changes only, don't panic

googlemock/include/gmock/gmock-spec-builders.h
```diff
@@ -1471,7 +1471,8 @@ class FunctionMockerBase : public UntypedFunctionMockerBase {
   virtual ~FunctionMockerBase()
         GTEST_LOCK_EXCLUDED_(g_gmock_mutex) {
     MutexLock l(&g_gmock_mutex);
-    VerifyAndClearExpectationsLocked();
+    if(!std::uncaught_exception())
```
BillyDonahue: This is deprecated in C++17. I wonder if that will affect us.

rgolovanov: Sorry, it is not truth. In C++17 the return type is changed only:
[cppreference.com](http://en.cppreference.com/w/cpp/error/uncaught_exception)

```
bool uncaught_exception();
(1)	(deprecated in C++17)
int uncaught_exceptions();
(2)	(since C++17)
```
(2) Detects how many exceptions in the current thread have been thrown or rethrown and not yet entered their matching catch clauses.

So the proposed change is correct.

lor-anonymous-spirit: name changed: uncaught_exception**s**

rgolovanov: oh, yep. I missed it first time. Should I modify this fix using ```GTEST_GCC_VER_``` to support both versions of C++?

lor-anonymous-spirit: imho it's better to use __cplusplus macro(std, §16.8)

rgolovanov: done, see update

rgolovanov:
@BillyDonahue, @lor-anonymous-spirit  could you please advice somebody who can review and accept this pull request.

thejcannon:
As a general rule of thumb, you shouldn't rely on std::uncaught_exceptions().
See [http://www.gotw.ca/gotw/047.htm](http://www.gotw.ca/gotw/047.htm) for an explanation of why it can lead to buggy code.

googlemock/include/gmock/gmock-spec-builders.h
```diff
@@ -1471,7 +1471,14 @@ class FunctionMockerBase : public UntypedFunctionMockerBase {
...
         GTEST_LOCK_EXCLUDED_(g_gmock_mutex) {
     MutexLock l(&g_gmock_mutex);
-    VerifyAndClearExpectationsLocked();
+   
+#if __cplusplus < 201703L
+    const bool inside_uncaught_exception = std::uncaught_exception();
+#else
+    const bool inside_uncaught_exception = std::uncaught_exceptions() > 0;
+#endif
+    if(!inside_uncaught_exception)
```
BillyDonahue: space after 'if'

googlemock/include/gmock/gmock-spec-builders.h
```diff
@@ -1471,7 +1471,14 @@ class FunctionMockerBase : public UntypedFunctionMockerBase {
   virtual ~FunctionMockerBase()
         GTEST_LOCK_EXCLUDED_(g_gmock_mutex) {
     MutexLock l(&g_gmock_mutex);
-    VerifyAndClearExpectationsLocked();
+   
+#if __cplusplus < 201703L
```
BillyDonahue: where is this magic number from? I can't find a settled numeric code for C++17.

googlemock/include/gmock/gmock-spec-builders.h
```diff
@@ -1471,7 +1471,14 @@ class FunctionMockerBase : public UntypedFunctionMockerBase {
...
+    const bool inside_uncaught_exception = std::uncaught_exception();
+#else
+    const bool inside_uncaught_exception = std::uncaught_exceptions() > 0;
```
BillyDonahue: would be more elegant as a function than as a variable.

googlemock/include/gmock/gmock-spec-builders.h
```diff
@@ -1471,7 +1471,14 @@ class FunctionMockerBase : public UntypedFunctionMockerBase {
...
+#endif
+    if(!inside_uncaught_exception)
+        VerifyAndClearExpectationsLocked();
```
BillyDonahue: indent by 2, not 4.

googlemock/include/gmock/gmock-spec-builders.h
BillyDonahue: Can you explain why it's okay to not `VerifyAndClearExpectationsLocked` if we're inside an unhandled exception? I'm concerned that this conditional evaluation will make an invariant more complex.

googlemock/include/gmock/gmock-spec-builders.h
thejcannon: This isn't really the intended use of std::uncaught_exceptions(). To use std::uncaught_exceptions() appropriately, you should capture the number of uncaught exceptions on object construction and check again at destruction. If the number at destruction is different (it should always differ by 0 or 1) then you know you can't throw.
See [the proposal](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2014/n4152.pdf).

Also worth mentioning, in the proposal is implementations for std::uncaught_exceptions() for older versions of MSVC and gcc/clang. Perhaps googletest should attempt to use those instead of defaulting to std::uncaught_exception.
See [this github repo](https://github.com/evgeny-panasyuk/stack_unwinding)

gennadiycivil:
@rgolovanov  Sincere apologies for the long wait. I see unfinished discussion.  If this i still relevant could you please rebase and we can try to move forward from here/

PR #1101, opened 2017-05-23
Make the failure messages from EXPECT_EQ and friends actually symmetric,

DarthGandalf:
instead of reading more like reversing the former "expected" and "actual"
roles of the LHS and RHS arguments.


gennadiycivil:

PR #1104, opened 2017-05-26
Add a table of content to GoogleMock CookBook.md

wizmer:


wizmer:
Google CLA signed

gennadiycivil:
Thank you for this work. It is useful however I can not accept just the TOC without a way to maintain its freshness, the maintenance is a problem. 
Looking for suggestions on how to maintain a TOC with a script or similar semi-automated method to ensure that its up to date.

PR #1109, opened 2017-05-31
Avoid warning C4619 in MSVC 2017.

davidben:
C4800 has since been removed in MSVC 2017, so trying to silence it throws warning C4619 when enabled.

I could also wrap the whole thing in an `_MSC_VER < 1900` ifdef or so. Whichever you prefer.

### DIFF
--- a/googletest/include/gtest/gtest.h
+++ b/googletest/include/gtest/gtest.h
@@ -259,7 +259,9 @@ class GTEST_API_ AssertionResult {
   // Used in EXPECT_TRUE/FALSE(assertion_result).
   AssertionResult(const AssertionResult& other);
 
+#if defined(_MSC_VER) && _MSC_VER < 1910
   GTEST_DISABLE_MSC_WARNINGS_PUSH_(4800 /* forcing value to bool */)
+#endif
 
   // Used in the EXPECT_TRUE/FALSE(bool_expression).
   //
@@ -276,7 +278,9 @@ class GTEST_API_ AssertionResult {
           /*enabler*/ = NULL)
       : success_(success) {}
 
+#if defined(_MSC_VER) && _MSC_VER < 1910
   GTEST_DISABLE_MSC_WARNINGS_POP_()
+#endif
 
   // Assignment operator.
   AssertionResult& operator=(AssertionResult other) {


gennadiycivil:
Please show a demonstrable before/after test output that tests the problem this PR is fixing.

davidben:
Here's what I get in BoringSSL without this patch:

```
C:\Users\davidben\Code\boringssl>ninja -C build gtest
ninja: Entering directory `build'
[0/1] Re-running CMake...
-- Configuring done
-- Generating done
-- Build files have been written to: C:/Users/davidben/Code/boringssl/build
[1/2] Building CXX object CMakeFiles\gtest.dir\third_party\googletest\src\gtest-all.cc.obj
FAILED: CMakeFiles/gtest.dir/third_party/googletest/src/gtest-all.cc.obj
C:\PROGRA~2\MICROS~2\2017\COMMUN~1\VC\Tools\MSVC\1410~1.250\bin\HostX64\x64\cl.exe  /nologo /TP -DBORINGSSL_IMPLEMENTATION -DNOMINMAX -DWIN32_LEAN_AND_MEAN -D_CRT_SECURE_NO_WARNINGS -D_HAS_EXCEPTIONS=0 -D_STL_EXTRA_DISABLED_WARNINGS="4774 4987" -I..\third_party\googletest -I..\third_party\googletest\include -Wall -WX  -wd4061 -wd4100 -wd4127 -wd4200 -wd4204 -wd4221 -wd4242 -wd4244 -wd4245 -wd4267 -wd4371 -wd4388 -wd4296 -wd4350 -wd4365 -wd4389 -wd4464 -wd4510 -wd4512 -wd4514 -wd4548 -wd4610 -wd4623 -wd4625 -wd4626 -wd4668 -wd4706 -wd4710 -wd4711 -wd4800 -wd4820 -wd5026 -wd5027  -w44265 /MDd /Zi /Ob0 /Od /RTC1 /showIncludes /FoCMakeFiles\gtest.dir\third_party\googletest\src\gtest-all.cc.obj /FdCMakeFiles\gtest.dir\gtest.pdb /FS -c ..\third_party\googletest\src\gtest-all.cc
C:\Users\davidben\Code\boringssl\third_party\googletest\include\gtest/gtest.h(262): error C2220: warning treated as error - no 'object' file generated
C:\Users\davidben\Code\boringssl\third_party\googletest\include\gtest/gtest.h(262): warning C4619: #pragma warning: there is no warning number '4800'
ninja: build stopped: subcommand failed.
```

The issue is MSVC's 4800 warning was really obnoxious, hence why you all had to suppress it, and they removed it in MSVC 2017. But that means trying to disable it triggers warning C4619, which is that you tried to disable a warning that doesn't exist. This warning isn't actually enabled by default, but BoringSSL builds with a pretty aggressive set of warnings. C4619 seems useful to catch typos in warning suppressions, so I've left it enabled in BoringSSL. We could build GTest under a different set of warnings than BoringSSL, but this one was easy enough to fix.

The options are:
1. Suppress C4619 in that block as well.
2. Wrap the two GTEST_DISABLE_MSC_WARNINGS* across in _MSC_VER < 1900.
3. Declare that GTest does not support C4619.

Your call on which you prefer.

gennadiycivil:
Logically , this problem only happens on VS2017. Therefore it seems that best solution for this scope only and not affecting others would be to wrap this into VS2017 only.

gennadiycivil:
Discussion is welcomed of course

davidben:
Happy to switch it to ifdefs. That's what I had originally, but then I realized this was a little shorter. But it's a somewhat more targeted change which is nice. (Though note that macro is already MSVC-only.)

jasjuang:
Ran into the same exact problem. What is the current status of this PR?

davidben:
Replaced with an ifdef.

PR #1113, opened 2017-06-08
Fixing float comparison broken link

jorgehb:
Updating the link with a valid one. The old one was considered obsolete by the publisher.


jorgehb:
I signed it!

PR #1118, opened 2017-06-14
Improving naming convention for interfaces

jesshack10:
Since originally Turtle object is just an interface, it is more readable for developers to indicate that in its name so we know it is a pure interface just with the fact of reading its name, and not a concrete class.


jesshack10:
I signed it!

gennadiycivil:
Note that this coding style is not compliant with the coding style for this project as specified in 
https://github.com/google/googletest/blob/48ee8e98abc950abd8541e15550b18f8f6cfb3a9/googletest/docs/DevGuide.md#coding-style

You can see interfaces specification [coding style](https://google.github.io/styleguide/cppguide.html#Interfaces)

I recommend closing this merge request

Issue #1116, opened 2017-06-11
Death test child under MinGW should not invoke Windows Error Reporting


tanzislam:
I noticed that a death test (e.g. `EXPECT_DEATH(...)`) on Windows 7 triggers Windows Error Reporting when the child process crashes, if the code is built with MinGW-w64. In the parent process, the test hangs until the user cancels or otherwise satisfies the Windows Error Reporting popup.

This simple code reproduces the problem:

```cpp
#include <gtest/gtest.h>
#include <cassert>

TEST(DeathTestTest, AssertionCrashesProgram)
{
    EXPECT_DEATH(assert(false), "");
}
```

I am using:

 - Windows 7 SP1
 - GCC v7.1.0 (part of mingw-w64)
 - Latest "master" branch of Google Test (commit 42bc671f47b)

BillyDonahue:
ok, I see how this is annoying but I don't see how to keep it from happening.

tanzislam:
I believe the issue is with [this section of code](https://github.com/google/googletest/blob/master/googletest/src/gtest.cc#L4217-L4254) being guarded by the `GTEST_HAS_SEH` macro, which [never gets set on MinGW](https://github.com/google/googletest/blob/master/googletest/include/gtest/internal/gtest-port.h#L929-L948).

That section of code doesn't seem to do anything that requires SEH support. I intend to raise a pull request to remove the `GTEST_HAS_SEH` guard from around it; getting permission from my employer first ...

PR #1120, opened 2017-06-19
Allow death test child to bypass Windows Error Reporting under MinGW

tanzislam:
The mechanics of suppressing debugger trapping and Windows Error Reporting for the crashed child process in a death test are currently guarded under the `GTEST_HAS_SEH` macro. This seems unnecessary, as the logic does not call any APIs related to Structured Error Handling.

Replace the guarding macro with the more permissive `GTEST_OS_WINDOWS`, so that Windows toolchains without SEH support (e.g. MinGW) can benefit from it.

Fixes: #1116 

NOTE: The `gtest_break_on_failure_unittest` test was previously failing on my setup (Windows 7 + GCC 7.1 + CMake 3.8.2 + Python 3.6.1), and is still failing now. According to [the comments inside the test code](https://github.com/google/googletest/blob/master/googletest/test/gtest_break_on_failure_unittest_.cc#L39-L40), it is to be invoked via the accompanying Python script and not directly (as it is expected to fail). However, the CMake build output didn't seem to heed this. I've left this existing issue to be addressed by a later change.


tanzislam:
OK so outright removal of the macro was a mistake on non-Windows. Will place `GTEST_OS_WINDOWS` in its place.

gennadiycivil:
@tanzislam Aplogies for not getting to this sooner. Could you please provide clear results "before and after" why this change is helpful.
Thanks again

tanzislam:
I have this test program containing a [death test](https://github.com/google/googletest/blob/master/googletest/docs/advanced.md#death-tests):
```cpp
#include <gtest/gtest.h>
#include <cassert>

TEST(DeathTestTest, AssertionCrashesProgram)
{
    EXPECT_DEATH(assert(false), "");
}
```
I built it against Google Test on MinGW-w64. (I'm using the MSYS2 distribution on Windows 7 SP1 x64, and I built Google Test itself using the makefile at `googletest/make/Makefile`.)

**Before this change:** Running the test program causes a "Windows Error Reporting" popup when the child process crashes as part of the death test. The parent process hangs at the death test until I cancel the popup manually.

This behavior is annoying/interruptive if the test program contains many death tests. Also, the behavior (or any similar popup indicating program malfunction) only happens with the GCC (v8.2.0 right now) in MinGW-w64 on Windows, but not on other compilers or OSs. I've confirmed this with:
 - Visual Studio Community v15.8.1 (Visual C++ / `cl.exe` v19.15.26726) on Windows 7 SP1 x64 and 10 x64 (using the project files in `googletest/msvc/2010`).
 - GCC v8.0.1 on Ubuntu 18.04.
 - Clang (Apple LLVM v9.1.0, `clang-902.0.39.2`) on macOS 10.13.6.
 - Embarcadero C++Builder v10.2.3 Starter (compiler v7.30, `bcc32c` v3.3.1), using my patch in #1217 / #1411 on Windows 7 SP1 x64.

**After this change:** The "Windows Error Reporting" popup is avoided for the crashed child process during the death test. Operation on other OSs is unaffected.

gennadiycivil:
cl 209413815

PR #1127, opened 2017-06-20
Fix background color in ColoredPrintf

zulkarnine:
Re-use existing background color for Widows' console window.
Before fix:
before

After fix:
after

This fixes a problem where the background color for ColoredPrintf would be BLACK even if the user's console is using a different BG color.

### DIFF
--- a/googletest/src/gtest.cc
+++ b/googletest/src/gtest.cc
@@ -2979,13 +2979,15 @@ void ColoredPrintf(GTestColor color, const char* fmt, ...) {
   CONSOLE_SCREEN_BUFFER_INFO buffer_info;
   GetConsoleScreenBufferInfo(stdout_handle, &buffer_info);
   const WORD old_color_attrs = buffer_info.wAttributes;
-
+  // Let's reuse the BG
+  const WORD existing_bg = old_color_attrs & 0x00F0;
+  
   // We need to flush the stream buffers into the console before each
   // SetConsoleTextAttribute call lest it affect the text that is already
   // printed but has not yet reached the console.
   fflush(stdout);
   SetConsoleTextAttribute(stdout_handle,
-                          GetColorAttribute(color) | FOREGROUND_INTENSITY);
+                          GetColorAttribute(color) | existing_bg | FOREGROUND_INTENSITY);
   vprintf(fmt, args);
 
   fflush(stdout);
--- a/googletest/src/gtest.cc
+++ b/googletest/src/gtest.cc
@@ -2980,7 +2980,8 @@ void ColoredPrintf(GTestColor color, const char* fmt, ...) {
   GetConsoleScreenBufferInfo(stdout_handle, &buffer_info);
   const WORD old_color_attrs = buffer_info.wAttributes;
   // Let's reuse the BG
-  const WORD existing_bg = old_color_attrs & 0x00F0;
+  const WORD background_mask = BACKGROUND_BLUE | BACKGROUND_GREEN | BACKGROUND_RED | BACKGROUND_INTENSITY;   
+  const WORD existing_bg = old_color_attrs & background_mask;
   
   // We need to flush the stream buffers into the console before each
   // SetConsoleTextAttribute call lest it affect the text that is already
--- a/googletest/src/gtest.cc
+++ b/googletest/src/gtest.cc
@@ -2894,6 +2894,30 @@ WORD GetColorAttribute(GTestColor color) {
   }
 }
 
+int GetBgOffset(WORD background_mask) {
+  if (background_mask == 0) return 0; //let's not fall into infinite loop
+
+  int bitOffset = 0;
+  while((background_mask & 1) == 0) {
+    background_mask >>= 1;
+    ++bitOffset;
+  }
+  return bitOffset;
+}
+
+WORD GetNewColor(const GTestColor color, const WORD old_color_attrs) {
+  // Let's reuse the BG
+  static const WORD background_mask = BACKGROUND_BLUE | BACKGROUND_GREEN | BACKGROUND_RED | BACKGROUND_INTENSITY;
+  static const WORD foreground_mask = FOREGROUND_BLUE | FOREGROUND_GREEN | FOREGROUND_RED | FOREGROUND_INTENSITY;
+  const WORD existing_bg = old_color_attrs & background_mask;
+
+  WORD new_color = GetColorAttribute(color) | existing_bg | FOREGROUND_INTENSITY;
+  static const int bg_bitOffset = GetBgOffset(background_mask); //it does not change 
+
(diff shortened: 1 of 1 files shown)


zulkarnine:
I signed it! :)

googletest/src/gtest.cc
```diff
@@ -2979,13 +2979,15 @@ void ColoredPrintf(GTestColor color, const char* fmt, ...) {
   CONSOLE_SCREEN_BUFFER_INFO buffer_info;
   GetConsoleScreenBufferInfo(stdout_handle, &buffer_info);
   const WORD old_color_attrs = buffer_info.wAttributes;
-
+  // Let's reuse the BG
+  const WORD existing_bg = old_color_attrs & 0x00F0;
```
BillyDonahue: I'm uncomfortable with the `0xf0` magic mask. Can this be programmatically determined?
Does MSVC provide a mask constant to apply here?

BillyDonahue: I think we should probably define a local mask that includes these fields from the [character attributes documentation](
https://msdn.microsoft.com/en-us/library/windows/desktop/ms682088(v=vs.85).aspx#_win32_character_attributes)

    const WORD background_mask =
      BACKGROUND_BLUE |       // Background color contains blue.
      BACKGROUND_GREEN |      // Background color contains green.
      BACKGROUND_RED |        // Background color contains red.
      BACKGROUND_INTENSITY;   // Background color is intensified.

So we can at least give a name and a single maintenance point to this concept.

zulkarnine: That sounds good. I will add the mask and push the changes.

googletest/src/gtest.cc
```diff
@@ -2979,13 +2979,16 @@ void ColoredPrintf(GTestColor color, const char* fmt, ...) {
...
+  const WORD background_mask = BACKGROUND_BLUE | BACKGROUND_GREEN | BACKGROUND_RED | BACKGROUND_INTENSITY;   
+  const WORD existing_bg = old_color_attrs & background_mask;
+  
   // We need to flush the stream buffers into the console before each
   // SetConsoleTextAttribute call lest it affect the text that is already
   // printed but has not yet reached the console.
   fflush(stdout);
   SetConsoleTextAttribute(stdout_handle,
-                          GetColorAttribute(color) | FOREGROUND_INTENSITY);
+                          GetColorAttribute(color) | existing_bg | FOREGROUND_INTENSITY);
```
BillyDonahue: Let's group the foreground bits together.
So:

    GetColorAttribute(color) | FOREGROUND_INTENSITY | existing_bg

Should we try to detect foreground==background here and tweak the color to avoid invisible text?

zulkarnine: Ok. I'll group them together. :)

That's a good question. So, if foreground == background, should we just invert the color bits (or just invert the FOREGROUND_INTENSITY bit)? That will hopefully work.

BillyDonahue: sounds good inverting FOREGROUND_INTENSITY sounds right.

googletest/src/gtest.cc
```diff
@@ -2894,6 +2894,30 @@ WORD GetColorAttribute(GTestColor color) {
   }
 }
 
+int GetBgOffset(WORD background_mask) {
+  if (background_mask == 0) return 0; //let's not fall into infinite loop
```
BillyDonahue: omit comment. Reader can figure that out.

googletest/src/gtest.cc
```diff
@@ -2894,6 +2894,30 @@ WORD GetColorAttribute(GTestColor color) {
...
+}
+
+WORD GetNewColor(GTestColor color, WORD old_color_attrs) {
+  // Let's reuse the BG
+  static const WORD background_mask = BACKGROUND_BLUE | BACKGROUND_GREEN | BACKGROUND_RED | BACKGROUND_INTENSITY;
+  static const WORD foreground_mask = FOREGROUND_BLUE | FOREGROUND_GREEN | FOREGROUND_RED | FOREGROUND_INTENSITY;
+  const WORD existing_bg = old_color_attrs & background_mask;
+
+  WORD new_color = GetColorAttribute(color) | existing_bg | FOREGROUND_INTENSITY;
+  static const int bg_bitOffset = GetBgOffset(background_mask); //it does not change
```
BillyDonahue: omit comment. That's what static const already means.

googletest/src/gtest.cc
```diff
@@ -2894,6 +2894,30 @@ WORD GetColorAttribute(GTestColor color) {
...
+WORD GetNewColor(GTestColor color, WORD old_color_attrs) {
+  // Let's reuse the BG
+  static const WORD background_mask = BACKGROUND_BLUE | BACKGROUND_GREEN | BACKGROUND_RED | BACKGROUND_INTENSITY;
+  static const WORD foreground_mask = FOREGROUND_BLUE | FOREGROUND_GREEN | FOREGROUND_RED | FOREGROUND_INTENSITY;
+  const WORD existing_bg = old_color_attrs & background_mask;
+
+  WORD new_color = GetColorAttribute(color) | existing_bg | FOREGROUND_INTENSITY;
+  static const int bg_bitOffset = GetBgOffset(background_mask); //it does not change 
+
+  if (((new_color & background_mask) >> bg_bitOffset) == (new_color & foreground_mask)) new_color ^= FOREGROUND_INTENSITY; //revert intensity
```
BillyDonahue: line too long. <80 cols please.

zulkarnine:
@BillyDonahue, Could you please take a look at the new patches?

gennadiycivil:
@zulkarnine , could you please provide evidence of testing
1) How you tested this change
2) What platforms
3) Any relevant results

zulkarnine:
@gennadiycivil, Thanks for asking.
1. Testing process:
a. Change the background color of the cmd to any color other than Black.
b. Run any unit test.
It can be seen that the background for every colored texts are black. Like this:
nocolorpatch
c. After applying these patches, run the same tests again and the background color will be same as the selected color in step-a
aftercolorpatch
For example, I ran **sample1_unittest** in this case

2. I have tested on **Windows 7** only (This patch affects Windows platform only). I haven't tested on any other version of windows.

3. Please check the screenshots attached.

gennadiycivil:
Thank you for testing, looks good. It would be important to test on one more Windows platform, something more recent and relevant that Windows7. Do you have ability to test on newer Windows?

zulkarnine:
I will try to run this on Windows 10 today and upload the test results once it is done.

zulkarnine:
I ran the same test on Windows 10 and it works correctly.
Here is the test result:

**Unpatched:**
unpatched

**Patched:**
patched

zulkarnine:
Thanks. :)

PR #1130, opened 2017-06-26
rvalue-only object streaming to Message

WillerZ:
Some objects are designed to exist only as temporaries while a single statement is evaluated. They typically capture constructor parameters by reference and do other things which are not safe for long-lived objects.

One way to make it more difficult for these objects to be used incorrectly is to define all the public methods/operators/etc such that only `&&` and `const&&` parameters are permitted. In this case the expected streaming operator declaration for such a `Type` is `std::ostream& operator<<(std::ostream&, Type const&&)`. For correctness, `std::ostream& operator<<(std::ostream&, Type const&)` would not be implemented.

This change allows such objects to be used with googletest Message objects.

### DIFF
--- a/googletest/include/gtest/gtest-message.h
+++ b/googletest/include/gtest/gtest-message.h
@@ -154,6 +154,25 @@ class GTEST_API_ Message {
     }
     return *this;
   }
+
+#if __cplusplus >= 201103L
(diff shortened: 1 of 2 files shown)


gennadiycivil:
No Cla

PR #1159, opened 2017-07-27
Add function name to exception if there's no default action

ShadowIce:
Adds the name of the mocked method to the exception that is thrown on NiceMock objects if there is no default action for the return type.


ShadowIce:
I signed it!

gennadiycivil:
This still needs  a CLA for one or more commit authors. Closing, please re-submit when ready

PR #1164, opened 2017-07-30
remove documentation of former versions - it's available in the commit history

hethi:
As former versions can easily be found by checking out the appropriate git tags, I see no need to keep the old documentation in parallel to the current documentation.

There is also some reference to SVN trunk, which is obsolete, and one link in the googlemock documentation to the pump documentation can be fixed.


gennadiycivil:
Sadly , looks out of date, please fix the merge conflicts

googletest/docs/Documentation.md
```diff
@@ -1,6 +1,8 @@
-This page lists all documentation wiki pages for Google Test **(the SVN trunk version)**
+This page lists all documentation markdown files for Google Test **(the
+current git version)**
 -- **if you use a released version of Google Test, please read the
```
gennadiycivil: I would not say "released", I would say "an older version"

gennadiycivil:
I agree in principle. Could you please change the wording a bit, be careful with "released version" terminology,. In essence we are saying that  " former versions can easily be found by checking out the appropriate git tags" so lets be consistent and say "former versions" everywhere and not go into the "released" versions

googletest/docs/Documentation.md
hethi: I will take "former version" as you mentioned in the discussion.

hethi:
The term "released version" was already in the documentation before my changes, but you're right, "former version" is better.

Issue #948, opened 2016-11-29
Allow MOCK_METHOD without specifying the argument count


ugoren:
Currently, you must use the correct MOCK_METHODn macro:
```
MOCK_METHOD2(foo, void(int, int));
MOCK_METHOD3(bar, void(int, int, int));
```
I suggest a variation that works without the number:
```
MOCK_METHOD(foo, void, (int, int));
MOCK_METHOD(bar, void, (int, int, int));
```
Note that the return type is now separate from the argument types.

Implementing this isn't complicated. It's basically this:
```
#define MOCK_METHOD(name, ret, args) MOCK_METHOD##PP_NARG(args)(name, ret args)
```
Where `PP_NARG` is define as in [this StackOverflow answer](http://stackoverflow.com/a/2124385/1111371).
It will be limited to a maximum number of arguments - but so are existing `MOCK_METHODn` macros.
Perhaps there's a more elegant implementation, instead of reducing it to existing macros.

I'm willing to do the development (if I think there's a chance that it will be accepted).

sbenzaquen:
On Tue, Nov 29, 2016 at 10:32 AM, ugoren <notifications@github.com> wrote:

> Currently, you must use the correct MOCK_METHODn macro:
>
> MOCK_METHOD2(foo, void(int, int));
> MOCK_METHOD3(bar, void(int, int, int));
>
> I suggest a variation that works without the number:
>
> MOCK_METHOD(foo, void, (int, int));
> MOCK_METHOD(bar, void, (int, int, int));
>
> Note that the return type is now separate from the argument types.
>
> Implementing this isn't complicated. It's basically this:
>
> #define MOCK_METHOD(name, ret, args) MOCK_METHOD##PP_NARG(args)(name, ret args)
>
> Where PP_NARG is define as in this StackOverflow answer
> <http://stackoverflow.com/a/2124385/1111371>.
> It will be limited to a maximum number of arguments - but so are existing
> MOCK_METHODn macros.
> Perhaps there's a more elegant implementation, instead of reducing it to
> existing macros.
>
> I'm willing to do the development (if I think there's a chance that it
> will be accepted).
>

We have been toying with the idea of one macro to rule them all.
Right now we have a combinatoric explosion of arity * const * template *
calltype.
This is one of the reasons we have not added C++11 support here for
override, final, reference qualification, etc.
Supporting all those combinations would give us hundreds of macros and make
the problem worse.

We discussed a few different approaches last time we talked about this
problem.
[...]

PR #1168, opened 2017-07-31
[Fix 546] Added general macros MOCK_QUALIFIED_METHOD*. (--rebased--)

Bagira80:
This is the same pull-request as in #794 but based on current master branch.

Former discussion can be found in the original pull-request.

### Text from original pull-request:

This fixes issue #546.

These new macros `MOCK_QUALIFIED_METHODx` are a generalized version of the
former macros `MOCK_METHODx` and `MOCK_CONST_METHODx`. They allow any
combination of function qualifiers (const, volatile, noexcept).

Versions for different call-types do also exist
(`MOCK_QUALIFIED_METHODx_WITH_CALLTYPE`) as well as templated versions
(`MOCK_QUALIFIED_METHODx_T` and `MOCK_QUALIFIED_METHODx_T_WITH_CALLTYPE`).

Example:

```
struct Example {
  virtual void func() = 0;
  virtual int constFunc(double) const = 0;
  virtual int constNoexceptFunc() const noexcept = 0;
};

struct ExampleMock : public Example {
  MOCK_QUALIFIED_METHOD0(func, , void());
  MOCK_QUALIFIED_METHOD1(constFunc, const, int(double));
  MOCK_QUALIFIED_METHOD0(constNoexceptFunc, const noexcept, int());
};
```

**Note**
This mainly applies the patches (from Edgar Ellis) referenced from issue #546 and fixes the errors which occurred while doing so (because the code evolved since the time the original patches were created).

### DIFF
--- a/googlemock/include/gmock/gmock-generated-function-mockers.h
+++ b/googlemock/include/gmock/gmock-generated-function-mockers.h
@@ -320,6 +320,16 @@ class FunctionMocker<R(A1, A2, A3, A4, A5, A6, A7, A8, A9, A10)> : public
   }
 };
 
+
+template <typename T>
+T& RemoveVolatileCast(T& v) {
+  return v;
+}
+template <typename T>
+T& RemoveVolatileCast(T volatile& v) {
+  return const_cast<T&>(v);
+}
+
 }  // namespace internal
 
 // The style guide prohibits "using" statements in a namespace scope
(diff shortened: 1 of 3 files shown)


irwand:
I just want to say again that this patch is very useful to us. Here's what I posted on issue #794 
--
National Instruments, the company I work for, has also been using this successfully since Sept 2016. We'd appreciate it if this patch makes it into the master branch :-)
Thank you much!
--
So 👍 for me
And thank you again Bagira80!

gennadiycivil:
I see 2 failed CI builds. I m going to close this and re-open just to trigger the builds.

gennadiycivil:
I still see failed [AppVeyor](https://ci.appveyor.com/project/GoogleTestAppVeyor/googletest/build/545) build on Windows. Please fix

evgen48:
@gennadiycivil  may we know when will be next time you decide to look at this PR?
Originally when it was submitted 28 days ago against same compilation issue for previous PR there were no errors, obviously after 1 month they can appear after merge from master. So if you are going to come back in say 1 month it doesn't make sense to fix it right away.

Bagira80:
OK, I might have found the issue and have a fix. But as I do not have any Windows, I cannot verify it. (I would prefer AppVeyor doing this for me.)

**Nevertheless, the real problem is that I cannot reopen this pull-request** because I do not have the required access rights to the *googletest* repository.
Pushing my fix without this pull-request being open is possible but then this pull-request might not be able to be re-opened at all. (See [this](https://github.com/isaacs/github/issues/361) for a reason.) And therefore my changes will not appear here.

@gennadiycivil Could you please re-open this pull-request, so that I can fix the issues?

Thanks,
Deniz

PS: I do not see any value in creating a third pull-request with the same changes...

evgen48:
> PS: I do not see any value in creating a third pull-request with the same changes... 

Vote up for it - new PR will go into top of remaining 60 PRs, and will be checked again significantly later. So it is better to reopen that one and give you time to fix. Not sure why it was closed, and how it was supposed to be fixed then.

gennadiycivil:
I ll re-open it to trigger the builds, no problem. We need clean builds before this can be looked at.

Bagira80:
OK, I rebased my branch and the new commit which includes the fixes is in commit 392c0e5e718b1c69e3518413d06eadd4b8c92eed.

I hope, the builds will be triggered automatically for that commit, too.
(They seemed to have started when re-opening this ticket and before I pushed my changes.)

gennadiycivil:
Thank you very much again for this contribution.

After review, it is useful however we do not want to add more combination of the macros. This work would add dozens of new macros.

There was a recent design idea of having a single macro MOCK_METHOD() that worked for everything.
I suggest looking into #948 , that has better potential going forward.

Thank you again

gennadiycivil:
We would appreciate working on #948 and moving it forward with new design conversations and implementations

evgen48:
I think that #948 PP_NARG() suggestion is absolutely must have feature,  however not sure whether mock_qualified proposal can be implemented in same way as number of arguments is proposed to be hidden, @gennadiycivil thanks for letting us know.

PR #1177, opened 2017-08-05
ignore QtCreator's CMakeLists.txt.user local settings

hethi:
While it's nice to be able to user cmake projects with QtCreator on Linux (and others), it's a little bit annoying that QtCreator's local settings in file CMakeLists.txt.user make git signalling a change, so ignore this type of files.


mathstuf:
IMO, these kinds of things (along with editor backup files, etc.) belong in user-level ignore files, not per-project. Is there a reason that isn't sufficient here?

hethi:
Ah, okay. You're right - I was tempted to add this due to other project's habit, but it's really better to keep this user specific setting in a user level ignore file.

Issue #1085, opened 2017-05-08
Consider renaming term "test case" to "test suite"


ghost:
Google Test is using the term "test case" as sort of container of "test", as is written in [Primer.md](https://github.com/google/googletest/blob/master/googletest/docs/Primer.md#basic-concepts):

> A _test case_ contains one or many tests. You should group your tests into test cases that reflect the structure of the tested code. When multiple tests in a test case need to share common objects and subroutines, you can put them into a _test fixture_ class.

It seems to me, that contrary to this use of the term "_test case_", most quality assurance literature including the International Software Testing Qualifications Board (ISTQB) are using the term "_test suite_" for this purpose. E.g. ISTQB's glossary about [test suites](http://glossary.istqb.org/search/test%20suite):

> A set of several test cases for a component or system under test, where the post condition of one test is often used as the precondition for the next one.

And about [test case](http://glossary.istqb.org/search/test%20case):

> A set of input values, execution preconditions, expected results and execution postconditions, developed for a particular objective or test condition, such as to exercise a particular program path or to verify compliance with a specific requirement.

So the `TEST()` macro defines a _test case_ in my opinion, and multiple `TEST()` together form the _test suite_, and the documentation might be changed accordingly in [Primer.md](https://github.com/google/googletest/blob/master/googletest/docs/Primer.md#simple-tests). Currently it is

> TEST() arguments go from general to specific. The first argument is the name of the test case, and the second argument is the test's name within the test case.

In my opinion, the common use of _test case_ and _test suite_ would be reflected by changing this to

> TEST() arguments go from general to specific. The first argument is the name of the test suite, and the second argument is the test case's name within the test suite.

Then instead of `TEST(test_case_name, test_name)` it would be `TEST(test_suite_name, test_case_name)`.

Of course, the code would also need some changes to reflect this.

Maybe this was discussed already and I didn't find the issue. Would be interested to have some discussion on this issue. For my purpose, I just changed the code to output the number of "test suites" instead of "test cases" as shown in pull request #1084 to avoid confusion locally.

BillyDonahue:
Yes. Thanks for the ISTQB links.
I agree that this is a confusing violation of common conventional terminology.

It goes back to the beginning of gtest and will be very difficult to change, so I think everyone has just gotten used to the backwardsness. What can we really do about it?

hethi:
[...]

PR #1178, opened 2017-08-05
add note about different definitions of Test Case

hethi:
There are contradictory definitions of the term "test case", so prepare
new users in Primer.md to avoid confusion.

This pull request is related to issue #1085.

### DIFF
--- a/googletest/docs/Primer.md
+++ b/googletest/docs/Primer.md
@@ -43,7 +43,7 @@ others.
 
 The term _Test_ is commonly of broad enough sense, including ISTQB's
 definition of _Test Case_, so it's not much of a problem here. But the
-term _Test Case_ is of contradictory sense and thus confusing.
+term _Test Case_ as used in Google Test is of contradictory sense and thus confusing.
 
 Unfortunately replacing the term _Test Case_ by _Test Suite_ throughout
 the Google C++ Testing Framework is not easy without breaking dependent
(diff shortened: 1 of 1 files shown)


gennadiycivil:
Please stay tuned, this is being considered. 

Also it would be helpful as a follow up to re-visit [1084](https://github.com/google/googletest/pull/1084),

googletest/docs/Primer.md
```diff
@@ -23,6 +23,40 @@ So let's go!
...
+_Note:_ There might be some confusion of idea due to different
+definitions of the terms _Test_, _Test Case_ and _Test Suite_, so beware
+of misunderstanding these.
+
+Historically, the Google C++ Testing Framework started to use the term
+_Test Case_ for grouping related tests, i.e. in the same sense as
+current publications, including the International Software Testing
+Qualifications Board ([ISTQB](http://www.istqb.org/)) and different
+textbooks on Software Quality, are using the term _[Test
+Suite](http://glossary.istqb.org/search/test%20suite)_.
```
jdennett: This might be easier for humans to parse if written as "Historically, the Google C++ Testing Framework started to use the term Test Case for grouping related tests, whereas current publications including the International Software Testing Qualifications Board (ISTQB) and various textbooks on Software Quality use the term Test Suite for this."

That way the reader is alerted earlier that this is a distinction, not a similarity.

hethi: Thanks for the clarification. I gave my very best, but I'm not a native speaker :-)

PR #1183, opened 2017-08-08
Make mocks movable

marbub:
If you use templated mocks it can be very handy to create them in
factory functions. To archive it the mocks needs to be movable, which is
prevented by the definition of ~FunctionMockerBase which removes the move
contructor. So we added a default move constructor again.

mock.h
```
class MockSqliteWriteStatement
{
public:
    MOCK_METHOD2(write,
                 void (Utils::SmallStringView));

};
```
implementation.h
```
template<typename WriteStatement>
class SqliteStatementFactory
{
private:
    static WriteStatement createStatement()
    {
        SqlStatementBuilder builder;
        builder.setTableName("newSybols");

        return WriteStatement(builder.sqlStatement());
    }

    WriteStatement m_statement{createStatement()};
}
```


marbub:
I signed it!

marbub:
I tried to set my email address as alternative email address but it still does not match the CLA.

googlemock/include/gmock/gmock-spec-builders.h
```diff
@@ -1464,6 +1464,9 @@ class FunctionMockerBase : public UntypedFunctionMockerBase {
   typedef typename Function<F>::ArgumentMatcherTuple ArgumentMatcherTuple;
 
   FunctionMockerBase() : current_spec_(this) {}
+#if GTEST_LANG_CXX11
+  FunctionMockerBase(FunctionMockerBase&&) = default;
```
nyibbang: Visual Studio 2013 does not support defaulted move constructor/operator.

marbub: There is no feature testing macro. So we should exclude Visual Studio 2013?

nyibbang: It would be easier to implement it instead of using "= default" I think.

rburny:
Unfortunately, this will not work with the current GMock implementation.
GMock maintains the map of objects, stored by their address:

https://github.com/google/googletest/blob/a325ad2db5deb623eab740527e559b81c0f39d65/googlemock/src/gmock-spec-builders.cc#L549

This is used by `AllowUninterestingCalls`, `VerifyAndClearExpectations`, and other internal methods. If you move the mock object, these would stop working (in particular, Nice and StrictMock would not work correctly).

googlemock/include/gmock/gmock-spec-builders.h
KindDragon: `GTEST_LANG_CXX11` is not defined for Visual Studio 2013, so it's ok

gennadiycivil:
Closing , this approach wont work

Issue #521, opened 2015-08-25
Infinite Loop when calling a mock function that takes boost::filesystem::path as parameter


BillyDonahue:
_From @GoogleCodeExporter on August 24, 2015 22:40_

```
Minimal example:

---------------------
#include "gtest/gtest.h"
#include "gmock/gmock.h"
#include <boost/filesystem.hpp>

class MyMock {
public:
  MOCK_METHOD1(func, void(const boost::filesystem::path &path));
};

TEST(MyTest, mockTest) {
  MyMock mock;
  mock.func(boost::filesystem::path("/myfile"));
}
-------------------

When running this test (using gtest 1.7.0 and gmock 1.7.0 on linux 64bit with 
g++ (Ubuntu 4.9.1-16ubuntu6)), it aborts with a SIGSEV.
gdb shows that the SIGSEV is raised in malloc(), while the stack is very large 
and contains a short periodic set of functions calling each other in a circle. 
I assume there is an infinite loop.

The interesting part of the stacktrace is here:


#0  _int_malloc (av=0x7ffff703a760 <main_arena>, bytes=26) at malloc.c:3302
#1  0x00007ffff6cff1e0 in __GI___libc_malloc (bytes=26) at malloc.c:2891
#2  0x00007ffff72b7688 in operator new(unsigned long) () from 
/usr/lib/x86_64-linux-gnu/libstdc++.so.6
#3  0x00007ffff731b669 in std::string::_Rep::_S_create(unsigned long, unsigned 
long, std::allocator<char> const&) () from 
/usr/lib/x86_64-linux-gnu/libstdc++.so.6
#4  0x00007ffff731b6f4 in std::string::_M_mutate(unsigned long, unsigned long, 
unsigned long) () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#5  0x00007ffff731bd2e in std::string::_M_replace_safe(unsigned long, unsigned 
long, char const*, unsigned long) () from 
/usr/lib/x86_64-linux-gnu/libstdc++.so.6
#6  0x00007ffff731d1f9 in std::string::replace(unsigned long, unsigned long, 
char const*, unsigned long) () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#7  0x00007ffff7779821 in boost::filesystem::path::begin() const () from 
/usr/lib/x86_64-linux-gnu/libboost_filesystem.so.1.55.0
#8  0x00000000004c3cd4 in 
[...]

PR #1186, opened 2017-08-09
Prevent infinite loops for recursive containers like boost::filesystem::path

Dani-Hub:
This pull request introduces an additional detection for recursively defined containers (such as ``boost::filesystem::path`` or ``std::filesystem::path``) and prevents the possibility of an infinite loop within the output handler for containers.


Dani-Hub:
I signed the CLA.

gennadiycivil:
Please add tests so we have evidence that this in fact works as intended

Dani-Hub:
Can you give me a hint about the test guidelines in googletest? First: Where should these be added? Is it expected to add tests to both googletest and googlemock? Thanks.

Dani-Hub:
I looked into the existing googletest test directory and found that gtest-printers_test.cc is already the right test for this. I will add a corresponding test case to this pre-existing test file.

Dani-Hub:
I signed the CLA.

Dani-Hub:
Is there anything that I need to do to realize that the CI build will run again?

Dani-Hub:
Could someone please review the suggested patch? Thank you!

gennadiycivil:
Note that appVeyor build failed , could you please take a look what happened

Dani-Hub:
Yep, I'll fix it this evening.

Dani-Hub:
Note that the fixed had been applied and the previous build succeeded. I decided to update the branch yesterday (This IC is still in progress), but that seems to be an endless story, so I don't intend to apply further changes and instead await the review approval.

gennadiycivil:
Thank you for the updates. I fully support small PRs that are much easier to handle. For this PR, could you please add full test information as well as the test output before/after to demonstrate what we are fixing and how works now. Thank you

Dani-Hub:
I'm not sure that I understand your request, but the set of changes are intended to solve the problem described in issue #521. Does this clarify your question?

gennadiycivil:
Yes, this has information needed. 
We just need testing evidence, test *before* this PR and demonstrate the problem in action ... and after that test again "after* this PR to demonstrate that solution works

Dani-Hub:
The real-world test will require someone to validate that `boost::filesystem::path` and `std::filesystem::path` will cause normal output by the gtest printers, but I cannot make gtest itself dependent on these concrete types, therefore these types are not used in the test. The patch will work for every "recursive" container and that is validated by the added test in `gtest-printers_test.cc` by means of the most minimal recursive container example denoted as `PathLike`.

gennadiycivil:
How about using Alexhuszagh's Jun 19 suggestion (<experimental/filesytem>) ? 
We do need a real world test. The best would be adding a new unit test, with appropriate ifdefs to protect against old compilers

Dani-Hub:
No, `<experimental/filesystem>` is not a required header, so we cannot rely on the assumption that it is available. I don't understand the difference between the test type and a "real-world" type. I'm not willing to re-implement std::filesystem within a test just as a proof for *another* possible type in the world that behaves correctly similar to `PathLike`. What is wrong with `PathLike` alone?

SimonEbner:
PR fixed the issue nlohmann/json#709

gennadiycivil:
Thank you for detailed explanation. lgtm

PR #1187, opened 2017-08-09
Fix temp dir on Android.

yabinc:
/sdcard is no longer available for processes running in app
context starting from Android O. Instead, first try /data/local/tmp,
which is usually used as the temporary directory, then try the
current directory, which is usually accessible in app context.


gennadiycivil:
Can you please check with different versions of Android. "/sdcard is no longer available for processes running in app context starting from Android O." - can we ensure that this change does not break older version of Android?

gennadiycivil:
Ping

gennadiycivil:
There hasn't been much activity on it. Closing, but please re-open if it still occurs.

enh:
oops, looks like we dropped the ball on this. yes, this works fine on older releases. (we found that we'd not got this upstreamed on an internal bug today where someone wanted this fix in the google3 copy of the NDK, and asked me why the fix isn't in the public copy of the NDK, and i asked yabinc why the fix isn't in upstream googletest...)

enh:
yabinc: i don't have the power to reopen this, so can you, to make sure it's back on their radar?

yabinc:
I also can't reopen it, will ping gennadiycivil or create a new one.

gennadiycivil:
@enh  @yabinc  As per contribution guidelines , the best would be to make an internal CL, once that lands open a new PR and I will merge it.

yabinc:
Thanks, will do it.

PR #1189, opened 2017-08-10
Adding a flag option to change the default mock type

alyssawilk:
As discussed with Samuel Benzaquen


googlemock/src/gmock.cc
```diff
@@ -48,6 +48,13 @@ GMOCK_DEFINE_string_(verbose, internal::kWarningVerbosity,
                      "  warning - prints warnings and errors.\n"
                      "  error   - prints errors only.");
 
+GMOCK_DEFINE_int32_(default_mock_behavior, 1,
```
sbenzaquen: What happens if someone sets this to an invalid value?
Unless we prevent it, the static_cast above can have undefined behavior.

googlemock/src/gmock.cc
```diff
@@ -120,6 +127,19 @@ static bool ParseGoogleMockStringFlag(const char* str, const char* flag,
   return true;
 }
 
+static bool ParseGoogleMockIntFlag(const char* str, const char* flag,
```
sbenzaquen: Maybe we should just call testing::internal::ParseInt32Flag instead of rewriting the function.

googlemock/src/gmock.cc
alyssawilk: Fixed.

googlemock/src/gmock.cc
alyssawilk: ping!

alyssawilk: ping!

gennadiycivil: @alyssawilk is this upstreaming an internal change or brand new?

alyssawilk:
Brand new.

googlemock/src/gmock.cc
alyssawilk: deprecated in favor of upstream google3 cr/

alyssawilk: @gennadiycivil I think this is ready for another look!

PR #1191, opened 2017-08-10
Adding Bazel build to googletest (initial version ) 

gennadiycivil:
Added [bazel](https://bazel.build/) BUILDfiles and workspace
Minor Changes to samples


BUILD.bazel
```diff
@@ -0,0 +1,151 @@
...
+        "//conditions:default": [
+            "-pthread",
+        ],
+    }),
+)
+
+cc_library(
+    name = "gtest_main",
+    srcs = [
+        "googletest/src/gtest_main.cc",
```
sbenzaquen: If we are including both gTest and gMock in a single target, this should use gmock_main.cc, which calls InitGoogleMock.
InitGoogleMock calls InitGoogleTest, plus does a little more related to gMock itself.

googletest/test/BUILD.bazel
```diff
@@ -0,0 +1,142 @@
...
+#
+# Bazel BUILD for The Google C++ Testing Framework (Google Test)
+
+licenses(["notice"])
+
+""" gtest own tests """
+
+#on windows exclude gtest-tuple.h and gtest-tuple_test.cc
+filegroup(
+    name = "win_only_test_files",
```
sbenzaquen: can this rule be written in terms of default_test_files instead of copying it?

BUILD.bazel
gennadiycivil: Done

googletest/test/BUILD.bazel
gennadiycivil: I could not figure out a way to do this.
I tried
srcs = glob( 
include = [<...snip>], 
exclude = [<...snip>] + select({ 
":win": ["specific file1","specific file2" ], 
"//conditions:default": [], 
}), 
), 
select() does not work inside glob().

The problem is that default ( linux ) is MORE files not less. I could sitch default to be windows but this would be inconsistent. So I accepted that in this case I had to copy.

there may be a way to do this more efficiently but at a time I cant find it and dont think that it is important enough to delay

sbenzaquen: We can separate as portable files and nix-specific files, which includes portable files.

gennadiycivil: Done

PR #1193, opened 2017-08-10
Fix scoped class gmock-gen.py

joshbodily:
Addresses the issue of scoped enums causing the mock generator script to skip subsequent method declarations.

🙏 Thanks!


joshbodily:
I signed it!

joshbodily:
Bump

PR #1194, opened 2017-08-10
Fix scoped enum not working in gmock-gen.py

joshbodily:
Addresses the issue of scoped enums causing the mock generator script to skip subsequent method declarations.

🙏 Thanks!


gennadiycivil:
Could you please provide 
* more explanation on the problem that you are fixing
* clear tests before and after showing that the problem has been fixed. In other words - test case for this

Thank you

joshbodily:
You betcha

joshbodily:
So, the problem we were having was that a number of our interface classes that we generate mocks from have scoped enums `enum class`.  `gen_mocks.py` would not properly handle those and would generate an empty file.  Solution, or workaround, was to treat scoped enums as enums.

If I comment out my change to `ast.py`, the unit test I added fails w/ when running `python gmock_class_test.py`:
```
AssertionError:
class MockTest : public Test {
  public:
  MOCK_METHOD1(Bar,
    void(const FooType& test_arg));
};
!=
class MockTest : public Test {
};
```
(Note, I insterted the line breaks to make it easier to see the problem.)

joshbodily:
[gmock_class_test.txt](https://github.com/google/googletest/files/1226012/gmock_class_test.txt)

Raw test output w/ my unit test from this branch & my changes commented out.

gennadiycivil:
My apologies for the long wait.

joshbodily:
No worries.  Thanks!

PR #1199, opened 2017-08-14
Adding support for chrono instead of gettimeofday

gaspardpetit:
This is useful on systems which do not support gettimeofdays but have support for std::chrono


gaspardpetit:
The check to enable chrono is done differently in https://github.com/google/googletest/pull/1199/ - I can update this request once https://github.com/google/googletest/pull/1199/ is integrated.

bryanzim:
Pull request #1370 has this fix as well as monotonic clocks for duration.  If you would like to lend support to it.

gennadiycivil:
@gaspardpetit  Apologies for not getting to this sooner. The changes seem to depend on GTEST_HAS_CHRONO_. 
Where is it defined?
Thanks

googletest/src/gtest.cc
```diff
@@ -107,6 +107,10 @@
 # include <windows.h>  // NOLINT
 # undef min
 
+#elif GTEST_HAS_CHRONO_
```
bryanzim: <chrono> is part of the C++ 11 standard so instead of GTEST_HAS_CHRONO probably want to use GTEST_LANG_CXX11 which is already a defined macro variable

googletest/src/gtest.cc
```diff
@@ -868,6 +872,10 @@ TimeInMillis GetTimeInMillis() {
   struct timeval now;
   gettimeofday(&now, NULL);
   return static_cast<TimeInMillis>(now.tv_sec) * 1000 + now.tv_usec / 1000;
+#elif GTEST_HAS_CHRONO_
+	std::chrono::time_point<std::chrono::system_clock> now = std::chrono::system_clock::now();
+	auto duration = now.time_since_epoch();
```
bryanzim: time_since_epoch is not guaranteed to return a duration in milliseconds which is what this function is to return.  So one needs to cast it to what is expected.  

  const std::chrono::milliseconds nowMs{ std::chrono::duration_cast<std::chrono::milliseconds>(now) };
  return static_cast<TimeInMillis>(nowMs.count());

googletest/src/gtest.cc
```diff
@@ -868,6 +872,10 @@ TimeInMillis GetTimeInMillis() {
   struct timeval now;
   gettimeofday(&now, NULL);
   return static_cast<TimeInMillis>(now.tv_sec) * 1000 + now.tv_usec / 1000;
+#elif GTEST_HAS_CHRONO_
```
bryanzim: I don't believe that the current #if / #elif combination does what you want with regards to selecting chrono versus other options available here.

bryanzim:
As a last comment, pull request #1495 has this capability already in it along with monotonic clock for C++ compilers that don't support the new <chrono> library.

gennadiycivil:
I am closing this PR, please collaborate on #1495

PR #1201, opened 2017-08-14
Added support for customization of printf using new GTEST_PRINTF, GTEST_VPRINTF and GTEST_FLUSH_STDOUT macros

gaspardpetit:
My use case was to use a custom console on mobile devices which don't support stdout very well.  By redefining GTEST_PRINTF, GTEST_VPRINTF and GTEST_FLUSH_STDOUT, an application can output from googletest and render or route it differently.


gennadiycivil:
Thank you for this contributions.
Any new macros are discouraged, sorry.

PR #1203, opened 2017-08-14
User logger GTEST_LOG instead of printf in gtest.cc to log errors

gaspardpetit:
This, combined with #1201 attempts to remove raw printf so that the host application can catch them and decide what to do with them. The GTEST_LOG was already in place but wasn't used here to log fatal errors - I am not sure why - but here is a pull request which fixes that.


googletest/src/gtest.cc
```diff
@@ -3449,8 +3449,10 @@ class XmlUnitTestResultPrinter : public EmptyTestEventListener {
 XmlUnitTestResultPrinter::XmlUnitTestResultPrinter(const char* output_file)
     : output_file_(output_file) {
   if (output_file_.c_str() == NULL || output_file_.empty()) {
-    fprintf(stderr, "XML output file may not be null\n");
-    fflush(stderr);
+    {
```
gennadiycivil: For each log to fatal FATAL we can trust GTEST_LOG_ to abort(), so we can skip the subsequent call to exit()

gennadiycivil: So, it should just be GTEST_LOG_(FATAL) << "XML output file may not be null";
 and no need for { .. } scope

googletest/src/gtest.cc
```diff
@@ -3476,10 +3478,10 @@ void XmlUnitTestResultPrinter::OnTestIterationEnd(const UnitTest& unit_test,
     //   3. To interpret the meaning of errno in a thread-safe way,
     //      we need the strerror_r() function, which is not available on
     //      Windows.
-    fprintf(stderr,
-            "Unable to open file \"%s\"\n",
-            output_file_.c_str());
-    fflush(stderr);
+    { // scoped to ensure the log is flushed before we exit
```
gennadiycivil: Ditto

gennadiycivil:
@gaspardpetit please make changes as requested

PR #1206, opened 2017-08-16
Add function name to exception if there's no default action

ShadowIce:
Second try. I hope it works this time...You aren't making it easy to contribute to Google projects. 😞

### DIFF
--- a/googlemock/src/gmock-spec-builders.cc
+++ b/googlemock/src/gmock-spec-builders.cc
@@ -364,7 +364,7 @@ UntypedFunctionMockerBase::UntypedInvokeWith(const void* const untyped_args)
 
     if (!need_to_report_uninteresting_call) {
       // Perform the action without printing the call information.
-      return this->UntypedPerformDefaultAction(untyped_args, "");
+      return this->UntypedPerformDefaultAction(untyped_args, "Function call: " + std::string(Name()));
     }
 
     // Warns about the uninteresting call.
--- a/googlemock/test/gmock-nice-strict_test.cc
+++ b/googlemock/test/gmock-nice-strict_test.cc
@@ -79,6 +79,7 @@ class MockFoo : public Foo {
 
   MOCK_METHOD0(DoThis, void());
   MOCK_METHOD1(DoThat, int(bool flag));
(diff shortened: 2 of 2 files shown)


gennadiycivil:
Thank you for your contribution. Could you please provide testing evidence. We need a test case to demonstrate the functionality

ShadowIce:
What format do you need the test case in? I'm using the boost testing framework and Visual Studio 2017, so a very simple example would be:

```c++
struct Mock
{
	MOCK_METHOD0(GetValue, std::wstring());
};

BOOST_AUTO_TEST_CASE(TestMock)
{
	NiceMock<Mock> mock;
	mock.GetValue();
}
```

The output of this without my changes is:
```
Running 1 test case...
unknown location(0): fatal error: in "TestSuite/TestMock": class std::runtime_error:
    The mock function has no default action set, and its return type has no default value set.
Test.cpp(35): last checkpoint: "TestMock" entry.

*** 1 failure is detected in the test module "Unit Tests"
```

And with my changes it's:
```
Running 1 test case...
unknown location(0): fatal error: in "TestSuite/TestMock": class std::runtime_error: Function call: GetValue
    The mock function has no default action set, and its return type has no default value set.
Test.cpp(35): last checkpoint: "TestMock" entry.

*** 1 failure is detected in the test module "Unit Tests"
```

The part with "Function call..." is new. Without this it can get very difficult to find the method that was causing the error. It would also be good to know where this "unknown location" is but I haven't been able to fix that and I think it would just point me to where the exception was thrown which isn't helpful either.

gennadiycivil:
It would be great to have examples of tests running on command-line and without Boost, just to show that there is no regression problems introduced with this change. 
I would suggest to write a new test in unit tests to prove that a) this works and b) this does not break other things

ShadowIce:
I've added a test case.

There seems to be an inconsistency between the behavior with C++11 and without it. With C++11 any type that has a default-constructor will not throw an exception, and without C++11 only the types that have a BuiltInDefaultValue specialization will not throw an exception.

gennadiycivil:
Please take a look at small changes I am proposing @ShadowIce

ShadowIce:
Looks good. I just removed the unused exception string variable in the test.

PR #1219, opened 2017-08-21
Support x64 configuration for VS2010 projects

KindDragon:
VS2010 solution only to simplify old users (who used these solutions) upgrading to new gtest/gmock, new users should use CMake generated solutions. VS2010 solution can be opened in any new VS.

### DIFF
--- a/googlemock/msvc/2010/gmock.sln
+++ b/googlemock/msvc/2010/gmock.sln
@@ -10,21 +10,35 @@ EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Win32 = Debug|Win32
+		Debug|x64 = Debug|x64
 		Release|Win32 = Release|Win32
+		Release|x64 = Release|x64
 	EndGlobalSection
 	GlobalSection(ProjectConfigurationPlatforms) = postSolution
 		{34681F0D-CE45-415D-B5F2-5C662DFE3BD5}.Debug|Win32.ActiveCfg = Debug|Win32
 		{34681F0D-CE45-415D-B5F2-5C662DFE3BD5}.Debug|Win32.Build.0 = Debug|Win32
+		{34681F0D-CE45-415D-B5F2-5C662DFE3BD5}.Debug|x64.ActiveCfg = Debug|x64
+		{34681F0D-CE45-415D-B5F2-5C662DFE3BD5}.Debug|x64.Build.0 = Debug|x64
 		{34681F0D-CE45-415D-B5F2-5C662DFE3BD5}.Release|Win32.ActiveCfg = Release|Win32
 		{34681F0D-CE45-415D-B5F2-5C662DFE3BD5}.Release|Win32.Build.0 = Release|Win32
+		{34681F0D-CE45-415D-B5F2-5C662DFE3BD5}.Release|x64.ActiveCfg = Release|x64
+		{34681F0D-CE45-415D-B5F2-5C662DFE3BD5}.Release|x64.Build.0 = Release|x64
 		{F10D22F8-AC7B-4213-8720-608E7D878CD2}.Debug|Win32.ActiveCfg = Debug|Win32
 		{F10D22F8-AC7B-4213-8720-608E7D878CD2}.Debug|Win32.Build.0 = Debug|Win32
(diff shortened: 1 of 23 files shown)


gennadiycivil:
Thank you for new PR. 

I see that there is new directory googletest/msvc/2010 with VS2010 project files. 

I also see that you are removing googlemock/msvc/2015/gmock_main.vcxproj... 

could you please explain why?

The PR is about adding support for VS2010. ( I will not speculate how relevant this is anymore, assuming that it is ) . Not sure why we are removing 2015 project files from googlemock?

Thank you

KindDragon:
I think users should prefer to use CMake instead of MSVC projects. Therefore, we must either remove MSVC projects or reduce their number to a minimum. Therefore, I think that you need to leave only VS2010 projects because they can be easily opened even in VS2017 for those who do not want to use CMake.

gennadiycivil:
Thank you for the information. 
Consider this:

We are trying to *solve the issue". And the issue is defined as follows:
**" simplify old users (who used these solutions) upgrading to new gtest/gmock".**  This is well-defined problem and you are addressing it with including "2010" files. 

Later on you *also* start solving a different issu. That different issu may be defined as follows:
**"Some users are using VS2015 project files from googlemock. We want to deny this to them because they should be using CMake"**
I am not even discussing the merits of this as a issus. I am simply stating that these 2 issues should not be in the same PR. 
Please split the PRs and we can dicuss separately.

KindDragon:
Ok. I returned projects for VS2015

KindDragon:
> Later on you also start solving a different issue. That different issu may be defined as follows:
> "Some users are using VS2015 project files from googlemock. We want to deny this to them because they should be using CMake"

How they use them without VS2015 projects for gtest?

gennadiycivil:
Thank you for the changes, I am still a bit not sure of whats going on.
I see a whole bunch of files added under 2010 directory. I assume these are .sln and relarted VS2010 files.

However I also see other changes:
1)
googlemock/msvc/2015/gmock.vcxproj
googlemock/msvc/2015/gmock_main.vcxproj
googlemock/msvc/2015/gmock_test.vcxproj
It appears that in these files there is new X64 debug/release VS configuration. 

2)
There is a number of file moves. It appears that you are separating VS2005(?) 
googletest/msvc/gtest-md.sln → googletest/msvc/2005/gtest-md.sln
googletest/msvc/gtest-md.vcproj → googletest/msvc/2005/gtest-md.vcproj
googletest/msvc/gtest.sln → googletest/msvc/2005/gtest.sln
googletest/msvc/gtest.vcproj → googletest/msvc/2005/gtest.vcproj
googletest/msvc/gtest_main-md.vcproj → googletest/msvc/2005/gtest_main-md.vcproj
googletest/msvc/gtest_main.vcproj → googletest/msvc/2005/gtest_main.vcproj
googletest/msvc/gtest_prod_test-md.vcproj → ...etest/msvc/2005/gtest_prod_test-md.vcproj
googletest/msvc/gtest_prod_test.vcproj → googletest/msvc/2005/gtest_prod_test.vcproj

As previously described, we should try to do one things at a time. There are 37 changed files in this PR and from what I can many of them have nothing to do with the stated purpose or the PR.

Please understand that we need to be pedantic here, we need to do only one thing at a time and ONLY the thing that PR states it does, no more , no less. 

I recommend splitting this into separate PRs and clearly stating what each of them does. 
In this current form this PR is not acceptable I am afraid,

Thank you for your understanding

KindDragon:
> googlemock/msvc/2015/gmock_test.vcxproj
> It appears that in these files there is new X64 debug/release VS configuration.

This is because you do not want to delete VS2015 projects. I made these changes so that VS2015 projects match to the projects for VS2010.

I think we shouldn't have projects that do different things for VS2010 and VS2015

> There is a number of file moves. It appears that you are separating VS2005(?)
> googletest/msvc/gtest-md.sln → googletest/msvc/2005/gtest-md.sln

I moved VS2005 projects to sub folder to match project structure in googlemock (https://github.com/google/googletest/tree/master/googlemock/msvc). So that the user can understand where the projects are for VS2010, and where for VS2005. It makes no sense to store projects in the root directory `googletest\msvc`, after adding new projects there

gennadiycivil:
I am not questioning the motives or validity of the changes. I am simply saying that we should split the changes into separate, atomic PRs

KindDragon:
Ok, but others two will depend on this PR

PR #1223, opened 2017-08-23
change API and doc to "Test Suite" instead of "Test Case" (backward compatible)

hethi:
In continuation of the discussion in isue #1085 and complementing #1084 and #1178  the changes in this pull request change the names of Google Test's API from using the term "Test Case" to using the term "Test Suite".

To keep these changes backward compatible to projects using the "Test Case" based API names, the former API is kept available in parallel to the new API. To support a migration, the former API can be disabled by a CMake configuration option `gtest_disable_former_test_case` (see [CMakeLists.txt](1223/files#diff-7b12e50666f60ab6d4721a97c7b54868)), which is OFF by default. Setting this option to ON removes the compiler option `-DGTEST_HAS_TESTCASE=1`; usage of the old API results in compilation errors then.

The Google Test documentation (`googletest/docs`) contains a note regarding this API change ([Primer.md](1223/files#diff-d3e5412a8d2b5edded19ac9c3715dcf2)) and describes the new API only, so that new users are guided to use the new API. Current users should'nt suffer from this, as the API is identical except for few name changes.

The samples are changed too.

The internal implementation parts are not yet converted from using method names as `*test_case*` etc., this can be done after discussing and (hopefully) merging the user visible parts.

The most obvious changes are in [gtest.h](1223/files#diff-26580a56129b3aaf09631cef9f0a7bcb), e.g.:

- `class TestCase` renamed to `class TestSuite`;   
  backward compatibility is kept by `typedef TestSuite TestCase`
- additional static `TestSuite::SetUpTestSuite` and `TestSuite::TearDownTestSuite`   
  backward compatibility is kept by keeping the former `SetUpTestCase` and `TearDownTestCase`; all are called if available (though only the former or the new methods should be used)

Other affected method names and macros have been changed in a similar way, while keeping the former version (which usually just calls the new one).


hethi:
@gennadiycivil - please take a look at this. I'd like to use these changes in next semesters practice for my students :-)

gennadiycivil:
Thank you for this body of work. Unfortunately it is just too large , too many things to review, too many things to check. Logistically this is a good example where massive refactoring needs to become less massive and more manageable.
Think about how we can split this? We need to carefully plan and execute , keeping in mind that there may be many many people out there who depend on this to be the way it is ( right or wrong).

I suggest adding "new" way and deprecating the old way with the idea that it will still work for a while but eventually will stop working.

And we need to do this gradually, step by step

This is actually a good example to teach your students, what it takes to address technical debt in the real world

gennadiycivil:
Mainly lets see if we can get *small* and *manageable* piece of this work to get the reviews going and the process started

hethi:
Ok, I'll try to split this in several steps, i.e. several pull requests.

hethi:
I will wait for the discussion of #1228 now, as the further pieces depend on this.

gennadiycivil:
See https://github.com/google/googletest/projects/1#card-4441303

gennadiycivil:
Let me close this and lets start working on the design document ( sent you email)

PR #830, opened 2016-07-19
Speed up printing of characters which need hex escaping

brian-peloton:
This change speeds up the runtime of a value-parameterized test I have
which has lots of values with large strings full of unprintable
characters by 2x. I profiled it and traced most of the slowness during
googletest startup down to the way String::FormatHexInt was creating and
destroyed a stringstream for each character in the string for each
value.

PR #1225, opened 2017-08-28
Speed up printing of characters which need hex escaping

brian-peloton:
This change speeds up the runtime of a value-parameterized test I have
which has lots of values with large strings full of unprintable
characters by 2x. I profiled it and traced most of the slowness during
googletest startup down to the way String::FormatHexInt was creating and
destroyed a stringstream for each character in the string for each
value.

Originally submitted as google/googletest#830.

### DIFF
--- a/googletest/src/gtest-printers.cc
+++ b/googletest/src/gtest-printers.cc
@@ -180,7 +180,10 @@ static CharFormat PrintAsCharLiteralTo(Char c, ostream* os) {
         *os << static_cast<char>(c);
         return kAsIs;
       } else {
-        *os << "\\x" + String::FormatHexInt(static_cast<UnsignedChar>(c));
+        ostream::fmtflags flags = os->flags();
+        *os << "\\x" << std::hex << std::uppercase
+            << static_cast<int>(static_cast<UnsignedChar>(c));
+        os->flags(flags);
         return kHexEscape;
       }
   }


gennadiycivil:
Thank you for the new PR. Can you please include your profiling results as a prove that this works ( while waiting for the CI builds )

brian-peloton:
I've never used a profiling tool which is easy to paste output from (suggestions welcome), but I do have before/after times for a test program which is designed to hit this codepath a lot. My real code doesn't have this many cases, but it feeds much longer strings which are binary data so have a lot of unprintable characters in as parameters.

Before:
```console
$ bazel build -c dbg //:demo && time bazel-bin/demo > /dev/null
INFO: Found 1 target...
Target //:demo up-to-date:
  bazel-bin/demo
INFO: Elapsed time: 1.595s, Critical Path: 1.47s

real	0m1.917s
user	0m1.720s
sys	0m0.176s
```

After:
```console
$ bazel build -c dbg //:demo && time bazel-bin/demo > /dev/null
INFO: Found 1 target...
Target //:demo up-to-date:
  bazel-bin/demo
INFO: Elapsed time: 1.175s, Critical Path: 0.97s

real	0m1.070s
user	0m0.940s
sys	0m0.120s
```

Test code:
```patch
diff --git a/BUILD.bazel b/BUILD.bazel
index a442374..48f2534 100644
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -145,3 +145,11 @@ cc_test(
         ":gtest",
     ],
 )
+
+cc_binary(
+    name = "demo",
+    srcs = ["demo.cc"],
+    deps = [
+        ":gtest_main",
+    ],
+)
diff --git a/demo.cc b/demo.cc
index e69de29..b0ae0b2 100644
--- a/demo.cc
+++ b/demo.cc
@@ -0,0 +1,13 @@
+#include "gtest/gtest.h"
+
+class DemoTest
+    : public ::testing::TestWithParam<::std::tuple<uint32_t, ::std::string>> {};
+
+TEST_P(DemoTest, Nothing) {}
+
+INSTANTIATE_TEST_CASE_P(
+    DemoInstantiatedTest, DemoTest,
+    ::testing::Combine(::testing::Range<uint32_t>(0, 1 << 16, 1),
+                       ::testing::Values(::std::string(
+                           "\002\002\003\003\003\003\003\003\003\003\004\004"
+                           "\004\004\004\004\004\004\004\004\004\004"))));
```

PR #1228, opened 2017-08-29
introduce build switch to support transition from TestCase to TestSuite

hethi:
This PR is extracted from too-big #1223.

It introduces a cmake build option `gtest_disable_former_test_case` which is by default `OFF`, and so the former `TestCase` interfaces are kept, thus legacy code is not affected until at some time the default may be switched.

The idea is, that the new `TestSuite` interfaces are added in parallel to the former `TestCase` interfaces, and they are always available independent of the build option. New users of Google Test should use the `TestSuite` variant only, while old users can smoothly switch by using a mixture of legacy `TestCase` and new `TestSuite` variants.

If a Google Test user wants to transition completely to the new `TestSuite` interface, she can set the `gtest_disable_former_test_case` to `ON`, so a verification is possible that only the new interface is used.

Technically, the cmake build option (if `OFF`) defines a preprocessor macro `GTEST_HAS_TESTCASE` which will be used by following PRs to cover the former `TestCase` interfaces, as soon as the new `TestSuite` interfaces (which are always available and independent of the build switch) are added. If the build option is set to `ON`, the preprocessor macro disappears, and the former `TestCase` interface is not built anymore.


gennadiycivil:
The API refactoring is a large effort that needs to be coordinated properly. We appreciate your contributions  :+1: and want to move forward. I suggest the following process 
* Is there a GitHub "issue" for this? If not please create one
* Lets create a document where we can discuss the refactoring process and design. - could you please provide email that I can use to share the document with you. We will reference this document in the GitHub issue so other people who want to participate can ask for access and be included in the discussion
Let me know if this is Ok with you

gennadiycivil:
See https://github.com/google/googletest/projects/1#card-4441303

hethi:
The GitHub issue is #1085.

hethi:
E-mail is visible in my profile now (same as in git commits, anyway) - I appreciate getting your document :-)

hethi:
@gennadiycivil - not shure what to do with the card on the project board you linked above, seems I have no right to do there anything.

gennadiycivil:
Sorry, that card is just a To-Do item.

gennadiycivil:
Closing this for now, lets work on the design document as discussed

PR #1229, opened 2017-08-29
fix typo: xUnit

hethi:
According to [Wikipedia](https://en.wikipedia.org/wiki/XUnit), `xUnit` is spelled with a lower-case `x`.


hethi:
Trying to close and reopen this PR to restart AppVeyor (which aborted due to other configuration problems).

PR #1232, opened 2017-08-29
reduce travis buildjobs by 4/12

hethi:
Currently, Travis runs 12 build jobs:

(Linux | OSX) * (gcc | clang) * (googletest | googlemock | googlemock-c++11)

But the `googlemock` build includes `googletest` already, including all unit tests and samples of `googletest`, so a separate build of `googletest` doesn't provide additional knowledge.

By removing the `googletest` build, we can reduce the build number from 12 to 8 builds and thus get a much quicker response from Travis, without reducing the test coverage.


gennadiycivil:
I agree. Lets just wait a bit, I want Appveyor to catch up as well. ( I had Appveyor set up with duplicate builds, I fixed it now but need it to catch up)

PR #1242, opened 2017-09-01
Make the failure messages from EXPECT_EQ and friends actually symmetric (reopen of #1101)

DarthGandalf:
instead of reading more like reversing the former "expected" and "actual"
roles of the LHS and RHS arguments.

This patch is manually applied from internal version (125109873)

### DIFF
--- a/googletest/test/gtest_unittest.cc
+++ b/googletest/test/gtest_unittest.cc
@@ -2429,8 +2429,9 @@ TEST(StringAssertionTest, ASSERT_STREQ) {
   const char p2[] = "good";
   ASSERT_STREQ(p1, p2);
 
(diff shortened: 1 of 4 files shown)


googletest/test/gtest_unittest.cc
```diff
@@ -2430,7 +2430,7 @@ TEST(StringAssertionTest, ASSERT_STREQ) {
   ASSERT_STREQ(p1, p2);
 
   EXPECT_FATAL_FAILURE(ASSERT_STREQ("bad", "good"),
-                       "Expected: \"bad\"");
+                       "Expected equality of these values:\n  \"bad\"");
```
gennadiycivil: "  \"bad\"\n  \"good\""); ?

DarthGandalf:
Fixed.

PR #913, opened 2016-10-21
Remove gcc 6 misleading indentations warnings

Romain-Geissler:
This removes some simple warnings that I hit when using Gmock with gcc6 and -Wall -Wextra -Werror.

PR #1243, opened 2017-09-01
switch one Trevis build to Release mode

hethi:
Using Release mode turns on the compiler's optimization which allows the compiler to discover more problems and omit some more warnings.

This enhances the coverage of hidden problems in comparison to running debug mode builds only.

### DIFF
--- a/googletest/test/gtest_unittest.cc
+++ b/googletest/test/gtest_unittest.cc
@@ -2096,7 +2096,7 @@ class UnitTestRecordPropertyTestEnvironment : public Environment {
 };
 
 // This will test property recording outside of any test or test case.
-Environment* record_property_env =
+Environment* record_property_env GTEST_ATTRIBUTE_UNUSED_ =
     AddGlobalTestEnvironment(new UnitTestRecordPropertyTestEnvironment);
 
 // This group of tests is for predicate assertions (ASSERT_PRED*, etc)
@@ -4187,12 +4187,6 @@ TEST(AssertionSyntaxTest, WorksWithConst) {
 #endif  // GTEST_HAS_EXCEPTIONS
 
 }  // namespace
-
-// we don't use the variable further, just avoid compiler warning
-// by defining a function which uses it
-void dummy_use_of_record_property_env() {
-	(void) record_property_env;
-}
 
 namespace testing {
 
--- a/googletest/test/gtest_unittest.cc
+++ b/googletest/test/gtest_unittest.cc
@@ -2096,7 +2096,7 @@ class UnitTestRecordPropertyTestEnvironment : public Environment {
 };
 
 // This will test property recording outside of any test or test case.
-static Environment* record_property_env =
+Environment* record_property_env =
     AddGlobalTestEnvironment(new UnitTestRecordPropertyTestEnvironment);
 
 // This group of tests is for predicate assertions (ASSERT_PRED*, etc)
@@ -4187,6 +4187,12 @@ TEST(AssertionSyntaxTest, WorksWithConst) {
 #endif  // GTEST_HAS_EXCEPTIONS
 
 }  // namespace
+
+// we don't use the variable further, just avoid compiler warning
+// by defining a function which uses it
+void dummy_use_of_record_property_env() {
+	(void) record_property_env;
+}
 
 namespace testing {
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,7 +37,7 @@ script: ./travis.sh
 env:
   matrix:
     - SHARED_LIB=OFF STATIC_LIB=ON CMAKE_PKG=OFF BUILD_TYPE=Debug VERBOSE=1
-    - SHARED_LIB=OFF STATIC_LIB=ON CMAKE_PKG=OFF BUILD_TYPE=Debug VERBOSE=1 CXX_FLAGS=-std=c++11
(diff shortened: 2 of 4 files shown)


hethi:
Maybe running one build with optimization makes warnings detectable easier, similar to those reported in #913.

hethi:
Hmm, the [travis build 1243](https://travis-ci.org/google/googletest/builds/270970178?utm_source=github_status&utm_medium=notification) finished o.k., but it's not reported here in the PR - will close and reopen.

hethi:
Ok, [Travis build 911](https://travis-ci.org/google/googletest/builds/271183103?utm_source=github_status&utm_medium=notification) finished successful.

The release build on gcc is showing some warning, though, [at line 1557](https://travis-ci.org/google/googletest/jobs/271183105#L1557) in the log. It's in Google Test's unit test only, but shows that the release build options detect some more potential problems.

gennadiycivil:
I think this is very useful, however I dont think that we should modify gtest-param-test_test.cc for this to happen.
Can we figure out a way to keep modifications only to travis config files?

hethi:
The change of `gtest_param-test_test.cc` is required due to a warning which is emitted by GCC. If we treat warnings as errors (to avoid ignoring them), we need to get rid of this warning.

Option `-Wshadow` was part of the compiler switches before. By switching on optimization, GCC emitted the warning now.

The change is just renaming a local variable in unit test ... and either we test it there, or we disable the warning for this unit test completely in `CMakefiles.txt`. But disabling a warning for some code is not a good habit if we can avoid the warning by a small code change, is it?

The relevant part of the logfile of [Travis build job 914.2](https://travis-ci.org/google/googletest/jobs/271726255) (scroll to the end to see it there):
```
[ 45%] Building CXX object googlemock/gtest/CMakeFiles/gtest-param-test_test.dir/test/gtest-param-test_test.cc.o
cd /home/travis/build/google/googletest/build/googlemock/gtest && /usr/bin/g++-4.9    -std=c++11 -O3 -DNDEBUG -I/home/travis/build/google/googletest/googletest/include -I/home/travis/build/google/googletest/googletest    -std=c++11 -Wall -Wshadow -Werror -DGTEST_HAS_PTHREAD=1 -fexceptions -o CMakeFiles/gtest-param-test_test.dir/test/gtest-param-test_test.cc.o -c /home/travis/build/google/googletest/googletest/test/gtest-param-test_test.cc
In file included from /home/travis/build/google/googletest/googletest/include/gtest/gtest.h:62:0,
                 from /home/travis/build/google/googletest/googletest/test/gtest-param-test_test.cc:36:
/home/travis/build/google/googletest/googletest/test/gtest-param-test_test.cc: In lambda function:
/home/travis/build/google/googletest/googletest/test/gtest-param-test_test.cc:860:77: error: declaration of ‘const testing::TestParamInfo<std::basic_string<char> >& info’ shadows a parameter [-Werror=shadow]
                         [](const ::testing::TestParamInfo<std::string>& info) {
                                                                             ^
/home/travis/build/google/googletest/googletest/include/gtest/gtest-param-test.h:1427:10: note: in definition of macro ‘INSTANTIATE_TEST_CASE_P’
         (__VA_ARGS__)(info); \
          ^
/home/travis/build/google/googletest/googletest/include/gtest/gtest-param-test.h:1425:66: note: shadowed declaration is here
       const ::testing::TestParamInfo<test_case_name::ParamType>& info) { \
                                                                  ^
/home/travis/build/google/googletest/googletest/test/gtest-param-test_test.cc:857:1: note: in expansion of macro ‘INSTANTIATE_TEST_CASE_P’
 INSTANTIATE_TEST_CASE_P(CustomParamNameLambda,
 ^
make[2]: *** [googlemock/gtest/CMakeFiles/gtest-param-test_test.dir/test/gtest-param-test_test.cc.o] Error 1
```

googletest/test/gtest_unittest.cc
```diff
@@ -2096,7 +2096,7 @@ class UnitTestRecordPropertyTestEnvironment : public Environment {
 };
 
 // This will test property recording outside of any test or test case.
-static Environment* record_property_env =
+Environment* record_property_env =
```
gennadiycivil: Why are we removing "static"?

gennadiycivil:
Thank you , yes, I agree. Just another question, whats the rational of removing static  ( see my comment )?
Thanks again

googletest/test/gtest_unittest.cc
hethi: Oh, that was a try to resolve the warning in the first place, but didn't change anything because this `static` is redundant anyway: Line 2099 is part of a local namespace (which has the same semantic in C++ as the `static` keyword inherited from C in this use case). The local namespace starts at line 316 and ends at line 4189.

As `static` was redundant, I left it removed (minimal refactoring ...).

hethi:
Just added the comment in place at the file - `static` is redundant here (local namespace).

hethi:
Regarding the `-Wshadow` related problem discussed above: We also could consider removing this type of warning. [Here is a blog article against using it](http://tewha.net/2013/02/against-gcc_warn_shadow/).

googletest/test/gtest_unittest.cc
```diff
@@ -4188,6 +4188,12 @@ TEST(AssertionSyntaxTest, WorksWithConst) {
 
 }  // namespace
 
+// we don't use the variable further, just avoid compiler warning
+// by defining a function which uses it
+void dummy_use_of_record_property_env() {
```
hethi: Instead of introducing this dummy function, we could mark the variable with attribute 'unused' in line 2099 (at least for GCC which is complaining here) - more elegant?

PR #1245, opened 2017-09-04
Allow macros inside of parametrized test names.

sheepmaster:


sheepmaster:
Copied over from https://github.com/google/googletest/pull/706, which didn't list the CLA as accepted for some reason.

sheepmaster:
@gennadiycivil PTAL, thanks!

sheepmaster:
@gennadiycivil Ping?

gennadiycivil:
Could you please provide explanation _why_ allowing macros inside of parametrized test names is needed.
Not clear what problem we are trying to solve?
Thank you

sheepmaster:
A common pattern in Chromium code to conditionally disable tests (e.g. on certain platforms) is to use a macro to expand the test name to `DISABLED_...` when the condition is met. For example:
```
  #if defined(OS_MACOSX)
  #define MAYBE(X) DISABLED_X
  #else
  #define MAYBE(X) X
  #endif
  TEST(FooTest, MAYBE(BAR)) { ... }
```

That works with `TEST` or `TEST_F`, but not with `TEST_P`, but that is not obvious. Now, adding to that the fact that the most common use for disabling tests is to get the tree green again, trying to disable a parametrized test in this way would instead cause a compile failure. This PR just aligns the behavior for TEST_P with TEST_F and TEST.

gennadiycivil:
Looks sane to me... could you please do internal CL first? ( As per [DevGuide](https://github.com/google/googletest/blob/7f8fefabedf2965980585be8c2bff97458f28e0b/googletest/docs/DevGuide.md#submitting-patches))

gennadiycivil:
ping

sheepmaster:
Internal CL is up at cr/169519470.

PR #1246, opened 2017-09-04
enable Travis ccache to speed up CI builds

hethi:
Travis-CI supports [caching](https://docs.travis-ci.com/user/caching) of different build artefacts.

Using `ccache` speeds up builds especially in case of minor changes (doc etc.).

The first build for this pull request does not show any acceleration, as the cache needs to be filled first. But further builds can be faster in case the cache contains fitting objects.


hethi:
Hm, one OSX build was terminated in between with no visible reason; I'll trigger another build by closing and reopening the PR.

PR #1248, opened 2017-09-04
remove GTEST_HAS_PARAM_TESTS

### LINKED ISSUE #360: need to remove uses of GTEST_HAS_PARAM_TESTS
```
Now that all the platforms gtest supports work with value-parameterized tests, 
we should remove the uses of the GTEST_HAS_PARAM_TESTS macro from the codebase 
everywhere.
```

Original issue reported on code.google.com by `w...@google.com` on 10 Mar 2011 at 6:49

hethi:
As mentioned in issue #360:
"Now that all the platforms gtest supports work with value-parameterized
tests, we should remove the uses of the GTEST_HAS_PARAM_TESTS macro from
the codebase everywhere."

Fixes #360


gennadiycivil:
Also cl 179442080

PR #1250, opened 2017-09-05
remove obsolete link_directories command

hethi:
The link_directories() command not necessary, as the target_link_libraries command contains an absolute path already, and the path given doesn't exist anymore, leading only to linker warnings like:
```
ld: warning: directory not found for option '-L/Users/travis/build/google/googletest/build/googlemock/gtest/src'
```


hethi:
OSX builds at Travis seem to be unreliable currently, restarting by close & reopen ...

gennadiycivil:
Re-Started Travis build again

PR #1255, opened 2017-09-09
Fix ellipsis position in examples

yursha:


gennadiycivil:
I don't see what problem this is solving, could you please explain

yursha:
It solves consistency problem in using `...` in code examples.

PR #1259, opened 2017-09-14
Add explicit `CMAKE_DEBUG_POSTFIX` option

ly2048:
Enable generating different library name to be compatible with CMake's `FindGTest`.


ly2048:
I signed it!

gennadiycivil:
Thank you for your contribution
googletest at a moment requires cmake v2.6.4 or newer.
Can you please demonstrate , that this change is safe for older CMake versions

ly2048:
https://cmake.org/cmake/help/cmake2.6docs.html#variable:CMAKE_DEBUG_POSTFIX

That's the proof. Such option exists in CMake v2.6.

PR #1270, opened 2017-09-20
Make sure IsATTY does not clobber errno.

Romain-Geissler:
Hi,

I am having a trouble when migrating an existing code base from gtest 1.7 to 1.8 (+ a lot of other changes). It happens that my unittests are always run with stdout redirected to a file. In "src/gtest.cc" you do check if the output is a TTY, and if yes, enable colours. In my case, stdout is a file, so eventually "isatty" returns falls and sets errno to "ENOTTY". Later in my own tests, I initially expect errno to zero, while it is not.

This patch just makes sure that calling IsATTY doesn't have any side effect.

Cheers,
Romain


Romain-Geissler:
Ping.

Romain-Geissler:
Ping

gennadiycivil:
@Romain-Geissler Thank you for this contribution. Could you please provide test evidence/output before/after to demonstrate the result of this PR

gennadiycivil:
@Romain-Geissler ping

PR #1274, opened 2017-09-25
Increase the stack size afforded to clone

Peter-Levine:
When building googletest on Gentoo Linux, `gtest-death-test_test` fails and causes a segfault in libsandbox, a sandbox environment used by the Gentoo portage build system.  libsandbox works by being loaded through LD_PRELOAD and providing wrappers around potentially insecure system calls such as `execve`.  The segfault is caused because available stack space is exhausted in the wrapper's environment, causing a stack overflow.

In 'googletest/src/gtest-death-test.cc', in `ExecDeathTestSpawnChild`, only a page of memory is granted to clone() as stack space, presumably because it only has to execute `ExecDeathTestChildMain` which contains the call to `execve`.  But because that call is wrapped in a library function offered by libsandbox, the stack gets exhausted before the actual call to the real `execve` ever occurs.

Testing with a stack size equal to that of the parent process (139264 on my end; roughly 34 times the page size on my end) succeeded but is likely overkill.  Testing with a stack size that is 5 times the page size still segfaulted.  Testing with a stack size that is 10 times the pagesize succeeded.

See https://bugs.gentoo.org/629620 for details.


gennadiycivil:
Sincere aplogies for the long wait. I see that the parent issues are closed now

Peter-Levine:
@gennadiycivil commit message for https://github.com/google/googletest/pull/1274/commits/4a09774caa9458692dedeb06b0736034a4291666 is `Merge branch 'master' into master` but was merged to branch `9A681768AABE08D1EFA5CA77528236A4 `.  I assume that was intentional will eventually be merged to master?

Arfrever:
The original version of this pull request contains this change:
```
-    const size_t stack_size = getpagesize();
+    const size_t stack_size = getpagesize() * 10;
```
This change was never merged to master branch.

As of now, the relevant line of code in master branch is:
```
    const auto stack_size = static_cast<size_t>(getpagesize());
```
I assume that the forward-ported version of fix from this pull request would be:
```
-    const auto stack_size = static_cast<size_t>(getpagesize());
+    const auto stack_size = static_cast<size_t>(getpagesize() * 10);
```

PR #1285, opened 2017-10-05
Changed CMAKE_BINARY_DIR to CMAKE_CURRENT_BINARY_DIR in documentation.

mfoco:
Original code only worked if it was inserted in the main CMakeLists.txt, while using CMAKE_CURRENT_BINARY_DIR also works for a project added via add_subdirectory.


gennadiycivil:
NO CLA

PR #1289, opened 2017-10-08
Minor fix (proposal)

kant:


gennadiycivil:
Thank you for this proposal. 
Does not look that there is enough substance to this PR. I am going to close it but please feel free to open a new one and argue your proposal if you feel strongly about it.

Issue #1282, opened 2017-10-02
Small Documentation bug: incorrect macro used on Cheat Sheet for specifying calling convention


btpnlsl:
The documentation for specifying the calling convention of a mock method uses an incorrect macro.

Here is a link to the section: [https://github.com/google/googletest/blob/master/googlemock/docs/CheatSheet.md#specifying-calling-conventions-for-mock-functions](https://github.com/google/googletest/blob/master/googlemock/docs/CheatSheet.md#specifying-calling-conventions-for-mock-functions)

Here is the incorrect text:
`MOCK_METHOD_1_WITH_CALLTYPE(STDMETHODCALLTYPE, Foo, bool(int n));`

The text should read:
`MOCK_METHOD1_WITH_CALLTYPE(STDMETHODCALLTYPE, Foo, bool(int n));`

Thank you.

gennadiycivil:
I have been cleaning up older and inactive GitHub googletest issues. You may see an issue "closed" if it appears to be inactive/abandoned 
Thank you

PR #1295, opened 2017-10-16
Revert "#1282: Doc typo fix"

gennadiycivil:
Reverts google/googletest#1292


DariuszOstolski:
The problem seems to be related to ccache in travis environment. Maybe it is because this is docs only patch. In job log I can see:
```
$ ccache --zero-stats
/Users/travis/.travis/job_stages: line 57: ccache: command not found
```

gennadiycivil:
Travis is currently broken, this has nothing to do with the PR, I just tested this here.
See  https://github.com/travis-ci/travis-ci/issues/8552

PR #1316, opened 2017-10-30
Modernize Python 2 code to get ready for Python 3

cclauss:
Make the minimal, safe changes required to convert the repo's code to be syntax compatible with both Python 2 and Python 3. More work might be required to complete the port to Python 3 but this is a minimal, safe first step.

* [futurize stage 1](http://python-future.org/futurize_cheatsheet.html#step-1-modern-py2-code)
* file() --> open()
* Defined xrange() == range() in Python 3 only.


evelikov:
@cclauss is the series enough to get everything running with python3?
I not, for the time being we could tweak the code in `googletest/cmake/internal_utils.cmake` to explicitly request python2.7 via `find_package(PythonInterp 2.7)`

cclauss:
@evelikov I wrote:

> More work might be required to complete the port to Python 3 but this is a minimal, safe first step.

Was that statement ambiguous?

Why do we want googletest to only support Python 2 deployments when that version will EOL in two years?  It would be useful if googletest could survive the transition.

http://www.python3statement.org

evelikov:
I'm blind - apologies @cclauss
My suggestion for 2.7 was meant as interim step everything is python3 compatible. Growing number of distributions have python as a symlink to python3, causing some quite 'fun' experiences.

Issue #1319, opened 2017-10-31
possible strncmp NULL dereference in SkipPrefix


hazelnusse:
Scan build is reporting a potential null pointer dereference here:
https://github.com/google/googletest/blob/master/googletest/src/gtest.cc#L5027

This might be a false positive, but given that the path length is 18, it isn't obvious to me.

[strncmp_null_dereference.tar.gz](https://github.com/google/googletest/files/1432345/strncmp_null_dereference.tar.gz)

hazelnusse:
Hello?

MichaelVincentHubbard:
Hi, I might try to look at the files in cygwin.

I'll look at it in the morning ummm, forgeting exactly ummm seems super
interesting or code for not code.

How are you?  I don't work for Google or anything.

Michael

gennadiycivil:
Thank you very much for this report. The best way to approach this would be to create a proper PR and submit it for consideration.

PR #1321, opened 2017-11-02
Fix potential NULL dereference and silence clang_analyzer

hazelnusse:
Address the static analysis warnings identified by clang-check presented in #1319


hazelnusse:
Hello?

googletest/src/gtest.cc
```diff
@@ -4630,7 +4634,10 @@ void UnitTest::AddTestPartResult(
       // from removing. We use this rather than abort() or __builtin_trap() for
       // portability: Symbian doesn't implement abort() well, and some debuggers
       // don't correctly trap abort().
+
+#ifndef __clang_analyzer__
```
gennadiycivil: ... and if __clang_analyzer__ is defined ?
In general this would be the first time we would try to placate a static analysis tool. Not enough reason to make the change here

gennadiycivil:
@hazelnusse  Sincere apologies for the long wait. Please address comments

gennadiycivil:
Looks abandoned

googletest/src/gtest.cc
hazelnusse: @gennadiycivil If __clang_analyzer__ is defined, you will avoid seeing a warning about undefined behavior that is being intentionally invoked as a workaround for deficiencies of the symbian platform. A better thing to do would be to invoke abort() rather than invoke undefined behavior. I'm happy to do that, but I suspect this will break things downstream eventually (perhaps they should fix abort())?

PR #1324, opened 2017-11-07
Fix issue #826 by treating MinGW as "non-Windows" when determining colored output

### LINKED ISSUE #826: Coloured output under MinGW/MSYS2/MinTTY is broken
Even if `--gtest_color=yes` is specified, no colour is displayed when running under msys2 (MinTTY; `$TERM==xterm`).

Correct me if I'm wrong, but a brief search indicates that gtest always assumes that on Windows (but not cygwin?), it's running under a standard Windows console (i.e. cmd.exe) and hence uses WinAPI functions (`SetConsoleTextAttribute`) to change text colours.

However, this won't work if it's run under a terminal emulator such as mintty, which expects ANSI escape codes.

whame:
It seems like MinGW's `isatty()` always returns 0, one can therefore not use `--gtest_color=auto`. However,  `--gtest_color=yes` works with this patch and thus resolves #826.

### DIFF
--- a/googletest/src/gtest.cc
+++ b/googletest/src/gtest.cc
@@ -2884,7 +2884,7 @@ enum GTestColor {
 };
 
 #if GTEST_OS_WINDOWS && !GTEST_OS_WINDOWS_MOBILE && \
-    !GTEST_OS_WINDOWS_PHONE && !GTEST_OS_WINDOWS_RT
+    !GTEST_OS_WINDOWS_PHONE && !GTEST_OS_WINDOWS_RT && !GTEST_OS_WINDOWS_MINGW
 
 // Returns the character attribute for the given color.
 WORD GetColorAttribute(GTestColor color) {
@@ -2943,7 +2943,7 @@ bool ShouldUseColor(bool stdout_is_tty) {
   const char* const gtest_color = GTEST_FLAG(color).c_str();
 
   if (String::CaseInsensitiveCStringEquals(gtest_color, "auto")) {
-#if GTEST_OS_WINDOWS
+#if GTEST_OS_WINDOWS && !GTEST_OS_WINDOWS_MINGW
     // On Windows the TERM variable is usually not set, but the
     // console there does support colors.
     return stdout_is_tty;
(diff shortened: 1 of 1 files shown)


whame:
I signed it!

gennadiycivil:
Could you please provide test output "before" and "after"

whame:
Compiling and running `sample1_unittest` with `auto` (the default `gtest_color` option) and `yes` before:

before

Same thing after:

after

I hope that this was what you were asking for!

jtanx:
Wouldn't this break coloured output if you ran a mingw binary under a 'real' console (cmd.exe)?

whame:
Although Windows 10 now seems to support ANSI color codes, I agree with you. Something like `is_cygpty` would probably be a better design choice.

For now, this fix satisfies my needs and hopefully other MinGW users as well.

berryzplus:
i think, this pull-request is wrong.

The MinGW-tools provide unix-tools for Windows without any unix environmental things(includes a TTY).
If you want to run your gtest module on a TTY(like a `Git Bash`), you should link your gtest module with cygwin or msys.

this pull-request changed our gtest module to depend on a TTY. I think, something is wrong.

tanzislam:
@berryzplus @jtanx I'm fixing with #3049.

@whame, regarding:
> It seems like MinGW's `isatty()` always returns 0

`isatty()` returns zero when running under MinTTY, because this program directly reads our stdout/stderr with a pipe (similar to a shell pipeline) in order to receive ANSI escape sequences for its own rendering. `isatty()` would return non-zero when stdout/stderr is connected to the "real" Windows Console ([`conhost.exe`/`condrv.sys`](https://devblogs.microsoft.com/commandline/windows-command-line-inside-the-windows-console/#so-whats-inside-the-windows-console)). As that article explains, most other 3rd-party consoles (before ConPTY was introduced in Windows 10 1809) work by hiding and screen-scraping the Windows Console window, which is less efficient but maintains the truthfulness of the `isatty()` function to the application.

PR #1332, opened 2017-11-27
Resolve compiler warnings when using -Wundef

evelikov:
GCC and Clang provide a compiler flag Wundef for a few years now.

The flag is used to highlight cases where a macro is evaluated (via #if XXX) while the macro itself is not defined. This can be used to catch typos and missing header inclusions.

This pull request addresses the warnings.


evelikov:
Something has gone horribly wrong while forward porting these. Closing for now - will check that CI is happy before resubmitting.

PR #1333, opened 2017-11-27
Use system includes

sam-lunt:
Using system includes for the CMake `INTERFACE_INCLUDE_DIRECTORIES` property silences some warnings that cannot be silenced otherwise. For example, when compiling with `-Wpedantic`, `INSTANTIATE_TEST_CASE_P` generates a warning `ISO C++11 requires at least one argument for the "..." in a variadic macro`. I also had some other warnings about signed-unsigned comparison when using `{ASSERT,EXPECT}_EQ` with an integral constant as the second parameter.

This also conforms to the behavior when `find_package(GTest)` is used, which will facilitate changing from a system-wide install to building googletest as part of the project (as the FAQ recommends)

### DIFF
--- a/googlemock/CMakeLists.txt
+++ b/googlemock/CMakeLists.txt
@@ -101,8 +101,8 @@ cxx_library(gmock_main
 # to the targets for when we are part of a parent build (ie being pulled
 # in via add_subdirectory() rather than being a standalone build).
 if (DEFINED CMAKE_VERSION AND NOT "${CMAKE_VERSION}" VERSION_LESS "2.8.11")
-  target_include_directories(gmock      INTERFACE "${gmock_SOURCE_DIR}/include")
-  target_include_directories(gmock_main INTERFACE "${gmock_SOURCE_DIR}/include")
+  target_include_directories(gmock      SYSTEM INTERFACE "${gmock_SOURCE_DIR}/include")
+  target_include_directories(gmock_main SYSTEM INTERFACE "${gmock_SOURCE_DIR}/include")
 endif()
 
 ########################################################################
--- a/googletest/CMakeLists.txt
+++ b/googletest/CMakeLists.txt
@@ -103,8 +103,8 @@ target_link_libraries(gtest_main gtest)
 # to the targets for when we are part of a parent build (ie being pulled
 # in via add_subdirectory() rather than being a standalone build).
 if (DEFINED CMAKE_VERSION AND NOT "${CMAKE_VERSION}" VERSION_LESS "2.8.11")
-  target_include_directories(gtest      INTERFACE "${gtest_SOURCE_DIR}/include")
-  target_include_directories(gtest_main INTERFACE "${gtest_SOURCE_DIR}/include")
+  target_include_directories(gtest      SYSTEM INTERFACE "${gtest_SOURCE_DIR}/include")
+  target_include_directories(gtest_main SYSTEM INTERFACE "${gtest_SOURCE_DIR}/include")
 endif()
 
 ########################################################################


sam-lunt:
I signed it!

gennadiycivil:
@sam-lunt . could you please provide test output / evidence before and after the proposed changes

sam-lunt:
Minimal complete example:

`CMakeLists.txt`:
``` cmake
cmake_minimum_required(VERSION 2.8.12)                                                                                                                                                                            
                                                                                                                                                                                                                  
set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -pedantic")                                                                                                                                                    
                                                                                                                                                                                                                  
add_subdirectory(googletest)                                                                                                                                                                                      
                                                                                                                                                                                                                  
add_executable(foo_test test.cpp)                                                                                                                                                                                 
target_link_libraries(foo_test gtest_main)
```

`test.cpp`:
``` cpp
#include <gtest/gtest.h>                                                                                                                                                                                          

class FooTest : public ::testing::TestWithParam<int>
{};
                                                                                                                                                                                                                  
TEST_P(FooTest, IsNonZero)
{                                                                                                                                                                                                                 
    EXPECT_NE(GetParam(), 0);
}
                                                                                                                                                                                                                  
INSTANTIATE_TEST_CASE_P(Numbers, FooTest, ::testing::Values(1, 2, 3));
```

With master, the following warnings are generated:
using g++ 6.2.1:
```
test.cpp:11:69: warning: ISO C++11 requires at least one argument for the "..." in a variadic macro
 INSTANTIATE_TEST_CASE_P(Numbers, FooTest, ::testing::Values(1, 2, 3));
                                                                     ^
```
using g++ 4.8.5:
```
test.cpp:11:69: warning: ISO C99 requires rest arguments to be used [enabled by default]
 INSTANTIATE_TEST_CASE_P(Numbers, FooTest, ::testing::Values(1, 2, 3));
                                                                     ^
```
With the change proposed here, there are no warnings generated with either version of g++.

Unfortunately, clang 5.0.0  generates this warning in both cases:
```
test.cpp:11:69: warning: must specify at least one argument for '...' parameter of variadic macro [-Wgnu-zero-variadic-macro-arguments]
INSTANTIATE_TEST_CASE_P(Numbers, FooTest, ::testing::Values(1, 2, 3));
                                                                    ^
googletest/include/gtest/gtest-param-test.h:1421:10: note: macro 'INSTANTIATE_TEST_CASE_P' defined here
# define INSTANTIATE_TEST_CASE_P(prefix, test_case_name, generator, ...) \
         ^
```

Based on the warning name (`-Wgnu-zero-variadic-macro-arguments`), I'm guessing that clang is attempting to emulate the behavior of gcc and the warning might be considered incorrect when the files are included using `-isystem` instead of `-I`

grizag:
CMake 2.8.11 does not support `SYSTEM `option of `target_include_directories()` command - see [CMake Version Compatibility Matrix](https://cmake.org/Wiki/CMake_Version_Compatibility_Matrix/Commands#cite_note-76).

So that build fails with cmake 2.8.11:
```
CMake Error at .../googletest-src/googletest/CMakeLists.txt:108 (target_include_directories):
  target_include_directories called with invalid arguments

CMake Error at .../googletest/CMakeLists.txt:109 (target_include_directories):
  target_include_directories called with invalid arguments

CMake Error at .../googletest-src/googlemock/CMakeLists.txt:113 (target_include_directories):
  target_include_directories called with invalid arguments

CMake Error at .../googlemock/CMakeLists.txt:114 (target_include_directories):
  target_include_directories called with invalid arguments
```
Please advance VERSION_LESS condition to "2.8.12".

grizag:
It appears that advancing version to 2.8.12 solves gtest build problem but brakes user configuration later on. It leads to undefined reference during test build with cmake 2.8.11:
```
undefined reference to `testing::internal::MakeAndRegisterTestInfo(char const*, char const*, char const*, char const*, void const*, void (*)(), void (*)(), testing::internal::TestFactoryBase*)'
```

PR #1335, opened 2017-11-28
Resolve compiler warnings when using -Wundef, take 2

evelikov:
GCC and Clang provide a compiler flag Wundef for a few years now.

The flag is used to highlight cases where a macro is evaluated (via #if XXX) while the macro itself is not defined. This can be used to catch typos and missing header inclusions.

This pull request addresses the warnings.


evelikov:
Humble fortnightly ping?

evelikov:
Is there anything I can help to get this moving forward?

evelikov:
It's been 2 months without a single comment - humble ping?
Can anyone take a look letting me know their concerns?

Thanks

evelikov:
@gennadiycivil can you please take a look?

gennadiycivil:
@evelikov thank you for this contrubution. Please explain a bit more, so you tried to compile with -Wundef , got warnings and then try to address these warnings?
Is my understanding correct?

1ace:
Yes. We use gtest in [Mesa](https://mesa3d.org), and would like to turn on the `-Wundef` warning as it is very useful to catch typos, but we can't do that currently as gtest produces huge amounts of warnings, hiding any useful information.

gennadiycivil:
@evelikov In general yes, we should work to get googletest cleanly compiling with Wundef. However in this PR I see that you chose to remove some macros and define others as 0 - lets be consistent . Please don't remove any macros but rather treat all of them the same .

googletest/include/gtest/internal/gtest-port.h
```diff
@@ -434,10 +434,7 @@ typedef struct _RTL_CRITICAL_SECTION GTEST_CRITICAL_SECTION;
 # endif
 #endif
 
-#if GTEST_USES_PCRE
```
gennadiycivil: Why are we removing GTEST_USES_PCRE? We should treat it just like any other currently undefined macros.

gennadiycivil:
In addition, please add a test that would exercise this compilation mode so we can prove that the changes work as intended. I would like to see the test output before ( failing ) and after ( succeeding )

gennadiycivil:
* consider this change and once it is of acceptable quality ... merge it
Thanks
G

gennadiycivil:
@evelikov  More importantly - this solution would only work if your project ( Mesa ) ONLY uses #if and not #ifdef ... consider the implications  for other existing projects using gogletest -- this proposed solution would break those users who happens to be using googletest in other projects. There are situations where people would write #ifdef when they *really* meant #if .

So, while the idea is appealing, the proposed implementation is unacceptable unfortunately.

evelikov:
First of all, thanks for having a look @gennadiycivil. I was about to give up after 2 months of total silence. Closing this w/o a single comment is a bit [hn]asty. Serves me right for taking a week off ;-(

Back on topic: if you look closely nearly all the instances addressed are internal-only macros. And as said in the documentation - "Code outside Google Test MUST NOT define them.". The user provided ones are also safe, as described below.

Going through the list:
 - GTEST_OS_* - internal
 - GTEST_USES_*_RE  - internal
 - GTEST_USES_PCRE - undocumented, unused, likely internal
 - GTEST_HAS_*_ - user, safe - user will provide a value and we'll set to 0 otherwise
 - GTEST_HAS_DOWNCAST_ - undocumented, unused, likely internal
 - GTEST_DONT_DEFINE_* - user, safe, explicitly documented to set to 1 in order to omit $FOO()

gennadiycivil:
@evelikov Let me see if I understand your logic, I am going to re-state it in a different way and tell me if I am correct
Goal - compile with -Wundef and not have noisy warnings
Solution: Find the undefined macros and define them to 0
Note: The macros addressed are all internal, or undocumented , or appear to be unused. Note the emphasis on **appear to be**. Moreover, the documentation says "dont use them, dont define them, they are only internal". So by this logic, we can #define them to be 0 and if some project happens to use them despite explicitly told not to do so, well that is unfortunate but not really a problem.

Did I get this right? Please correct me if not

evelikov:
Precisely. It anyone is using the macros, this series will flag up existing bugs in their codebase. The documentation is pretty clear about that, so one should really fix them ASAP.

Orthogonal to that, about the used bits:
I would gladly double-check the codebase for mistakes on my end. Yet if a undocumented macro is set/used by then it is again in a user bug. We can add an small check to make it more obvious, say:
```
#ifdef GTEST_FOO
#error GTEST_FOO is an internal, undocumented macro. Your code must not rely upon it.
#endif
```

I can see why you are might feel uneasy about the series, but in all honesty googletest cannot guard against all user mistakes. If anything the series helps people make less mistakes ;-)

evelikov:
Please don't be alarmed that projects may misuse gtest. Hundreds if not thousands of projects in the wild had this problem and have sorted it out. One of the simplest ways is to:
To inform:
 - add a bunch of warnings - s/error/warn/ in the example plus a few others
 - roll a release with the warnings and a clear note - that future versions will break for misbehaving users

For nice projects:
 - potentially guard the warnings behinds a HIDE_INTERNAL_SYMBOLS_FOO
Hence projects that are doing the right thing can simply hide the annoying warnings ;-)

That said - I'm currently busy with other work. Will re-spin the patches as soon as I find some time.

evelikov:
Updated the series done. Changes:
 - error if the user provides invalid (!= 1) value for GTEST_DONT_DEFINE_*
 - error if the user sets the undocumented/internal GTEST_HAS_DOWNCAST_. was: remove the macro
 - error if the user sets the undocumented/internal GTEST_USES_PCRE. was: remove the macro
 - remove unreachable GTEST_USES_PCRE hunks
 - error if the user sets the internal GTEST_OS_*

Note: if GTEST_HAS_DOWNCAST_ or GTEST_USES_PCRE is used internally in Google please provide some documentation and example users.

I've opted to keep these as error instead of warnings for a few reasons
 - no standard way to handling MSVC vs GCC #warning
 - people can easily suppress/ignore the warnings

evelikov:
@gennadiycivil in case you've missed it - updated series is available.
Please let me know what I can do to get this moving again.

evelikov:
Another ping? @gennadiycivil I'm open to any suggestions - even if that involves pulling the top X github users of gtest and testing them against the branch.

evelikov:
@gennadiycivil any comments? All of your previous feedback should be addressed, plus I'd gladly perform extra testing to ensure users don't break with these changes.

evelikov:
@gennadiycivil is there any plan to look at this or you're simply short on time?

gennadiycivil:
@evelikov Thank you for working on this. As I mentioned before not only this seems problematic with existing projects but also not clear what benefits it would provide beyond your usage.
There are much higher priority work that is going on with googletest, see https://github.com/google/googletest/issues/1467

evelikov:
Thought I mentioned it before, but it seems like I've completely missed it. You want to use -Wundef since it catches genuine bugs as the following:
 - Mistype a macro - FOO vs FOO_ or similar [less obvious] typos.
 - Missing header inclusion - FOO is defined in bar.h
 - Incorrect `#undef FOO` in the compilation unit - say in some header X levels deep in the include chain

The above are fairly common if the codebase is in the millions LOC.

I have alluded previously, that I'm not planning to land this and disappear. I'm happy to look into and address projects that use gtest - both before landing and after this series. A man needs a name, or in this case a repo - I'll gladly check and send patches.

@gennadiycivil would the above make sense?

Issue #776, opened 2016-05-16
Building out of source directory with autotools build fails


meconlen:
I'm using googletest and googlemock inside another autotools project, so I'm preferring the autotools build. When configuring outside the source directory the googlemock configure does not configure googletest and exits with failure. 

This appears to be because of 

AC_CONFIG_SUBDIRS([../googletest])

which doesn't appear valid (though I haven't yet found a definitive source on. There seems to be two options to recover
1. Make googletest a subdirectory of googlemock
2. Manage each configure from the parent directory so that a parent configure.ac generates a configure script that recurses and builds into each project separately.

PR #1341, opened 2017-12-03
Run autoconf from top-level directory.

coryan:
This is part (hopefully all) of the fixes for #776.  The top-level
configure.ac configures googletest first and then googlemock.
With these changes it is possible to embed googletest into another
project that uses autoconf.  For an example (though it is WIP), see
the commits (and soon PR) referenced from google/protobuf#236.

For example:
https://github.com/coryan/protobuf/tree/fix-issue-236-gmock-as-submodule


gennadiycivil:
@coryan , could you please provide before/after test evidence that demonstrates how the issue is fixed. Thank you again for this contribution

coryan:
PTAL.  With the latest commit I introduce a Travis build that works in this branch.  As shown [here](https://travis-ci.org/coryan/googletest/builds/321059480) the equivalent commit in `master` breaks.

.travis.yml
```diff
@@ -24,10 +24,11 @@ matrix:
       script: ./ci/build-linux-bazel.sh
     - os: linux
       compiler: gcc
-      env: BUILD_TYPE=Debug VERBOSE=1
+      install: ./ci/install-linux.sh && ./ci/log-config.sh
+      script: ./ci/build-linux-autotools.sh
     - os: linux
       compiler: gcc
-      env: BUILD_TYPE=Release VERBOSE=1 CXX_FLAGS=-std=c++11
+      env: BUILD_TYPE=Debug VERBOSE=1 CXX_FLAGS=-std=c++11
```
gennadiycivil: @coryan not sure why we are changing Release to Debug and how it is related to the running of automake from the top level dir?

googletest/Makefile.am
```diff
@@ -120,16 +121,16 @@ EXTRA_DIST += \
...
-  msvc/gtest-md.vcproj \
-  msvc/gtest.sln \
-  msvc/gtest.vcproj \
-  msvc/gtest_main-md.vcproj \
-  msvc/gtest_main.vcproj \
-  msvc/gtest_prod_test-md.vcproj \
-  msvc/gtest_prod_test.vcproj \
-  msvc/gtest_unittest-md.vcproj \
-  msvc/gtest_unittest.vcproj
+  msvc/2010/gtest-md.sln \
```
gennadiycivil: I would avoid adding vusual studion 2010 (!) ... why?

.travis.yml
coryan: That is just to keep the number of builds under control.  The autotools build already builds optimized Linux (Ubuntu) with g++, why have another build do the exact same thing?

googletest/Makefile.am
coryan: The files got moved from `msvc/*` to `msvc/2010/*` at some point, but `Makefile.am` was not changed at the same time.  If you think it would be best to remove the files altogether, then sure, but shouldn't that be a separate change?

.travis.yml
coryan: Happy to revert this change if you prefer.

gennadiycivil: No problem, we can leave them this way, thank you

googletest/Makefile.am
gennadiycivil: Since we are making a change, I just want to understand why. Could you please take a look and suggest why this happened and more importantly is this the general enough way to make this change for all googletest users?

gennadiycivil: @coryan

coryan: > Since we are making a change, I just want to understand why.
> Could you please take a look and suggest why this happened

**TL;DR;** since there was no test for automake builds, nothing detected the breakage.

Let's play code archeologist:

The last time those lines in `Makefile.am` were touched was around 9 years ago:

https://github.com/google/googletest/blame/master/googletest/Makefile.am

At that time the `msvc/*` directory was full of files:

https://github.com/google/googletest/tree/bcaf6f542fac67549528027c24e530e0e89f700b/msvc

The files were removed in this commit:

https://github.com/google/googletest/commit/45287f3dffa494d612c4ba94199aeaa40c99571f

Which I think you merged here:

https://github.com/google/googletest/commit/f25921565bbfb16b02919165669734a5bea7ee98


> and more importantly is this the general enough way to make this change for all
> googletest users?

Well, those particular lines in `Makefile.am` are simply there to include the MSVC files in any tarballs created for a "distribution".  Whether you want to continue to support tarballs or not is a different question.  Right now automake insists on creating the tarball and testing the tarball.  I think if you want to drop support for tarballs that should be a separate change.  Or you may want to drop the MSVC files altogether, that is an option too, but those should be separate changes from "getting automake to work again".

gennadiycivil: makes sense!

PR #1311, opened 2017-10-27
fix for VS2017 Preview deprecation of ::tr1::tuple

bryanzim:
change static_cast to ImplicitCast_ for consitency
fixes for building with path names containing spaces

PR #1343, opened 2017-12-06
Remove C4996 warning in VS2017, version 15.5

wmamrak:
Prevent `warning C4996: 'std::tr1': warning STL4002: The non-Standard std::tr1 namespace and TR1-only machinery are deprecated and will be REMOVED.

### DIFF
--- a/googletest/include/gtest/internal/gtest-port.h
+++ b/googletest/include/gtest/internal/gtest-port.h
@@ -643,8 +643,11 @@ typedef struct _RTL_CRITICAL_SECTION GTEST_CRITICAL_SECTION;
 // STLport, provided with the Android NDK, has neither <tr1/tuple> or <tuple>.
 #  define GTEST_HAS_TR1_TUPLE 0
 # else
+// Prevent `warning C4996: 'std::tr1': warning STL4002: The non-Standard std::tr1 namespace and TR1-only machinery are deprecated and will be REMOVED.`
+#  if defined(_MSC_VER) && (_MSC_VER < 1912)
 // The user didn't tell us not to do it, so we assume it's OK.
-#  define GTEST_HAS_TR1_TUPLE 1
+#   define GTEST_HAS_TR1_TUPLE 1
+#  endif
 # endif
 #endif  // GTEST_HAS_TR1_TUPLE
 


blizzard4591:
Please take a look at https://github.com/google/googletest/pull/1311 which is also (among others) trying to fix this. The other PR handles this differently and adds more comments - Maybe both can be improved/merged (at least with respect to TR1)?

Also: The chance of this being fixed could be higher if it only touches one aspect (i.e. the TR1 issue) - so polishing your PR with details from the other one could be beneficial.

wmamrak:
I assumed the head is free of any warnings and didn't check issues against that.
I admit the comments in both CMakeLists.txt files are useful and I missed them during my (very brief) examination of the source code. What I did realize, however, is there are at least three macros affecting the compilation wrt to tuple. The change proposed in 1311 is a bit more precise, because it turns out _MSC_VER 1910 does already deprecate tr1::tuple. Why hasn't it been approved? Solely because noone did split it into three separate commits?

googletest/include/gtest/internal/gtest-port.h
```diff
@@ -643,8 +643,11 @@ typedef struct _RTL_CRITICAL_SECTION GTEST_CRITICAL_SECTION;
 // STLport, provided with the Android NDK, has neither <tr1/tuple> or <tuple>.
 #  define GTEST_HAS_TR1_TUPLE 0
 # else
+// Prevent `warning C4996: 'std::tr1': warning STL4002: The non-Standard std::tr1 namespace and TR1-only machinery are deprecated and will be REMOVED.`
+#  if defined(_MSC_VER) && (_MSC_VER < 1912)
```
KindDragon: Maybe in `#else` better add `#define GTEST_HAS_TR1_TUPLE 0`
``` cpp
# elif defined(_MSC_VER) && (_MSC_VER >= 1912)
// Prevent `warning C4996: 'std::tr1': warning STL4002: The non-Standard std::tr1 namespace and TR1-only machinery are deprecated and will be REMOVED.`
#  define GTEST_HAS_TR1_TUPLE 0
# else
// The user didn't tell us not to do it, so we assume it's OK.
#  define GTEST_HAS_TR1_TUPLE 1
# endif
```

wmamrak: Yes, I missed it, according to the description it should be set.

PR #1347, opened 2017-12-07
Fix value pointed to by `_NSGetArgc()` on macOS.

### LINKED ISSUE #1346: Stripping of googletest arguments from argv can lead to crashes on macOS.
In [src/gtest.cc#L5326-L5341](https://github.com/google/googletest/blob/a868e618c0607259c63f37d948b72586a13922ff/googletest/src/gtest.cc#L5326-L5341), if an argument is a googletest argument, it is parsed and each successive element of `argv` is shifted left, and `argc` is decremented.

The signature of `main` takes `argc` by copy, but `argv` by pointer. However, on macOS, the functions `_NSGetArgc` and `_NSGetArgv` from `<crt_externs.h>` both return pointers (to `int` and `char**` respectively). The pointer returned by `_NSGetArgv` points to the same block of memory as the `argv` argument passed to main, but the integer pointed to by `_NSGetArgc` is of course not modified by modifying the `int` passed by value as `argc` to `main`.

If library code then tries to inspect the command line arguments via the pointers returned by `_NSGetArgc` and `_NSGetArgv` (such as [this code in Chromium](https://searchcode.com/codesearch/view/86186786/#l-54)), it will see the modified `argv` but `argc` is unchanged. Therefore, it can try to dereference `NULL`, based on the (not unreasonable) assumption that `argc` is a correct count of the elements in `argv`. This can lead to crashes of the entire test executable.

I'm happy to submit a pull request which corrects this behaviour by setting the `int` pointed to by `_NSGetArgc` to the same value as `argc` at the end of the loop in `ParseGoogleTestFlagsOnlyImpl`, only on macOS, and perhaps only if `argv` is equal to `*_NSGetArgv()`, to prevent oddities if a user has constructed their own `argc` and `argv`.

Burgch:
Fixes #1346


Burgch:
I signed it!

gennadiycivil:
Apologies for not getting to this sooner and thank you for your contribution. As part of release 1.8.x preparation I am closing all older PRs that have conflicts and/or failed CI. Please resolve conflicts.

Burgch:
Resolved, thanks.

Burgch:
See issue #1346 mentioned in the PR description for the reasoning. Suggestions for how / where to add an appropriate test are welcome!

PR #1348, opened 2017-12-07
Remove C4996 warning in VS2017

wmamrak:
Prevent `warning C4996: 'std::tr1': warning STL4002: The non-Standard std::tr1 namespace and TR1-only machinery are deprecated and will be REMOVED.
This commit originates from #1311

### DIFF
--- a/googlemock/CMakeLists.txt
+++ b/googlemock/CMakeLists.txt
@@ -74,6 +74,8 @@ include_directories("${gmock_SOURCE_DIR}/include"
 # <= VS 2010  <= 10        <= 1600         Use Google Tests's own tuple.
 # VS 2012     11           1700            std::tr1::tuple + _VARIADIC_MAX=10
 # VS 2013     12           1800            std::tr1::tuple
+# VS 2015     14           1900            std::tuple
+# VS 2017     15           >= 1910         std::tuple
 if (MSVC AND MSVC_VERSION EQUAL 1700)
   add_definitions(/D _VARIADIC_MAX=10)
 endif()
--- a/googletest/CMakeLists.txt
+++ b/googletest/CMakeLists.txt
@@ -83,6 +83,8 @@ include_directories(
 # <= VS 2010  <= 10        <= 1600         Use Google Tests's own tuple.
 # VS 2012     11           1700            std::tr1::tuple + _VARIADIC_MAX=10
 # VS 2013     12           1800            std::tr1::tuple
+# VS 2015     14           1900            std::tuple
+# VS 2017     15           >= 1910         std::tuple
 if (MSVC AND MSVC_VERSION EQUAL 1700)
   add_definitions(/D _VARIADIC_MAX=10)
 endif()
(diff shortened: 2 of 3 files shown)


KindDragon:
@gennadiycivil thank you

davidben:
FYI, this seems to have broken `testing::Combine` on VS2017. `GTEST_HAS_COMBINE`'s definition checks for `GTEST_HAS_TR1_TUPLE`, but it actually just needs `testing::tuple` to work, as far as I can tell. To be honest, I'm not sure what the right check is here as there appear to be a ton of different codepaths here. (`GTEST_ENV_HAS_STD_TUPLE_` and `GTEST_HAS_STD_TUPLE_`, some code to put `<tuple>` into `std::tr1`, but there's also `GTEST_TUPLE_NAMESPACE_`.)

Replacing the `GTEST_HAS_TR1_TUPLE` with `defined(GTEST_TUPLE_NAMESPACE_)` resolves it for me, but I've no clue if that's the actual correct fix. There's probably some cleanup to be done around the tuple stuff here.

@gennadiycivil, thoughts?

davidben:
Poking a little further, it looks like `GTEST_ENV_*` are part of auto-defining `GTEST_USE_OWN_TR1_TUPLE`, if the caller didn't define it themselves. If the caller defined it, then neither gets defined. That suggests that the use of `GTEST_ENV_HAS_STD_TUPLE_` later one isn't quite right? This was introduced by 2147489625ea8071ca560462f19b1ceb8940a229. I guess back then GTest really needed `tuple` to live at `std::tr1::tuple`. I suspect by now you all don't need that anymore? (Later on, ccf8e33bc59a26745753d494b2535a5f0a97acc5 introduced `GTEST_HAS_STD_TUPLE_` and `5df87d70b64dd8080ab9e3f1b0250e74715e2a60` added a macro for the namespace.)

Then there's `GTEST_USE_OWN_TR1_TUPLE` which pulls in GTest's own `tuple` implementation. That puts itself into `std::tr1::tuple`, but probably `testing::tuple` would be more prudent now? But that's odd because if one sets `GTEST_USE_OWN_TR1_TUPLE` and zeros `GTEST_HAS_TR1_TUPLE`, it seems `GTEST_TUPLE_NAMESPACE_` never gets defined and we never put anything into `testing::tuple`? Is it possible that mode broke and no one noticed because everyone has some kind of tuple now?

I think this means `defined(GTEST_TUPLE_NAMESPACE_)` is the right fix, but there may be something to be said for unwinding some of the other bits.

davidben:
Alternatively, elsewhere, things seem to check `#if GTEST_HAS_TR1_TUPLE || GTEST_HAS_STD_TUPLE_`. I think that would work equivalently? Though I'm still puzzled on how `GTEST_USE_OWN_TR1_TUPLE` still works.

davidben:
Okay, last comment. :-) The reason `GTEST_USE_OWN_TR1_TUPLE` works is its used in tandem with `GTEST_HAS_TR1_TUPLE`. So it's intended that it jam itself into `std::tr1::tuple`. This is potentially an ODR violation, but probably it's simplest to leave it alone until googletest can require C++11 support and unwind all this `std::tr1` mess entirely?

gennadiycivil:
See https://github.com/google/googletest/issues/1366. We are working towards requiring C++11 support.

davidben:
Oh, awesome! Is the master branch slated to become 1.8.1 or 1.9.0? I.e. can new changes assume C++11 support, or not quite yet? (If they can, https://github.com/google/googletest/pull/1375 can be even more aggressive.)

Issue #1350, opened 2017-12-08
CHECK_NEAR accepts meaninglessly small epsilon values


randomascii:
Chromium bug crbug.com/786046 was a logic bug in a test. This logic bug sat unnoticed for two years until subtle timing changes turned it into an occasionally flaky test. The problem is that EXPECT_NEAR was being passed an epsilon value of 1.0 when the doubles that were passed in were so huge that adjacent doubles were 512.0 apart. Thus, the epsilon value was pointless and EXPECT_NEAR was silently converted to EXPECT_EQUAL.

It is straightforward for EXPECT_NEAR (actually DoubleNearPredFormat) to detect the smallest practical epsilon value given the input values and then fail if abs_error is smaller than that. With this change the flaky test mentioned in crbug.com/786046 becomes this 100% reliable failure:

../../base/time/time_win_unittest.cc(292): error: The epsilon used to compare expected_microseconds_since_origin and converted_microseconds_since_origin is less than the minimum epsilon in this range, which is 512, where
expected_microseconds_since_origin evaluates to 3.640989196237484e+18,
converted_microseconds_since_origin evaluates to 3.640989196237484e+18, and
1.0 evaluates to 1.
ticks=9223372036854775807, to be converted via logic path: SAFE

gennadiycivil:
I have been cleaning up older and inactive GitHub googletest issues. You may see an issue "closed" if it appears to be inactive/abandoned 
Thank you

randomascii:
Note that this issue was ultimately fixed - using a slightly different technique than originally planned - with this recent commit:

https://github.com/google/googletest/commit/b5687db554a295e697f5d459cf6d3f343d2ca179

PR #1353, opened 2017-12-08
Make DoubleNearPredFormat fail on bad epsilons

randomascii:
DoubleNearPredFormat will happily accept epsilon values (abs_error) that
are so small that they are meaningless. This turns EXPECT_NEAR into a
complicated and non-obvious version of EXPECT_EQUAL.

This change modifies DoubleNearPredFormat) to calculate the smallest
meaningful epsilon value, given the input values, and then fail if
abs_error is smaller than that.

With this change the flaky test mentioned in crbug.com/786046 goes from
a flaky test to this 100% reliable failure:

../../base/time/time_win_unittest.cc(292): error: The epsilon used to
compare expected_microseconds_since_origin and
converted_microseconds_since_origin is less than the minimum epsilon in
this range, which is 512, where
expected_microseconds_since_origin evaluates to 3.640989196237484e+18,
converted_microseconds_since_origin evaluates to 3.640989196237484e+18, and
1.0 evaluates to 1.
ticks=9223372036854775807, to be converted via logic path: SAFE

This fixes issue #1350


randomascii:
This pull request makes EXPECT_NEAR stricter on the values that it will accept. Therefore it may cause new failures in any software which uses googletest. That is both unfortunate (nobody likes causing downstream breaks) and the entire point of this change (bad tests *should* fail).

I have verified that this change detects the known bad test in Chromium's base_unittests and that no other failures in that test binary were found. Once this lands I will test all of Chromium's other tests and make any necessary fixes.

randomascii:
How does this look? The change works now and should detect misuse of EXPECT_NEAR. Merge?

gennadiycivil:
@randomascii 
Thank you for this contribution! In general for googlers the [guidelines](https://github.com/google/googletest/blob/7f8fefabedf2965980585be8c2bff97458f28e0b/googletest/docs/DevGuide.md#submitting-patches) suggest CL - first. 
 Waiting for the corresponding internal CL to land. Once it is submitted please make sure that the CL and this PR are in sync ( there might be changes in the CL in process ) and then we can merge it. 
Thank you!

randomascii:
Acknowledged. I saw the guidance about creating internal CLs first only after I completed this one :-) but I will finish it and then copy it over to here.

gennadiycivil:
Closing for now, just for cleanup. Please come back when CL is ready

PR #1355, opened 2017-12-09
Implement bazel-based builds in Travis.

coryan:
This refactored test build scripts to enable bazel-build scripts in Travis.


coryan:
> What version of bazel will this install? The latest?

Yes.

> Will this be in addition to the existing builds?

This creates one new build: compiles with bazel, on Ubuntu:trusty as configured by Travis, with Clang.

> We should probably rationalize the entire matrix.

I agree.  I thought you wanted just the bazel build first and then we would rationalize the build matrix.  Happy to do both at the same time.

> i am concerned with the build times.

Correct, though Travis rarely has a backlog for Linux builds, and they are parallelized.  So this one build is not going to hurt too much.

> I dont see bazel on macOS

Correct.  I intentionally added only a build on Linux because (a) macOS builds on Travis tend to have a backlog, and (b) adding a bazel build on macOS won't add much in terms of coverage: if the bazel build works on Linux, and *some* build runs on macOS, chances are the bazel build works on macOS.

We can switch one of the existing macOS builds, but I think we should be trying to remove macOS builds in general.  One or two should be enough if the coverage is good elsewhere.

PR #1357, opened 2017-12-11
CMake fixes for paths with spaces

bryanzim:


bryanzim:
+1.  This pull request only involves this issue and no other changes.

bryanzim:
Is there anyone on the GoogleTest team that can take a quick look at this easy fix?

gennadiycivil:
@bryanzim Thank you for your contribution. 
Could you please provide test evidence, before and after the patch to demonstrate the issue being fixed and that the fix is in fact successful

bryanzim:
Thank you for taking a look.  I was unable to replicate the issue for which I made this patch.  The only difference between original submission and now was recreation of the build machine.  I tried on another build machine to replicate to no avail also.

Thus, I have removed the -J patch but I believe that the remaining patch should still be merged as it leads to consistency of quoting between and throughout the two CMake files, and works as intended.

bryanzim:
Thanks for the merge.

PR #1361, opened 2017-12-13
gtest: windows: use performance counters to measure test duration

Seraphime:
In our test runs on Windows we sometimes get negative duration for some tests.
Our analysis shows a combination of NTP synchronization
and using wall time (from _ftime64) to be the root cause of this issue.

Cure this by replacing _ftime64 (which is deprecated anyway)
by QueryPerformance{Counter,Frequency}.

This patch was run a few times on our test setup with no noticable
regression in the timestamps and no negative times (which are always
present without it).

### DIFF
--- a/googletest/src/gtest.cc
+++ b/googletest/src/gtest.cc
@@ -1,4 +1,5 @@
 // Copyright 2005, Google Inc.
+// Copyright (c) 2017, VMware, Inc.
 // All rights reserved.
 //
 // Redistribution and use in source and binary forms, with or without
@@ -824,17 +825,19 @@ TimeInMillis GetTimeInMillis() {
   }
   return 0;
 #elif GTEST_OS_WINDOWS && !GTEST_HAS_GETTIMEOFDAY_
-  __timeb64 now;
+  static LARGE_INTEGER sPerfFrequency;
+  LARGE_INTEGER perfCounter;
 
-  // MSVC 8 deprecates _ftime64(), so we want to suppress warning 4996
-  // (deprecated function) there.
-  // TODO(kenton@google.com): Use GetTickCount()?  Or use
-  //   SystemTimeToFileTime()
-  GTEST_DISABLE_MSC_WARNINGS_PUSH_(4996)
-  _ftime64(&now);
-  GTEST_DISABLE_MSC_WARNINGS_POP_()
+  if (!sPerfFrequency.QuadPart) {
+    assert(QueryPerformanceFrequency(&sPerfFrequency));
+  }
 
-  return static_cast<TimeInMillis>(now.time) * 1000 + now.millitm;
+  assert(QueryPerformanceCounter(&perfCounter));
+  // Do the division in two times to avoid integer division
+  // cutting off valuable milliseconds.
+  const TimeInMillis secs = perfCounter.QuadPart / sPerfFrequency.QuadPart;
+  const TimeInMillis rem = perfCounter.QuadPart % sPerfFrequency.QuadPart;
+  return secs * 1000 + (rem * 1000) / sPerfFrequency.QuadPart;
 #elif GTEST_HAS_GETTIMEOFDAY_
   struct timeval now;
   gettimeofday(&now, NULL);


googletest/src/gtest.cc
```diff
@@ -824,17 +824,19 @@ TimeInMillis GetTimeInMillis() {
   }
   return 0;
 #elif GTEST_OS_WINDOWS && !GTEST_HAS_GETTIMEOFDAY_
-  __timeb64 now;
+  static LARGE_INTEGER sPerfFrequency;
```
bryanzim: Note this is not initialized and may undefined behavior.

bryanzim:
In my opinion a better replacement is GetSystemTimeAsFileTime or GetSystemTimePreciseAsFileTime on mondern OS's.  Or if C++11 is available then std::chrono::high_resolution_clock

KindDragon:
Some test failing with this changes

bryanzim:
`  const unsigned long long epochTime(0x19DB1DED53E8000); // 100ns ticks between 1/1/1601 00:00 UTC and 1/1/1970 00:00 UTC

  // get the system time, convert to system time period and convert to number of 100ns ticks since 1/1/1970 00:00 UTC
  FILETIME now;
  GetSystemTimeAsFileTime(&now);
  unsigned long long fileTimeTicks((static_cast<unsigned long long>(now.dwHighDateTime) << (sizeof(DWORD) * 8)) +
                                     static_cast<  unsigned long long>(now.dwLowDateTime) - epochTime);

  return static_cast<TimeInMillis>(fileTimeTicks / 10000);`

Does that work?

Seraphime:
Sorry for not responding, I didn't receive emails notifications.
I'm not sure whether GetSystemTimeAsFileTime will do the job, AFAIU, it doesn't give a monotonic timestamp. I can spin a testround to see if it fixes the issue.
In my initial experiments, I did try a fix by std::chrono::steady_clock, but this can't do as it will break c++98 users.

Will look into error logs when it fires again.

Thanks for the review.

googletest/src/gtest.cc
Seraphime: static variables are zero initialized

bryanzim: LARGE_INTEGER is a union which is not default initialized.

bryanzim:
GetSystemTimeAsFileTime is the basis for std::chrono::high_resolution_clock in VS2015 and GetSystemTimeAsFileTimeEX is the basis in VS2017.  It is monotonic with respect to time zone changes and daylight savings time, since it is UTC.  However it is not monotonic with respect to changing the time/date on PC.

googletest/src/gtest.cc
Seraphime: No it is not default initialiazed. It is zero-initialized. From c++{11,98} standard 6.7.4:

The zero-initialization (8.5) of all block-scope variables with static storage duration (3.7.1) or thread storage duration (3.7.2) is performed before any other initialization takes place

Seraphime:
> GetSystemTimeAsFileTime is the basis for std::chrono::high_resolution_clock in VS2015 and GetSystemTimeAsFileTimeEX is the basis in VS2017. It is monotonic with respect to time zone changes and daylight savings time, since it is UTC. However it is not monotonic with respect to changing the time/date on PC.

Well, the issue we're having is related directly to the clock not being monotonic with respect to changing the time/date. Our machines don't really travel and daylight saving doesn't occur on every test run.
I think the best option for measuring the duration of a test is a monotonic clock.

googletest/src/gtest.cc
bryanzim: The issue with the union is that the standard says 

(6.3) — if T is a (possibly cv-qualified) union type, the object’s first non-static named data member is zeroinitialized
and padding is initialized to zero bits;

Depending on whether the compiler has 64bit support or not it could use a different part of the union which in C++ is called punning and leads to undefined behavior.

bryanzim:
If you are changing the date/time during the google tests then this would lead to undefined behavior.  If you are only changing between tests then the behavior with GetSystem... would work.

Seraphime:
> If  you are changing the date/time during the google tests then this would  lead to undefined behavior
Unless we use monotonic time to measure the duration of the test, which is exactly I'm suggesting.

bryanzim:
The best solution would be 

  #if GTEST_LANG_CXX11
  auto now{ std::chrono::steady_clock::now().time_since_epoch() };
  return std::chrono::duration_cast<std::chrono::milliseconds, TimeInMillis>(now).count();

else something like.  Note lack of union and Microsoft Visual Studio 64 bit integers

  static const __int64 _Freq = _Query_perf_frequency();	// doesn't change after system boot
  const __int64  _Ctr = _Query_perf_counter();
  const __int64 _Whole = (_Ctr / _Freq);
  const __int64 _Part = (_Ctr % _Freq);
  return static_cast< TimeInMillis >(_Whole * 1000 + (_Part * 1000) / _Freq);

gennadiycivil:
@Seraphime Thank you for your contribution. 

I am not convinced that this issue is in fact worth solving for overall googletest user community. 

From what I can see this is a narrow case where if a user is running on Windows and if there is an important test duration to be measured the measurements would go into Windows perf. counters.

I can be convinced though, could you please provide rationale why this is important enough to be included 
Thank you agan

Seraphime:
@gennadiycivil, thanks for taking the time to look at this.

I'm not convinced either that casual users of the framework would benefit from this. But this fix aims at the general correctness of the framework. Measuring duration with a wall clock (even if it is the most accurate clock in the world) is not a good idea. We got bitten by this, and we're really annoyed as sometimes test failure seemed to be related to negative time duration which eventually turned out not to be case, but we lost valuable hours debugging clock issue not related to our clock. As a provider of a test service, I think, the test framework shouldn't make debugging issue harder (but I may be wrong...). And yes - I do agree, that having a clock skew in a test is not the most common event, but that's what you get when you push somewhat the limits.

So please, do tell if this is motivation is not enough, I will drop the PR.

bryanzim:
I wish to contribute my two cents here.  After some thought, it is a common occurrence for machines to synchronize to external time sources using NTP or other methods.  When these synchronizations occur there could be 10ms+ changes in the clock which would give a negative test duration for short tests times.  Whether this occurs for a particular user would be dependent on the length of the tests and frequency of synchronization.  

So I do think this is a relevant issue to fix.  In my opinion the best approach would be to use the C++ 11 chrono library with platform independent monotonic clock if available, i.e. with

`#if GTEST_LANG_CXX11
  auto now{ std::chrono::steady_clock::now().time_since_epoch() };
  return std::chrono::duration_cast<std::chrono::milliseconds, TimeInMillis>(now).count();`

and to use something similar to QueryPerformanceCounters above for Windows OS's that support them.  This would also remove the deprecated _ftime.  For other OS's that don't have C++ 11 support an alternative besides gettimeofday should be used but maybe this isn't as important.

Seraphime:
@bryanzim, thanks for the support  !

Can someone point in me to the right direction ? How is the gtest_output_test_.cc supposed to be run ?

gennadiycivil:
@bryanzim The crux of the thing here is your statement "So I do think this is a relevant issue to fix"
Lets define what "this" is?

So far "this" has been "windows: use performance counters to measure test duration".  What exactly are we trying to achieve?

bryanzim:
@gennadiycivil I think a concrete example may serve purposes.

Suppose we are running a set of tests on Windows/Mac/Linux does not matter, We start a test A that takes 5ms but during execution the system clock is adjusted via a background synchronization service backwards 10ms.  Then with the current code because the clock is non monotonic the reported test duration is -5ms instead of correct duration of 5ms.

gennadiycivil:
@bryanzim I see. Well this is a corner case that we can debate but it is well out of scope for this proposed change because this one is specifically trying to address on Windows using OS specific performance counters.  A bit academic here.

gennadiycivil:
It is actually mostly broken on Windows, please take a look at AppVeyor [results](https://ci.appveyor.com/project/GoogleTestAppVeyor/googletest/build/851)

bryanzim:
You are correct that the pull request here is only for Windows and is not viable in current state.

Seraphime:
I looked at the result, but I can't figure out how to run the test locally. I managed to compile it, but the result is nonsensical.
And yes the PR addresses only the Windows side. Other OSes are as vulnerable, but I've not yet seen negative duration on other OSes or strange time related things.

Ifdefining the whole thing around c++11 chrono features, would fix the issue for the majority of the users I think. The others will need OS specific monotonic time hacks.

gennadiycivil:
I dont see a viable way forward with this PR. Closing and we can come back to this once there is more clarity how to proceed

bryanzim:
@Seraphime.  I have created a tentative cross platform patch for this issue which passes all tests.  It is pull request #1370.

Issue #1368, opened 2017-12-21
Build of unit tests fails with GCC 7.2


stefanosoffia:
When building Google Test regression with GCC 7.2.0, I get the following error:
`error: suggest explicit braces to avoid ambiguous ‘else’ [-Werror=dangling-else]`
on multiple files.

stefanosoffia:
See PR #1369.

PR #1369, opened 2017-12-21
Fix test build issue with gcc 7.2

stefanosoffia:
See Issue #1368

Despite of the compiler error message `suggest explicit braces to avoid ambiguous ‘else’`, the use of braces here could create _false positives_ in the tests. Therefore, consistently with what I see in the files, the solution is to add an `else ;` where needed.

### DIFF
--- a/googletest/test/gtest-death-test_test.cc
+++ b/googletest/test/gtest-death-test_test.cc
@@ -281,6 +281,8 @@ TEST_F(TestForDeathTest, SingleStatement) {
   if (AlwaysFalse())
     // This would fail if executed; this is a compilation test only
     ASSERT_DEATH(return, "");
+  else
+    ;  // NOLINT
 
   if (AlwaysTrue())
     EXPECT_DEATH(_exit(1), "");
@@ -291,6 +293,8 @@ TEST_F(TestForDeathTest, SingleStatement) {
 
   if (AlwaysFalse())
     ASSERT_DEATH(return, "") << "did not die";
+  else
+    ;  // NOLINT
 
   if (AlwaysFalse())
     ;
@@ -1376,6 +1380,8 @@ TEST(ConditionalDeathMacrosSyntaxDeathTest, SingleStatement) {
   if (AlwaysFalse())
     // This would fail if executed; this is a compilation test only
     ASSERT_DEATH_IF_SUPPORTED(return, "");
+  else
+    ;  // NOLINT
 
   if (AlwaysTrue())
     EXPECT_DEATH_IF_SUPPORTED(_exit(1), "");
@@ -1386,6 +1392,8 @@ TEST(ConditionalDeathMacrosSyntaxDeathTest, SingleStatement) {
 
   if (AlwaysFalse())
     ASSERT_DEATH_IF_SUPPORTED(return, "") << "did not die";
+  else
+    ;  // NOLINT
 
   if (AlwaysFalse())
     ;  // NOLINT
--- a/googletest/test/gtest-port_test.cc
+++ b/googletest/test/gtest-port_test.cc
@@ -233,6 +233,8 @@ TEST(GtestCheckSyntaxTest, BehavesLikeASingleStatement) {
   if (AlwaysFalse())
     GTEST_CHECK_(false) << "This should never be executed; "
                            "It's a compilation test only.";
+  else
+    ;  // NOLINT
 
   if (AlwaysTrue())
     GTEST_CHECK_(true);
--- a/googletest/test/gtest_unittest.cc
+++ b/googletest/test/gtest_unittest.cc
@@ -4066,6 +4066,8 @@ TEST(AssertionSyntaxTest, BasicAssertionsBehavesLikeSingleStatement) {
   if (AlwaysFalse())
     ASSERT_TRUE(false) << "This should never be executed; "
                           "It's a compilation test only.";
+  else
+    ;  // NOLINT
 
   if (AlwaysTrue())
     EXPECT_FALSE(false);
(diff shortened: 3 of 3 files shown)


gennadiycivil:
@stefanosoffia  thank you fro your contribution. Please provide reference links to support your statement "consistently with what I see in the files, the solution is to add an else ; where needed."

gennadiycivil:
Please also provide test output that illustrates the problem this PR is fixing and the before/after output

stefanosoffia:
@gennadiycivil, these are the build errors obtained when building `gtest-port_test.cc` with gcc7.2.
```
Scanning dependencies of target gtest_dll_test_                                                                                                                                                                    
[ 49%] Building CXX object googlemock/gtest/CMakeFiles/gtest_dll_test_.dir/test/gtest_all_test.cc.o                                                                                                                
In file included from /home/gip/googletest/googletest/test/gtest_all_test.cc:40:0:                                                                                                          
/home/gip/googletest/googletest/test/gtest-port_test.cc: In member function ‘virtual void testing::internal::GtestCheckSyntaxTest_BehavesLikeASingleStatement_Test::TestBody()’:            
/home/gip/googletest/googletest/test/gtest-port_test.cc:233:6: error: suggest explicit braces to avoid ambiguous ‘else’ [-Werror=dangling-else]                                             
   if (AlwaysFalse())                                                                                                                                                                                              
      ^                                                                                                                                                                                                            
In file included from /home/gip/googletest/googletest/test/gtest_all_test.cc:46:0:                                                                                                          
/home/gip/googletest/googletest/test/gtest_unittest.cc: In member function ‘virtual void {anonymous}::AssertionSyntaxTest_BasicAssertionsBehavesLikeSingleStatement_Test::TestBody()’:      
/home/gip/googletest/googletest/test/gtest_unittest.cc:4066:6: error: suggest explicit braces to avoid ambiguous ‘else’ [-Werror=dangling-else]                                             
   if (AlwaysFalse())                                                                                                                                                                                              
      ^                                                                                                                                                                                                            
/home/gip/googletest/googletest/test/gtest_unittest.cc:4075:6: error: suggest explicit braces to avoid ambiguous ‘else’ [-Werror=dangling-else]                                             
   if (AlwaysFalse())                                                                                                                                                                                              
      ^                                                                                                                                                                                                            
/home/gip/googletest/googletest/test/gtest_unittest.cc: In member function ‘virtual void {anonymous}::AssertionSyntaxTest_ExceptionAssertionsBehavesLikeSingleStatement_Test::TestBody()’:  
/home/gip/googletest/googletest/test/gtest_unittest.cc:4100:6: error: suggest explicit braces to avoid ambiguous ‘else’ [-Werror=dangling-else]                                             
   if (AlwaysFalse())                                                                                                                                                                                              
      ^                                                                                                                                                                                                            
/home/gip/googletest/googletest/test/gtest_unittest.cc:4108:6: error: suggest explicit braces to avoid ambiguous ‘else’ [-Werror=dangling-else]                                             
   if (AlwaysFalse())                                                                                                                                                                                              
      ^                                                                                                                                                                                                            
/home/gip/googletest/googletest/test/gtest_unittest.cc:4116:6: error: suggest explicit braces to avoid ambiguous ‘else’ [-Werror=dangling-else]                                             
   if (AlwaysFalse())                                                                                                                                                                                              
      ^                                                                                                                                                                                                            
```
Please notice the `all warnings being treated as errors`.
The same error is issued also for the other test files included in the commit.

About the fix, consider e.g. this [statement](https://github.com/google/googletest/blob/master/googletest/test/gtest-port_test.cc#L239) from the master branch. It shows an `if-then-else` with an empty `else`. No curly braces are used here to avoid the ambiguity, instead the empty `else` is used. The same pattern occurs over and over in this and other test files. AFAIK, the rationale to avoid braces is that in these cases they might hide potential issues with the macros definitions.

After the fix, all tests are successfully compiled and pass.
```
100% tests passed, 0 tests failed out of 59 
Total Test time (real) =  16.11 sec    
```

I also tried to understand what exactly happened with GCC. It seems that `-Wparentheses` was implied by `-Wall` as far as as [version 4.9](https://gcc.gnu.org/onlinedocs/gcc-4.9.0/gcc/Warning-Options.html), i.e. the one currently used with Travis. Though, for [GCC7](http://gcc.gnu.org/gcc-7/changes.html), _the new -Wdangling-else command-line option has been split out of -Wparentheses and warns about dangling else_.

gennadiycivil:
@stefanosoffia seems that there is a misunderstanding with regards to the code. The [statement](https://github.com/google/googletest/blob/master/googletest/test/gtest-port_test.cc#L239) you are referring to from the master branch ... take a look a[ few lines above](https://github.com/google/googletest/blob/b3a2048beb0f8d6cbb2d5c95e7f639780eff5805/googletest/test/gtest-port_test.cc#L232) - we are specifically testing that all permutations will work  - this is by design. 

[-Werror=dangling-else]  is not something we can support.

stefanosoffia:
@gennadiycivil, thanks for the clarification about the test code. Since this compile error shows up with the default `googletest` flags(*) on GCC7, I propose update them so as to keep this specific warning but not the error (see the updated commit.)

(*) Specifically,  `${cxx_base_flags}` includes `-Wall`, which implies `-Wparentheses`, which in turn implies `-Wdangling-else`.

mbrucher:
Isn't his warning because of the missing braces?
So it would be:
```
if (AlwaysTrue())
{
    GTEST_CHECK_(true);
}
  else
{
    ;  // NOLINT
}

```
It's not about the missing else;, it's about the fact that adding an instruction after the if would lead to a dangling else (if you have an else) or a bug (see the infamous Apple bug about braces).

PR #1371, opened 2017-12-21
static_cast to BiggestInt

kongy:
Fix compilation error in Clang ToT:
  error: implicit conversion from 'const _Complex long double' to 'const internal::BiggestInt' (aka 'const long long') is not permitted in C++


kongy:
After looking into the code more carefully, this is actually a compiler bug. Abandoning this change.

PR #1375, opened 2018-01-02
Clean up now unneccessary std::tr1::tuple undefined behavior.

davidben:
This is just some cleanup. It doesn't add or fix functionality. I also could have done my archaeology wrong, so this warrants some careful testing.

---

Originally, gtest relied on std::tr1::tuple for some features. Later on,
GTEST_USE_OWN_TR1_TUPLE was introduced which provided an alternate
std::tr1::tuple if one was not provided. (This was injected into
std::tr1 which I believe is undefined behavior, but this CL doesn't
attempt to resolve it.)

Later, 2147489625ea8071ca560462f19b1ceb8940a229 introduced some logic to
use std::tuple on platforms which had std::tuple and not
std::tr1::tuple. Since the rest of gtest expected the tuple to live at
std::tr1::tuple, it just injected std::tuple into std::tr1::tuple, which
is I believe also undefined behavior.

Later work (ccf8e33bc59a26745753d494b2535a5f0a97acc5 and
5df87d70b64dd8080ab9e3f1b0250e74715e2a60) removed the need for this
hack. Instead, gtest now no longer depends on tuples existing at
std::tr1::tuple. It includes pretty-printers for std::tuple or
std::tr1::tuple, whichever exists, and exports the one it needs
internally in testing::tuple.

Thus, the hack to inject std::tuple into std::tr1::tuple is no longer
necessary. Remove it. This does have a slight subtlety in that, if
std::tr1::tuple does not exists, but std::tuple does, we should not set
GTEST_USE_OWN_TR1_TUPLE as later code will otherwise include the header
when it shouldn't.

### DIFF
--- a/googletest/include/gtest/internal/gtest-port.h
+++ b/googletest/include/gtest/internal/gtest-port.h
@@ -671,14 +671,9 @@ typedef struct _RTL_CRITICAL_SECTION GTEST_CRITICAL_SECTION;
 #  define GTEST_ENV_HAS_TR1_TUPLE_ 1
 # endif
 
-// C++11 specifies that <tuple> provides std::tuple. Use that if gtest is used
-// in C++11 mode and libstdc++ isn't very old (binaries targeting OS X 10.6
-// can build with clang but need to use gcc4.2's libstdc++).
-# if GTEST_LANG_CXX11 && (!defined(__GLIBCXX__) || __GLIBCXX__ > 20110325)
-#  define GTEST_ENV_HAS_STD_TUPLE_ 1
-# endif
-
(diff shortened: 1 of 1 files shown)


slonopotamus:
Also see #1461

spiridenok:
Are there plans to merge this MR? I also run into compiler issues when compiling gtest with MS VS 2017. Is there any other solution for this problem?

gennadiycivil:
Apologies for not getting to this sooner. As a preparation for 1.8.x release I am cleaning up older PR that are in CI failed status and/or have merge conflicts.
Please submit a new PR if you think this is still important/relevant.

PR #1378, opened 2018-01-03
upstreaming cl 124976692

gennadiycivil:
gtest/src/gtest-internal-inl.h -- the preprocessor was being used as an
ad-hoc header file visibility mechanism. the mechanism feels outdated and quixotic.


sbenzaquen:
Accepted. Merge when tests are green.

PR #1379, opened 2018-01-04
Support JSON output format in addition to XML

petrhosek:
This change allows emitting output in JSON format in addition to the already supported XML format. The implementation as well as the file structure is very similar to the XML one.


gennadiycivil:
Closing as discussed, will do the internal change first and upstream it when ready

PR #1389, opened 2018-01-08
Rename BUILD.bazel to BUILD.

darrengarvey:
Allow using gtest directly from other bazel projects without having to copy
the BUILD.bazel file.

With this change the repo can be used with `http_archive()` and `git_repository()` bazel rules.


darrengarvey:
Travis build seems to have failed due to network errors.

gennadiycivil:
Bazel build has to be build.bazel not to conflict with internal implementations

darrengarvey:
Do you mean internal to google? I don't see anything in this repo that doesn't like this file being called BUILD. Could you clarify a bit please?

The main reason, as mentioned above is that AFAICT there is no way to use googletest as a bazel project without having to copy-paste the BUILD.bazel file into any project that wants to use it. I might be missing something but that seems like a big hindrance.

gennadiycivil:
@darrengarvey Please take a look at [this example](https://github.com/abseil/abseil-cpp/blob/master/WORKSPACE) how to include googletest as dep in a Bazel build

darrengarvey:
Thanks @gennadiycivil. I was just missing the `strip_prefix` and thought it couldn't work. Glad it does. :)

PR #1410, opened 2018-01-18
Check whether _MSC_VER is defined when detecting presence of cxxabi.h…

pcc:
… under libc++.

If _MSC_VER is defined, it means that we are using the Microsoft
ABI, so cxxabi.h (which is associated with the Itanium ABI) will not
be available.


googletest/include/gtest/internal/gtest-port.h
```diff
@@ -980,7 +980,7 @@ using ::std::tuple_size;
 #endif
 
 // _LIBCPP_VERSION is defined by the libc++ library from the LLVM project.
-#if defined(__GLIBCXX__) || defined(_LIBCPP_VERSION)
+#if defined(__GLIBCXX__) || (defined(_LIBCPP_VERSION) && !defined(_MSC_VER))
```
gennadiycivil: Will this work for any MSVC? what about older versions?

pcc: This macro has been around for a long time (since at least Visual Studio 6 ref https://software.intel.com/en-us/forums/intel-c-compiler/topic/309948) so I think we can rely on it.

PR #1423, opened 2018-01-25
Use _CPPUNWIND instead of _HAS_EXCEPTIONS with MSVC.

pcc:
_HAS_EXCEPTIONS is specific to the MSVC STL and defining it to 0 causes
problems with libc++, so libc++ users may leave it undefined. This can
cause GTEST_HAS_EXCEPTIONS to be defined incorrectly if the user has
disabled exceptions via the compiler, which can lead to build errors.

_CPPUNWIND is a builtin macro provided by the compiler so it should
work with both STLs.


gennadiycivil:
@pcc Thank you very much for the contribution. For Googlers, could you please submit a CL first as per contributing guidelines. Once that CL lands we can merge this PR

gennadiycivil:
187072725

pcc:
Submitted 187194353

PR #1426, opened 2018-01-25
Fix test build issue with GCC7.2 (with -Wno-error=)

stefanosoffia:
Restarting from the closed PR #1369 (Issue #1368).

@gennadiycivil, thanks for the clarification about the test code. The compile error described in PR #1369  shows up with the default GoogleTest flags(*) on GCC7. In this PR I propose to fix it by updating the compile flags so as to keep this specific warning but not the error.

(*) Specifically,  GoogleTest `${cxx_base_flags}` includes `-Wall` and `-Werror`. `-Wall` implies `-Wparentheses`, which in turn implies `-Wdangling-else`, which causes the compile error when `-Werror` is also used.

Notice that `-Wdangling-else` is introduced with GCC7. We must do a version check before disabling it, otherwise we would get a compile error with older GCC versions.


googletest/cmake/internal_utils.cmake
```diff
@@ -95,6 +95,9 @@ macro(config_compiler_and_linker)
     set(cxx_no_rtti_flags "-GR-")
   elseif (CMAKE_COMPILER_IS_GNUCXX)
     set(cxx_base_flags "-Wall -Wshadow -Werror")
+    if(NOT CMAKE_CXX_COMPILER_VERSION VERSION_LESS 7.0.0)
```
gennadiycivil: What will this do for clang?

PR #1439, opened 2018-02-03
Add ability to throw from ASSERT

DarthGandalf:
while not losing benefits of EXPECT, and not killing the whole test,
as with --gtest_throw_on_failure.

183822976


gennadiycivil:
@DarthGandalf ... does not compile, could you pls take a look

DarthGandalf:
Fixed. I forgot to update include.

PR #1450, opened 2018-02-10
Fix std::iscntrl use in gtest-printers.cc

pwnall:
ContainsUnprintableControlCodes() in gtest-printers.cc passes a char argument to std::iscntrl. Although its argument is an int, std::iscntrl produces undefined behavior if its argument is not representable as an unsigned char. The standard library on Windows asserts that the argument is an unsigned char, resulting in an assertion crash on debug builds.

Source for my claim: http://en.cppreference.com/w/cpp/string/byte/iscntrl
I ran into this issue because it crashes Chrome's unittests on Windows. I'd be really grateful for a quick review.


pwnall:
I forgot to mention -- my patch is a combination between the current code in ContainsUnprintableControlCodes() and the code in IsValidUTF8(), which is 20 lines down in the same file. I hope this helps with the review!

(Also, I don't expect a response until tomorrow, given the weekend... sorry my earlier comment sounds demanding and ungrateful.)

pwnall:
@gennadiycivil  Thank you very much!

PR #1453, opened 2018-02-13
Guard possibly unused static functions to avoid -Werror=unused-function.

EricWF:
Most GCC versions pass -Wunused-function when -Wall is specified.
This results in build failures on OS X w/ GCC. Specifically in
gtest-death-test.cc.

This patch fixes the warning by correctly guarding the unused
functions inside conditional compilation blocks.

### DIFF
--- a/googletest/src/gtest-death-test.cc
+++ b/googletest/src/gtest-death-test.cc
@@ -983,6 +983,7 @@ static int ExecDeathTestChildMain(void* child_arg) {
 }
 #  endif  // !GTEST_OS_QNX
 
+#if !GTEST_ON_QNX && GTEST_HAS_CLONE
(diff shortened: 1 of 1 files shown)


EricWF:
I think the auto-merge I just did is incorrect. I'll remerge correctly in a bit.

EricWF:
Nevermind. The merge looks correct.

gennadiycivil:
@EricWF , Thank you very much for you contribution. This is addressing the same issue as #1466 , it might make sense to dicsuss with @pwnall

pwnall:
@EricWF Please let me know if #1466 doesn't address your needs. Happy to discuss and submit a CL on your behalf.

gennadiycivil:
I ll close this because #1466 should be addressing it

PR #1464, opened 2018-02-16
Add preprocessor macro for default death test style.

pwnall:
Google Test has recently (02/09/2018) switched the default death test style from "fast" to "threadsafe" in https://github.com/google/googletest/commit/ec7faa943d7817c81ce7bdf71a21ebc9244dc8de

Threadsafe death tests have been used internally for a while, and are proven to be a better default.

However, adopting this better default can be challenging for large projects with a significant investment in custom infrastructure built on top of Google Test.  The same custom infrastructure can make it difficult for large projects to switch back to the old default by passing in `--gtest_death_test_style=fast`.

For this reason, the default switch is considered too disruptive, and this PR reverts it. This PR also introduces the `GTEST_DEFAULT_DEATH_TEST_STYLE` preprocesor macro, which replaces the hard-coded default. The macro can be defined in `gtest/internal/custom/gtest-port.h` by projects that are ready to migrate to thread-safe death tests.

### DIFF
--- a/googletest/include/gtest/internal/gtest-port.h
+++ b/googletest/include/gtest/internal/gtest-port.h
@@ -107,6 +107,12 @@
 //   GTEST_CREATE_SHARED_LIBRARY
 //                            - Define to 1 when compiling Google Test itself
 //                              as a shared library.
+//   GTEST_DEFAULT_DEATH_TEST_STYLE
+//                            - The default value of --gtest_death_test_style.
+//                              The legacy default has been "fast" in the open
+//                              source version since 2008. The recommended value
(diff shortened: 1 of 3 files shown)


pwnall:
Submitting this PR to see if it passes CI for now.

gennadiycivil:
@pwnall Thank you for the contribution. Instead of trying to deduct the platforms where the correct death style might be "fast" , it may make sense to allow the author to overwrite the default with a command line flag. 
WDYT?
Thanks again

pwnall:
@gennadiycivil I have updated this PR and internal CL 185978798

PR #1471, opened 2018-02-22
TEST() arguments are invalid in an example

aleksejkozin:
Both names must be valid C++ identifiers, and they should not contain underscore (`_`)


aleksejkozin:
I signed it!

gennadiycivil:
@aleksejkozin Thank you for this contribution. Could you please clarify why the names should not contain underscores? Reference https://google.github.io/styleguide/cppguide.html#General_Naming_Rules

aleksejkozin:
Hi @gennadiycivil,
The next sentence in the text says
"Both names must be valid C++ identifiers, and they should not contain underscore (`_`)."
It feels like a contradiction, mb I'm wrong

PR #1472, opened 2018-02-23
Removed trailing comma in enum

AndersSundmanTobii:
Trailing commas are allowed since C99 and C++ standards since. It is not allowed in C++ 98. To make gtest  compatible with C++98 this has to go. It causes issues for some less common compilers.

### DIFF
--- a/googletest/include/gtest/internal/gtest-internal.h
+++ b/googletest/include/gtest/internal/gtest-internal.h
@@ -1050,7 +1050,7 @@ class NativeArray {
  private:
   enum {
     kCheckTypeIsNotConstOrAReference = StaticAssertTypeEqHelper<
-        Element, GTEST_REMOVE_REFERENCE_AND_CONST_(Element)>::value,
+        Element, GTEST_REMOVE_REFERENCE_AND_CONST_(Element)>::value
   };
 
   // Initializes this object with a copy of the input.


AndersSundmanTobii:
Travis fail is not related to code change. Is failure in build agent provisioning.

gennadiycivil:
@AndersSundmanTobii please provide more explanation and evidence. Also please note that googletest can not have a goal of supporting all possible compilers in the C++98 mode. The C++ 98 mode will be removed *very soon*

AndersSundmanTobii:
The grammar for enums in C++98 (section 7.2 in the standard) has this:
```
  enum-specifier:
    enum identifier opt { enumerator-list opt }
  enumerator-list:
    enumerator-definition
    enumerator-list , enumerator-definition
```
So technically it's not allowed. I wouldn't nitpick about it if it wasn't for the case that some compilers give errors (that can't be suppressed with flags).

On a side note, I think it would be a big mistake to abandon support for C++98 in googletest since lot's of project still use it (almost everything in the embedded space). Doing that will only create googletest98 fork.

gennadiycivil:
@AndersSundmanTobii , thanks again for the explanation. Could you please provide actual compiler output and the compilers in question

AndersSundmanTobii:
The compiler `xt-xc++` for the (xtensa/cadence/tensilica) cpu family gives the error: `foo.cpp:8: error: comma at end of enumerator list`

I also noticed that gcc actually also complains about this if you compile with the `-pedantic` option `foo.cpp:8:81: warning: comma at end of enumerator list [-Wpedantic]` and we do that for our x86/64 builds.

Unfortunately I can't disclose what other compilers we use since they are specific to CPU's and I can't disclose what CPU's we work with (except xtensa). 

Currently we maintain an inhouse fork of gtest that we patch with this comma removal every time we upgrade to a new version of gtest.

Issue #1476, opened 2018-02-23
signed vs. unsigned wchar_t on ARM


sgraham:
https://github.com/google/googletest/blob/master/googlemock/test/gmock-actions_test.cc#L107

fails to compile on Fuchsia with -Wsign-compare on aarch64:

```
FAILED: obj/third_party/gtest/gtest/googlemock/test/gmock_all_test.gmock-actions_test.o 
../../third_party/fuchsia/clang/linux-amd64/bin/clang++ -MMD -MF obj/third_party/gtest/gtest/googlemock/test/gmock_all_test.gmock-actions_test.o.d -DGUNIT_NO_GOOGLE3=1 -I../../third_party/gtest/gtest/googlemock -I../../third_party/gtest/gtest/googletest -I../../third_party/gtest/gtest/googlemock/include -I../../third_party/gtest/gtest/googletest/include -Wall -Wendif-labels -Werror -Wextra -Wno-missing-field-initializers -Wno-unused-parameter -Wsign-compare -fno-exceptions -fno-rtti -fno-strict-aliasing -fobjc-call-cxx-cdtors -fstack-protector-all -fvisibility-inlines-hidden -fvisibility=hidden -Wheader-hygiene -Wnewline-eof -Wstring-conversion -fPIC -Wimplicit-fallthrough --sysroot=../../third_party/fuchsia/sdk/linux-amd64/sysroot/aarch64-fuchsia --target=aarch64-fuchsia -Wexit-time-destructors -Wno-unused-private-field -std=c++14 -Wno-inconsistent-missing-override -c ../../third_party/gtest/gtest/googlemock/test/gmock-actions_test.cc -o obj/third_party/gtest/gtest/googlemock/test/gmock_all_test.gmock-actions_test.o
In file included from ../../third_party/gtest/gtest/googlemock/test/gmock-actions_test.cc:36:
In file included from ../../third_party/gtest/gtest/googlemock/include/gmock/gmock-actions.h:46:
In file included from ../../third_party/gtest/gtest/googlemock/include/gmock/internal/gmock-internal-utils.h:46:
../../third_party/gtest/gtest/googletest/include/gtest/gtest.h:1421:11: error: comparison of integers of different signs: 'const int' and 'const wchar_t' [-Werror,-Wsign-compare]
  if (lhs == rhs) {
      ~~~ ^  ~~~
[...]

PR #1477, opened 2018-02-23
Try to handle unsigned wchar_t (arm) a bit better

sgraham:
This is an attempt at fixing #1476. I don't know where __WCHAR_UNSIGNED__ is and isn't defined, so I tried to make it so that it not being defined maintains existing behaviour.


gennadiycivil:
@sgraham Thank you very much for the contribution. For Googlers, could you please submit a CL first as per contributing guidelines. Once that CL lands we can merge this PR

gennadiycivil:
cl 187794354

PR #1479, opened 2018-02-26
Support JSON output format in addition to XML

petrhosek:
This change allows emitting output in JSON format in addition to the
already supported XML format. The implementation as well as the file
structure is intentionally modelled after the XML one.


gennadiycivil:
@petrhosek Please see the build failures, this version of googletest is still compliant with C++98, you need to add C++11 guards or use C++98 features.
googlest will *very soon* stop 98 support but not just yet

petrhosek:
Done, it was only a few cases so I replaced them with the C++98 syntax.

gennadiycivil:
@petrhosek  Thank you, please update the CL as well so the code is in sync

PR #1481, opened 2018-02-26
Use DEBUG_POSTFIX instead of CMAKE_DEBUG_POSTFIX

### LINKED ISSUE #1268: GoogleTest messes with global library postfix
Google Test sets CMAKE_DEBUG_POSTFIX which is a global user variable.
This makes it impossible to use in a simple add_subdirectory case as it will break all the library names afterwards.

Instead, it should set DEBUG_POSTFIX for the gtest/gmock targets.

### LINKED ISSUE #1334: GoogleTest is modifying my CMake based environment
Why does adding Google Test modify my environment.

For example after I added it to my CMake project I get:
warning C4530: C++ exception handler used, but unwind semantics are not enabled. Specify /EHsc

CMAKE_DEBUG_POSTFIX  is also set to "d"

dneto0:
CMAKE_DEBUG_POSTFIX is a global configuration parameter, and
changing it pollutes the configuration space for other projects
that enclose this project.

DEBUG_POSTFIX is better to use since it is a target-specific poperty.

Fixes #1334
Fixes #1268


dneto0:
I've tested this on Linux with a Release and Debug build.  As expected, the .a files for the Debug build have the "d" suffix, and the .a files for the Release build do not.

dneto0:
This is a followup to https://github.com/google/googletest/pull/1259 which added the use of CMAKE_DEBUG_POSTFIX in the first place.

The DEBUG_POSTFIX target property exists in CMake 2.6.  Proof: https://cmake.org/cmake/help/cmake2.6docs.html#prop_tgt:DEBUG_POSTFIX

PR #1495, opened 2018-03-06
Monotonic clock

bryanzim:
For anyone who has interest.  This pull requests uses the C++ 11 steady_clock(monotonic) for compilers that support it; otherwise it implements the calls for Win/Linux/Mac to use a monotonic clock.

The monotonic clock is for timing of google tests so that test times can never be negative due to a change in the system clock through a background process such as ntp.


bryanzim:
Hmmm, that is a new one.  This pull request only contains the latest master plus changed files by myself only.  I have only been rebaselining with latest merged changes via other authors; so author consent is implicit.

KindDragon:
Maybe rebase and force-push will help with this problem

bryanzim:
@gennadiycivil this should fix the issues with linking librt as the CMake files now mark librt as transitive, ie. needed for dependent projects.

gennadiycivil:
@bryanzim , thank you very much for this contribution. I am reading that this change would be potentially useful in the "corner case" when the system clock is changing while the test is running. Correct? It would be helpful to include test results, to illustrate the difference "before" and "after" .

gennadiycivil:
I suggest delaying this until googletest drops C++98 support. I can not promise any dates but it should be * soon * . This way the code can be much simpler.

bryanzim:
I'm sure there are many people using pre c++ 11 compilers.  I understand that googletest is moving to C++ 11 requirement.  However, if you could please test this to see if it fixes the issue you had with the test compilation.  I believe I found the issue with CMake not marking libraries as required for dependent projects.  If it still fails then I would agree with your assessment.

gennadiycivil:
@bryanzim Could you please explain, which "issue you had with the test compilation" ? 
(I see that this PR passed both Travis and Appveyor builds)
Thank you again.

bryanzim:
@gennadiycivil I didn't have a problem with any of the tests.  My comment was with regards to last years issue with some testing by yourself and not being able to compile one of the internal tests.

gennadiycivil:
@bryanzim I merged the master branch, CI tests passed

gennadiycivil:
I suggest delaying until C++11.  This way the code can be much simpler, most of the code will actually disappear.

gennadiycivil:
@bryanzim

googletest/cmake/internal_utils.cmake
```diff
@@ -176,13 +176,11 @@ function(cxx_library_with_type name type cxx_flags)
   endif()
   if (DEFINED GTEST_HAS_PTHREAD)
     # add pthread to this library and all librariers which link to this library if needed
-    set_property(TARGET ${name} APPEND PROPERTY LINK_LIBRARIES ${CMAKE_THREAD_LIBS_INIT})
-    set_property(TARGET ${name} APPEND PROPERTY INTERFACE_LINK_LIBRARIES ${CMAKE_THREAD_LIBS_INIT})
+    target_link_libraries(${name} PUBLIC ${CMAKE_THREAD_LIBS_INIT})
```
KindDragon: Current required cmake version 2.6.4

bryanzim: So it was target_link_libraries before but I added the PUBLIC option.  Is that what you are alluding to when referencing V2.6.4?  If so would preference be to use old target_link_libraries with an addition set_property or use two set_properties?

KindDragon: Never mind, I thought that `target_link_libraries` was added to CMake 3.0

gennadiycivil:
@bryanzim ping. Feel free to remove everything C+98. This should simplify this work sugnificantly

gennadiycivil:
Closing for now, looks stale

PR #1498, opened 2018-03-08
Add missing using statement in mock CookBook

stevenleigh:
Example from docs wont compile without the DoAll using statement which is a major part of the example


stevenleigh:
I signed it!

PR #1499, opened 2018-03-08
Add missing using statement in mock CookBook

stevenleigh:
Example doesn't compile without this using statement, and it is a major part of the example


stevenleigh:
Actually, I was wrong.  This example doesn't seem to need that using statement to compile.

PR #1500, opened 2018-03-09
Include cstring istead of string.h for compatibility

AndersSundmanTobii:
One specific compiler we use (for an embedded platform) does not have a string.h in it's library. It does have cstring. 

Using cstring is the normal C++ way to do it anyway, so I seems a reasonable thing to use it instead of string.h - it will probably increase the probability of gtest being portable to other compilers/libraries as well. 

I know it is not a priority of gtest to support all platforms, but this should be a zero cost change for others while benefiting the embedded space (since it will help us not having to maintain an inhouse fork of the project).


AndersSundmanTobii:
Sorry, bug in the PR. Will update with the right change.

PR #1501, opened 2018-03-09
Remove hard-coding of template typenames

Osram-Lux:
Hi,

Changing arity on line 5 in gmock-generated-actions.h.pump
0005    $var n = 10  $$ The maximum arity we support.
will brake compilation, since one template type list is hard-coded

This pull request should solve the problem


Osram-Lux:
I've signed it!

googlemock/include/gmock/gmock-generated-actions.h
```diff
@@ -1,4 +1,6 @@
-// This file was GENERATED by a script.  DO NOT EDIT BY HAND!!!
+// This file was GENERATED by command:
```
gennadiycivil: It would be best to keep the formatting so the changes are limited to the actual purpose

Osram-Lux: Hi,

I do not understand this request. I did the "real" change in `.../gmock-generated-actions.h.pump`, and then I regenerated the `.h` file with:
```sh
$ ./googletest/scripts/pump.py googlemock/include/gmock/gmock-generated-actions.h.pump
```
Any diff in the generated `.h` file is either result of (other) changes in `pump.py` or due to my changes in `.../gmock-generated-actions.h.pump`. Inspection of the main function in `pump.py` indeed shows that pump.py is responsible for the changed output.

The reason I regenerated the header file is twofold:
* There is no Cmake support for generating the header files.
* To show that dynamic template types where correctly generated (this seems to have been misleading however).

It seems that the root cause of the unwanted change in the header file(s) is that they where not regenerated when pump.py was changed. The optimal solution would be to generate the header files from Cmake, so that error prone manual steps are eliminated. I could have easily done this in make, but unfortunately I have never written CMake scripts

Please advice if I should keep the re-generated header file (i.e. not update the pull request) or revert the generation and just keep the change in the `.pump` file?

gennadiycivil: @Osram-Lux Sorry for the confusion, you are correct. There is no need to invest in making pump changes in CMake/make at this point.

gennadiycivil:
@Osram-Lux could you please provide test evidence, show compilation run with the maximum arity we support "before" - I assume broken and "after" - proving successful fix. 
Thank you again

gennadiycivil:
@Osram-Lux , Also there appears to be some tool confusion , see "The bad news is that it appears that one or more commits were authored by someone other than the pull request submitter. We need to confirm that all authors are ok with their commits being contributed to this project. Please have them confirm that here in the pull request."

Please confirm that you are in fact the author of these commits and if not please confirm who the authors are and that they are ok with their commits being contributed to this project. 

Thank you!

Osram-Lux:
Hi,
@gennadiycivil I will not be able to provide a test evidence before tomorrow, but setting n>10 triggers compile errors (and possibly also n<10, but I have not tried that, since I needed to increase arity)
 I'm also not sure sure how to handle the "thumbs down" from the google-bot. The bot does not mention which commit(s) it thinks have been authored by someone else and what mismatch or other circumstance makes the verification to fail. 
Git asked for email and name before I was able to commit - will a `git commit --amend` be enough with name and e-mail set to what the bot expects.
I've also set github to hide my e-mail - can this prevent an author consent verification?

gennadiycivil:
@Osram-Lux , as far as googlebot. it will be sufficient if you state and confirm that you are in fact the author of 100% commits in this OR and if not please confirm who the authors are and that they are ok with their commits being contributed to this project.

Osram-Lux:
@gennadiycivil , I definitely am the author of the changes in `googlemock/include/gmock/gmock-generated-actions.h.pump`. `googlemock/include/gmock/gmock-generated-actions.h` was generated by `pump`, so there should not be any copyright issues.

I unfortunately had less luck with adding a test-case. For the proof/test-case you asked for i have had less luck. To make a reasonable test-case, it is most elegant to let CMake take care of the generation. I know that I should use `add_custom_command` in the `CMakeLists.txt`, but I have not yet found out how to tie this command to the right target (right now the pump command does not even en up in the `Makefile`).

Please have patience with me while I learn CMake a little bit better.

PR #1520, opened 2018-03-28
Improve test case name and test name in code examples

aaronang:
In the Google Test Primer documentation, it is mentioned that test case name and test name should not have any underscores. However, in one code example, it does use underscores which implicitly causes confusion. Therefore, in this patch, I made sure that it is consistent with respect to this rule.


gennadiycivil:
@aaronang please resolve conlicts

PR #1523, opened 2018-03-28
provide alternative for DebugBreak()

leissa:
This uses asm("int3") for clang/gcc on x86 as alternative for DebugBreak()

### DIFF
--- a/googletest/src/gtest.cc
+++ b/googletest/src/gtest.cc
@@ -4549,6 +4549,9 @@ void UnitTest::AddTestPartResult(
       // when a failure happens and both the --gtest_break_on_failure and
       // the --gtest_catch_exceptions flags are specified.
       DebugBreak();
+#elif (defined(__clang__) || defined(__GNUC__)) && (defined(__x86_64__) || defined(__i386__))
+      // with clang/gcc we can acchieve the same effect on x86 by invoking int3
+      asm("int3");
 #else
       // Dereference NULL through a volatile pointer to prevent the compiler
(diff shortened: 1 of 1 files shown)


gennadiycivil:
@leissa Thank you for this contribution. Could you please explain what problem is this solving and why this would be advantageous for the master branch?

leissa:
This patch increases the user experience for platforms which don't support ```DebugBreak()```. ```asm("int3")``` does the same for x86-based systems. This means, on my linux system the following happens without the patch:
```
gdb ./my_unit_tests --gtest_break_on_failure
```
1. On the first failure I'll get a segmentation fault because of the ```*static_cast<volatile int*>(NULL) = 1;```
2. Then in the debugger I can look around, inspect values, and so forth but I cannot ```continue``` program execution.

With this patch, I can ```continue``` in ```gdb``` because ```int3``` is just a usual break point.

gennadiycivil:
@leissa Could you please confirm consent of ALL commit author(s). For some reason the bot is having trouble . Did you write ALL code in this PR and if not could you confirm that all authors are OK wit this PR to be merged into googletest?
Thanks

leissa:
@gennadiycivil I was puzzled by that, too. I'm the only author and this patch only takes 3 lines of code with two commits:
* 11855d8ef08567c7961fb3c6562fbe6f9b479092 is the original 3-line patch
* d04bf07fadd204881df4c7114906fa63b1b80c48 fixes a typo in a comment.
That's it.

Issue #1366, opened 2017-12-20
C++11 


gennadiycivil:
An update from googletest team: it's decided starting from version 1.9.0 googletest will require C++11 to compile. A 1.8.1 branch will be created for pre-C++11 compilers but only critical bug fixes will be accepted in the branch.

In addition starting with 1.9.0 version googletest will be following Abseil compatibility  policies outlined [HERE](https://abseil.io/about/compatibility).

Thank you

ninjaCoder:
Great news, when 1.9.0 except to be released ?

PR #1524, opened 2018-03-29
DRY in function mockers pumpfile

scraimer:
The pump file has a repeated bit of code: `MOCK_CONST_METHOD` and `MOCK_METHOD`. If we could generate both the CONST and non-CONST versions for all the `MOCK_*_METHOD` macros from the same source, we could cut he size of the pump file in half.

To do that, I refactored the code in `gmock-generated-function-mockers.h.pump` so that
both CONST and non-CONST mocks are generated from the same line.

(When reviewing, please ignore the generated `gmock-generated-function-mockers.h` file, and only look at the changes in the `gmock-generated-function-mockers.h.pump`)


gennadiycivil:
@scraimer Thank you for your contribution. We are not really accepting changes to pump anymore because it is about to become irrelevant when googletest drops C++98 support. See https://github.com/google/googletest/issues/1366

PR #1532, opened 2018-04-01
Allow MOCK_METHOD without specifying the argument count

### LINKED ISSUE #948: Allow MOCK_METHOD without specifying the argument count

ugoren:
Fix #948


ugoren:
I signed it!

ugoren:
I signed it!

wizardsd:
@ugoren Could merge from master branch, please?

ugoren:
@wizardsd, done, but there's some problem with Travis-CI at present.

wizardsd:
@ugoren could you trigger the build one more time?

ugoren:
@wizardsd, done.

gennadiycivil:
Sincere apologies for the long wait. We are actively discouraging any new macros.

ugoren:
Hello Gennadiy,
I would appreciate it if I could get a more elaborate reason for closing.

The change is not meant to add more macros, but to replace the existing
ones with fewer macros, which would be easier to use. The old ones remain
just for backward compatibility.

When I started working on it, the maintainer (https://github.com/sbenzaquen)
looked favourably on this change, so I believed it can be accepted if
implemented well.

I really hoped this change could help make Google Mock more convenient to
use, and I know it fixes a painful point for many users. But if you think
it will not benefit the product, this is of course your right.

Thanks, Uri Goren.

gennadiycivil:
@ugoren My bad, re-opening

ugoren:
Thanks.

ivserna:
Hi @ugoren, 

Interestingly, the idea of reducing the number of macros, by allowing the macro to identify arity and allow override/final, calltype and cv qualification within the same macro is being suggested at the moment.

Such API is a big change and internally requires a so called API review. 

Here are some examples:

MOCK_METHOD0(ReturnTypeWithComma, std::map<int, std::string>()); MOCK_METHOD((std::map<int, std::string>), ReturnTypeWithComma, (), ());

MOCK_CONST_METHOD0(OverloadedOnConstness, char());
MOCK_METHOD(char, OverloadedOnConstness, (), (override, const));

MOCK_METHOD0_T(InTemplate, int());
MOCK_METHOD(int, InTemplate, (), ());

MOCK_METHOD0_T_WITH_CALLTYPE(STDMETHODCALLTYPE, Pop, void());
MOCK_METHOD(void, Pop, (), (Calltype(STDMETHODCALLTYPE), override));

It would be cool if you can have a look at my implementation.

Cheers.

Ivan

ugoren:
Sure, I'll take a look. I agree that these macro changes are related and
should be considered together.
Just note that I'm not a Google Test maintainer, so I'm not sure it counts
as an API review.

ivserna:
Thanks.

https://github.com/google/googletest/compare/master...ivserna:patch-1

Let me know what your initial impressions are, before I create a pull request. 

I see a lot of overlapping functionality. This patch has also an _Spec tuple and just defines the macro without pump files.

I would go as far as to define all previous macros in terms of GMOCK_INTERNAL_MOCK_METHOD_IMPL. 

An API review is an internal process. It is still pending. There is also a doc explaining the rationale for some design choices.

ugoren:
@ivserna,
I started looking at it, and have some comments. But I think they're off-topic on this pull request.
Perhaps you can create a pull request into my fork (https://github.com/ugoren/googletest)? Then we can use it to discuss further.

Getting rid of the `_CONST`, `_T` and `_CALLTYPE` variations is certainly an important goal. In my pull request I tried to make the lower level macros flexible for such additions, but didn't expose it in any way to the caller. This would be a good addition.

One important thing I noticed is that I think your macros don't seem to support MSVC. I realized it when I enabled AppVeyor CI (I recommend enabling both AppVeyor and Travis on your fork, and add/modify the internal tests to use your new macros). The MSVC preprocessor is a nightmare - it deviates from the standard in subtle but important points, and I never found any documentation explaining its logic. Your macros seem to be written for standard C/C++ but not for MSVC.

Regarding the API - I like the `MOCK_METHOD(int, foo, (int x));` usage, without the `_Spec` parameter. But of course we need a way to pass `const`, `override` and such. Perhaps we can make it optional, and possibly even allow multiple specs:
```
MOCK_METHOD(int, foo, (int x));
MOCK_METHOD(int, foo, (int x), Const);
MOCK_METHOD(int, foo, (), Const, Override);
```

ivserna:
An optional Spec is part of the proposed design. Wrt MSVC, the written macros are expected to be portable, although I just haven't been able to test them in  _MSC_VER <= 140. Pending on the response I get next week, I will create a pull request in your fork.

Thanks for your time.

sbenzaquen:
On Sun, Sep 2, 2018 at 9:30 AM ivserna <notifications@github.com> wrote:

> An optional Spec is part of the proposed design. Wrt MSVC, the written
> macros are expected to be portable, although I just haven't been able to
> test them in _MSC_VER <= 140. Pending on the response I get next week, I
> will create a pull request in your fork.
>
Note that this week we released 1.8.1, which was our last release that
supported C++98.
We are raising to the floor of compatible compilers to at least C++11. We
will be following the compatibility guarantees of Abseil.
For MSVC it means at least _MSC_VER>=1900, iirc.

ugoren:
Hi @ivserna,
I wrote some comments on your commits, you should be able to see them in your clone.

ivserna:
Hi @ugoren,
Thank you very much! I addressed most comments, I will prepare the changes and wait for the internal review. Once that is done, I will work directly on your repository.

ugoren:
Glad to know my work was put to actual use.
One of my goals was to provide a base that will allow passing more
attributes, such as noexcept. But I didn't actually add noexcept support.
I'm not sure what broke.
Const is a bit more than an attribute - it allows overloading, so it
requires attention when creating members according to the function name, yo
prevent duplicates. This isn't required for override and noexcept. This
extra mechanism seems to get in your way.

I'll try to see if I can find an easy way to add attributes.

pepe82sh:
That'd be awsome! #667 has been closed, but was (around some corners) referenced to the issue leading to this pull request.
< off topic>I find it a bit weird that `c++98` will be dropped soon, while something like `noexcept` is not supported yet.</off topic>

ugoren:
@pepe82sh,
As a simple hack, `#define AdjustConstness_noexcept AdjustConstness_` should fix it.
It should be good enough to support `noexcept`, but not `noexcept` and `const` together.

pepe82sh:
Unfortunately the simple hack does not work. The compiler throws me a `error: expected ',' or '...' before 'noexcept'`. Do you have any other idea regarding that issue? I'm tempted to reopen the issue in googletest.

ugoren:
@pepe82sh, I created a noexcept branch in my clone, with some support.
The temporary usage is `MOCK_METHOD_ATTRS(ret, method, args, noexcept)`.
It supports `overrire` too.

pepe82sh:
Thank you very much @ugoren! Works like one would expect, even using multiple specifiers:
`MOCK_METHOD_ATTRS(int, ConstNoexceptFunc, (), const noexcept override);`
🥇

ugoren:
I would guess that the problem is with const specifically. i did yest
multiple attributes (noexcept+override). It shouldn't be hard to write a
MOCK_CONST_METHOD_ATTR macro.

pepe82sh:
Oh maybe we have a misunderstanding: There is no problem anymore, not even with `const noexcept override`.

ugoren:
Oh, I misread 'even' as 'except'.

pepe82sh:
If this PR goes through, the branch might be a worthy fix for #1821.

ugoren:
@pepe82sh,
Yes, it should be included, in some form. I don't think the macros are good as they are.
@ivserna did it more nicely in [his fork](https://github.com/ivserna/googletest/tree/patch-1), though he only supports specific attributes.

I think both can be combined to make it ideal. Something like this:
```
MOCK_METHOD(int, foo, (int));                // No attributes
MOCK_METHOD(int, foo, (int), override);      // Or noexcept or whatever
MOCK_METHOD(int, foo, (int), const);         // Identical API for const, though it gets special treatment
MOCK_MOTHOD(int, foo, (int), const, override, noexcept);   // Multiple attributes
```

Issue #822, opened 2016-07-14
How to test MPI applications?


floli:
Hello,

we want to do unit und integration testing of our scientific application that uses MPI. MPI uses different processes within one communication domain (`MPI_COMM_WORLD`) to implement parallelism. A MPI application is usally started like `mpirun -n 4 ./application` starting 4 instances in one comunication domain. Each process is assigned a rank = {0, 1, 2, ..., n-1}

For testing it using googletest I see two challenges:
- Some tests only run sequentially (some other use MPI to communicate between instances of the same test case), therefore they should be skipped when the rank != 0. For that we need some means of runtime skipping, like the boost.test prondition decorator (http://www.boost.org/doc/libs/1_61_0/libs/test/doc/html/boost_test/tests_organization/enabling.html). There is issue #490 which is the same, as far as I can tell.
- Started with the command above each process is connected to the same stdout, so the mesages mingle. For that we need some kind of message aggregation technique, so we can have a unified output from just one process at the end.

I'm not fixed to these requirements, what is your approach for testing MPI applications?

Thanks!

AntoineD:
Hi floli,

I'm currently writing MPI tests as well, but unfortunately I do not know how to help you with your concerns.
So far I created an gtest environment to init and finalize the MPI context:
```
#include "gtest/gtest.h"

#include "mpi.h"

using namespace ::testing;

// This class deals with the MPI context,
// it shall be done globally for the test program.
// see https://github.com/google/googletest/blob/master/googletest/docs/AdvancedGuide.md#global-set-up-and-tear-down
class TestEnvironment : public Environment {
protected:

  virtual void SetUp() {
    char** argv;
    int argc;
    int mpiError = MPI_Init(&argc, &argv);
    ASSERT_FALSE(mpiError);
  }

  virtual void TearDown() {
    int mpiError = MPI_Finalize();
    ASSERT_FALSE(mpiError);
  }

  virtual ~TestEnvironment() {};
};

Environment* const foo_env = AddGlobalTestEnvironment(new TestEnvironment);
```

I'm also trying to figure out how to handle death test within MPI.

norci:
hi @floli 

For this question:
Started with the command above each process is connected to the same stdout, so the mesages mingle. For that we need some kind of message aggregation technique, so we can have a unified output from just one process at the end.

[...]

Issue #490, opened 2015-07-28
Plase add a way to skip/disable some tests during runtime.


GoogleCodeExporter:
```
Hello,

It would be nice to have a possibility to skip some tests basing on criteria 
available during runtime. This subject has been alreade mentioned few years 
ago, but nothing really happened since this time. Threads on mailing list about 
the case are available:
 * https://groups.google.com/d/msg/googletestframework/Nwh3u7YFuN4/-hWG7KhT0bMJ
 * https://groups.google.com/d/msg/googletestframework/0dZM4b-5Lp8/nsLj2TxZwQEJ

Kind regards
    Artur
```

Original issue reported on code.google.com by `artu...@gmail.com` on 14 May 2015 at 12:37

sirzooro:
Any update on this? I am looking for this too.

ermilindwalekar:
Looking for the same. I fee a macro should be there "Conditional TEST_C" which can take 3 parameters wherein the 3rd parameter gives the control, the other 2 being the conventional ones.

gennadiycivil:
GTEST_SKIP is available

Issue #160, opened 2015-07-28
In addition to PASS & FAIL, also add test result class SKIPPED (or similar)?


GoogleCodeExporter:
```
We are using GTEST for very many tests, however all test in our test case 
definition are linked together and executed. Unfortunately, some tests can 
only be executed if the environment is correctly set up (must be done 
manually). Therefore these tests are always failing, but in fact they could 
not be executed.

Would it be possible to the test case result "SKIPPED" as a new class to a 
test case (maybe colored yellow ;-) to abort a test if it determines it 
cannot be executed in this environment?
```

Original issue reported on code.google.com by `constant...@gmail.com` on 2 Jul 2009 at 5:30

CryPtiC128:
Is there any update on this request or any need in the community to require such an update?

In my opinion, a third state ("inconclusive" / "skipped" / ...) in testing is very important: Every test above unit test level (or at least most of them)
- answers a question and
- has some kind of preconditions (= the arrange part).
If the test environment cannot be brought to match the preconditions, the question cannot be answered. Thus the test did not fail.

In many cases, the arrange part of a test does not even concern the aspect under test. If there was an error in such an arrange part, we would not have a problem in our aspect under test, but no information.

LazyWolfLin:
I am looking for this too.

I use gtest in my work, and it rely on the remote server. So I want some test is skipped if the remote server is break.

Issue #1079, opened 2017-05-04
Next release?


brandon-arnold:
Hello,

It looks like [master](https://github.com/google/googletest) is the only branch that builds on all systems for me, and not the most recent 1.8 release around 7 months ago. Is there plan to tag a new release any time soon?

-Brandon

alpha-gj:
I want to know too. 
Thanks 

GJ

4brunu:
I'm also interested in this.
Maybe @BillyDonahue can help.
Thanks

SylvainCorlay:
A `1.8.1` including the fix for mingw support (https://github.com/google/googletest/pull/721) which was merged right after the last release would be greatly appreciated.

petko:
Please do something about it. Right now it's quite confusing for people not monitoring closely the development of GTest which commits are stable enough for production use. I use 1.8, but it references some TR1 items in the Standard Library and I started getting deprecation warnings in the latest compilers..

madscientist:
If GTest decides to go to a "continuous release" model (or really, even if it doesn't) I really would love to see a NEWS file or similar updated in commits that add major new features, important bug fixes, etc. with a description.  The Git history doesn't really help: it's a long list of "merge pull request" commit messages, mostly without information about what was actually changed.

I really have no idea what changes I'm getting if I take a cut of the Git repository.

Thanks!

gennadiycivil:
See https://github.com/google/googletest/issues/1366

madscientist:
Thanks for the pointer!  I'm afraid I'm a little too slow today to grok it fully.  Are you saying that all the changes being made are related to C++11 support and no other changes to GTest are happening, so no NEWS-type file is necessary because there aren't any functional improvements (only rewrites to C++11)?

Or are you saying that the description of how GTest is going to be developed and released going forward is defined by the comment discussing Abseil compatibility policies in the second paragraph?

In either case I still think it would be a good practice to maintain a NEWS or CHANGES file.  It may currently have only one entry ("Convert GTest to C++11"), but it will give a good place to add information in the future.  If there are no more releases, there won't be release notes or whatever.  It's good if the documentation is up to date and we can always get the latest code (due to compatibility guarantees), but scanning the documentation to see what improvements might have been made since the last time we used GTest, that we can take advantage of, is not really feasible.

There is already a CHANGES file available, but it doesn't seem to have been updated since 1.7.0.

[...]

PR #1544, opened 2018-04-04
Skipping tests at runtime with GTEST_SKIP()

aprelev:
### Description
Nothing fancy, plain and simple test skipping based on condition evaluated at runtime. One might want to skip
+ tests which are not supported within current environment,
+ tests which depend on some external resource with conditional availability, 
+ tests which require privileges when run under user with no such privileges, etc.

Feature has been requested several times, see #160, #490 and #822. Works as demonstrated in [sample test](https://github.com/aprelev/googletest/blob/a62e58697979146eb700501f31f0b2266ee4111f/googletest/test/gtest_skip_test.cc#L35):

```C++
TEST(SkipTest, DoesSkip) {
  SKIP();
  EXPECT_EQ(0, 1);
}
```

yields

```
[==========] Running 1 test from 1 test case.
[----------] 1 test from SkipTest
[ RUN      ] SkipTest.DoesSkip
[  SKIPPED ] SkipTest.DoesSkip (1 ms)
[----------] 1 test from SkipTest (1 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test case ran. (1 ms total)
[  PASSED  ] 0 tests.
[  SKIPPED ] 1 test, listed below:
[  SKIPPED ] SkipTest.DoesSkip
```

Skipped tests are displayed in green, and are not considered neither successful nor failed since they have never been completely run. 

### Further work
There are some conceptual issues which are not yet resolved, but I would love to hear you thoughts on them.
1. Listing all skipped tests on teardown might be not such a good idea; it is at least reasonable to add compile option to control this behaviour.
2. If test is skipped at any point, all preceding failures, which occurred in this test, are ignored; this can be handled in test result assigning with checking for failures before setting `kSkip` result.
3. Skipping is not available from `SetUp()` and `TearDown()` methods, as well as from functions external to test body, since skipping aborts only current function; this can be resolved by adding additional `AssertHelpers` in fashion similar to handling `FAIL()`.
4. Additional improvements shall be made in loggers to support this additional test result.


gennadiycivil:
@aprelev Thank you for your contribution. On the surface I am tempted to say that the ability to skip tests is already available with [disabling](https://github.com/google/googletest/blob/18d270e335c420d69df0a7db17d9068056165459/googletest/docs/AdvancedGuide.md#temporarily-disabling-tests) . Could you please explain why this is different and better? Thank you very much

aprelev:
Sure! The key here is ability to skip tests at **run-time** (based on a run-time condition); this feature is introduced **in addition** to compile-time disabling, rather than as replacement.

Consider a test, which can fail or succeed, but is only supported (i.e. can be executed) when certain condition is met, and testing the condition can only be carried out in run-time; otherwise test is not supported. As of now, there is no pretty way to handle this:
+ test cannot be disabled with `DISABLED_` keyword because it is not known in advance (at compile-time) whether the test will be supported,
+ test can be manually set to fail or succeed (with `FAIL()` or `return` in test body) if condition proves to be false, but this would indicate that test ran with definitive result type, such as _failure_, or _success_, albeit result of such test shall be _inconclusive_.

### My use-case

#### Test description

Consider a test which validates correctness of hardware-accelerated AES encryption (e.g. with AES-NI); such test can only be executed on processors with support for AES-NI, and it is not known in advance if target machine has such processor.

#### Environment constraints

Tests binary is pre-built and distributed onto several target machines, and none of target machines have C++ compiler installed.

#### Solution with `SKIP()`

```C++
TEST(SampleSuite, EncryptionTest) {
  if (!is_available(AES_NI))
    SKIP();

  test_encryption();
}
```

### Other use-cases

A few cases where described in referenced issues:

> [#160] We are using GTEST for very many tests, however all test in our test case 
definition are linked together and executed. Unfortunately, some tests can 
only be executed if the environment is correctly set up (must be done 
manually). Therefore these tests are always failing, but in fact they could 
not be executed. Would it be possible to the test case result "SKIPPED" as a new class to a 
test case (maybe colored yellow ;-) to abort a test if it determines it 
cannot be executed in this environment?

> [#160] Typical use case would be testing a library that 
might do things that require administrative privileges. If the tests are run as 
a limited user, then the requiring-root tests would have a guard marking the 
test as skipped if administrative privileges are absent.

> [#160] I use gtest in my work, and it rely on the remote server. So I want some test is skipped if the remote server is break.

> [#490] It would be nice to have a possibility to skip some tests basing on criteria 
available during runtime.

> [#822] we want to do unit und integration testing of our scientific application that uses MPI. <...> Some tests only run sequentially (some other use MPI to communicate between instances of the same test case), therefore they should be skipped when the rank != 0. For that we need some means of runtime skipping, like the boost.test prondition decorator 

More examples can be found elsewhere:

> [25999057](https://stackoverflow.com/questions/25999057/how-to-signal-to-gtest-that-a-test-wants-to-skip-itself) I have a set of typed test cases in google test. However, some of these test cases are simply not applicable for a specific type parameter.

Such feature has been implemented in other testing frameworks as `throw SkipException`, `Assert.Inconclusive()`, etc.

aprelev:
I will look into it. What is your take on problems mentioned earlier?

> 1. Listing all skipped tests on teardown might be not such a good idea; it is at least reasonable to add compile option to control this behaviour.
> 2. If test is skipped at any point, all preceding failures, which occurred in this test, are ignored; this can be handled in test result assigning with checking for failures before setting `kSkip` result.
> 3. Skipping is not available from `SetUp()` and `TearDown()` methods, as well as from functions external to test body, since skipping aborts only current function; this can be resolved by adding additional `AssertHelpers` in fashion similar to handling `FAIL()`.
> 4. Additional improvements shall be made in loggers to support this additional test result.

gennadiycivil:
@aprelev 
My comments:

>>Listing all skipped tests on teardown might be not such a good idea; it is at least reasonable to add >>compile option to control this behaviour.

Could you please clarify why you feel that this may not be a good idea? I don't immediately see issues with it

>>If test is skipped at any point, all preceding failures, which occurred in this test, are ignored; this can >>be handled in test result assigning with checking for failures before setting kSkip result.

Skipping preceding failure seems to be logical. Skip is a user over-ride, meaning "I know better that this test can not run so lets skip it". Moreover, deciding to skip the test should be done in the very beginning of the test. 

>>Skipping is not available from SetUp() and TearDown() methods, as well as from functions external to >>test body, since skipping aborts only current function; this can be resolved by adding additional >>AssertHelpers in fashion similar to handling FAIL().

Dont see a need to complicate this way. Skipping a test is a user-override for this test. It should be decided on the test level and not outside it

Dont forget the docs!

Thank you again
G

aprelev:
Okay, I've looked into implementing skipping with function.

As far as I understand, fatal failure, `FAIL()`, escapes current scope by expanding into `return` statement, while non-fatal failure, `ADD_FAILURE()`, does not. That is why test skipping was introduced as a __macro__:
 
```C++
#define SKIP() \
  return AssertHelper() = Message()
  ^^^^^^

TestBody() {
  SKIP();
  FAIL(); // < This will not be executed.
}
```

Any __function__, on the other hand, fails to escape current scope, unless exception is thrown or long jump is performed, therefore rendering skipping useless.

```C++
TestBody() {
  ::testing::SkipTest();
  FAIL(); // < This will be executed anyway.
}
```
Could you be more specific on how test skipping could be implemented with function; and what was the rationale behind your decision?

aprelev:
Ping.

gennadiycivil:
@aprelev Sorry for the delay. It will be some time before I can pay my full attention to this. The current priority is to produce 1.8.1 release, see https://github.com/google/googletest/issues/1079. This is still important but takes a second priority , thank you again

aprelev:
@gennadiycivil Okay, no worries :) poke me here once you are done with important stuff

dturner:
Any update here? `SKIP()` seems like a useful function for my use case as well. I am writing tests for Android NDK code and some tests are dependent on features introduced in a specific API. For me it makes sense to write something like: 

    if (androidAPI() < 26){
         SKIP();
    } else {
         ASSERT_EQ(...);
    }

aprelev:
I am waiting for @gennadiycivil for his attention on this feature and implementation mechanics.

aidancully:
I don't know if this needs to be in _this_ PR, but it'd be useful to me if we could skip all tests from a test fixture. Something like:
```c++
class MyTestFixture : public ::testing::Test {
protected:
  virtual void SetUp() override {
    if (runtime_check_failed) {
      SkipAll(); // defined in base class
      return;
    }
  }
```

aprelev:
Ping.

aprelev:
Ping.

gennadiycivil:
@aprelev please resolve all conflicts. Thank you

aprelev:
Ping.

aprelev:
Ping.

gennadiycivil:
@aprelev I made the changes in your branch. Please add documentation and ping me when ready
Thank you again

zrphercule:
@aprelev Thanks! Your PR really helps a lot. Hope it can be merged soon =)

gennadiycivil:
This PR is ready to be merged. I am waiting for the docs

gennadiycivil:
@aprelev given that others are waiting,  
I made changes internally that fix small issues you have in the code, same changes that I made in your branch before. 
I will push the changes shortly.
You might see that this PR is "Closed" and not "Merged" but when you look at the actual commits you should see your name as the author.

gennadiycivil:
@aprelev Ping on documentation?

Apteryks:
@sandy5842 given that the last tagged release (1.8.1 at time of writing) dates back to the end of August 2018, I'd say it is still yet unreleased unfortunately.

gennadiycivil:
It is in the master branch and has been for a while .

samsta:
Hi Guys, does this also put `<skipped/>` in the [XML output](https://stackoverflow.com/a/9691131/2884694) ? This would be useful for CI systems like Jenkins.

PR #1547, opened 2018-04-05
Matchers testing

gennadiycivil:


googlemock/include/gmock/gmock-more-matchers.h
```diff
@@ -43,6 +43,15 @@
 
 namespace testing {
 
+// Silence C4100 (unreferenced formal
+// parameter) for MSVC
+#ifdef _MSC_VER
+# pragma warning(disable:4100)
+#if (_MSC_VER == 1900)
+# pragma warning(disable:4800)
```
KindDragon: Why not just cast to `bool`?

PR #1552, opened 2018-04-07
Sync gmock-generated-nice-strict.h.pump with gmock-generated-nice-strict.h

pwnall:
Commit fe402c27790ff1cc9a7e17c5d0aea4ebe7fd8a71 published the changes in internal CL 156157936, but missed the diff in gmock-generated-nice-strict.h.pump. This makes it difficult to reason about the change, because the .pump file is more concise than the generated file.

This PR was tested by re-generating the .h file using the command below and checking the git diff.

```bash
./googletest/scripts/pump.py googlemock/include/gmock/gmock-generated-nice-strict.h.pump
```


gennadiycivil:
@pwnall Thank you

PR #1563, opened 2018-04-12
fixed comparison between linked_ptr and nullptr in clang with c++11

s-kenji:
Hi, 
I cannot run a unit test using gtest in clang with c++11.
The following error has occurred.
```
In file included from /Users/s-kenji/workspace/my_projct/test/jsonparser.cc:5:
In file included from /Users/s-kenji/workspace/my_projct/build/googletest-src/googlemock/include/gmock/gmock.h:58:
/Users/ksugimur/workspace/my_projct/build/googletest-src/googlemock/include/gmock/gmock-actions.h:388:18: error: use of overloaded operator '==' is ambiguous (with operand types 'const internal::linked_ptr<ActionInterface<void
      (const basic_json<std::map, std::vector, basic_string<char>, bool, long long, unsigned long long, double, std::allocator, adl_serializer> &, bool *, int *)> >' and 'nullptr_t')
    return impl_ == nullptr && fun_ == nullptr;
           ~~~~~ ^  ~~~~~~~
```
The cause of this error is that overloaded operator '==' is not defined for comparison of linked_ptr and nullptr, So I defined the comparison of linked_ptr and nullptr.

### DIFF
--- a/googlemock/include/gmock/gmock-more-matchers.h
+++ b/googlemock/include/gmock/gmock-more-matchers.h
@@ -46,8 +46,11 @@ namespace testing {
 // Silence C4100 (unreferenced formal
 // parameter) for MSVC
 #ifdef _MSC_VER
+# pragma warning(push)
 # pragma warning(disable:4100)
 #if (_MSC_VER == 1900)
+// and silence C4800 (C4800: 'int *const ': forcing value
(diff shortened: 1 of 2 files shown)


s-kenji:
I signed it!

ksugimur:
I signed it!

PR #1574, opened 2018-04-14
Test that mocks can be memcpy-ed in memory.

pwnall:


gennadiycivil:
@pwnall I still see a failed Windows build...

pwnall:
@gennadiycivil Sorry for the noise!

Turns out what we were doing (shown in the test added by this PR) is undefined behavior in C++. We fixed Chrome's test code.

PR #1581, opened 2018-04-18
Introduce parameterless expectations

dnsunderland:
This is a clone of internal change cr/176529515, which introduces parameterless expectations. Copied the description of that change here:

Add support for setting "any args" in ON/EXPECT_CALL without explicitly specifying number of matchers. For example,

```
ON_CALL(mock, Method(_, _, _)).WillByDefault(…)
EXPECT_CALL(mock, AnotherMethod(_, _, _, _)).WillOnce(…)
```
may instead be written as:

```
ON_CALL(mock, Method).WillByDefault(…)
EXPECT_CALL(mock, AnotherMethod).WillOnce(…)
```

This syntax is only available for non-overloaded methods; attempting to use the parameterless macro form with an overloaded method results in compiler error. It is conceivable that this approach could be extended to support some kinds of expectations on overloaded methods, but that is beyond the scope of this change.


dnsunderland:
I'm going to clean up my change on a new branch and resubmit the PR.

PR #1589, opened 2018-04-27
Add no-exception and no-rtti to build matrix

sgraham:
(I'm not sure if this is right yet, I'm hoping creating a pull request will run a Travis build?)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,6 +37,9 @@ matrix:
       group: deprecated-2017Q4
       compiler: clang
       env: BUILD_TYPE=Release VERBOSE=1 CXX_FLAGS=-std=c++11
+    - os: linux
+      compiler: clang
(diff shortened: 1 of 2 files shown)


googletest/cmake/internal_utils.cmake
```diff
@@ -94,6 +94,7 @@ macro(config_compiler_and_linker)
     set(cxx_no_exception_flags "-EHs-c- -D_HAS_EXCEPTIONS=0")
     set(cxx_no_rtti_flags "-GR-")
   elseif (CMAKE_COMPILER_IS_GNUCXX)
+    message("in here")
```
gennadiycivil: I think it may be better to say message ("CMAKE COMPILER IS GNUCXX")

sgraham: Oh, sorry, I wasn't planning to land any of the message()s, I was just trying to figure out why I had to manually pass CMAKE_COMPILER_IS_GNUCXX since it seemed like the sort of thing that should be automatic. I'll revert them out.

sgraham:
Removed message() calls.

PR #1590, opened 2018-05-01
Avoid using environ on FreeBSD

lantw44:
WebKit passes -Wl,--no-undefined to the compiler when linking shared
libraries. This is usually a good practice, but it causes gtest to fail
to link because of the use of environ. FreeBSD puts environ symbol in
crt1.o instead of libc.so, so the linker is unable to find environ when
linking a shared library.

Fortunately, there is an easy way to fix it. Since the only use of
environ in gtest is to pass it to execve, we can simply replace execve
with execv to avoid the need of environ. execv is usually a simple
wrapper around execve, so it should be safe to use.

This problem was found and reported more than three years ago. The same
fix is already committed to WebKit and we don't see any problem reports
caused by the change.

https://bugs.webkit.org/show_bug.cgi?id=138420
https://trac.webkit.org/changeset/194501/webkit
https://groups.google.com/forum/#!topic/googletestframework/wrrMj_fmXMc

### DIFF
--- a/googletest/src/gtest-death-test.cc
+++ b/googletest/src/gtest-death-test.cc
@@ -940,21 +940,9 @@ struct ExecDeathTestArgs {
   int close_fd;       // File descriptor to close; the read end of a pipe
 };
 
-#  if GTEST_OS_MAC
-inline char** GetEnviron() {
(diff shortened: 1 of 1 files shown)


gennadiycivil:
No CLA

Issue #1554, opened 2018-04-09
Broken with -fno-rtti


sgraham:
Build currently fails when -fno-rtti is specified on the command line. git bisect says it started at

https://github.com/google/googletest/commit/0cd6a4f5f585f4a58a24254dc4cc81675edba16c

```
$ git bisect log
git bisect start
# bad: [7529698fa378528ecc8415ac9ea0c04060281409] Merge pull request #1552 from pwnall/mock-pump
git bisect bad 7529698fa378528ecc8415ac9ea0c04060281409
# good: [11855d8ef08567c7961fb3c6562fbe6f9b479092] provide alternative for DebugBreak()
git bisect good 11855d8ef08567c7961fb3c6562fbe6f9b479092
# good: [98687b7847eb357b5bea1c7114bc35703b36f8e1] Merge branch 'master' into int3
git bisect good 98687b7847eb357b5bea1c7114bc35703b36f8e1
# good: [f7098a2a4dc2e3233251c4f8e8001ee773cd5a61] Merge branch 'master' into master
git bisect good f7098a2a4dc2e3233251c4f8e8001ee773cd5a61
# bad: [e0f4cf05614a90133b5912f149c1f59fdaf3aff6] fixing MCVS warn
git bisect bad e0f4cf05614a90133b5912f149c1f59fdaf3aff6
# bad: [47d15bceea62696b856e203c236cdd95b285d73c] Merge branch 'master' of github.com:google/googletest
git bisect bad 47d15bceea62696b856e203c236cdd95b285d73c
# good: [2cf2a1f8ce9e25a2c0d053a51a3e67b3621f025b] Merge pull request #1545 from gennadiycivil/master
git bisect good 2cf2a1f8ce9e25a2c0d053a51a3e67b3621f025b
# bad: [0cd6a4f5f585f4a58a24254dc4cc81675edba16c] Merging matchers test
git bisect bad 0cd6a4f5f585f4a58a24254dc4cc81675edba16c
# first bad commit: [0cd6a4f5f585f4a58a24254dc4cc81675edba16c] Merging matchers test
```

PR #1591, opened 2018-05-02
Fix gmock not building when -fno-rtti

sgraham:
Fixes issue #1554. This is internal cl/195020996.


sgraham:
Hi, sorry to bug you, could I trouble you to take a look at this PR? I believe it's identical to the upstream internal CL.

PR #1593, opened 2018-05-03
Add death test support for Fuchsia.

Steelskin:
This add support for Fuchsia death tests and ensures all tests are passing on Fuchsia.
In addition, this implements GetThreadCount() for Fuchsia.


sgraham:
Hi, the #if around g_in_fast_death_test_child also needs to be !GTEST_OS_FUCHSIA similar to Windows, otherwise, the Fuchsia build errors out with an unused variable warning.

gennadiycivil:
@Steelskin pls take a look

Steelskin:
Ah I'm not sure why I didn't catch it when doing a chromium build with this change.
Here is the change: https://github.com/google/googletest/pull/1603

PR #1601, opened 2018-05-11
Std lib versioning

jdennett:
Add support for versioned standard libraries.

This canonicalizes demangled names by omitting a nested inline namespace within namespace std if the name of the nested namespace begins with a double underscore.  This improves compatibility with libc++.


jdennett:
This is ready for review.

PR #1606, opened 2018-05-23
#1595: fix compiler error with Visual Studio 2017 on Win10 JP.

m-tmatma:
#1595: fix compiler error with Visual Studio 2017 on Win10 JP.

* remove non-ASCII charactors.


Kuniwak:
I faced the same problem too. What is the status of this PR?

m-tmatma:
> What is the status of this PR?

At the time of my PR creation, all builds were successful.
but after 9080833, the travis builds have failed.

m-tmatma:
I made an empty commit to reinvoke CI builds.

m-tmatma:
@gennadiycivil 

Could you review this PR?

m-tmatma:
@janisozaur

> We faced a similar issue in another project. We solved it by enabling utf-8-by-default mode in MSVC (-utf-8) which is much nicer way deal to with this deficiency of MSVC compiler rather than preventing the use of standard UTF8. It is 2018 and this is not unreasonable expectation.

If you want to fix this issue by your method,  please send a PR to add encoding option to CMakeLists.txt of googletest.

But your solution may affect broader area of codes.
My solution doesn't need to be cared about the influence of the change.

I think b539167cf0254f521b791e908f6d3a5ff3f30245 doesn't have  enough
reason to change charactor setting of a compiler.

> You could also clean up this PR by containing it to a single commit.

I rebased this PR to remove the empty commit.

m-tmatma:
I changed the commit log.

berryzplus:
This request will replace a character ```'...'(\u2026)``` to a string ```"..."``` because vs2017 miss-recognize encoding of a file [```gmock-generated-function-mockers.h```](https://github.com/google/googletest/blob/b539167cf0254f521b791e908f6d3a5ff3f30245/googlemock/include/gmock/gmock-generated-function-mockers.h) that has no Byte Order Mark.

I hope you to accept this request and merge it. You can also fix the generator of ```gmock-generated-function-mockers.h``` to generate Byte Order Mark.

m-tmatma:
Before b539167cf0254f521b791e908f6d3a5ff3f30245, gmock-generated-function-mockers.h and gmock-generated-function-mockers.h.pump didn't have non-ASCII charactors.

Since b539167cf0254f521b791e908f6d3a5ff3f30245, they have UTF-8 charactors.
But as @berryzplus explained at the above messsage, they don't have Byte Order Mark.

Japapanese version of Visual Studio can't recognize them as UTF-8,
tries to interpret them as Shift-JIS, and find unrecognized charactors as Shift-JIS.

This PR will remove non-ASCII charactors from them because the charactors are used in a comment
and there is not enough reason to use non-ASCII charactors.

But if you want to use the charactors, you can save them as UTF-8 BOM.

m-tmatma:
@gennadiycivil

Could you check this PR?

PR #1609, opened 2018-05-24
std references shouldn't be fully qualified

laixer:


gennadiycivil:
@laixer could you please resolve conflicts

PR #1619, opened 2018-05-31
Remove launchpad dependency from Fuchsia.

Steelskin:


Steelskin:
Internal CL is at http://cl/198652750

PR #1620, opened 2018-06-04
Adress, #1616, add  printer for std::nullptr_t

gennadiycivil:
Add printer for std::nullptr_t, 
Addressing #1616 
Upstream, cl/199129756


andschwa:
May I ask, why is this guarded with `GTEST_LANG_CXX11`? Doesn't this fix need to be applied for 14, 17, and anything newer (e.g. MSVC's "latest") too?

PR #1627, opened 2018-06-12
Reduce the number of strcmp calling while initialization

atetubou:
When we do parallel test execution with a process for a test,
initialization of gtest become performance bottleneck when the test
binary contains many testcases.

Especially, some parameterlized test in chromium browser affected by
largely when address sanitizer is enabled.
Address sanitizer does not allow using optimized strcmp function and
test addition in parameterized test require lookup of test case using
strcmp.

This patch reduces the number of strcmp, it is called when registering
parameterized test. Using reverse iterator improves the time to find
registered tests in such case.

Some tests for chromium browser using address sanitizer finished 2x
faster with this patch.


nico:
(I didn't understand why this helped. https://bugs.chromium.org/p/chromium/issues/detail?id=794372#c28 explains it.)

PR #1629, opened 2018-06-12
fix #907: add support for std::string_view

### LINKED ISSUE #907: better std::experimental::string_view support
std::experimental::string_view has operator<< overload for std::ostream but still printed as collection of chars. The only way to make gtest print string_view value as string is add PrintTo into std::experimental::fundamentals_v1 namespace which is not really great practice.

dermojo:
Implementation is identical to absl::string_view.


dermojo:
Wrong username in the commit - conflicts with the CLA. Will update and create a new PR.

PR #1630, opened 2018-06-12
fix #907: add support for std::string_view for C++17

### LINKED ISSUE #907: better std::experimental::string_view support

dermojo:
Implementation is identical to absl::string_view.


gennadiycivil:
Apologies for the long wait. googletest can conditionally depend on absl:: so absl:string_view would be available

KindDragon:
@gennadiycivil This PR for **stl**:string_view. Please check PR again

gennadiycivil:
Aplogies for not explaining this clear-er. absl::string_view IS std:string_view under C++ 17
See https://abseil.io/blog/20170926-welcome-to-abseil. 
Thanks again

gennadiycivil:
... and https://abseil.io/about/design/dropin-types

dermojo:
Thanks for taking the time to explain. Maybe I should have done the same.
The fact that both types are identical doesn't seem to help - when using std::string_view with GTest (e.g. in an ASSERT_EQ), it will be printed as a plain array of int's. Maybe this works when absl is actually used and included, but it doesn't if not.

Sample code:
```
#include <gtest/gtest.h>
#include <string_view>

TEST(Foo, StringView)
{
   std::string_view s1 = "Hello";
   std::string_view s2 = "World";
   ASSERT_EQ(s1, s2);
}
```
This results in the following test error:
```
ex.cpp:8: Failure
      Expected: s1
      Which is: { 'H' (72, 0x48), 'e' (101, 0x65), 'l' (108, 0x6C), 'l' (108, 0x6C), 'o' (111, 0x6F) }
To be equal to: s2
      Which is: { 'W' (87, 0x57), 'o' (111, 0x6F), 'r' (114, 0x72), 'l' (108, 0x6C), 'd' (100, 0x64) }
```

But it should look like this (same output as using std::string):
```
ex.cpp:8: Failure
      Expected: s1
      Which is: "Hello"
To be equal to: s2
      Which is: "World"
```

dermojo:
@gennadiycivil Bump (not sure if my latest comment triggered a notification)

gennadiycivil:
@dermojo You are correct, I have not seen your previous notification. I ll re-open this PR , I have to apologize in advance because it will be some time before I can get back to it, we have higher priorities currently with 1.8.x release and cleanup work.

dermojo:
Thanks! Don't worry, take your time.

Jonny007-MKD:
@dermojo: You could instead implement this PR for absl::string_view, so that anybody who is using absl::string_view with pre-C++17 compilers would also have nice output in gtest, whereas everybody with C++17 would in fact be having code for std::string_view (see https://abseil.io/about/design/dropin-types)

dermojo:
@Jonny007-MKD: Sorry, but I don't understand your comment - there already is support for absl::string_view in GoogleTest.

Jonny007-MKD:
@dermojo: In this case std::string_view is also already supported, because absl::string_view == std::string_view if available.

dermojo:
@Jonny007-MKD: So if I understand correctly, you're requesting to merge the support code for both string_view types? That would be totally reasonable :)
I just need to make sure that all conditional compilations still work (with/without ABSL, pre/post-C++17).

dermojo:
@Jonny007-MKD : I updated the implementatio to alias either absl::string_view or std::string_view into the ::testing namespace, so that the rest of the code doesn't need to distinguish. Hope this now looks as expected.

googletest/include/gtest/gtest-printers.h
```diff
@@ -117,8 +117,28 @@
...
+#if GTEST_HAS_STD_STRINGVIEW
+#include <string_view>
+#endif  // GTEST_HAS_STD_STRINGVIEW
+
 namespace testing {
 
+// create an alias for string_view, wherever it comes from
+#if GTEST_HAS_ABSL
+
+using absl::string_view;
```
shahzadlone: This is a neat trick! but is this a good idea to do this in a .h file ? hope this doesn't cause future contributors confusion or additional linking problems.

dermojo: That saves a lot of `#if's`, but I get your point. How about assigning the typedef to a dedicated name? E.g.:
`using gtest_string_view = /*...*/`

shahzadlone: perhaps that would be better

dermojo:
@shahzadlone Updated the typedef to use `StringView` instead. This should avoid any confusion (I hope)

kamolsrisan:
Good

Oooocean:
#-   

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

dermojo:
Related issue: #907

davidzchen:
I am still getting the same issue when comparing two `absl::string_view` instances. Also added a comment on #2437

PR #1645, opened 2018-06-27
FIX: Compilation warning. Initialize MutexBase's owner_ field to 0.

tisi1988:
This fixes a warning from GCC (using Qt Creator and Qt 5.9) that complains about a member not being initialised. Since the owner_ attribute is an unsigned long int redefined as pthread_t we can set the value to 0.


tisi1988:
I signed it!

PR #1646, opened 2018-06-27
FIX: Compilation warning with GCC regarding a non-initialised member …

tisi1988:
This fixes a warning from GCC (using Qt Creator and Qt 5.9) that complains about a member not being initialised. Since the owner_ attribute is an unsigned long int redefined as pthread_t we can set the value to 0.


gennadiycivil:
Is this still needed after https://github.com/google/googletest/pull/1646?

PR #1652, opened 2018-07-09
Cast the tr1::tuple_element template parameter to int

medithe:
Because in `std::tr1::tuple_element` the first template parameter should be of type int (https://gcc.gnu.org/onlinedocs/libstdc++/libstdc++-html-USERS-4.4/a00547.html), but the code inserts a size_t, the first template parameter should be casted to int before, to get rid of the following errors:

googletest-src/googletest/include/gtest/gtest-printers.h:957:60: error: conversion from ‘long unsigned int’ to ‘int’ may change value [-Werror=conversion]
   struct tuple_element : ::std::tr1::tuple_element<I, Tuple> {};

and 

googletest-src/googletest/include/gtest/gtest-printers.h:961:56: error: conversion from ‘long unsigned int’ to ‘int’ may change value [-Werror=conversion]
       const typename ::std::tr1::tuple_element<I, Tuple>::type>::type get(

compiled with g++8.1 with the flag `-Wconversion`.


medithe:
I signed it!

Issue #1654, opened 2018-07-11
gmock_ex_test.cc fails if exceptions are disabled


AdrianMoranMontes:
If preprocessor macro GTEST_HAS_EXCEPTIONS is set to 0, corresponding test does not compile.

PR #1655, opened 2018-07-11
Fix issue #1654.

AdrianMoranMontes:
Fix issue #1654.

Signed-off-by: Adrian Moran <amoran@ikerlan.es>

### DIFF
--- a/googlemock/test/gmock_ex_test.cc
+++ b/googlemock/test/gmock_ex_test.cc
@@ -31,15 +31,14 @@
 
 // Tests Google Mock's functionality that depends on exceptions.
 
+#if GTEST_HAS_EXCEPTIONS
 #include "gmock/gmock.h"
 #include "gtest/gtest.h"
 
 namespace {
 
 using testing::HasSubstr;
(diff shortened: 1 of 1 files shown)


AdrianMoranMontes:
Hi @gennadiycivil, the cahnge you requested is done and pushed.

googlemock/test/gmock_ex_test.cc
```diff
@@ -31,6 +31,7 @@
 
 // Tests Google Mock's functionality that depends on exceptions.
 
+#if GTEST_HAS_EXCEPTIONS
```
gennadiycivil: Please move the #if after the #includes. After all the GTEST_HAS_EXCEPTIONS is defined in [the header](https://github.com/google/googletest/blob/41f0e24335d8874cc87420a6349bb1844e2a15de/googletest/include/gtest/internal/gtest-port.h#L475) included in the header file, otherwise it will never be defined

Thank you again

AdrianMoranMontes:
You are rigth sorry, now it is fixed.

PR #1658, opened 2018-07-11
New command line parameter: Allows to see the location where tests are defined

### LINKED ISSUE #1657: Add Information about test location in list_tests output
In order to improve the UX of tools written over GoogleTest, it would be useful to know what the location of a specific test is (the file in which it is defined, and the line number). GoogleTest already has this information, but doesn't expose it to the user. In order not to break existing uses of GoogleTest, we could provide a new command line switch (list_tests_with_location for example) to display this information.

DavidSchuldenfrei:
New command line parameter: list_tests_with_location.
Allows to see the location where tests are defined.
Fixes #1657

### DIFF
--- a/googletest/include/gtest/gtest.h
+++ b/googletest/include/gtest/gtest.h
@@ -116,6 +116,11 @@ GTEST_DECLARE_string_(filter);
 // are actually run if the flag is provided.
 GTEST_DECLARE_bool_(list_tests);
 
+// This flag causes the Google Test to list tests with their location 
+// (file and line number). None of the tests listed
+// are actually run if the flag is provided.
+GTEST_DECLARE_bool_(list_tests_with_location);
+
 // This flag controls whether Google Test emits a detailed XML report to a file
 // in addition to its normal textual output.
 GTEST_DECLARE_string_(output);
--- a/googletest/src/gtest-internal-inl.h
+++ b/googletest/src/gtest-internal-inl.h
@@ -84,6 +84,7 @@ const char kCatchExceptionsFlag[] = "catch_exceptions";
 const char kColorFlag[] = "color";
 const char kFilterFlag[] = "filter";
 const char kListTestsFlag[] = "list_tests";
+const char kListTestsWithLocationFlag[] = "list_tests_with_location";
 const char kOutputFlag[] = "output";
 const char kPrintTimeFlag[] = "print_time";
 const char kPrintUTF8Flag[] = "print_utf8";
@@ -165,6 +166,7 @@ class GTestFlagSaver {
     filter_ = GTEST_FLAG(filter);
     internal_run_death_test_ = GTEST_FLAG(internal_run_death_test);
     list_tests_ = GTEST_FLAG(list_tests);
+    list_tests_with_location_ = GTEST_FLAG(list_tests_with_location);
     output_ = GTEST_FLAG(output);
     print_time_ = GTEST_FLAG(print_time);
     print_utf8_ = GTEST_FLAG(print_utf8);
@@ -187,6 +189,7 @@ class GTestFlagSaver {
     GTEST_FLAG(filter) = filter_;
     GTEST_FLAG(internal_run_death_test) = internal_run_death_test_;
     GTEST_FLAG(list_tests) = list_tests_;
+    GTEST_FLAG(list_tests_with_location) = list_tests_with_location_;
     GTEST_FLAG(output) = output_;
     GTEST_FLAG(print_time) = print_time_;
     GTEST_FLAG(print_utf8) = print_utf8_;
@@ -209,6 +212,7 @@ class GTestFlagSaver {
   std::string filter_;
   std::string internal_run_death_test_;
   bool list_tests_;
+  bool list_tests_with_location_;  
   std::string output_;
   bool print_time_;
   bool print_utf8_;
@@ -722,7 +726,7 @@ class GTEST_API_ UnitTestImpl {
   int FilterTests(ReactionToSharding shard_tests);
 
   // Prints the names of the tests matching the user-specified filter flag.
-  void ListTestsMatchingFilter();
+  void ListTestsMatchingFilter(bool list_location);
 
   const TestCase* current_test_case() const { return current_test_case_; }
   TestInfo* current_test_info() { return current_test_info_; }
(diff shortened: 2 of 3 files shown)


DavidSchuldenfrei:
I signed it!

gennadiycivil:
Could you please provide detailed usage example.

DavidSchuldenfrei:
I wrote an extension for VsCode (https://marketplace.visualstudio.com/items?itemName=DavidSchuldenfrei.gtest-adapter) which shows all google tests in a tree. The ability to "Go to Test" where I would open the file at the line where the test is defined, from a specific leaf was requested as an enhancement. This PR would allow me to provide this feature.

gennadiycivil:
Thank you for the explanation.
We agree that ability to pro grammatically tell the interested tools about the tests meta information such as location is a good addition to googletest.
However the implementation could be tweaked. Please consider the following:
Currently you are using --gtest_list_tests that outputs the test info on the command-like. It it clearer and more maintainable to use structured format (XML) for this purpose 

Please implement new option using --gtest_list_tests and exising flag gunit_output. Currently gtest_list_tests ignores this flag but you can make the change so it can be called like so:

--gtest_list_tests --gunit_output=xml:/tmp/xxx.txt that would spit out structured form where the tools can parse it.

Thank you
G

DavidSchuldenfrei:
I thought of such an approach. However, if someone has a GTEST_OUTPUT variable set up, to always output to xml, this approach would be a breaking change, as querying for existing tests, via list_tests would no longer output anything to the console. How do you suggest we handle such an issue?

gennadiycivil:
We could output XML **in addition to** outputting to the console not **instead of**. In other words, existing behaviour would not change....
Can you see issues with this?

DavidSchuldenfrei:
If someone has env setup so that each run outputs to the same file, then running with list_tests would unexpectedly delete the data in that file.

gennadiycivil:
Think this:

The output of the XML file that googletest produces already have some meta info about the testsuite, for example if I run bazel test gtest-options_test --gtest_output=xml:/tmp/xxx.txt 
Here is the snippet for the file produced:
Note the "<testsuite " node. 

All you would have to do is to add more information into the node. This would re-use already existing ability to generate XML , not adding any new flags ( there are already too many of them)

<?xml version="1.0" encoding="UTF-8"?>
<testsuites tests="12" failures="0" disabled="0" errors="0" timestamp="2018-07-11T15:56:48" time="0.009" name="AllTests">
  <testsuite name="XmlOutputTest" tests="5" failures="0" disabled="0" errors="0" time="0.004"   **ADD MORE HERE such as path**   >
    <testcase name="GetOutputFormatDefault" status="run" time="0.001" classname="XmlOutputTest" />
 .....

 You argument "running with list_tests would unexpectedly delete the data in that file."  - 
If you run googletest with 
--gunit_list_tests --gunit_output=xml:/tmp/xxx.txt it would only list tests. Its either list tests or run the tests , not both

DavidSchuldenfrei:
My concern was, that if someone has an environment variable set, for example GTEST_OUTPUT=xml,
 as the help states: 

`
you can alternatively set the corresponding
environment variable of a flag (all letters in upper-case). For example, to
disable colored text output, you can either specify --gtest_color=no or set
the GTEST_COLOR environment variable to no`
Then when running **without** specifying any output but just with --gtest_list_tests would erase the data in the file test_details.xml.
If you think that this is an edge case, which could just be documented, (I agree), I have just submitted a PR, which solves the issue using your suggestion.
I currently don't output any extra information when outputting run information, not to break any tests.
I can add the line/file information in all cases in a later commit.

gennadiycivil:
Thank you again for working on this. I dont belive that googletest is making a promise to preserve any state output information between the runs. As such each run may modify the output.
However I agree that  we should include a paragraph in the documentation that would describe this situation, could you please add it
Also I think we need a unit test.
Thanks again

DavidSchuldenfrei:
Also, do you have any idea why the Travis CI is failing? I am not familiar with Travis

DavidSchuldenfrei:
Also, regarding unit tests, the way that the file is displayed is OS/Environment specific (windows has \, others use /), and may depend on where the project is compiled.
Do you have any smart suggestions on how to test the otput of JSON/xml? using Regex to modify the output? Or do you have another idea?

gennadiycivil:
1) Unit test , please see already existing XML test [here](https://github.com/google/googletest/blob/ba96d0b1161f540656efdaed035b3c062b60e006/googletest/test/BUILD.bazel#L310), I think you can follow the same strategy
2) Documentation, lets put aside for now
Thank you again

DavidSchuldenfrei:
Thank you, I will try to take a look at it tomorrow

DavidSchuldenfrei:
I added unit tests

DavidSchuldenfrei:
@gennadiycivil I now have unit tests. Do you need anything else to review/merge to main?
Thanks,

gennadiycivil:
@DavidSchuldenfrei Could you please resolve conflicts, the source has moved
Thank you again

DavidSchuldenfrei:
@gennadiycivil I resolved all conflicts.

DavidSchuldenfrei:
@gennadiycivil I saw you closed this. Would you mind sharing why you won't merge? Is there anything I can do to fix any issues? Thanks

gennadiycivil:
@DavidSchuldenfrei It is actually merged. Take a look at the commit https://github.com/google/googletest/commit/603ce4b81df48a268b84d0e4f2b999b89b903250 - they are mostly your changes with the small format and restucturing. You should also see that the changes are properly attributed to you as an author.  The "Closed" tag is misleading in this case and I am hoping to find a better way shortly

DavidSchuldenfrei:
@gennadiycivil Thank you SO much. May I suggest, for future such cases, adding a comment giving the relevant commit to avoid disappointments.
Also, what are the policies if I want to update the Readme.md file (https://github.com/google/googletest/blob/master/README.md), the section "Related Open Source Projects" to link to an extension that I wrote for VsCode (https://marketplace.visualstudio.com/items?itemName=DavidSchuldenfrei.gtest-adapter), which will now use this new feature.
Should I create a new PR, or do you prefer not to clutter this Readme with too many "advertising pitches"?

gennadiycivil:
@DavidSchuldenfrei Sorry for the confusion about Closed/Merged and thank you for your understanding.

Re: Related Open Source projects - I don't see why not, please send a PR. The only thing I would suggest, use proper name "googletest" , GTest is not really a correct name I think

G

PR #1664, opened 2018-07-12
vs2017 solution and project files

c-w-m:
add VS2017 support


gennadiycivil:
The VS soluition would be gener

PR #1665, opened 2018-07-13
Disable MSVC function deprecation when using Clang

rongjiecomputer:
`clang-cl` does not understand `/wd4996` or `#pragma warning(disable:4996)`, so we use the usual clang way to disable deprecation warning.


gennadiycivil:
Thank you for your contribution. Can you please explain your use case? The MSVC mcro should only be relevatn when compiling with MSVC

rongjiecomputer:
`clang-cl` will consume MSVC's C headers and STL headers. MSVC's C headers add deprecation warnings which `clang-cl` understand, but `clang-cl` so far does not understand `#pragma warning(disable:4996)` or `/wd4996` (in fact, `clang-cl` just silently ignore all `/wd*` flags), so those warnings are not disabled for `clang-cl`. `clang-cl` needs `-Wdeprecated-declarations` and `-Wdeprecated-implementations`.

This prevents `clang-cl` users from enabling `/WX`.

rongjiecomputer:
Ping @gennadiycivil

PR #1669, opened 2018-07-17
Ignore .DS_Store file

syohex:
I use `googletest` as Xcode subproject(git submodule). Sometimes `.DS_Store` file is generated in `googletest/xcode` directory and then git submodule status becomes dirty as below. 

ds_store

I think that there is no problem to add `.DS_Store` into `.gitignore` because no one needs it.

Thanks for great product.
Regards.


syohex:
Thanks!!

PR #1675, opened 2018-07-20
Fixed some URLs and changed others to HTTPS

einsteinsfool:
Runned `grep -rnw . -e 'http:/'` in the repo directory and changed all occurrences of "http://" to "https://" if the page was accessible via HTTPS.

Checked all URLs manually. Tried both with and without "www". Skipped licences. Didn't change URLs when only HTTP version was working. Sometimes the page was returning 404 so I used the latest version from [Internet Archive](https://web.archive.org). I also tried googling some text from the Internet Archive version to see if there's a live one and if there was, I used its URL.

TODO after this is merged (so I don't forget):
- in [gmock_main.cc](https://github.com/google/googletest/blob/master/googlemock/src/gmock_main.cc#L36) the condition is probably no longer needed since the bug was fixed by Microsoft in Visual Studio 2010
- make an issue for someone to update [XcodeGuide.md](https://github.com/google/googletest/blob/master/googletest/docs/XcodeGuide.md) since it's using svn instead of git and links to non-existing svn repo
- make an issue for someone to fix [this URL](https://github.com/google/googletest/blame/master/googletest/docs/faq.md#L710) and [this one](https://github.com/google/googletest/blob/master/googletest/src/gtest.cc#L5171) as I was not able to fix a good replacement

### DIFF
--- a/googlemock/docs/CookBook.md
+++ b/googlemock/docs/CookBook.md
@@ -2247,7 +2247,7 @@ enum class AccessLevel { kInternal, kPublic };
 
 class Buzz {
  public:
-  explicit Buzz(AccessLevel access) { … }
+  explicit Buzz(AccessLevel access) { ... }
   ...
 };
 
@@ -2320,7 +2320,7 @@ Note that `ByMove()` is essential here - if you drop it, the code won’t compil
 
 Quiz time! What do you think will happen if a `Return(ByMove(...))` action is
 performed more than once (e.g. you write
-`….WillRepeatedly(Return(ByMove(...)));`)? Come think of it, after the first
+`.WillRepeatedly(Return(ByMove(...)));`)? Come think of it, after the first
 time the action runs, the source value will be consumed (since it’s a move-only
 value), so the next time around, there’s no value to move from -- you’ll get a
 run-time error that `Return(ByMove(...))` can only be run once.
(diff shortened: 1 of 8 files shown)


einsteinsfool:
I signed it!

einsteinsfool:
So, some links were breaked. Which means that they were leading to 404 or redirecting to the main page. Why have links in the code that don't lead to the place indended by the person who created them?

As for the URLs that worked, I changed them to HTTPS because it looks more professional. Imo we should start thinking that 'http://*' = bad, and switch the such occurrences whenever possible. Another argument is that some of the URLs I changed don't support HSTS, so if somebody goed to HTTP version, they wouldn't be redirected to HTTPS. And on HTTP websites malicious scripts can be runned with MITM attack (like mining a cryptocurrency, displaying a porn ad or even changing some URLs to phishing websites).

I don't see any reason not to make these changes.

gennadiycivil:
There are unrelated things in this change.
1) Broken links - yes this is useful and thank you for noticing and taking the time to fix
1) HTTPs changes, completely unrelated and different dicussion.

Could you please remove HTTPs changes from this PR and just live broken links fixes . Then we can review

einsteinsfool:
So I should divide this PR into 2 separate ones: fixed URLs and URLs changed to HTTPS?
Would both be accepted if no problems are finded or would the HTTPS PR be ignored as it's not important?

gennadiycivil:
While broken links are useful and important the HTTPs changes need to be looked at more closely. I would advise to do the broken links first .
Thank you

Issue #1628, opened 2018-06-12
Seed information within output files for reporting


stoesselt:
It would be great if the random seed information could be output in the XML and json formats as it is already available in the console output. 

In this way, the output files would contain the seed information in order to repeat the test run exactly in that order (assuming shuffled execution is activated). Use case is reporting and archiving issues. I use a script taking json output file as input plus some configuration of the executing environment and generate a PDF report which is archived. I need to be able to repeat the test run in exactly the same order.

stoesselt:
OK, at least for json format this is already implemented. For my concerns, this is enough and sorry for this nonsense issue.

PR #1682, opened 2018-07-24
Improve CMake exported targets

dakotahawkins:
I _think_ this represents some of the "best practices" for exporting
targets. They'll be available in a `googletest::` namespace (e.g.
`googletest::gmock`) with non-namespaced `ALIAS` targets.

- Added GOOGLETEST_VERSION variable
- Use `CMakePackageConfigHelpers`, bump minimum CMake version to 2.8.8

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,22 +1,16 @@
-cmake_minimum_required(VERSION 2.6.4)
+cmake_minimum_required(VERSION 2.8.8)
 
 if (POLICY CMP0048)
   cmake_policy(SET CMP0048 NEW)
 endif (POLICY CMP0048)
 
-project( googletest-distribution )
+project(googletest-distribution)
+set(GOOGLETEST_VERSION 1.9.0)
 
 enable_testing()
 
 include(CMakeDependentOption)
-if (CMAKE_VERSION VERSION_LESS 2.8.5)
-  set(CMAKE_INSTALL_BINDIR "bin" CACHE STRING "User executables (bin)")
-  set(CMAKE_INSTALL_LIBDIR "lib${LIB_SUFFIX}" CACHE STRING "Object code libraries (lib)")
-  set(CMAKE_INSTALL_INCLUDEDIR "include" CACHE STRING "C header files (include)")
-  mark_as_advanced(CMAKE_INSTALL_BINDIR CMAKE_INSTALL_LIBDIR CMAKE_INSTALL_INCLUDEDIR)
-else()
-  include(GNUInstallDirs)
-endif()
+include(GNUInstallDirs)
 
 option(BUILD_GTEST "Builds the googletest subproject" OFF)
 
@@ -26,8 +20,31 @@ option(BUILD_GMOCK "Builds the googlemock subproject" ON)
 cmake_dependent_option(INSTALL_GTEST "Enable installation of googletest. (Projects embedding googletest may want to turn this OFF.)" ON "BUILD_GTEST OR BUILD_GMOCK" OFF)
 cmake_dependent_option(INSTALL_GMOCK "Enable installation of googlemock. (Projects embedding googlemock may want to turn this OFF.)" ON "BUILD_GMOCK" OFF)
 
+if(WIN32)
+  set(INSTALL_CMAKE_DIR "cmake" CACHE PATH "CMake exported targets")
+else()
+  set(INSTALL_CMAKE_DIR "${CMAKE_INSTALL_LIBDIR}/cmake/gtest" CACHE PATH "CMake exported targets")
+endif()
+
+set(googletest_install_targets)
 if(BUILD_GMOCK)
   add_subdirectory( googlemock )
 elseif(BUILD_GTEST)
   add_subdirectory( googletest )
+endif()
+
+if(googletest_install_targets)
+  include(CMakePackageConfigHelpers)
+  configure_package_config_file(
+    "${CMAKE_CURRENT_LIST_DIR}/cmake/googletestConfig.cmake.in"
+    "${CMAKE_CURRENT_BINARY_DIR}/googletestConfig.cmake"
+    INSTALL_DESTINATION "${INSTALL_CMAKE_DIR}")
+  write_basic_package_version_file(
+    ${CMAKE_CURRENT_BINARY_DIR}/googletestConfigVersion.cmake
+    VERSION "${GOOGLETEST_VERSION}"
+    COMPATIBILITY SameMajorVersion)
+  install(FILES
+    ${CMAKE_CURRENT_BINARY_DIR}/googletestConfig.cmake
+    ${CMAKE_CURRENT_BINARY_DIR}/googletestConfigVersion.cmake
+    DESTINATION "${INSTALL_CMAKE_DIR}")
 endif()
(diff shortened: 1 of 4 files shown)


gennadiycivil:
@dakotahawkins Thank you for this contribution. Could you please explain why this change would be beneficial.

dakotahawkins:
@gennadiycivil Sure, I can try.

From a usability perspective, exporting the targets in this way should be beneficial in a few ways:
- `find_package` will work without a custom find module, as long as it finds `googletestConfig.cmake`. I believe these config files are "preferred" over find modules these days. Using `CMakePackageConfigHelpers` gets you those plus related and standard version config files, etc.
- Using namespaces for your targets is probably a good idea (for all the reasons it's a good idea in code).  Personally, I like using namespaced targets more than non-namespaced targets and it gives me some indication that the CMake providing the targets is probably reasonably OK.
- "Laundering" the for-use `googletest::` namespaced targets through `googletest_` namespaces is a little harder for me to explain... I first saw this in XercesC and had a Kitware CMake trainer say it was a good idea when I asked about it. I _think_ it prevents users from accessing more than they need to through the `googletest::` targets (since they're `INTERFACE IMPORTED`) and maybe that's all there is to it.

Then from a maintainability perspective I think it's just closer to "modern" CMake than it was, with could be a good thing in the future.

dakotahawkins:
Looking at #1338, perhaps the namespace should be changed to `GTest::` to be consistent with [`FindGTest.cmake`](https://gitlab.kitware.com/cmake/cmake/blob/master/Modules/FindGTest.cmake#L206).

TODO:

- [ ] Use `GTest::` namespace and make sure exported target names are consistent with CMake's find module

gennadiycivil:
the project name is googletest ... not sure what GTest is. The #1338 needs to have conflicts resolved very shortly

CMakeLists.txt
```diff
@@ -1,22 +1,16 @@
-cmake_minimum_required(VERSION 2.6.4)
+cmake_minimum_required(VERSION 2.8.8)
 
 if (POLICY CMP0048)
   cmake_policy(SET CMP0048 NEW)
 endif (POLICY CMP0048)
 
-project( googletest-distribution )
+project(googletest-distribution)
+set(GOOGLETEST_VERSION 1.9.0)
```
gennadiycivil: Where did 1.9.0 come from?

CMakeLists.txt
```diff
@@ -1,22 +1,16 @@
-cmake_minimum_required(VERSION 2.6.4)
+cmake_minimum_required(VERSION 2.8.8)
```
gennadiycivil: What is the rational of changing the version?

dakotahawkins:
@gennadiycivil 

> the project name is googletest ... not sure what GTest is. The #1338 needs to have conflicts resolved very shortly

`GTest` came from whoever wrote the CMake find module for google test. If this project didn't have anything to do with it, and in a vacuum, I'd prefer `googletest` (and then I wouldn't have to change anything).

It might be worthwhile attempting to maintain compatibility with the CMake find module, however, since it may minimize the impact to anybody relying on those target names and CMake variables.

googletest/CMakeLists.txt
```diff
@@ -44,7 +44,7 @@ if (CMAKE_VERSION VERSION_LESS 3.0)
   project(gtest CXX C)
 else()
   cmake_policy(SET CMP0048 NEW)
-  project(gtest VERSION 1.9.0 LANGUAGES CXX C)
+  project(gtest VERSION ${GOOGLETEST_VERSION} LANGUAGES CXX C)
```
dakotahawkins: @gennadiycivil This is where 1.9.0 came from

CMakeLists.txt
gennadiycivil: I see these - 
"googletest/CMakeLists.txt:  project(gtest VERSION 1.9.0 LANGUAGES CXX C)
googlemock/CMakeLists.txt:  project(gmock VERSION 1.9.0 LANGUAGES CXX C)
"
Since you are replacing the hard-coded version with the variable, shouldnt the variable be used everywhere?

dakotahawkins: If it isn't being used everywhere, then I made a mistake!

dakotahawkins: I see only this one:

```
$ git grep -nIF '1.9.0'
CMakeLists.txt:8:set(GOOGLETEST_VERSION 1.9.0)
```

stefanosoffia:
@dakotahawkins thanks for bringing the attention back to this feature.

#1338 started in a similar fashion and then integrated many contributions from @ruslo, among which the main are:
* The use of the `configure_package_config_file` utility to properly wrap the exported targets (which seems to be a CMake suggested best-practice).
* Factoring out some common logic between the `gtest` and `gmock` targets.

Personally I'm favorable to keep things simple. Just consider if any of the contributions from #1338 are of interest.

dakotahawkins:
@stefanosoffia Thanks! I think that like you did I tried to do more and had to revert some of it because of the existing project structure. It seems like everything could be cleaned up a lot more by bumping to a much newer CMake version and removing use of the custom (e.g. `cxx_library`) CMake functions.

I wonder if my CMake config file needs to handle the Threads dependency like you did... do you know when `CMakeFindDependencyMacro` was added?

stefanosoffia:
@dakotahawkins good point, the `CMakeFindDependencyMacro` dates back at least to [3.0](https://cmake.org/cmake/help/v3.0/module/CMakeFindDependencyMacro.html).
The `configure_package_config_file` seems [older](https://cmake.org/cmake/help/v2.8.12/cmake.html).

CMakeLists.txt
gennadiycivil: I still see 2 of them
googletest/CMakeLists.txt:  project(gtest VERSION 1.9.0 LANGUAGES CXX C) - https://github.com/google/googletest/blob/b88511ef640c046ec9c7553b383a2aba64c80967/googletest/CMakeLists.txt#L47

and ggooglemock/CMakeLists.txt:  project(gmock VERSION 1.9.0 LANGUAGES CXX C) - https://github.com/google/googletest/blob/b88511ef640c046ec9c7553b383a2aba64c80967/googlemock/CMakeLists.txt#L40

dakotahawkins: @gennadiycivil That's apparently a `cleanup-cmake` branch, not this branch.

gennadiycivil: @dakotahawkins Sorry , we have a disconnect. 
Just look in the master brach, https://github.com/google/googletest/blob/master/googlemock/CMakeLists.txt line 40. 
Thanks
G

dakotahawkins: Those have been replaced in this commit with `${GOOGLETEST_VERSION}`. See [here](https://github.com/google/googletest/pull/1682/files#diff-98eb25640483d40b7d2663b41473906aR40) and [here](https://github.com/google/googletest/pull/1682/files#diff-7b12e50666f60ab6d4721a97c7b54868R47).

gennadiycivil: Thank you

gennadiycivil:
looks good to me

stefanosoffia:
As a follow-up of the discussion from #1338, please can you clarify about the following:

* What is the rationale behind `INSTALL_CMAKE_DIR`? Why is it initialized differently between Windows and Linux?
* What is the rationale behind the creation of 4 export files (one for each exported target) as opposed to one single export file that includes all the exported libraries?

dakotahawkins:
@stefanosoffia I think you linked the wrong issue in #1338 (#1628 ?), so I got confused and explained my reasoning for those things and a couple of others there. We can continue the discussion in whichever you'd prefer :)

PR #1699, opened 2018-08-03
Suppress default exception handling

drwez:
At present the exception conditions deliberately created by death-tests are not handled by the death-test framework, so the system default exception handler will receive them, causing each death-test to trigger a full stack-trace to be dumped.

This CL makes the following changes:
1. Register an exception port on the child process, and wait on both that and on process termination. If an exception is observed then kill the process directly, to prevent the exception being handled.
2. Fix the death-test implementation for Fuchsia to correctly close the sub-process handle, where previously each death-test leaked a (zombie) process until the parent process exited.
3. Fix initialization of the fdio_spawn_action_t not to use the C-only struct member initialization syntax.

This corresponds to cl/207346696

### DIFF
--- a/googletest/src/gtest-death-test.cc
+++ b/googletest/src/gtest-death-test.cc
@@ -67,6 +67,7 @@
 #  include <lib/fdio/spawn.h>
 #  include <zircon/processargs.h>
 #  include <zircon/syscalls.h>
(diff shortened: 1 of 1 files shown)


drwez:
Re CLA: Was not in Google org, am now.

PR #1706, opened 2018-08-07
Changed some URLs to HTTPS

einsteinsfool:
Runned `grep -rnw . -e 'http:/'` in the repo directory and changed all occurrences of "http://" to "https://" if the page was accessible via HTTPS.

Checked all URLs manually. Tried both with and without "www". Skipped licences. Didn't change URLs when only HTTP version was working.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 # Build matrix / environment variable are explained on:
 # https://docs.travis-ci.com/user/customizing-the-build/
 # This file can be validated on:
-# http://lint.travis-ci.org/
+# https://lint.travis-ci.org/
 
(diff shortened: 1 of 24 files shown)


gennadiycivil:
Thank you for this contribution. It is not Googletest job to dictate HTTP vs HTTPs. Modern browsers would do this on their own anyway.

einsteinsfool:
I specifically asked [here](https://github.com/google/googletest/pull/1675) if you'll ignore this PR if I divide my #1675 into two PRs. You could have saved me the time wasted producing this PR...

What is the problem with merging this PR? Nothing will break, so why not?

Issue #1717, opened 2018-08-09
Don't use CMAKE_BINARY_DIR for pkg-config files


eliasdaler:
Hello
I'm currently building Google Test as a part of my project (via `add_subdirectory`) . The main project is built in `project/build/` and GoogleTest is built in `project/build/_deps/googletest-build`. 

Unfortunately, GoogleTest puts the following files into `project/build` during the build:

- gmock_main.pc
- gmock.pc
- gtest_main.pc
- gtest.pc

See [here](https://github.com/google/googletest/blob/051fe2fa05af4ee761904aaee5314d67d5a15720/googlemock/CMakeLists.txt#L128) and [here](https://github.com/google/googletest/blob/e887b03e5a1f6095a2cd08c28db8a63d4c016231/googletest/CMakeLists.txt#L122).

As you can see, CMAKE_BINARY_DIR causes the problem. So, can CMAKE_CURRENT_BINARY_DIR be used there instead?

PR #1734, opened 2018-08-13
Change location of pkg-config files from CMAKE_BINARY_DIR to gmock/gtest_BINARY_DIR

eliasdaler:
Hello.

This is a simple patch which fixes (#1717). In short, if building Google Test as part of a CMake project (by adding it via add_subdirectory), it generated some pkg-config files into main project's binary dir. But now they're generated into gtest/gmock respective binary dirs, so the main build dir is left clean. This doesn't affect where the files get installed afterwards.


gennadiycivil:
Thank you for this contribution. Could you please provide test output "before" and "after" to show positive results.
Also does this have a potential to break other use cases?

eliasdaler:
# Before

## Scenario 1

Build directory - ~/googletest-build
Source directory - ~/googletest-src

gtest.pc, gtest_main.pc, gmock.pc and gmock_main.pc are put in ~/googletest-build after generation

## Scenario 2

Some CMake project is built in ~/build and Google Test is added via `add_subdirectory(${HOME}/build/googletest-src ${HOME}/build/googletest-build)`

gtest.pc, gtest_main.pc, gmock.pc and gmock_main.pc are put in ~/build after generation

# After

## Scenario 1

Build directory - ~/googletest-build
Source directory - ~/googletest-src

gtest.pc, gtest_main.pc are put into ~/googletest-build/googlemock/gtest. gmock.pc and gmock_main.pc are put in ~/googletest-build/googlemock after generation

## Scenario 2

Some CMake project is built in ~/build and Google Test is added via `add_subdirectory(${HOME}/build/googletest-src ${HOME}/build/googletest-build)`

gtest.pc, gtest_main.pc are put into ~/build/googletest-build/googlemock/gtest. gmock.pc and gmock_main.pc are put in ~/build/googletest-build/googlemock after generation

# Installing
"make install" puts them into <install-prefix>/lib/pkgconfig in both cases - this behaviour is unchanged.

# Breaking use cases

Low chance. The only thing I can think about is user modifying pkg-config files manually before running "make install", but this is rarely done, and will only break previous behaviour if user expected "*.pc" files to always be in the root of the build folder after generation. Usually pkg-config files are used after "make install," but there's no change of behaviour there.

Issue #1584, opened 2018-04-19
Failed to build with gcc-7 (-Werror=unused-function)


xkszltl:
Build from master branch with cmake/gcc-7 and get the following -Werror:

```
[172/191] Building CXX object googlemock/CMakeFiles/shared_gmock_test_.dir/test/gmock-spec-builders_test.cc.o
FAILED: googlemock/CMakeFiles/shared_gmock_test_.dir/test/gmock-spec-builders_test.cc.o 
ccache /opt/rh/devtoolset-7/root/usr/bin/c++  -DGTEST_LINKED_AS_SHARED_LIBRARY=1 -I../googlemock/include -I../googlemock -I../googletest/include -I../googletest -g -O3 -DNDEBUG   -g -Wall -Wshadow -Werror -Wno-error=dangling-else -DGTEST_HAS_PTHREAD=1 -fexceptions -MD -MT googlemock/CMakeFiles/shared_gmock_test_.dir/test/gmock-spec-builders_test.cc.o -MF googlemock/CMakeFiles/shared_gmock_test_.dir/test/gmock-spec-builders_test.cc.o.d -o googlemock/CMakeFiles/shared_gmock_test_.dir/test/gmock-spec-builders_test.cc.o -c ../googlemock/test/gmock-spec-builders_test.cc
In file included from ../googlemock/test/gmock-spec-builders_test.cc:36:0:
../googlemock/include/gmock/gmock-spec-builders.h:1083:23: error: ‘testing::Expectation testing::internal::TypedExpectation<F>::GetHandle() [with F = {anonymous}::NonDefaultConstructible()]’ declared ‘static’ but never defined [-Werror=unused-function]
   virtual Expectation GetHandle() {
                       ^~~~~~~~~
../googlemock/include/gmock/gmock-spec-builders.h:1083:23: error: ‘testing::Expectation testing::internal::TypedExpectation<F>::GetHandle() [with F = void(bool, int, std::basic_string<char>, void*, const {anonymous}::Printable&, {anonymous}::Unprintable)]’ declared ‘static’ but never defined [-Werror=unused-function]
../googlemock/include/gmock/gmock-spec-builders.h:903:11: error: ‘testing::internal::TypedExpectation<F>::~TypedExpectation() noexcept [with F = {anonymous}::NonDefaultConstructible()]’ declared ‘static’ but never defined [-Werror=unused-function]
   virtual ~TypedExpectation() {
           ^
[...]

PR #1749, opened 2018-08-18
Add GCC8 to Travis build matrix

### LINKED ISSUE #1722: Help Wanted: Please add GCC8 to Travis build matrix
We are missing it now and it already led to https://github.com/google/googletest/issues/1711#issuecomment-411788837

danghai:
I make PR to add GCC8 to Travis build matrix, and fix some warning
Fix #1722

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -53,6 +53,24 @@ matrix:
     - os: osx
       env: BUILD_TYPE=Release VERBOSE=1 CXX_FLAGS=-std=c++11
       if: type != pull_request
+    - os: linux
+      compiler: gcc
+      addons:
+        apt:
+          sources:
+            - ubuntu-toolchain-r-test
+          packages:
+            - g++-8
+            - gcc-8
+      env: CXX=g++-8 CC=gcc-8 AR=gcc-ar-8 RANLIB=gcc-ranlib-8 BUILD_TYPE=Release VERBOSE=1 CXX_FLAGS=-std=c++11
+    - os: osx
+      compiler: gcc
+      addons:
+        apt:
+          packages:
+            - g++-8
+            - gcc-8
+      env: CXX=g++-8 && CC=gcc-8 AR=gcc-ar-8 RANLIB=gcc-ranlib-8 BUILD_TYPE=Release VERBOSE=1 CXX_FLAGS=-std=c++11
 
 # These are the install and build (script) phases for the most common entries in the matrix.  They could be included
 # in each entry in the matrix, but that is just repetitive.
@@ -74,7 +92,7 @@ addons:
     - ubuntu-toolchain-r-test
     - llvm-toolchain-precise-3.7
     packages:
-    - g++-4.9
+    - g++-8
     - clang-3.7
 
 notifications:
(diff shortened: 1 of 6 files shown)


danghai:
I signed it!

danghai:
This patch provides a travis configuration with GCC 8,So it could automatically catch error:
```
/home/travis/build/google/googletest/googlemock/include/gmock/gmock-spec-builders.h:1084:23: error: ‘testing::Expectation testing::internal::TypedExpectation<F>::GetHandle() [with F = {anonymous}::NonDefaultConstructible()]’ declared ‘static’ but never defined [-Werror=unused-function]
   virtual Expectation GetHandle() {
                       ^~~~~~~~~
/home/travis/build/google/googletest/googlemock/include/gmock/gmock-spec-builders.h:1084:23: error: ‘testing::Expectation testing::internal::TypedExpectation<F>::GetHandle() [with F = void(bool, int, std::basic_string<char>, void*, const {anonymous}::Printable&, {anonymous}::Unprintable)]’ declared ‘static’ but never defined [-Werror=unused-function]
/home/travis/build/google/googletest/googlemock/include/gmock/gmock-spec-builders.h:904:11: error: ‘testing::internal::TypedExpectation<F>::~TypedExpectation() noexcept [with F = {anonymous}::NonDefaultConstructible()]’ declared ‘static’ but never defined [-Werror=unused-function]
   virtual ~TypedExpectation() {
```
For example:
#1584

.travis.yml
```diff
@@ -53,6 +53,22 @@ matrix:
...
       if: type != pull_request
+    - os: linux
+      compiler: gcc
+      addons:
+        apt:
+          sources:
+            - ubuntu-toolchain-r-test
+          packages:
+            - g++-8
+      env: CXX=g++-8 && CC=gcc-8 BUILD_TYPE=Release VERBOSE=1 CXX_FLAGS=-std=c++11
```
tanzislam: You probably want to also set `AR=gcc-ar-8 && RANLIB=gcc-ranlib-8`.

danghai: @tanzislam what is the reason why I need to set `AR=gcc-ar-8 && RANLIB=gcc-ranlib-8`?

tanzislam: [This StackOverflow answer](https://stackoverflow.com/a/48815662/1433768) explains that the `gcc-`-prefixed versions get shipped with the compiler rather than the OS. (We can't change the system default to GCC 8 here because [that would require sudo](https://github.com/travis-ci/travis-ci/issues/3668).)

I'm not sure exactly how the GCC-shipped ones are different from the OS-shipped ones. (It may well be that right now they only matter for using the Link Time Optimization feature, which we don't use here, and thus it may not make a difference after all.) It just seems safer to use tools from the same package/installation.

gennadiycivil:
I see CI errors https://travis-ci.org/google/googletest/jobs/418082136

.travis.yml
danghai: @tanzislam Thanks. Do you know how to fix the error from `Travis`? The error shows above comment.

gennadiycivil:
@tanzislam @danghai could you please take a look

gennadiycivil:
Looks abandoned

PR #862, opened 2016-08-30
Suppress MSVC warnings 4251 (and fix 4275) in header files to fix his…

KindDragon:
… appearance when include one of that header files

This warnings really annoying when you include gtest/gmock header files with BUILD_SHARED_LIBS=ON

PR #1750, opened 2018-08-18
Suppress MSVC warnings 4251 (and fix 4275) in header files

KindDragon:
These warnings really annoying when you include gtest/gmock header files with BUILD_SHARED_LIBS=ON

Continuation of #862


gennadiycivil:
This has been merged via a different commit. I am still testing ability to automatically close PRs with proper attribution and in this case this did not work. But the code is there, thank you for your contribution

Issue #1735, opened 2018-08-14
Explicitly using CMAKE_CXX_FLAGS in cxx_exception breaks build


jschloss-swift:
Hi,

I'm trying to build gtest as a nested subrepository while cross-compiling with some custom CXX_FLAGS predefined.

cxx_exception explicitly appends to CMAKE_CXX_FLAGS here

https://github.com/google/googletest/blob/8bc11c040a1dcbe50c340b21d78175d152dd3837/googletest/cmake/internal_utils.cmake#L141

However, this actually double-appends, as cmake will automatically prepend CMAKE_CXX_FLAGS. This breaks my build, as I am specifying some flags that must only be specified once (otherwise gcc will refuse to compile), such as -specs=nosys.specs.

doing this to googletest/cmake/internal_utils.cmake

-  set(cxx_exception "${CMAKE_CXX_FLAGS} ${cxx_base_flags} ${cxx_exception_flags}")
+  set(cxx_exception "${cxx_base_flags} ${cxx_exception_flags}")

allows gtest to build, and still has all the correct cflags (checked by looking at make VERBOSE=1).

I can provide more info if needed.

Thanks,

Jacob

PR #1767, opened 2018-08-22
Issue #1735 It seems like CMAKE_CXX_FLAGS is getting double appended when buildin…

jschloss-swift:
Related to Issue #1735 

It seems like CMAKE_CXX_FLAGS is getting double appended when building as a nested cmake project,
which breaks my build as gcc does not allow -specs=nosys.specs to be called multiple times.
Removing this fixes it and seems to keep the same compile options, just removes the duplicate definition.


jschloss-swift:
I signed it!

jschloss-swift:
I signed it!

gennadiycivil:
No CLA

jschloss-swift:
OK, we got confirmation now that the CLA went through ~12PM today.

PR #1784, opened 2018-08-27
New Readme.md

DavidSchuldenfrei:


DavidSchuldenfrei:
@gennadiycivil Thank you for your help in my previous PR. 

I cannot rename an existing extension (internal) name, but I believe that I changed all GTest to GoogleTest in the description/readme.md of the extension

PR #1785, opened 2018-08-28
Improve CMake exported targets (revisited)

dakotahawkins:
# FOR REVIEW!
- [ ] Review this
---

See notes and discussion in #1682 

Attempts to recombine the changes made in #1682 and #1338

Differences:

- More control of process through cache variables/options
- CMake export install defaults to lib dir, with no project-specific
  subdirectory


dakotahawkins:
Not sure if AppVeyor failures are mine or not...

gennadiycivil:
I see AppVeyor errors

dakotahawkins:
@gennadiycivil Yes, `gmock_output_test` failed for 64-bit VS2017 and 32-bit VS2015. I didn't immediately see why that test would fail in those configurations due to my changes, so I wonder if that test is flaky.

dakotahawkins:
I see the difference between the actual output and the golden output is:

```
diff --git a/googlemock/test/gmock_output_test_golden.txt b/googlemock/test/gmock_output_test_golden.txt
index dbcb211..e95138c 100644
--- a/googlemock/test/gmock_output_test_golden.txt
+++ b/googlemock/test/gmock_output_test_golden.txt
@@ -292,7 +292,7 @@ Stack trace:
 FILE:#: Failure
 Value of: (std::pair<int, bool>(42, true))
 Expected: is pair (is >= 48, true)
-  Actual: (42, true) (of type std::pair<int, bool>)
+  Actual: (42, true)
 [  FAILED  ] GMockOutputTest.PrintsMatcher
 [  FAILED  ] GMockOutputTest.UnexpectedCall
 [  FAILED  ] GMockOutputTest.UnexpectedCallToVoidFunction
```

So it's missing ` (of type std::pair<int, bool>)`, which I think means `GTEST_HAS_RTTI` is undefined...

ccvca:
> CMake export install defaults to lib dir, with no project-specific
subdirectory

Why? Maybe it is the best to choose any of the options that are compatible with unix and windows

https://cmake.org/cmake/help/v3.8/command/find_package.html?highlight=i
```
<prefix>/                                                       (W)
<prefix>/(cmake|CMake)/                                         (W)
<prefix>/<name>*/                                               (W)
<prefix>/<name>*/(cmake|CMake)/                                 (W)
<prefix>/(lib/<arch>|lib|share)/cmake/<name>*/                  (U)
<prefix>/(lib/<arch>|lib|share)/<name>*/                        (U)
<prefix>/(lib/<arch>|lib|share)/<name>*/(cmake|CMake)/          (U)
<prefix>/<name>*/(lib/<arch>|lib|share)/cmake/<name>*/          (W/U)
<prefix>/<name>*/(lib/<arch>|lib|share)/<name>*/                (W/U)
<prefix>/<name>*/(lib/<arch>|lib|share)/<name>*/(cmake|CMake)/  (W/U)
```

dakotahawkins:
@ccvca Because that's what I see most often, though I forgot about that table.

My actual preference is to use a different default for Windows, as in my original change from #1682 [here](https://github.com/google/googletest/pull/1682/files#diff-af3b638bc2a3e6c650974192a53c7291R23):

```
if(WIN32)
  set(INSTALL_CMAKE_DIR "cmake" CACHE PATH "CMake exported targets")
else()
  set(INSTALL_CMAKE_DIR "${CMAKE_INSTALL_LIBDIR}/cmake/gtest" CACHE PATH "CMake exported targets")
endif()
```

But you could say (and you'd be right) that mostly the point of that is to give me the default I happen to like on the platform I happen to use :)

I guess I'm OK with changing it to whatever everybody wants as long as it works for Windows and Unix and I can override it with a cache variable.

Do you have a preference?

dakotahawkins:
In the VS2017 failing configuration, `gmock_leak_test` and `gmock_output_test` were run (`gmock_output_test` fails), while in the same configuration for #1792 those two tests did not run.

Anybody know why that is?

ccvca:
I'm fine with every version that works out of the box, so the version above is fine.
My personal favourite are the versions with a package name in it, because you could remove an installed package very easy (just remove the folder, no need to search all related files)

PS: If I get it right, the Windows case above could be removed, as the default case (Unix) does also works on Windows.

gennadiycivil:
Ping

dakotahawkins:
@gennadiycivil I've not abandoned it, but I am busy with non-c++ work (boo) right this second. I'd like to get this done, but I could use some guidance:

I do not understand the test failures -- and as I mentioned [here](https://github.com/google/googletest/pull/1785#issuecomment-416773978) I also don't understand why the failing cases ran for my branch but not for #1792. I did do some investigation, but at least as far as I understood the problem I could not relate it to my changeset.

Do you have any insight into that?

gennadiycivil:
@dakotahawkins I merged the latest branch in - take a look at the CI . I don't have much to say when it comes to CMake nuances I am afraid.

ccvca:
There is a minor difference in the output:
https://www.diffchecker.com/3WLCIBhU
This is missing in the output: ` (of type std::pair<int, bool>)`
The expected extra seems like a missing option like "Show type name"

gennadiycivil:
@dakotahawkins ping? (I suggest closing it for now)

gennadiycivil:
Closing for now

PR #1791, opened 2018-08-28
Silence use-after-free warning from clang-tidy

antoinedesbois:
Inside the AddNewExpectation() function, the expectation object owner is
pushed in a vector effectively incrementing the owner count. This prevents
the next destructor call (the closing bracket) from deleting the object.

### DIFF
--- a/googlemock/include/gmock/gmock-spec-builders.h
+++ b/googlemock/include/gmock/gmock-spec-builders.h
@@ -1282,7 +1282,7 @@ class MockSpec {
     const std::string source_text(std::string("EXPECT_CALL(") + obj + ", " +
                                   call + ")");
     LogWithLocation(internal::kInfo, file, line, source_text + " invoked");
-    return function_mocker_->AddNewExpectation(
+    return function_mocker_->AddNewExpectation( // NOLINT
         file, line, source_text, matchers_);
   }
 


antoinedesbois:
I signed it!

googlemock/include/gmock/gmock-spec-builders.h
```diff
@@ -1282,7 +1282,7 @@ class MockSpec {
     const std::string source_text(std::string("EXPECT_CALL(") + obj + ", " +
                                   call + ")");
     LogWithLocation(internal::kInfo, file, line, source_text + " invoked");
-    return function_mocker_->AddNewExpectation(
+    return function_mocker_->AddNewExpectation( // NOLINT
```
pavelkryukov: You may use `NOLINT(
use-after-free)` to silence a specific checker

gennadiycivil:
@antoinedesbois thank you for your contribution.
If this is a real issue we specifically don't want to silence the warning but rather we would want to fix it.

antoinedesbois:
```

**Comment by antoinedesbois** [comment]
```
  TypedExpectation<F>& AddNewExpectation(
      const char* file,
      int line,
      const string& source_text,
      const ArgumentMatcherTuple& m)
          GTEST_LOCK_EXCLUDED_(g_gmock_mutex) {
    Mock::RegisterUseByOnCallOrExpectCall(MockObject(), file, line);
    TypedExpectation<F>* const expectation =
        new TypedExpectation<F>(this, file, line, source_text, m);
    const linked_ptr<ExpectationBase> untyped_expectation(expectation);
    untyped_expectations_.push_back(untyped_expectation);

    // Adds this expectation into the implicit sequence if there is one.
    Sequence* const implicit_sequence = g_gmock_implicit_sequence.get();
    if (implicit_sequence != NULL) {
      implicit_sequence->AddExpectation(Expectation(untyped_expectation));
    }

    return *expectation;
  }
```
In this specific case, the destruction of `untype_expectation` won't free `expectation` because the owner count is incremented by `untyped_expectations_.push_back(untyped_expectation);
`
Returning a reference to `expectation` is okay in this case because the owner count is at least 1. This fix will not change the way cases like the one below are handled by clang-tidy. (true use after free)
```
auto& expectation = EXPECT_CALL( mock, myFunc() );
expectation.Times( 1 );
Mock::VerifyAndClearExpectations( &mock );
expectation.WillOnce( Return( myResult ) );
```

gennadiycivil:
@antoinedesbois Thank you so much. I suggest instead of commenting on a closed pull request, please propose the actual change in a new PR, We can discuss it there when it is actually working, passed CI, etc.

PR #1792, opened 2018-08-28
Fix MSVC warning C5046 in gmock-matchers.h

dakotahawkins:
C5046: Symbol involving type with internal linkage not defined

This warning appeared in VS2017 15.8, and is (so far) undocumented.

The workaround described in the following link should be harmless:

https://developercommunity.visualstudio.com/content/problem/288509/view.html

### DIFF
--- a/googlemock/include/gmock/gmock-matchers.h
+++ b/googlemock/include/gmock/gmock-matchers.h
@@ -145,6 +145,9 @@ class MatcherDescriberInterface {
 template <typename T>
 class MatcherInterface : public MatcherDescriberInterface {
  public:
+  // https://developercommunity.visualstudio.com/content/problem/288509/view.html
+  virtual ~MatcherInterface() {};
+
   // Returns true iff the matcher matches x; also explains the match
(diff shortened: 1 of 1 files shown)


gennadiycivil:
Thank you for this contribution.
I am not sure that googletest should be making this change in this way. If and when this is actually documented we may consider it.

dakotahawkins:
@gennadiycivil It's treated as an error currently, would you prefer to ignore the warning?

dakotahawkins:
@gennadiycivil As far as I can tell it is brand new, it surprised me. There's next to no info about it on the internet (except for the MS developer community link). I'd bet it goes away in a VS update, but right this second it blocks me from building successfully.

gennadiycivil:
@dakotahawkins  Please take a look at master branch now. I just merged a PR that among other things has code to silence this warning. 
Hope this helps. I ll close this PR now.

dakotahawkins:
@gennadiycivil Yeah that should do the trick, thanks!

PR #1801, opened 2018-09-02
pkgconfig: Unconditionally depend on GTest when using GMock

SoapGentoo:
* GTest is a required dependency for GMock, hence
  we always need to pull it in.

@gennadiycivil could we also get this into the `1.8.0` branch? Having the wrong pkg-config files there will set it in stone and make them quasi useless.


gennadiycivil:
@SoapGentoo Agreed. Thank you!

PR #1815, opened 2018-09-06
Fix Compile error, and warning in Visual Studio 2013

DavidSchuldenfrei:
The commit [b50b2f7](https://github.com/google/googletest/commit/b50b2f775ed0268af9c83a96b285e296778d0743) causes the project build to fail in Visual Studio 2013. This PR fixes this, by reverting the breaking change for older versions of VS. 
Also, VisualStudio 2013 doesn't support pragma warning 5046 and gives a warning/error when run on the current version of GooglewMock. This PR also fixes this issue.

### DIFF
--- a/googlemock/include/gmock/gmock-matchers.h
+++ b/googlemock/include/gmock/gmock-matchers.h
@@ -56,11 +56,17 @@
 # include <initializer_list>  // NOLINT -- must be after gtest.h
 #endif
 
+#if _MSC_VER >= 1900
 GTEST_DISABLE_MSC_WARNINGS_PUSH_(
     4251 5046 /* class A needs to have dll-interface to be used by clients of
                  class B */
     /* Symbol involving type with internal linkage not defined */)
-
+#else //Pragma 5046 doesn't exist in version of MSC prior to 1900
+GTEST_DISABLE_MSC_WARNINGS_PUSH_(
+    4251 /* class A needs to have dll-interface to be used by clients of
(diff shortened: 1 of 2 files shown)


gennadiycivil:
Visual Studio 2013 can not be officially supported in the master branch. This is a very old compiler, sorry.

DavidSchuldenfrei:
@gennadiycivil I understand that. However as we are writing an multi-platform SDK which is used by several vendors to extend their exisitng applications, and some of these applications are written in MS-VS2013, we need to also support this environemnt (It is only used by our build machine, which is why I hadn't even noticed it before). 
I don't suggest bending backwards to support it, but as 1.8.x should work with pre-C++ 11 compilers, and this is a minor change, I think that this change could be worth while.

Thanks

nicolastagliani:
google benchmark is not passing all appveryor builds for that same issue (since google test is a dependecy)...

gennadiycivil:
@DavidSchuldenfrei I can understand the 1.8.x logic. I think it might be reasonable to request a change in 1.8.x branch but definitely not in a master branch. This PR was asking for a change in the master brunch.

In addition we cant just revert a commit https://github.com/google/googletest/commit/b50b2f775ed0268af9c83a96b285e296778d0743 and potentially break someone else who requested this commit to start with. 

Thank you for your understanding

gennadiycivil:
@nicolastagliani

Issue #1810, opened 2018-09-04
Trying to build GTest on MSVC 12


uilianries:
Hi!

I'm trying to build google test using MSVC 12, but I get an error during the execution:

> C:\Users\appveyor.conan\data\gtest\1.8.1\bincrafters\testing\package\f84b2d17334e7b23564f7244508491e451d80357\include\gtest/gtest-printers.h(1036): error C2770: invalid explicit template argument(s) for 'AddReference<const ::std::tr1::tuple_element<I,std::tuple<bool,int>>::type>::type testing::internal::TuplePolicy::get(const std::tuple<bool,int> &)' [C:\projects\conan-gtest\test_package\build\ffafa1fe0219c44dbe68adf8ffbfd78683edb626\test_package.vcxproj]

I tried using some custom settings, like GTEST_HAS_TR1_TUPLE and GTEST_USE_OWN_TR1_TUPLE, but they did not work.

I'm forgetting something?

gennadiycivil:
This  (MSVC++ 12.0  _MSC_VER == 1800 (Visual Studio 2013 version 12.0)) is an old compiler. We wont 'support' it, you are on your own using it, sorry

uilianries:
No problem, thanks for your answer.

Regards.

PR #1817, opened 2018-09-08
Fix Compile error, and warning in Visaul Studio 2013

DavidSchuldenfrei:


DavidSchuldenfrei:
@gennadiycivil @nicolastagliani Same as PR 1815, but to branch 1.8.x.

googletest/include/gtest/gtest-printers.h
```diff
@@ -983,7 +983,7 @@ struct TuplePolicy {
 
   template <size_t I>
   static typename AddReference<const typename ::std::tr1::tuple_element<
-      static_cast<int>(I), Tuple>::type>::type
+      I, Tuple>::type>::type
```
gennadiycivil: As I explained in https://github.com/google/googletest/pull/1815 - that change was made for a reason. We can not just roll it back. You may want to consider using macros to only do this when this version of MSVC is used .

DavidSchuldenfrei:
@gennadiycivil Changed as requested.

gennadiycivil:
@DavidSchuldenfrei ping.

BrukerJWD:
Thanks for fixing this! I searched quite some time until I found this here. So for others struggling with this error (which was also described in #1810) I try to point Google to this PR here. Error was:
> error C2770: invalid explicit template argument(s) for 'const ::std::tr1::tuple_element<I,std::tuple<<unnamed-symbol>>>::type testing::internal::TuplePolicy<TupleT>::get(const std::tuple<<unnamed-symbol>> &)'

@DavidSchuldenfrei: Shouldn't we also fix the `tuple_element` a few lines above?

PR #1831, opened 2018-09-12
version: fix declared version to be in sync with CMakeLists.txt

anthraxx:


anthraxx:
No offense intended, but it would be awesome if the version numbers in this project could be kept better in sync when doing releases.
1.8.0 had a broken version as well: https://github.com/google/googletest/commit/fa892afcb9532fe559a067b49f31952c7a80c3c7
1.8.1 had CMakeLists.txt version 1.9.0 and autoconf version 1.8.0

This potentially breaks downstream projects depending on certain features or behaviour by relying on the version reported by gtest-config --version

anthraxx:
I signed it!

Issue #1841, opened 2018-09-17
MINGW64: Fatal error: can't write 286 bytes to section .text, gmock-matchers_test.cc.o: too many sections (80261)


matlo607:
It seems gmock's unittests are not buildable under msys.
Is it a known issue ?
```
[ 22%] Building CXX object googlemock/CMakeFiles/gmock-matchers_test.dir/test/gmock-matchers_test.cc.o
cd /c/Users/mlongo/Documents/googletest/build-dir/googlemock && /mingw64/bin/g++.exe   -isystem /c/Users/mlongo/Documents/googletest/googlemock/include -isystem /c/Users/mlongo/Documents/googletest/googlemock -isystem /c/Users/mlongo/Documents/googletest/googletest/include -isystem /c/Users/mlongo/Documents/googletest/googletest  -Wall -Wshadow -Werror -Wno-error=dangling-else -DGTEST_HAS_PTHREAD=0 -fexceptions -o CMakeFiles/gmock-matchers_test.dir/test/gmock-matchers_test.cc.o -c /c/Users/mlongo/Documents/googletest/googlemock/test/gmock-matchers_test.cc
C:/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/8.2.0/../../../../x86_64-w64-mingw32/bin/as.exe: CMakeFiles/gmock-matchers_test.dir/test/gmock-matchers_test.cc.o: too many sections (80261)
C:\Users\mlongo\AppData\Local\Temp\ccEOllKN.s: Assembler messages:
C:\Users\mlongo\AppData\Local\Temp\ccEOllKN.s: Fatal error: can't write 286 bytes to section .text of CMakeFiles/gmock-matchers_test.dir/test/gmock-matchers_test.cc.o: 'file too big'
C:/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/8.2.0/../../../../x86_64-w64-mingw32/bin/as.exe: CMakeFiles/gmock-matchers_test.dir/test/gmock-matchers_test.cc.o: too many sections (80261)
C:\Users\mlongo\AppData\Local\Temp\ccEOllKN.s: Fatal error: can't close CMakeFiles/gmock-matchers_test.dir/test/gmock-matchers_test.cc.o: file too big
```

```
C:\msys64\mingw64\bin\g++.exe --version
g++.exe (Rev1, Built by MSYS2 project) 8.2.0
Copyright (C) 2018 Free Software Foundation, Inc.
[...]

PR #1844, opened 2018-09-18
[msys] pass big object file option to assembler for target gmock-matchers_test

matlo607:
fix issue #1841


matlo607:
Hello @gennadiycivil,
The reason for the failing build is the below (I guess some connectivity issues):
```
apt-get install failed
The command "sudo -E apt-get -yq --no-install-suggests --no-install-recommends $(travis_apt_get_options) install g++-4.9 clang-3.9" failed and exited with 100 during .
```
Should I update the date of the commit to force a new build ?
Or can you still merge if your review is OK ?

KindDragon:
Maybe better modify script here https://github.com/google/googletest/blob/master/googletest/cmake/internal_utils.cmake#L207-L210

matlo607:
Hello @KindDragon,
As I said in the issue #1841,
> Passing the option -mbig-obj to gas through GCC with -Wa allows GCC to build the object file but in return, **highly increases the build time of the target** (I would say by **2 or 3 minutes** depending on the CPU of your machine).

There are around 60 unit tests with both gtest and gmock. I don't want to specify this flag globally because the build time would take 60 * 3 minutes = 3 hours !
I don't know the reason why the build time is so much increased. I did some research but could not find anything relevant.
For now, I want to have a minimal impact and specify it only for the needed target.

gennadiycivil:
@matlo607 I restarted failed travis build

PR #1847, opened 2018-09-20
Revert "Added special catch for std::exception in GTEST_TEST_NO_THROW_"

gennadiycivil:
Reverts google/googletest#1832. Style violations did not catch before Will provide more feedback


gennadiycivil:
@Jonny007-MKD @JonathanDCohen . I have to apologize. Due to a code sync bug the changes appear to have been overwritten. I will make this revert to document.

Would you mind terribly creating a new PR with the same changes? Also please address 80 chars wide limit and dont use "value" as a name. Thank you so much and please accept my apologies

Issue #1842, opened 2018-09-18
[msys] unit test ColoredOutputTest.UsesColorsWhenTermSupportsColors failing


matlo607:
When the test is compiled with /mingw64/bin/g++.exe, we should pass in the section "non-Windows platforms". However `GTEST_OS_WINDOWS` needs to be equal to 1 so that Windows headers are used.

```
ctest -V -R gtest_unittest
UpdateCTestConfiguration  from :/c/Users/mlongo/Documents/googletest/build-dir/DartConfiguration.tcl
UpdateCTestConfiguration  from :/c/Users/mlongo/Documents/googletest/build-dir/DartConfiguration.tcl
Test project /c/Users/mlongo/Documents/googletest/build-dir
Constructing a list of tests
Done constructing a list of tests
Updating test list for fixtures
Added 0 tests to meet fixture requirements
Checking test dependency graph...
Checking test dependency graph end
test 22
    Start 22: gtest_unittest

22: Test command: /c/Users/mlongo/Documents/googletest/build-dir/googlemock/gtest/gtest_unittest.exe
22: Test timeout computed to be: 10000000
22: Running main() from C:/Users/mlongo/Documents/googletest/googletest/src/gtest_main.cc
22: [==========] Running 425 tests from 103 test cases.
22: [----------] Global test environment set-up.

...

22: [ RUN      ] ColoredOutputTest.UsesColorsWhenTermSupportsColors
22: C:/Users/mlongo/Documents/googletest/googletest/test/gtest_unittest.cc:6838: Failure
22: Value of: ShouldUseColor(true)
22:   Actual: false
22: Expected: true
22: C:/Users/mlongo/Documents/googletest/googletest/test/gtest_unittest.cc:6841: Failure
22: Value of: ShouldUseColor(true)
22:   Actual: false
22: Expected: true
22: [  FAILED  ] ColoredOutputTest.UsesColorsWhenTermSupportsColors (1 ms)
22: [----------] 6 tests from ColoredOutputTest (1 ms total)

...

22:
22: [----------] Global test environment tear-down
22: [==========] 425 tests from 103 test cases ran. (663 ms total)
22: [  PASSED  ] 424 tests.
22: [  FAILED  ] 1 test, listed below:
22: [  FAILED  ] ColoredOutputTest.UsesColorsWhenTermSupportsColors
22:
22:  1 FAILED TEST
22:   YOU HAVE 13 DISABLED TESTS
22:
[...]

PR #1853, opened 2018-09-24
[mingw] enable the unittests for gmock and gtest again

matlo607:
Unittest were disabled because of #1841 and #1842.
https://github.com/google/googletest/blob/ed6e84ccef7f3c0f31caca60a47077a46d72a806/appveyor.yml#L70-L71
They can be enabled again since #1843 and #1844.


ekene90:
How does it work

PR #1858, opened 2018-09-25
Use correct minor version for 1.8 branch

### LINKED ISSUE #1816: Tagged 1.8.1 has GOOGLETEST_VERSION 1.9.0
First thank you for releasing a long awaited update!

As I am building googletest, I noticed that in the CMake, this package is showing up as `1.9.0` which is definitely not `1.8.1` ref: 
https://github.com/google/googletest/blob/release-1.8.1/CMakeLists.txt#L8. 

This seems to have been set this way in this commit: https://github.com/google/googletest/commit/759ef7c4e9662321548d1c30528c78ecdba2a05d

Should the tag have been 1.9.0 instead of 1.8.1?

SoapGentoo:
Fixes #1816

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,7 +5,7 @@ if (POLICY CMP0048)
 endif (POLICY CMP0048)
 
 project(googletest-distribution)
-set(GOOGLETEST_VERSION 1.9.0)
+set(GOOGLETEST_VERSION 1.8.2)
 
 enable_testing()
 
--- a/configure.ac
+++ b/configure.ac
@@ -1,5 +1,5 @@
 AC_INIT([Google C++ Mocking and Testing Frameworks],
-        [1.9.0],
+        [1.8.2],
         [googlemock@googlegroups.com],
         [googletest])
 


gennadiycivil:
@SoapGentoo Thank you for the patch. As discussed in the #1816 there is no robust way for testing this , we dont use cmake internally.  In addition please take a look at already merged https://github.com/google/googletest/pull/1831. If anything please make sure that the versions are the same

SoapGentoo:
@gennadiycivil synced up the version numbers. In general I suggest we remove Autotools. Given that the CMake build system is well behaved, there's no real reason to have the Autotools version around (especially because you don't bootstrap the tarballs, meaning we have to do it on our end).

gennadiycivil:
@SoapGentoo thank you again for preparing this change. 
I am still looking for a "real, not theoretical" reason to merge it. Could you provide one?

SoapGentoo:
@gennadiycivil sorry for the late reply. Say I have a C++98 codebase that requires version 1.8 to work. Currently with pkg-config I get

    $ pkg-config --libs "gtest < 1.9"
    Requested 'gtest < 1.9' but version of gtest is 1.9.0
    You may find new versions of gtest at https://github.com/google/googletest

With my fix I get

    $ pkg-config --libs "gtest < 1.9"
    -L/home/soap/gitz/googletest/test/lib64 -lgtest -pthread

That is, without the correct version I cannot constrain my dependencies correctly.

gennadiycivil:
@SoapGentoo thank you for the information.

PR #1859, opened 2018-09-25
Add another way to disable test case

louxiu:
Provide anther way to disable all tests in a test case.


louxiu:
I signed it!

gennadiycivil:
@louxiu Thank you for your contribution. Could you please explain why "another way" is helpful.

gennadiycivil:
@louxiu Thank you again however there does not appear to be enough merit to accept it

PR #1850, opened 2018-09-23
Added special catch for std::exception in GTEST_TEST_NO_THROW_

Jonny007-MKD:
Identical PR to #1832.
Is used the code from master when the PR was merged, shortened some lines and renamed the value member (I have to add that I think value is a pretty good name for the member of this class ;) )

PR #1862, opened 2018-09-25
Googletest export (and revert #1850)

gennadiycivil:
Googletest export. 
Note that this also reverts previously merged https://github.com/google/googletest/pull/1850. (causes breakage that needs to be investigated) 
@Jonny007-MKD @JonathanDCohen


Jonny007-MKD:
Sorry, I am confused. To what kind of breakages does #1850 lead to? Can we put them in unit tests?

gennadiycivil:
@Jonny007-MKD there is internal code that was broken by this change. The breakages were "a few" of many thousand tests that execute internally under many different conditions.

@JonathanDCohen will tackle this internally , When ready we will upstream and @JonathanDCohen will include proper attribution to mention your name

PR #1867, opened 2018-09-26
Use std::forward instead of std::move for MockFunction

marbub:
If you use std::move you get an compile error for MockFunction<void(Foo&)>.


gennadiycivil:
@marbub Thank you for this contribution. This looks mostly good, however could you please re-implement this in the pump file. As specified in the beginning of the file this file is generated.
Thank you again

gennadiycivil:
> forward<A$j> ?

yes!
Thank you

marbub:
Your welcome :-)

gennadiycivil:
@marbub Could you please provide test run before and after with output showing the successful results

marbub:
@gennadiycivil I don't understand? Which test run? My tests?

gennadiycivil:
@marbub So there was a piece of code that used to not compile as you mentioned in #1868 . Now that this you put together this change this piece of code works.. 

The easiest approach to show the "success" is to make a unit test out of that code. I recommend adding this new Unit test to googlemock/test/gmock-generated-function-mockers_test.cc.

Thank you again
G

marbub:
I messed it up. I simply have no experience this github. ;-(

marbub:
I feel really sorry, the workflow feels really different from gerrit. Thank you for your help!

PR #1870, opened 2018-09-27
Added RemoveDuplicateTests method to UnitTest

JordiGR:
In an effort to integrate googletest as a plug-in for Unreal Engine editor, the need to allow the removal of duplicate tests (by name) arose.
The Unreal Engine editor supports hot reloding of project modules (DLLs). If these modules contain googletest files, when they are reloaded, the tests get appended to the set of existing tests of the living singleton instance that gets created when the googletest plug-in gets loaded for the first time.


JordiGR:
I signed it!

gennadiycivil:
@JordiGR Thank you for this contribution. Could you please add unit tests.

gennadiycivil:
Looks stale

JordiGR:
Sorry for the delay. No, it's not stale. I'll create a new one.

PR #1878, opened 2018-10-02
Use std::forward instead of std::move for MockFunction

marbub:
If you use std::move you get an compile error for MockFunction<void(Foo&)>.


gennadiycivil:
@marbub  This is a duplicate of #1867 . I realize that #1867 is a bit messed up, no worries. I ll close this for now and we will deal with the original

PR #1889, opened 2018-10-08
Explicitly include crtdbg.h

adierking:
On Windows, MemoryIsNotDeallocated is accidentally relying upon crtdbg.h being
pulled in through one of the C++ headers. If Google Test is built with an
alternative C++ standard library (such as LLVM libc++), then this will not
happen and compilation will fail. Ensure that crtdbg.h is always included for
MSVC-compatible builds.


googletest/src/gtest-port.cc
```diff
@@ -41,6 +41,9 @@
 # include <io.h>
 # include <sys/stat.h>
 # include <map>  // Used in ThreadLocal.
+# ifdef _MSC_VER
```
gennadiycivil: would it make sense to do this on Windows and not just on MSVC? Or is this specific to MSVC? ( I am thinking about MinGW )

adierking: This header is from the MSVC runtime, so I'm not sure it would be present in MinGW. The code that uses this header is also gated with `#ifdef _MSC_VER`: https://github.com/google/googletest/blob/d7dd003c6d393545003f805efd69ed486d7a3a6d/googletest/src/gtest-port.cc#L314-L326.

gennadiycivil:
this has been merged

Issue #1883, opened 2018-10-03
Broken gmock - variadic templates 


jonbronson:
It looks like some recent change has broken the gmock build. We've began receiving this failure on Mac HighSierra 10.13.6:

[ 37%] Building CXX object googlemock/CMakeFiles/gmock.dir/src/gmock-all.cc.o
In file included from .../googletest/src/googletest/googlemock/src/gmock-all.cc:39:
In file included from .../googletest/src/googletest/googlemock/include/gmock/gmock.h:62:
In file included from .../googletest/src/googletest/googlemock/include/gmock/gmock-generated-function-mockers.h:44:
In file included from .../googletest/src/googletest/googlemock/include/gmock/gmock-spec-builders.h:71:
.../googletest/src/googletest/googlemock/include/gmock/gmock-matchers.h:5173:19: warning: variadic templates are a C++11 extension [-Wc++11-extensions]
template <typename... Args>
                  ^
.../googletest/src/googletest/googlemock/include/gmock/gmock-matchers.h:5174:77: error: a space is required
      between consecutive right angle brackets (use '> >')
internal::ElementsAreMatcher<tuple<typename std::decay<const Args&>::type...>>
                                                                            ^~
                                                                            > >
.../googletest/src/googletest/googlemock/include/gmock/gmock-matchers.h:5177:54: error: a space is required
      between consecutive right angle brackets (use '> >')
      tuple<typename std::decay<const Args&>::type...>>(
                                                     ^~
                                                     > >
.../googletest/src/googletest/googlemock/include/gmock/gmock-matchers.h:5181:19: warning: variadic templates
      are a C++11 extension [-Wc++11-extensions]
template <typename... Args>
                  ^
.../googletest/src/googletest/googlemock/include/gmock/gmock-matchers.h:5183:52: error: a space is required
      between consecutive right angle brackets (use '> >')
    tuple<typename std::decay<const Args&>::type...>>
                                                   ^~
                                                   > >
.../googletest/src/googletest/googlemock/include/gmock/gmock-matchers.h:5186:54: error: a space is required
      between consecutive right angle brackets (use '> >')
      tuple<typename std::decay<const Args&>::type...>>(
                                                     ^~
                                                     > >
2 warnings and 4 errors generated.
make[5]: *** [googlemock/CMakeFiles/gmock.dir/src/gmock-all.cc.o] Error 1
make[4]: *** [googlemock/CMakeFiles/gmock.dir/all] Error 2

[...]

PR #1890, opened 2018-10-09
Skip some warnings for clang-cl

peterjc123:


gennadiycivil:
@peterjc123 thank you for your contribution. Could you please explain what problem this is solving and why this is beneficial

peterjc123:
@gennadiycivil Sorry for the missing of the context. Without this change, building `googletest` with clang-cl will cause a warning `-std=c++11` is not recognized. With the flag `-WX` specified, the warning will become an error and cause the build to fail. This PR tries to fix this by turn off warnings for unrecognized compiler options and thus enable clang-cl build for `googletest`.

gennadiycivil:
@peterjc123 this seems to be an explosion of conflicting flags . Why not just specify Wno-unknown-argument on the command line when building?

gennadiycivil:
ping

peterjc123:
@gennadiycivil  Yes, I've tried your idea and it works. However, I think the analysis of this issue is still necessary. The build error is caused by using `-std=c++11` and '-WX' at the same time. The first one is not a recognized option and will trigger a warning in clang-cl, then '-WX' turns it into an error.
What do you think if I remove `-std=c++11` for MSVC? Or it may be a better idea to file a patch at the CMake side since it is `set(CMAKE_CXX_STANDARD 11)` that caused this error.

gennadiycivil:
@peterjc123 I have no opinion on CMake , we dont use it. I see some people in the community experiencing similar issues so I ll ping them here @whiterabbit963, @jonbronson , @goatshriek. 
https://github.com/google/googletest/issues/1883

MaEtUgR:
On Cygwin specifying `set(CMAKE_CXX_EXTENSIONS OFF)` breaks the build exactly like described in https://stackoverflow.com/questions/18784112/googletest-1-6-with-cygwin-1-7-compile-error-fileno-was-not-declared-in-this?rq=1

- `add_definitions(-std=c++11)` also breaks it.
- Removing the whole block of compile std defines builds fine

@peterjc123 Is `set(CMAKE_CXX_EXTENSIONS OFF)` really necessary for you? If yes I'll create a pr to only exclude the line for Cygwin to makes sure it builds out of the box on that platform as well.

peterjc123:
@MaEtUgR I used `set(CMAKE_CXX_EXTENSIONS OFF)` just to ensure that it works the same as `add_definitions(-std=c++11)`. So actually you could do that if you are sure that is the cause on Cygwin.

MaEtUgR:
Thanks for your quick answer. Makes sense, now that you explain. `-std=c++11` doesn't define the necessary file IO symbols under Cygwin, that's what's explained under the stack overflow link I posted. Here's the pr to fix the Cygwin build: #2101

PR #1893, opened 2018-10-09
For older CMake, avoid $<IF:...> syntax

### LINKED ISSUE #1892: CMakeLists.txt using $<IF:...> syntax when not supported by that version of CMake
Example of a broken build when using CMake 3.7:

https://source.cloud.google.com/results/invocations/c18bbd62-5117-4837-8aaa-5665b8e1b3ee/log 

Excerpt:

```
-- Google Mock found - building tests
-- Configuring done
CMake Error:
  Error evaluating generator expression:
    $<IF:$<CONFIG:Debug>,gmockd,gmock>
  Expression did not evaluate to a known generator expression
CMake Error:
  Error evaluating generator expression:
    $<IF:$<CONFIG:Debug>,gmock_maind,gmock_main>
  Expression did not evaluate to a known generator expression

```

The $<IF:...> syntax was introduced in CMake 3.8.
Insteda of $<IF:expr,A,B> we can use $<expr:A>$<$<NOT:expr>:B> construction instead.

dneto0:
The $<IF:...> syntax was introduced by CMake 3.8.

Fixes #1892

### DIFF
--- a/googletest/cmake/internal_utils.cmake
+++ b/googletest/cmake/internal_utils.cmake
@@ -352,7 +352,7 @@ function(install_project)
         get_target_property(t_pdb_name_debug ${t} COMPILE_PDB_NAME_DEBUG)
         get_target_property(t_pdb_output_directory ${t} PDB_OUTPUT_DIRECTORY)
         install(FILES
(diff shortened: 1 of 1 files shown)


dneto0:
Rebased against master

PR #1903, opened 2018-10-11
Use pcfiledir in the pkg-config file

fandjelo:
Using absolute paths in the pkg-config file makes it not relocatable and
leads to problems, when trying to use it with precompiled cross
toolchains. Setting prefix to relative path based on pcfiledir makes it
more reliable for such cases.


fandjelo:
I signed it!

PR #1905, opened 2018-10-11
Use pcfiledir for prefix in pkgconfig file

fandjelo:
Using absolute paths in the pkg-config file makes it not relocatable and
leads to problems, when trying to use it with precompiled cross
toolchains. Setting prefix to relative path based on pcfiledir makes it
more reliable for such cases.


gennadiycivil:
What is the before/after effect this would create and how is this beneficial?  "leads to problems" is not sufficient

fandjelo:
Thank you for the reply. Currently generated pkg-config (f.e. gtest.pc) file contains absolute paths to the includes and libraries directories. When building a package, which final usage location is unknown during the compilation, the paths could be incorrect, so using pkg-config through the environment variable PKG_CONFIG_PATH would not work. This is often needed, when using cross compiler environment with precompiled cross libraries (like gtest). To solve this problem all paths in the pc file can be generated relative to the installation path, using prefix set to ${pcfiledir} variable. This way the built package can be located and work anywhere. If it helps, following text explains it a little bit more in detail: https://wp.colliertech.org/cj/?p=164

PR #1906, opened 2018-10-12
Added RemoveDuplicateTests method to UnitTest (2)

JordiGR:
In an effort to integrate googletest as a plug-in for Unreal Engine editor, the need to allow the removal of duplicate tests (by name) arose.
The Unreal Engine editor supports hot reloding of project modules (DLLs). If these modules contain googletest files, when they are reloaded, the tests get appended to the set of existing tests of the living singleton instance that gets created when the googletest plug-in gets loaded for the first time.


JordiGR:
Ping @gennadiycivil

JordiGR:
@gennadiycivil ping
This PR comes from the https://github.com/google/googletest/pull/1870 with the requested changes.

gennadiycivil:
@JordiGR lets discuss a bit the use-case - because the solution certainly has a feeling of trying to cover-up a larger problem.
The "hot reloading" of the project modules you are mentioning and the fact that " the tests get appended to the set of existing tests of the living singleton instance" would lead me to believe that it should be addressed on the loading code to not allow duplicates, not trying to clean up after the fact.

Could you please comment on that
Thanks

JordiGR:
Hi @gennadiycivil. Thanks for your input.
I thought about checking the duplicate tests in the loading code. However, I chose to do it as a post-process because the solution solves a very specific use case (hot-reloading of Unreal modules), and I didn't want to impose the overhead of searching for duplicates to everyone using Google Test. Since the tests are not ordered alphabetically, the search for duplicate names is a linear search. Since some projects might have thousands of tests, I thought that the impact of performing this search for every single added test could not be negligible.

Bu11etmagnet:
Perhaps there should be a way to prevent re-registering the tests when the DLL is re-loaded, if first-time and subsequent loads can be distinguished.

gennadiycivil:
@JordiGR I understand. Still - think about logically what this PR is proposing - "we will load this twice and dedup the resulting messy conditions"..... this seems a bit messy and not really googletest's responsibility to support.

JordiGR:
@Bu11etmagnet, actually is the new set of tests that must be reatined over the existing ones, since the body of any of them might have been modified an the previous tests are obsolete (and potentially wrong).
@gennadiycivil, the solution I propose is trying to solve a real-life problem in a world-widespread videogames engine like Unreal, in a way that it does not affect at all existing test codebases. It would be a pitty to close the path of googletest into the engine and to miss the oportunity to promote TDD in a programming industry where it is still not as widespread as in other industries.

gennadiycivil:
While this effort is commendable the proposed implementation can not be accepted as-is. 
Feel free to propose another solution.

Issue #1737, opened 2018-08-14
Add function to enable the mocked object to ask if it is Nice or Strict


BrukerJWD:
I have two mocked objects: A factory and a child. I can only pass the factory to the code I want to test, so I told the FactoryMock that it shall return a ChildMock by default:
```
FactoryMock() {
    using namespace ::testing;
    ON_CALL(*this, GetChild()).WillByDefault(Return(std::make_shared<ChildMock>()));
}
```
Now I would like to also pass the NiceMock/StrictMock "tag". This would be easy if there were functions like `testing::Mock::IsNice(void*)` and `testing::Mock::IsStrict(void*)` that would expose the internal from `g_uninteresting_call_reaction`, so I could do
```
FactoryMock() {
    using namespace ::testing;
    ON_CALL(*this, GetChild()).WillByDefault(Invoke([this]() {
        Mock::IsNice(this)   ? return std::make_shared<NiceMock  <ChildMock>>() :
        Mock::IsStrict(this) ? return std::make_shared<StrictMock<ChildMock>>() :
       std::make_shared<ChildMock>();
    }));
}
```

gennadiycivil:
Thank you very much for this report. The best way to approach this would be to create a proper PR and submit it for consideration.

PR #1911, opened 2018-10-16
Add function to enable the mocked object to ask if it is Nice or Strict

BrukerJWD:
PR for issue #1737


BrukerJWD:
I signed it!

BrukerJWD:
Seems like I didn't have the latest master locally, but pushed into the current master :/

PR #1915, opened 2018-10-17
Fix a race condition in Fuchsia death tests.

Steelskin:
The exception port binding was set after the child process was
started, resulting in a race condition where the child process was
crashing before the exception port binding was set. Child processes
are now started in another job, whose exception port is bound before
the child process creation.
In addition, this CL changes the stderr redirection for the child
process to be done via Zircon sockets.


gennadiycivil:
@Steelskin Thank you for the change. There appears to be 2 changes here one for the race condition and the other for the stederr capture. 
Could you please describe their relationship. 
Thank you

gennadiycivil:
@Steelskin ping

Steelskin:
Sorry, this fell off my radar as I was fighting other fires. As for your question, we do have 2 changes, for 2 separate possible race conditions. The stderr redirection was another race condition where the output for the child process was closed too early and we risked missing some.
I can make 2 separate changes if you want, I need to get started on the internal change anyway.

gennadiycivil:
@Steelskin Thank you - yes I think it would be better to have 2 separate changes. Also if you have ability to do the internal change that works just fine, after it is submitted I can push it to OSS.
Please let me know which way you prefer.
Thanks

Steelskin:
Great, I'll make the internal change in 2 CLs then.

PR #1929, opened 2018-10-24
add documentation of manual c++11 specification

goatshriek:
Per #1883, builds of Google Test may fail if the version C++ is not manually
set to C++11 during the build process.


gennadiycivil:
@goatshriek this is very reasonable. Thank you

PR #1936, opened 2018-10-25
Add ASSERT_THROW_MATCH to examine the thrown exception

BrukerJWD:
The most tests I write are to check the error handling which usually involes throwing exceptions. All I currently can do with gtest is checking **whether** an exception of a specified type is thrown. But what I want to do is also testing whether this exception matches my criteria (message, member, ...).

Example:

    struct CustomException {
        int Member;
        CustomException(int member) : Member(member) {}
    };
    void ThrowException(int member) { throw CustomException(member); }
    
    ASSERT_THROW(ThrowException(1337), CustomException); // is the exception thrown? --> yes.
    // but does it also have the Member == 1337?

This is why I want to add an `ASSERT_THROW_MATCH` macro which applies the gmock matchers onto the catched exception to achieve this.

      ASSERT_THROW_MATCH(ThrowException(1337), CustomException,
          Field("Member", &CustomException::Member, Eq(1337))),

I don't like the naming but couldn't come up with something better. We can also move the code to other files, but I didn't what to put it into `gtest-matchers.h` like the ASSERT_THAT things (because it is no matcher). As soon as you say you like it I can also write something in the docs.

### DIFF
--- a/googlemock/include/gmock/gmock-gtest-fusion.h
+++ b/googlemock/include/gmock/gmock-gtest-fusion.h
@@ -1,0 +1,89 @@
+// Copyright 2007, Google Inc.
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are
+// met:
+//
+//     * Redistributions of source code must retain the above copyright
+// notice, this list of conditions and the following disclaimer.
+//     * Redistributions in binary form must reproduce the above
+// copyright notice, this list of conditions and the following disclaimer
+// in the documentation and/or other materials provided with the
+// distribution.
+//     * Neither the name of Google Inc. nor the names of its
+// contributors may be used to endorse or promote products derived from
+// this software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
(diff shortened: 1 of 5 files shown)


gennadiycivil:
Thank you for this report. We are actively discouraging adding new macros

BrukerJWD:
@gennadiycivil, @JonathanDCohen: This is really cool concept for testing exception safety and I would like to use it right away :)

But my macro here has a different scope, it is not about exception safety. Perhaps it gets clearer when I explain my use case:
I am writing library code against an error-code driver API and if an error occurs I throw an exception. In some cases I am not only throwing specific exception types (which `ASSERT_THROW` could test) but these instances also have some properties like an enum that explains which specific part went wrong. This information is important for our application but I do not want to create a specific exception type for each one of it because then the try-catch gets really extensive. Thus I want to test that the thrown exception really has the enum member value that I expect after I picked on my code by giving it an error code through gmock.

BTW: May I ask why you are discouraging adding new macros?

BrukerJWD:
@ping gennadiyvicil

gennadiycivil:
Sorry for the delay.

Firstly - this new proposed functionality would be useful.

Secondly - we are discouraging new macros as a rule , however in this case
it might be possible instead of creating all new macro, to modify an
existing macro. Consider adding an optional parameter - If the optional
parameter is used then the macro applies  gmock matchers onto the catched
exception.

Thank you again

PR #1950, opened 2018-11-01
Updating Sample Makefile to Specify C++11

benjamincarman:
Added a line to the sample Makefile in googletest/googletest/make to specify the use of C++11 in CXXFLAGS as required by googletest. This flag is required on some systems to make sure C++11 is used and it should be automatically included in the sample Makefile in order for it to run out of the box.


gennadiycivil:

benjamincarman:
If someone were to run the sample unit test directly with this Makefile, it would likely result in a string of errors due to the lack of specified C++11 compilation. The first few might appear as follows:

firstgtesterrors
etc. etc.
gtestfatalerror

Simply adding the C++11 flag to this makefile allows new users of Googletest the ability to quickly run its sample unit tests out of the box through this given Makefile, without any compilation errors.

Issue #676, opened 2016-01-05
Missing table of contents for docmentation


apbarrero:
I am a heavy user of Google Test documentation reference and have always promoted its use among colleagues claiming its extensive and complete documentation as a key feature.

Since the project home has been moved to Github, the table of contents for the docs has been lost. It was much easier to navigate through the docs before.

I guess Github's markdown can't automatically generate the TOC and keeping it hardcoded would be a maintenance pain and eventually would lead to inconsistent contents.

Would moving/linking the docs to the project Wiki help?

CJCombrink:
I agree with @apbarrero, since the TOC is missing it is much harder to navigate the documentation.

Wouldn't it be possible to add a TOC to the Wiki pages of github and perhaps move the documentation there...

apbarrero:
@BillyDonahue, are there any plans regarding the docs that you could share here?

I can help to moving the docs to the wiki if you find that a good solution.

BillyDonahue:
Moving to wiki is not my favorite way to go. I like having the docs checked in as part of the project.
Perhaps GitHub has or is developing a way to generate a ToC from Markdown?

apbarrero:
Just found [this answer is SO](http://stackoverflow.com/a/18244520/198011), and there are no references to ToC in [github's markdown reference](https://guides.github.com/features/mastering-markdown/).

I'm not confident that they might be implementing ToC automatic generation any soon.

AlekseiCherkes:
+1

I think, It's not a big burden to generate the first version of the TOC by script and then update it manually.

[https://github.com/google/googletest/pull/646](url)

Farwaykorse:
These are the main reference documents.
They are way too long and unorganized. You have to go through all to content to find a (potential) relevant section. And remember your way around.

I prefer a good Markdown documentation over the wiki, since it comes with the code and stays in sync with the code version you are using.  
Adding a TOC is essential to navigate these documents. But in-document linking is not a feature. Headers don't add have the option to add the required id tags.

#646 relies on GitHub making headers accessible by linking. Not entirely portable, but it is at least readable as plain-text.

edit: added a suggestion to #646 to make it portable.

PR #1951, opened 2018-11-02
Add table of contents and convert documentation to asciidoc

charlesjwhite:
The goal is to improve documentation readability and accessibility.
Many people are struggling to jump right to specific chapter quickly because of the lack of table of contents.

I found no way to add a dynamically generated table of contents in markdown format.
Though, it seems to be possible using asciidoc format. So I converted markdown *.md files to asciidoc *.adoc files and added a table of contents on large pages.

Some people had similar questions but were closed as "outdated": https://github.com/google/googletest/issues/676 https://github.com/google/googletest/pull/646


gennadiycivil:
Thank you very much for this contribution. Googletest standardized on markdown both internally and externally.
We realize that the doc set needs is better navigation and we would welcome another solution that would stay with markdown.

charlesjwhite:
would such a solution be acceptable for you: run a command line to manually update an automatically generated toc (https://github.com/ekalinin/github-markdown-toc)?

PR #1952, opened 2018-11-03
Remove `-Werror` from the build flags if built as a dependency

### LINKED ISSUE #1943: GoogleTest should not define -Werror or equivalent by default
As mentioned in the now closed https://github.com/google/googletest/issues/1373#issuecomment-355572722, it would probably be a good idea to not enable `-Werror` (or equivalently `/WX`). This can unnecessarily break the build for new compiler versions.

https://github.com/google/googletest/blob/b9347b31c338851879d0105f0fe32d09007f0433/googletest/cmake/internal_utils.cmake#L102

https://github.com/google/googletest/blob/b9347b31c338851879d0105f0fe32d09007f0433/googletest/cmake/internal_utils.cmake#L70

heinrich5991:
This makes GoogleTest more future-proof against new compilers that might
introduce new warnings, causing dependent projects to fail their build.

Fixes #1943.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,23 +25,23 @@ matrix:
       compiler: gcc
       install: ./ci/install-linux.sh && ./ci/log-config.sh
       script: ./ci/build-linux-autotools.sh
-      env: VERBOSE=1 CXXFLAGS=-std=c++11
+      env: VERBOSE=1 CXXFLAGS="-Werror -std=c++11"
     - os: linux
       group: deprecated-2017Q4
       compiler: gcc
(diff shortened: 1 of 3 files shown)


heinrich5991:
I'm not going to send Google my address and phone number. Why is that required to send a patch and why do you suggest people to "just send a patch" if that's a prerequisite? :(

https://github.com/google/googletest/issues/1943#issuecomment-434526323

Feel free to include this patch under whatever license you want, I grant Google all rights to it.

gennadiycivil:
As per https://github.com/google/googletest/blob/master/CONTRIBUTING.md the CLA is required.

PR #1959, opened 2018-11-10
Remove old MSVC workarounds

robinlinden:
Since the readme states that MSVC 2015 or later is a requirement, it might make sense to delete all of the workarounds for older versions and bump the version check in `googlemock/include/gmock/internal/gmock-port.h` to match the readme.

I also noticed that the pragma warning disable in `googlemock/test/gmock-actions_test.cc` has different conditions for when to push and when to pop, so now they're the same (`_MSC_VER == 1900`). If you want to keep the workarounds for older MSVC versions, I can open a PR with just the pragma warning fix.

There's also some mention of Windows CE which I wasn't sure what to do with. It should probably also be deleted since that platform's been end-of-life for a long time now.


robinlinden:
I've signed the CLA.

robinlinden:
The AppVeyor CI system is currently not passing due to something unrelated. I opened a PR fixing it here: #1960

PR #1961, opened 2018-11-13
Update README.md

coppered:
added -std=c++11

otherwise it shows during a build:

> /googletest/include/gtest/internal/gtest-port.h:891:12: error: 
      no member named 'make_tuple' in namespace 'std'
using std::make_tuple;
      ~~~~~^


coppered:
I signed it!

Issue #1955, opened 2018-11-07
THREADS_PREFER_PTHREAD_FLAG


lukasm91:
https://github.com/google/googletest/blob/master/googletest/cmake/internal_utils.cmake#L59 

On this line we set `THREADS_PREFER_PTHREAD_FLAG` to ON. This may trigger strange behaviour and also errors in certain cases:

- If you link a CUDA-enabled language against a gtest on a machine where CMake detects `-pthread`, this will not compile because the flags is propagated to the CUDA target, but the flag is not recognized by nvcc. On this machine, we don't want the pthread flag.  This is probably not that important as this is a general CMake problem.
- If you do `find_package(Threads)` already before adding googletest, this will lead to strange behaviour: The first cmake pass will do the `find_package(Threads)` with the option disabled  (or whatever the user specified), the second cmake pass will add the flag and find `-pthread`.

gennadiycivil:
Thank you very much for this report. The best way to approach this would be to create a proper PR and submit it for consideration.

gennadiycivil:
ping

PR #1977, opened 2018-11-22
Issue #1955: Remove THREADS_PREFER_PTHREAD_FLAG

lukasm91:
I think Googletest should not set `THREADS_PREFER_PTHREAD_FLAG` on its own. 

- If you link a CUDA-enabled language against a gtest on a machine where CMake detects -pthread, this will not compile because the flags is propagated to the CUDA target, but the flag is not recognized by nvcc. On this machine, we don't want the pthread flag. This is probably not that important as this is a general CMake problem and will be fixed with new CMake Versions (there are many issues, e.g. https://gitlab.kitware.com/cmake/cmake/merge_requests/2529 for MPI and many others).
- More importantly, if you do `find_package(Threads)` already before adding googletest, this will lead to strange behaviour: The first cmake pass will do the find_package(Threads) with the option disabled (or whatever the user specified), the second cmake pass will add the flag and find -pthread.

See also #1955

### DIFF
--- a/googletest/cmake/internal_utils.cmake
+++ b/googletest/cmake/internal_utils.cmake
@@ -56,7 +56,6 @@ macro(config_compiler_and_linker)
   unset(GTEST_HAS_PTHREAD)
   if (NOT gtest_disable_pthreads AND NOT MINGW)
     # Defines CMAKE_USE_PTHREADS_INIT and CMAKE_THREAD_LIBS_INIT.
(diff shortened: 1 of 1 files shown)


SoapGentoo:
We will probably have to set this flag downstream then, when packaging GTest. In general, for most compilation targets, `-pthread` also affects macros and stuff. But it's fine, go ahead.

Issue #1967, opened 2018-11-17
Add WaitForAndClearExpectations() for testing async side effects


calder:
Would it be reasonable to add the following helper to [`gmock-spec-builders`](https://github.com/google/googletest/blob/master/googlemock/src/gmock-spec-builders.cc) for testing functions with asynchronous side effects?

```cpp
bool Mock::WaitForAndClearExpectations(void* mock_obj, std::chrono::nanoseconds timeout);
```

gennadiycivil:
@calder Thank you for this proposal. Could you please provide more details, potential use case, and most importantly how are you accomplishing this "today"? Thanks again

calder:
@gennadiycivil: Thanks! Here's an example use case:

Let's say we want to test the behavior of `Foo::DoSomething()`.
* `Foo` takes a `Bar` in its constructor.
* `Foo::DoSomething()` calls `Bar::DoSomethingElse()` asynchronously (via `std::async` for example).

If `Bar::DoSomethingElse()` were called synchronously, we would just create a mock `Bar`, pass it to `Foo`'s constructor, set up expectations on the `Bar` mock, and call `Foo::DoSomething()`.

Instead, because it's asynchronous, after calling `Foo::DoSomething()` we have to either:
* Sleep long enough for `Bar::DoSomethingElse()` to probably have been called. This is undesirable because it slows down tests significantly, makes them brittle to timing changes, and introduces potential flakes.
* Modify production code to signal some synchronization primitive when the async operation completes. This is even less desirable since it impairs readability of production code.

So far I have usually opted for "sleep synchronization" with all its associated problems.

calder:
@gennadiycivil: Does https://github.com/abseil/googletest/pull/1991 look like roughly the right direction?

leunMar:
@calder @gennadiycivil This would be a very useful feature for testing in ROS, where we have asynchronous callbacks (http://wiki.ros.org/roscpp/Overview/Publishers%20and%20Subscribers).

Actually, there is another workaround:
You can wake up a sleeping thread when the mock function is called defining an appropriate action.
https://github.com/PilzDE/pilz_robots/blob/5a969f06e6b7626e2ddb3c485f0a5ca645bbaff4/prbt_hardware_support/test/include/prbt_hardware_support/async_test.h

gennadiycivil:
See #1991

PR #1991, opened 2018-11-29
Issue #1967: Add Mock::WaitForAndClearExpectations() for async effect testing.

calder:
See #1967 for background.

### DIFF
--- a/googlemock/docs/CheatSheet.md
+++ b/googlemock/docs/CheatSheet.md
@@ -562,6 +562,11 @@ Mock::VerifyAndClearExpectations(&mock_obj);
 // also removes the default actions set by ON_CALL();
 // returns true iff successful.
 Mock::VerifyAndClear(&mock_obj);
+...
+// Waits up to 1s for all expectations to be satisfied on
+// mock_object; returns true iff successful.
+// Always removes all expectations regardless of success.
+Mock::WaitForAndClearExpectations(&mock_obj, std::chrono::seconds(1));
 ```
 
 You can also tell Google Mock that a mock object can be leaked and doesn't
--- a/googlemock/docs/CookBook.md
+++ b/googlemock/docs/CookBook.md
@@ -2576,6 +2576,31 @@ and nothing should happen between the two check points. The explicit
 check points make it easy to tell which `Bar("a")` is called by which
 call to `Foo()`.
 
+## Verifying Asynchronous Side Effects ##
+
+Sometimes you want to verify that an asynchronous side effect has occured. You could sleep long enough that the effect will _probably_ occur, but this:
+1. Breaks thread sanitizers like TSan which (correctly) do not treat this sleep as a synchronization boundary.
+2. Makes tests slower since sleeps always take the maximum duration.
+
+Instead, you can use `Mock::WaitForAndClearExpectations(&mock_object, duration)`:
+
+```cpp
+using ::testing::MockFunction;
+
+TEST(FooTest, InvokesFunctionAsynchronously) {
+  MockFunction<void(string)> func;
+  EXPECT_CALL(func, Call("1"));
+
+  std::thread thread{[&](){
+    std::this_thread::sleep_for(std::chrono::milliseconds(10));
+    func.Call("1");
+  }};
+
+  Mock::WaitForAndClearExpectations(&func, std::chrono::seconds(1));
+  thread.join();
+}
+```
+
 ## Mocking Destructors ##
 
 Sometimes you want to make sure a mock object is destructed at the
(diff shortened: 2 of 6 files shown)


gennadiycivil:
ping

calder:
@gennadiycivil: I got some time to look at this. All builds are now passing.

calder:
@gennadiycivil: Friendly ping.

gennadiycivil:
@calder 

Looking at your example it does not appear that this should be part of the mocking framework.

>>>
Foo takes a Bar in its constructor.
Foo::DoSomething() calls Bar::DoSomethingElse() asynchronously (via std::async for example).
>>>

so, continuing here once you call Bar::DoSomethingElse() asynchronously via std::async, you get std::future back where you could call std::future::get to get the result https://en.cppreference.com/w/cpp/thread/future/get

This just does not appear to be withing mocking framework scope.

calder:
@gennadiycivil: Yeah, that works great for functions that actually return a future. Some functions don't (and shouldn't) return a future or provide any way of determining when asynchronous side effects have completed.

As an example, this crops up a lot in eventually consistent open loop controllers like network programmers.

sbenzaquen:
On Sun, Dec 30, 2018 at 2:53 AM Calder Coalson <notifications@github.com>
wrote:

> @gennadiycivil <https://github.com/gennadiycivil>: Yeah, that works great
> for functions that actually return a future. Some functions don't (and
> shouldn't) return a future or provide any way of determining when
> asynchronous side effects have completed.
>
If there is no way to know when the side effects have completed, then there
is no way to wait for them.
Waiting on some mock method to be called is not enough. The caller of that
method is still running and waiting for the result.

Even if you want to notify the test when the mock method is called, you can
easily do that from your end without extra support from the testing
framework.
Just specify the action on the mock method to signal some notification.

Eg, the example in
https://en.cppreference.com/w/cpp/thread/condition_variable is more or less
what you want, except that the notify_one happens in the action, like:

EXPECT_CALL(mock, Foo()).WillOnce([&] {
  std::unique_lock<std::mutex> lk(m);
  done = true;
  cv.notify_one();
});

This could be simpler if you have a higher level abstraction,
like absl::Notification, like:

absl::Notification notification;
EXPECT_CALL(mock, Foo()).WillOnce([&] { notification.notify(); });
notification.WaitForNotification();

> As an example, this crops up a lot in eventually consistent open loop
> controllers like network programmers.
>
Sometimes when the background system is really async you just have to
poll-and-sleep waiting for the results. This is a last ditch option, but
works.
Like:
while (!IsSideEffectDone()) absl::SleepFor(absl::Microseconds(10));

Obviously, whatever `IsSideEffectDone` is checking has to be properly
synchronized. You can't look at unsynchronized data that some other thread
might write because that has undefined behavior.

calder:
@sbenzaquen: That works, but adds a lot of boilerplate to every async-satisfied expectation.

See https://github.com/google/googletest/issues/1967#issuecomment-441458616 for previous discussion of alternatives.

calder:
@gennadiycivil: Friendly ping.

gennadiycivil:
@calder Thank you again for this effort. As discussed, this extra support for the async execution is outside of what we deem reasonable for the testing/mocking framework.
Appreciate your effort and good design conversation but we can not accept it.

calder:
@gennadiycivil: Sorry, I don't think I understand that reasoning.

It seems like 2 features are pretty fundamental to any mocking framework:
1. Simplify injection of behavior for testing how production code responds in certain scenarios.
2. Simplify assertions about the behavior of production code under these scenarios.

Both of these are *possible* without a mocking framework; mocking frameworks just make expressing them *simpler and more readable*.

@leunMar and I are saying that as users of GMock, a `WaitForAndClearExpectations()` helper would make assertions about a specific class of behavior (asynchronous side effects) *simpler and more readable*. If (2) is indeed a goal of any mocking framework, then this change falls squarely within that goal.

So while it's fair to reject this feature as too niche to be worth the burden of the increased API surface area, I don't think it's fair to say it's out of scope for a mocking framework.

# Alternative

If the GMock team decides this feature is indeed too niche to support, we could ideally still implement it in our own code bases. This is currently impossible because there's no public interface for testing if a Mock's expectations have been met. Exposing a `Mock::AreExpectationsMet()` function would enable this.

calder:
Thanks for the resonse @sbenzaquen. I can definitely see the potential for misuse, I just also see a decent potential for valid (non-racy) use.

A simple example is a class whose destructor drains all outstanding asynchronous operations, but whose individual methods do not provide an API for determining when their asynchronous side effects have completed.

If you all think the cost of potential misuse exceeds the benefit to valid users, I'll suck it up and continue to maintain a private fork.

Issue #1996, opened 2018-12-04
I don't believe GTEST_IS_THREADSAFE should depend on GTEST_HAS_SEH


gpakosz:
Hello,

I don't believe `GTEST_IS_THREADSAFE` should be defined inside `GTEST_HAS_SEH`.

I tried to fix that by submitting #935 but it was neglected 🤷‍♂️.

Can someone please comment on this?

gpakosz:
Thanks for merging

PR #1997, opened 2018-12-04
Do not define GTEST_IS_THREADSAFE within GTEST_HAS_SEH

gpakosz:
See #1996.


gennadiycivil:
@gpakosz 
What is the before/after effect this would create and how is this beneficial? Your explanation I don't believe GTEST_IS_THREADSAFE should be defined inside GTEST_HAS_SEH." is no sufficient.

Thanks again
G

gpakosz:
I can't remember in which situation I needed this change. But in any case the concept of [SEH](https://docs.microsoft.com/en-us/cpp/cpp/structured-exception-handling-c-cpp?view=vs-2017) has nothing to do with threading.

gpakosz:
Thanks for the merge

PR #2000, opened 2018-12-05
Issue #1659: Add Support for PlatformIO and Arduino

### LINKED ISSUE #1659: [Feature Request] Support for Arduino / PlatformIO
It would be nice if GTest was supported in the PlatformIO framework for Arduino.

One of my projects runs on both desktop OSes and IoT (ESP32 and ESP8266).  I use GTest for Windows/Mac/Linux and Aunit with an adapter to use the GTest interface.

Native support would be nice if possible.  Thanks.

ciband:
This is an initial effort to add PlatformIO/Arduino support to GTest and GMock.

This PR is not meant to be a final version, just the starting point to the discussion to gain feedback.

Is this general approach sound and what changes would you like to see?

I would like to target ESP8266 and ESP32 boards initially.

This would ultimately also be registered in the PIO Library Manager.

Fixes #1659 

Thank you.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,7 @@ matrix:
       dist: trusty
       sudo: required
       group: deprecated-2017Q3
+      before_install: chmod -R +x ./ci/*platformio.sh 
       install: ./ci/install-platformio.sh
       script: ./ci/build-platformio.sh      
     - os: linux
@@ -49,9 +50,6 @@ matrix:
     - os: osx
       env: BUILD_TYPE=Release VERBOSE=1 CXX_FLAGS=-std=c++11
       if: type != pull_request
-
-before_install:
-    - chmod -R +x ./ci/*.sh 
 
 # These are the install and build (script) phases for the most common entries in the matrix.  They could be included
 # in each entry in the matrix, but that is just repetitive.
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,12 @@ language: cpp
 matrix:
   include:
     - os: linux
+      dist: trusty
+      sudo: required
+      group: deprecated-2017Q3
+      install: ./ci/install-platformio.sh
+      script: ./ci/build-platformio.sh      
+    - os: linux
       compiler: gcc
       sudo : true
       install: ./ci/install-linux.sh && ./ci/log-config.sh
@@ -43,6 +49,9 @@ matrix:
     - os: osx
       env: BUILD_TYPE=Release VERBOSE=1 CXX_FLAGS=-std=c++11
       if: type != pull_request
+
+before_install:
+    - chmod -R +x ./ci/*.sh 
 
 # These are the install and build (script) phases for the most common entries in the matrix.  They could be included
 # in each entry in the matrix, but that is just repetitive.
--- a/googlemock/src/gmock_main.cc
+++ b/googlemock/src/gmock_main.cc
@@ -34,8 +34,12 @@
 
 #ifdef ARDUINO
 void setup() {
-  int argc = 0;
-  char** argv = nullptr;
+  // Since Arduino doesn't have a command line, fake out the argc/argv arguments
+  int argc = 1;
+  const auto arg0 = "PlatformIO";
+  char* argv0 = const_cast<char*>(arg0);
+  char** argv = &argv0;
+  
    // Since Google Mock depends on Google Test, InitGoogleMock() is
   // also responsible for initializing Google Test.  Therefore there's
   // no need for calling testing::InitGoogleTest() separately.
--- a/googlemock/src/gmock_main.cc
+++ b/googlemock/src/gmock_main.cc
@@ -32,6 +32,20 @@
 #include "gmock/gmock.h"
 #include "gtest/gtest.h"
 
+#ifdef ARDUINO
+void setup() {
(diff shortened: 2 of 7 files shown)


ciband:
I signed it!

.travis.yml
```diff
@@ -44,6 +50,9 @@ matrix:
       env: BUILD_TYPE=Release VERBOSE=1 CXX_FLAGS=-std=c++11
       if: type != pull_request
 
+before_install:
```
gennadiycivil: This looks like it would apply to all entries in the build matrix. ( Correct me if I am wrong ). It would be preferable to contain the platformio changes only to platformio builds

googlemock/src/gmock_main.cc
```diff
@@ -32,6 +32,20 @@
 #include "gmock/gmock.h"
 #include "gtest/gtest.h"
 
+#ifdef ARDUINO
```
gennadiycivil: Where will this be defined and how?

googletest/src/gtest_main.cc
```diff
@@ -30,8 +30,22 @@
 #include <stdio.h>
 #include "gtest/gtest.h"
 
+#ifdef ARDUINO
```
gennadiycivil: Where will this be defined and how?

gennadiycivil:
@ciband Thank you for this proposal. Please take a look

googlemock/src/gmock_main.cc
ciband: For supporting the Arduino frameworks, the setup()/loop() entry points are what is done.  The ARDUINO define is set by the Arduino framework itself.  As far as I know it is guaranteed by the framework to be there.

googletest/src/gtest_main.cc
ciband: See comment above

.travis.yml
ciband: It would apply to all builds.  I can look at applying only to PIO.  I am not sure why I need it.  The other builds execute scripts out of the ci directory without the explicit +x.  I would prefer to do it that way but my Travis CI skills are not advanced.

Any ideas there?

ciband:
If you are open to this concept then I will do some more polishing on this and push an update.

I'm not sure how to handle argc/argv as Arduino doesn't have the int main(..) signature.  What does gtest/gmock main use the args for?

.travis.yml
gennadiycivil: I just thing that if we are adding PIO specific builds then all changes should be self-contained to PIO only. I believe you can just move this into the PIO section in the travis.yml file

googlemock/src/gmock_main.cc
gennadiycivil: sounds logical.

gennadiycivil:
I think we should get the CI builds working before actual registration can
happen.
I would wait until we have tighten up this PR to the point when we are
'ready'

ciband:
Sounds good.  

Based on my past experience with PIO it has taken me a couple tries to get the library.json file correct such that all the right files get in all the right places when the user downloads the library from the PIO package manager.  I just wanted to set that expectation for the development.

Please let me know anything else you need from me and I'll get it turned around.

Thanks.

ciband:
Here is how PlatformIO handles versioning:
https://docs.platformio.org/en/latest/librarymanager/config.html#libjson-version

To answer your question, I think, yes, we can.
From the link: using case 1 in the versioning tag where we give an explicit version (e.g. 1.8.1) then we can update the library.json when a new release happens to ensure the platformIO entry is locked to the latest release.

I hope I understood the question and I hope that makes sense.

Thank you.

googlemock/src/gmock_main.cc
```diff
@@ -32,6 +32,20 @@
...
+void setup() {
+  int argc = 0;
+  char** argv = nullptr;
```
gennadiycivil: Arduino doesn't have a main() - but testing::InitGoogleMock expects proper argv/argc values.
argc==0 and argv==nullptr are not valid.

Here is one example how to fake this:

 int argc = 1;
  const char *arg0 = "ANYNAME"; // A dummy name for the executable. Any name will do.
  char *argv0 = (char *)arg0;
  char **argv = &argv0;
  ::testing::InitGoogleMock(&argc, argv);

gennadiycivil: Arduino doesn't have a main() - but testing::InitGoogleMock expects proper argv/argc values.
argc==0 and argv==nullptr are not valid.

Here is one example how to fake this:

int argc = 1;
const char *arg0 = "ANYNAME"; // A dummy name for the executable. Any name will do.
char *argv0 = (char *)arg0;
char **argv = &argv0;
::testing::InitGoogleMock(&argc, argv);

ciband: Ok.  I can make that change.  I looked at the code and it appeared to short circuit and return early if argc <= 0 with no change to use/dereference argv.

I will make this change.  Thank you.

ciband: done

gennadiycivil:
@ciband we are going to accept this next time we merge. 
Please leave it in this state, no more changes for now until it is merged. Once this is in you can take it forward with the platofrm

Issue #1614, opened 2018-05-26
ASSERT_EQ segfault with two boost::filesystem::path


graywolf:
ASSERT_EQ segfaults while trying to print in case of not equal if you use two instances for boost::filesystem::path. Minimal example:

```
TEST(a, b) {
    ASSERT_EQ(boost::filesystem::path("a"), boost::filesystem::path("b"));
}
```

when ran under gdb gives you:

```
Program received signal SIGSEGV, Segmentation fault.
0x00007ffff68f50e4 in std::__ostream_insert<char, std::char_traits<char> > (__out=..., __s=__s@entry=0x7fffff7ff06c "{", __n=__n@entry=1)
    at /build/gcc/src/gcc-build/x86_64-pc-linux-gnu/libstdc++-v3/include/bits/ostream_insert.h:76
76	/build/gcc/src/gcc-build/x86_64-pc-linux-gnu/libstdc++-v3/include/bits/ostream_insert.h: No such file or directory.
```

Stacktrace is 116468 levels deep, so attached as attachement (25MB): 
[gdb.txt](https://github.com/google/googletest/files/2041453/gdb.txt)

pepsiman:
Duplicate of #521.

graywolf:
I would think so, but the fix was merged into master and yet this still happend to me. So maybe the fix is not complete?

graywolf:
Anyway #521 is still open, why? Wasn't https://github.com/google/googletest/pull/1186 supposed to fix it?

ixSci:
This issue is not duplicate because the original issue was fixed (I don't understand why you didn't close #521 ?). Then in this [commit](https://github.com/google/googletest/commit/a3c2e107aeb7e821b76164cea366094ce420b656) the issue was brought up again but with different cause. In the aforementioned commit the following code has been added:
`template<typename T>
struct VoidT {
    typedef void value_type;
};

template <typename T, typename = void>
struct HasValueType : false_type {};
template <typename T>
struct HasValueType<T, VoidT<typename T::value_type> > : true_type {
};
`
Which is totally wrong. I do not know why it was added in the first place, but at least making it working it should look like this:
`template<typename T>
struct Void {
    typedef void type;
};

template<typename T>
using VoidT = typename Void<T>::type;

template <typename T, typename = void>
struct HasValueType: false_type {};
template <typename T>
struct HasValueType<T, VoidT<typename T::value_type>>: true_type {
};`

[...]

PR #2002, opened 2018-12-06
Issue #1614: Fix IsRecursiveContainer implementation to avoid infinite recursion

machida-mn:
Bug fix of #1614

[As ixSci mentioned](https://github.com/abseil/googletest/issues/1614#issuecomment-395414407), `HasValueType::value` will be always false and does not make any sense. Also the condition when `IsRecursiveContainer::value` will be true does not exactly match the condition when infinite recursion will be caused. infinite loop is still caused when the `const_iterator` object does not have `value_type` type or `value_type` is cv-qualified (because the result of is_same will be false).

Here is the exact condition when the infinite loop will be caused.

```cpp
// C: template argument of IsRecursiveContainer

is_same_v<remove_const_t<remove_reference_t<*(C::const_iterator())>>, C> == true
```

I changed the `IsRecursiveContainer` implementation so that infinite recursion will not be caused along `PrintTo` function. Also I updated the test to simulate `std::filesystem::path` and `boost::filesystem::path` object more accurately.

This fix eliminated the segfault on gcc 8.2.0 and Clang 6.0.1 (libstdc++) on Ubuntu 18.10.

### DIFF
--- a/googletest/include/gtest/internal/gtest-port.h
+++ b/googletest/include/gtest/internal/gtest-port.h
@@ -2127,6 +2127,32 @@ struct IteratorTraits<const T*> {
   typedef T value_type;
 };
 
+// remove const and reference qualifier
+template <typename T>
+struct remove_constref {
+  typedef T type;
+};
+template <typename T>
+struct remove_constref<T&> {
+  typedef T type;
+};
+template <typename T>
+struct remove_constref<T&&> {
+  typedef T type;
+};
+template <typename T>
+struct remove_constref<const T> {
+  typedef T type;
+};
+template <typename T>
+struct remove_constref<const T&> {
+  typedef T type;
+};
+template <typename T>
+struct remove_constref<const T&&> {
+  typedef T type;
+};
+
 #if GTEST_OS_WINDOWS
 # define GTEST_PATH_SEP_ "\\"
 # define GTEST_HAS_ALT_PATH_SEP_ 1
--- a/googletest/include/gtest/internal/gtest-port.h
+++ b/googletest/include/gtest/internal/gtest-port.h
@@ -2127,32 +2127,6 @@ struct IteratorTraits<const T*> {
   typedef T value_type;
 };
 
-// remove const and reference qualifier
-template <typename T>
-struct remove_constref {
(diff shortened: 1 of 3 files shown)


machida-mn:
I found that the case of **cyclic iterator type** still causes an infinite recursion. For example:

```cpp
class A {
  struct iterator {
    B& operator*();
  }

  typedef iterator const_iterator;

  iterator begin();
  iterator end();
};

class B {
  struct iterator {
    A& operator*();
  }

  typedef iterator const_iterator;

  iterator begin();
  iterator end();
};
```

This is very hardly detected (as cyclic reference of `std::shared_ptr`). Probably it is preferable that the case which causes an infinite recursion will be documented.

googletest/include/gtest/internal/gtest-port.h
```diff
@@ -2127,6 +2127,32 @@ struct IteratorTraits<const T*> {
   typedef T value_type;
 };
 
+// remove const and reference qualifier
```
gennadiycivil: It would be preferable to use standard traits

machida-mn: I've splitted `remove_constref` to `remove_const` and `remove_reference`. In addition, I removed the `remove_reference` in gmock-internal-util.h because it is duplicated.

gennadiycivil: Please completely remove your own implementation and use https://en.cppreference.com/w/cpp/types/remove_reference and related

machida-mn: Including `<type_traits>` requires C++11 compiler. Is it OK?

EricWF:
Yes. Google Test requires C++11 now

googletest/include/gtest/internal/gtest-port.h
machida-mn: Done.

gennadiycivil:
merged

gennadiycivil:
Looks like I have to manually rebase and merge this. The push process did not work.

gennadiycivil:
This has been manually merged

PR #2006, opened 2018-12-12
packageconfig: Don't add libs to Cflags

wak-google:
This makes clang unhappy because it gets a bunch of -lpthread during
non-linking phases.

### DIFF
--- a/googlemock/cmake/gmock.pc.in
+++ b/googlemock/cmake/gmock.pc.in
@@ -8,4 +8,4 @@ Version: @PROJECT_VERSION@
 URL: https://github.com/abseil/googletest
 Requires: gtest
 Libs: -L${libdir} -lgmock @CMAKE_THREAD_LIBS_INIT@
-Cflags: -I${includedir} @GTEST_HAS_PTHREAD_MACRO@ @CMAKE_THREAD_LIBS_INIT@
+Cflags: -I${includedir} @GTEST_HAS_PTHREAD_MACRO@
--- a/googlemock/cmake/gmock_main.pc.in
+++ b/googlemock/cmake/gmock_main.pc.in
@@ -8,4 +8,4 @@ Version: @PROJECT_VERSION@
 URL: https://github.com/abseil/googletest
 Requires: gmock
 Libs: -L${libdir} -lgmock_main @CMAKE_THREAD_LIBS_INIT@
-Cflags: -I${includedir} @GTEST_HAS_PTHREAD_MACRO@ @CMAKE_THREAD_LIBS_INIT@
+Cflags: -I${includedir} @GTEST_HAS_PTHREAD_MACRO@
--- a/googletest/cmake/gtest.pc.in
+++ b/googletest/cmake/gtest.pc.in
@@ -7,4 +7,4 @@ Description: GoogleTest (without main() function)
 Version: @PROJECT_VERSION@
 URL: https://github.com/abseil/googletest
 Libs: -L${libdir} -lgtest @CMAKE_THREAD_LIBS_INIT@
-Cflags: -I${includedir} @GTEST_HAS_PTHREAD_MACRO@ @CMAKE_THREAD_LIBS_INIT@
+Cflags: -I${includedir} @GTEST_HAS_PTHREAD_MACRO@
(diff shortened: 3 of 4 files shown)


gennadiycivil:

wak-google:
Does the updated message make it more clear?

gennadiycivil:
@wak-google I dont see any updates... just "This makes clang unhappy because it gets a bunch of -lpthread during
non-linking phases."

 We need a lot more information to consider the PR.

wak-google:
So I updated the commit message to show the output of the package config file and explain why that doesn't belong in the cflags.

This is the clang-tidy output:
```
warning: -Wl,--end-group: 'linker' input unused [clang-diagnostic-unused-command-line-argument]
warning: -Wl,--start-group: 'linker' input unused [clang-diagnostic-unused-command-line-argument]
warning: -lpthread: 'linker' input unused [clang-diagnostic-unused-command-line-argument]
```

That's because our object generation stages are being invoked like:
```c++ -Itest/9f86d08@@exception@exe -Isrc -I../src -I/usr/local/lib/pkgconfig/../../include -fdiagnostics-color=never --coverage -pipe -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -Wnon-virtual-dtor -std=c++17 -g -Wl,--start-group -lpthread -Wl,--end-group -DGTEST_HAS_PTHREAD=1  -MD -MQ 'test/9f86d08@@exception@exe/exception.cpp.o' -MF 'test/9f86d08@@exception@exe/exception.cpp.o.d' -o 'test/9f86d08@@exception@exe/exception.cpp.o' -c ../test/exception.cpp```

SoapGentoo:
This is fallout from https://github.com/google/googletest/pull/1977
`-lpthread` is a pure linker flag, `-pthread` is also a compiler flag that affects code generation. Honestly, I don't see a good solution this modulo tons of conditional magic to discriminate `-lpthread` from `-pthread` and make the insertion conditional. @gennadiycivil is there any code in GTest that, for instance, depends on `-D_REENTRANT`?

gennadiycivil:
@wak-google @SoapGentoo let me see if I understand , is the purpose of this change is to remove the linker warnings?
warning: -Wl,--end-group: 'linker' input unused [clang-diagnostic-unused-command-line-argument]
warning: -Wl,--start-group: 'linker' input unused [clang-diagnostic-unused-command-line-argument]
warning: -lpthread: 'linker' input unused [clang-diagnostic-unused-command-line-argument]

SoapGentoo:
@gennadiycivil correct (in that we want to avoid linker flags on a compiler invocation). I'd still like to solve this catering to everyone's usecases. Does GTest require `-D_REENTRANT` stuff?

gennadiycivil:
@SoapGentoo I dont have an opinion on cmake-related changes - we don't use it internally. I will delegate to the community

gennadiycivil:
Ping.

SoapGentoo:
@gennadiycivil you still haven't answered whether GTest depends on features enabled by `-D_REENTRANT`?

gennadiycivil:
@SoapGentoo I did not indeed and with a good reason - I dont know.

Quick search indicates (https://blogs.oracle.com/solaris/mtreentrant-v2) that this has been obsolete for some time.
More search indicates that Use g++ -pthread, it is equivalent to g++ -pthread -D_REENTRANT 

As far as PTREAD goes you can see [here](https://github.com/google/googletest/blob/a83cc11abe4856a60d92ceba2d65af8236cc3500/googletest/include/gtest/internal/gtest-port.h#L551)

HTH

P.S All this triggers my gut instinct that these changes are just not worth it. I would be happy to be proven otherwise.

SoapGentoo:
what about injecting `-D_REENTRANT` into the compile line (and pkg-config file) if pthreads was detected successfully? Not exactly a beautiful solution, but caters to this use-case, the cuda one and distros.

gennadiycivil:
Ping.

wak-google:
It's unclear to me how to even reproduce the original bug and it seems like a bug in nvcc / toolchain used. cmake does check to make sure that -pthread is supported by the compiler before building googletest with -DTHREADS_PREFER_PTHREAD_FLAG=ON, so I would imagine that user is using different parameters for building the tested project than what are used for the googletest build (maybe not even using nvcc but just clang?).

Anyway, it still isn't correct to be passing link stage flags as compile only flags so we should fix the package config file. Looking at FindPthreads.cmake and comparing it to the autoconf-archive/ax_pthread.m4 it looks like the cmake script doesn't care at all about determining the correct CFLAGS. I assume nowadays this really doesn't matter and we basically don't need to define _REENTRANT anymore.

gennadiycivil:
Based on the discussion I am closing this

PR #2017, opened 2018-12-18
Add snapshot testing

cristiklein:
This commit adds `EXPECT_EQ_SNAPSHOT(val)` to assert that the value should be equal to a snapshot. Snapshots can be generated (the first time) or updated by setting `GTEST_UPDATE_SNAPSHOT=1` in the environment.

Snapshot testing is useful when writing equality assertions is tedious, but checking that the output of your code "by hand" is rather quick. For example, your code might generate images, HTML, SQL statements, etc. If such output changes rather rarely, e.g., you do a lot of refactoring, but the output must not be affected, then snapshot testing might work for you.

Currently, only `std::string` can be checked against a snapshot, however, in case there is demand, the implementation can be extended to other types.


gennadiycivil:
@cristiklein Thank you for this , looks interesting. It may be clear-er to see exactly how this works when you add unit tests - could you please add unit tests and then we could look at it
Thanks

cristiklein:
@gennadiycivil Good point! Please see updated PR.

cristiklein:
Hi @gennadiycivil ,

Sorry for that, somehow missed that all copyright should be attributed to Google. I took the liberty to add myself to the contributors list. Feel free to remove me if you feel that my work does not qualify as non-trivial.

Regards,
Cristian

gennadiycivil:
Looking at the design and the purpose of this proposal. It is not clear that this change is adds significant new functionality and convenience that can not be achieved with the existing framework. 

Using current code a user can generate snapshots as needed as part of the their own test fixture ( for example ) and do the comparison work in the testing.
In addition, the snapshots could be saved in the golden lint file ( if practical ) and used that way.

cristiklein:
Hi @gennadiycivil 

First of all, thank you for your time looking into this PR.

I'm not sure what you mean by "golden lint file". Is there a similar feature in Google Test that I am unaware of? Could you point me to a documentation to read more about it?

Snapshot testing is becoming quite common in other testing frameworks, for example:
* [Jest](https://jestjs.io/docs/en/snapshot-testing)
* [Mocha](https://www.npmjs.com/package/expect-mocha-snapshot)
* [Snapshot Testing for Java](https://json-snapshot.github.io/)

The PR simply brings the convenience of snapshot testing to C++. Although, I agree with you that each developer could write code and eventually call `EXPECT_EQ`, having to write such boilerplate risks giving developers an excuse of not properly unit testing their code. Hence, this PR adds snapshot testing in the very core of Google Test, to avoid lack of convenience becoming a barrier to good testing.

I was also considering creating a library on top of Google Test. However, this has two problems:
(1) As above, it would create an extra barrier. Snapshot testing should (eventually) be a first-class citizen of all testing frameworks.
(2) There are certain Google Test internals that need to be accessed in order to make the name of the snapshot file unique but stable.

Please let me know what are your thoughts about bringing snapshot testing to the C++ world.

gennadiycivil:
@cristiklein There is a number of assumtions that this code is making that may not be true in all cases.
In particular the code makes an assumtion that when it is running it has permission to create and remove files on the filesystem. - which is incorrect is many cases.

gennadiycivil:
The biggest fault in all this is that it introduces dependency on the sequence of execution of the unit tests. This is a no-no.

cristiklein:
Hi @gennadiycivil and Happy New Year,

Thank you for your feedback. I refactored the snapshot testing code to address the issues you raised.

Let me know what you think.

gennadiycivil:
@cristiklein Thank you again and I just want to make sure I am not wasting your time. 
Snapshot testing very fundamentally depends on the sequence of execution of the unit tests. Even when you fix all the other issues, this will be hard to prove because googletest believes very firmly that there should be no dependency on the particular order of execution.

cristiklein:
@gennadiycivil Snapshot testing really consists of two features:
1. Checking that the currently produced value of the UUT is equal to the one in the snapshot: This is the core feature and is now thoroughly tested, with tests being independent.
2. Updating the snapshots.

The latter can be considered sugaring, i.e., the users can update the snapshot files themselves by copy-pasting the output of a failed snapshot test. The output contains both the expected value and the file where the expected value should be stored.

To make the tests independent, I opted not to test snapshot updating in my second PR. I hope this is okey, since snapshot updating is not the core feature, but just sugaring.

Let me know what you think.

gennadiycivil:
@cristiklein So lets say you only leave one feature - check against existing snapshot. Essentially this would mean using something like `EXPECT_EQ_SNAPSHOT(val)`. `val` is then compared to a snapshot stored in a file. 

... Which is no ***really different*** from EXPECT_EQ with a large hardcoded string. 

There is just not enough value here for the amount of change, the core difference is not hardcoding a large string but hardcoding a file name with some content inside.  

Thank you again for the good attempt and good discussion.

gennadiycivil:
I do think that we should keep talking, you mentioned that other testing frameworks implement snapshot testing. It might be useful to try to define a concrete use-case that would prove that having snapshot testing is really critical and the existing framework is not sufficient.

cristiklein:
@gennadiycivil Thank you for your patience.

You convinced me that conveniently updating the snapshot adds a lot of value to snapshot testing and should be considered a first-class citizen. Therefore, I added a test for that too. The snapshots must be written to the source tree -- as highlighted before, this is an essential feature for snapshot testing. The test only runs with Bazel, which provides sandboxing. Hence, the test does not write to the _original_ source tree, but only to the sandbox.

I understand there is a tension between the desire to keep a workspace untouched -- as required for correct builds -- and to write snapshots back to the workspace -- as other snapshot testing frameworks do. I am unsure how to solve this conflict. IMHO, an exception should be made for the "update snapshot" operation: This should not be considered a build-time action, but rather a convenience to the developer, who basically asks the code to write to the source tree. I understand that there is [some resistance to such ideas](https://github.com/bazelbuild/bazel/issues/3376).

I see advantages to using `EXPECT_EQ_SNAPSHOT(val)`, when compared to plain `EXPECT_EQ(val1, val2)`, when one or more of the following is true:
* When the tested value is large.
* When the tested value changes (albeit slowly) throughout the project's lifetime.
* When it is tedious to manually write a test from scratch, but it is easy to manually validate the tested value.

Basically `EXPECT_EQ_SNAPSHOT` expresses the intention of the developer that a given value should not change without the knowledge of the developer, however, the exact value is not set in stone.

I find snapshot testing very useful with:

* Images: For example, an SVG to PNG converter could be tested with snapshots. The snapshot might need updating when the font engine is updated or the anti-aliasing algorithm is improved. However, the snapshot should not change if the library is optimized for a new processor. A future version of Google Test might even open an "image diff" in an image viewer, when a test fails, as other frameworks do.

* User interfaces for webapps: Minor UI changes (e.g., title of a dialog) can be check by looking at the diff of the generated HTML. Other, larger changes, need a human-in-the-loop. Snapshot testing helps QA focus on only those UI components that changed and quickly update them if they are look okey.

* Between components: I'm not sure this is a sanctioned usage, but I like to use snapshot testing between components. For example, in my current project, I need some C++ code to output JSON for the next stage, which is more tedious to test. Snapshot testing helps me focus on changing the C++ code and "alerts" me when I also need to double-check the output with the next stage.

googletest/test/googletest-snapshot-test_.cc
```diff
@@ -0,0 +1,79 @@
...
+  return "Hello " + whom;
+}
+
+// Parameters to test
+std::vector<std::string> GetParamsToTest() {
+  GTEST_DISABLE_MSC_DEPRECATED_PUSH_(/* getenv: deprecated */)
+  const char *c_params = getenv("GREETERTEST_PARAMS");
+  GTEST_DISABLE_MSC_DEPRECATED_POP_()
+
+  std::string params = c_params ? c_params : "";
```
shahzadlone: perhaps a ```const```?

cristiklein: Fixed, thanks!

googletest/include/gtest/gtest-snapshot.h
```diff
@@ -0,0 +1,107 @@
...
+template <typename T>
+std::string SnapshotHelper(const T& val) {
+  const std::string snapshot_file = GetSnapshotFile();
+  
+  if (GTEST_FLAG(update_snapshot)) {
+    GTEST_LOG_(INFO) << "Updating snapshot " << snapshot_file;
+    std::ofstream snapshot(snapshot_file);
+
+    if (!snapshot.is_open())
+      return "Could not open snapshot for update: " + snapshot_file;
```
shahzadlone: would appreciate a braced ```if``` please.

googletest/include/gtest/gtest-snapshot.h
```diff
@@ -0,0 +1,107 @@
...
+
+    snapshot << val;
+
+    snapshot.close();
+  }
+
+  std::ifstream snapshot(snapshot_file);
+  if (!snapshot.is_open())
+    return "Could not open snapshot: " + snapshot_file;
```
shahzadlone: similar as above please.

gennadiycivil:
Upon additional review it is apparent that we are not going to accept this change into googletest proper. Thank you for good discussion .

PR #2026, opened 2018-12-21
Example/sample makefile improvements

justin6case:
Improvements have been made to the example/sample makefiles for both googlemock
and googletest.

Library files are now created and named like versions produced
by Cmake.

### DIFF
--- a/googlemock/make/Makefile
+++ b/googlemock/make/Makefile
@@ -19,6 +19,9 @@
 # a copy of Google Test at a different location.
 GTEST_DIR = ../../googletest
 
+# Points to the location of the Google Test libraries
+GTEST_LIB_DIR = .
+
 # Points to the root of Google Mock, relative to where this file is.
 # Remember to tweak this if you move this file.
 GMOCK_DIR = ..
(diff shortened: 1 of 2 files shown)


gennadiycivil:
Ping, no CLA

gennadiycivil:
@justin6case Unfortunately I dont see a signed CLA from you. I tried to verify manually with your GitHub username "justin6case". Perhaps you signed using different username?

justin6case:
I previously signed it with my email. I've changed it to my username now.

gennadiycivil:
Thank you for the CLA. Now we can actually talk about the merits of this change:
Could you please elaborate and provide more information on exactly the issue / problems this is trying to address.

justin6case:
This change only affects the sample or example Makefiles for googletest and googlemock.

Previously, the Makefiles seemed quite hastily put together. They produced libraries like gmock.a and gtest.a that were inconsistent with the names produced by cmake, libgmock.a and libgtest.a.

The gmock Makefile CXXFLAGS was also lacking -std=c++11, preventing older compilers from compiling these examples. 

Overall, these changes should make building the samples easier, and give users using these Makefiles as a baseline for their own a better place to start.

gennadiycivil:
are there any documentation changes that need to happen?
Thanks

PR #2038, opened 2018-12-29
Add noexcept variants to mock methods

### LINKED ISSUE #1821: Support "noexcept" function specifiers
I've been waiting for this feature for quite a while now, but the original issue #667 has been closed recently due to inactivity. Given the fact that googletest will drop support for c++98 soon, I'd expect that a language feature like `noexcept` should be supported.
#667 was referencing PR #681, which was then dropped in favour of PR #794 and PR #1168. PR #1168 has been dropped in favour of solutions discussed in issue #948, which lead to PR #1532. Unfortunately PR #1532 is close to be merged, but does not include support for noexcept anymore.

ciband:
There was a gap in the member function mocks that didn't allow for a mocked interface to have noexcept member functions.  This PR adds those methods.

Fixes #1821 

I tested this to the best of my knowledge but I am not an expert on this framework.  I welcome any feedback and help.

Thanks.


gennadiycivil:
@ciband Thank you very much for this proposal. This implementation will not do, we actively discourage new macros.  Please stay tuned there is an internal effort brewing that should among other things cover #1821.
Thank you again.

PR #2041, opened 2019-01-03
 fix: Remove Arduino entry points

ciband:
Improved flexibility by removing the Arduino entry points in favor of manual calls to setup/loop that the user can call from their entry point.  This is the more common use case for Arudino.

Also added the gtest/gmock_main files to the PlatformIO ignore list since we are not supporting that feature.

### DIFF
--- a/googlemock/src/gmock_main.cc
+++ b/googlemock/src/gmock_main.cc
@@ -32,22 +32,6 @@
 #include "gmock/gmock.h"
 #include "gtest/gtest.h"
 
-#ifdef ARDUINO
-void setup() {
-  // Since Arduino doesn't have a command line, fake out the argc/argv arguments
-  int argc = 1;
-  const auto arg0 = "PlatformIO";
-  char* argv0 = const_cast<char*>(arg0);
-  char** argv = &argv0;
-
-  // Since Google Mock depends on Google Test, InitGoogleMock() is
-  // also responsible for initializing Google Test.  Therefore there's
-  // no need for calling testing::InitGoogleTest() separately.
-  testing::InitGoogleMock(&argc, argv);
-}
-void loop() { RUN_ALL_TESTS(); }
-#else
-
 // MS C++ compiler/linker has a bug on Windows (not on Windows CE), which
 // causes a link error when _tmain is defined in a static library and UNICODE
 // is enabled. For this reason instead of _tmain, main function is used on
@@ -68,4 +52,3 @@ GTEST_API_ int main(int argc, char** argv) {
   testing::InitGoogleMock(&argc, argv);
   return RUN_ALL_TESTS();
 }
-#endif
--- a/googletest/include/gtest/gtest.h
+++ b/googletest/include/gtest/gtest.h
@@ -57,16 +57,16 @@
 #include <ostream>
 #include <vector>
 
-#include "gtest/internal/gtest-internal.h"
-#include "gtest/internal/gtest-string.h"
 #include "gtest/gtest-death-test.h"
 #include "gtest/gtest-matchers.h"
 #include "gtest/gtest-message.h"
 #include "gtest/gtest-param-test.h"
 #include "gtest/gtest-printers.h"
-#include "gtest/gtest_prod.h"
 #include "gtest/gtest-test-part.h"
 #include "gtest/gtest-typed-test.h"
+#include "gtest/gtest_prod.h"
+#include "gtest/internal/gtest-internal.h"
+#include "gtest/internal/gtest-string.h"
 
 GTEST_DISABLE_MSC_WARNINGS_PUSH_(4251 \
 /* class A needs to have dll-interface to be used by clients of class B */)
@@ -91,11 +91,10 @@ namespace testing {
 // Silence C4100 (unreferenced formal parameter) and 4805
 // unsafe mix of type 'const int' and type 'const bool'
 #ifdef _MSC_VER
-# pragma warning(push)
-# pragma warning(disable:4805)
-# pragma warning(disable:4100)
+#pragma warning(push)
+#pragma warning(disable : 4805)
+#pragma warning(disable : 4100)
 #endif
-
 
 // Declares the flags.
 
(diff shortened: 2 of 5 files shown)


library.json
```diff
@@ -31,6 +31,7 @@
             "googlemock/CMakeLists.txt",
             "googlemock/Makefile.am",
             "googlemock/configure.ac",
+            "googlemock/src/gmock_main.cc",
```
gennadiycivil: looks weird to read, could you please sort alphabetically

ciband:
There are changes in gtest.h @ line 2334.  I do not know what happened with the formatting.  I will correct.

I will alphabetize the file ignore list.

library.json
ciband: done

ciband:
Still working on getting the gtest.h resolved.  It is behaving like there is some auto-formatting happen during the check in.  The intent is just to add the lines at the bottom.  The other items should be done.

ciband:
@gennadiycivil 
For the Arduino builds, they require setup and loop functions.  In our project, we just made dummy functions for CI.  Would that be acceptable here?

The other option would be to put the setup/loop functions that we took out in gtest and gmock, back in, but only for CI.

A future feature could enable these for the PlatformIO framework as an opt-in feature.

Thoughts?

gennadiycivil:
CI failed

ciband:
@gennadiycivil I saw that CI failed.  I wasn't sure how best to proceed so I asked the question above.

Thoughts on this?  

My vote would probably be to put the original entry points back but leave them disabled in the PlatformIO framework by not including the *_main.cc file.

gennadiycivil:
Unfortunately I don't have an opinion because I am not familiar with the platform and its common use cases.

My take on the changes is to 
1) Keep them isolated to Arudino guarded inside #ifdefs, etc
2) keep the CI green so there are no obviouls breaks for the rest of the workd
3) Some general sanity check

ciband:
@gennadiycivil Do you need anything else on this issue.

I decided to put the functions in but only enable them for CI.  If we have to have the functions, we might as well make them do the right thing and give the option for later enabling in the PlatformIO package.

googletest/include/gtest/gtest.h
```diff
@@ -2499,6 +2499,20 @@ inline int RUN_ALL_TESTS() {
   return ::testing::UnitTest::GetInstance()->Run();
 }
 
+#ifdef ARDUINO
+inline void gtest_setup() {
+  // Since Arduino doesn't have a command line, fake out the argc/argv arguments
+  int argc = 1;
```
gennadiycivil: Let us Inot define these functions in the header. Too much noise.

Also, there is nothing interesting about this function.
If the user has to insert code to initialize GoogleTest, they might as well call testing::InitGoogleTest themselves.

At this point I think it might be simpler to provide an overload of InitGoogleTest that doesn't require argv/argc.

ciband: I like that idea better.  I pushed no argument versions of the init functions and cleaned up where appropriate.

googlemock/src/gmock.cc
```diff
@@ -58,8 +57,7 @@ namespace internal {
 // "=value" part can be omitted.
 //
 // Returns the value of the flag, or NULL if the parsing failed.
-static const char* ParseGoogleMockFlagValue(const char* str,
-                                            const char* flag,
+static const char* ParseGoogleMockFlagValue(const char* str, const char* flag,
```
gennadiycivil: Formatting seems to change without any actual changes. Could you please make sure that only actual changes are included

googlemock/src/gmock.cc
```diff
@@ -132,8 +130,8 @@ static bool ParseGoogleMockIntFlag(const char* str, const char* flag,
   if (value_str == nullptr) return false;
 
   // Sets *value to the value of the flag.
-  return ParseInt32(Message() << "The value of flag --" << flag,
-                    value_str, value);
+  return ParseInt32(Message() << "The value of flag --" << flag, value_str,
```
gennadiycivil: Same here, formatting

googlemock/src/gmock.cc
ciband: Yes I will fix this.  I think clang-format is ran automatically on the file edit.

googlemock/src/gmock.cc
ciband: Will fix.

googletest/src/gtest_main.cc
```diff
@@ -50,4 +44,6 @@ GTEST_API_ int main(int argc, char **argv) {
   testing::InitGoogleTest(&argc, argv);
   return RUN_ALL_TESTS();
 }
+
```
gennadiycivil: still extra new line

googletest/src/gtest_main.cc
```diff
@@ -50,4 +44,6 @@ GTEST_API_ int main(int argc, char **argv) {
...
 #endif
+
```
gennadiycivil: ... and here

googlemock/src/gmock_main.cc
```diff
@@ -68,4 +63,5 @@ GTEST_API_ int main(int argc, char** argv) {
   testing::InitGoogleMock(&argc, argv);
   return RUN_ALL_TESTS();
 }
+
```
gennadiycivil: ... and here

googlemock/src/gmock_main.cc
```diff
@@ -34,17 +34,12 @@
...
-  char* argv0 = const_cast<char*>(arg0);
-  char** argv = &argv0;
-
   // Since Google Mock depends on Google Test, InitGoogleMock() is
   // also responsible for initializing Google Test.  Therefore there's
   // no need for calling testing::InitGoogleTest() separately.
-  testing::InitGoogleMock(&argc, argv);
+  testing::InitGoogleMock();
 }
+
```
gennadiycivil: still extra new line

googletest/src/gtest_main.cc
ciband: Fixed.

googletest/src/gtest_main.cc
ciband: Fixed.

googlemock/src/gmock_main.cc
ciband: Fixed.

googlemock/src/gmock_main.cc
ciband: Fixed.

ciband:
@gennadiycivil  Sorry for the formatting issues.  Code should be good now.

Thanks.

PR #2044, opened 2019-01-04
Replace *_TEST_SUITE_* with *_TEST_CASE_* in Advanced.md doc

samolisov:
In fact, the googletest library uses the *_TEST_CASE_* macros for typed and value-parameterized tests, not the *_TEST_SUITE_* ones.

### DIFF
--- a/googletest/docs/advanced.md
+++ b/googletest/docs/advanced.md
@@ -1380,7 +1380,7 @@ TEST_P(FooTest, HasBlahBlah) {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


gennadiycivil:
As per https://github.com/google/googletest/blob/master/googletest/docs/primer.md#beware-of-the-nomenclature , "googletest recently started replacing the term Test Case by Test Suite The preferred API is TestSuite*. The older TestCase* API is being slowly deprecated and refactored away"

You should be able to use TEST_SUITE* APIs , if for some reason they dont work, please let us know

PR #2047, opened 2019-01-05
Only build with warnings-as-errors on travis

robinlinden:
This is so that -Werror doesn't break compilation for people on different platforms/newer compilers.


gennadiycivil:
This is a pretty fundamental change and would require community discussion before this can be accepted.

PR #2051, opened 2019-01-06
fixing build errors for unused parameter

enptfb55:


gennadiycivil:

enptfb55:
This change fixes build errors I encountered when building with the -Wextra compiler flag. All this change does is comment the parameters to avoid -Wunused-parameter errors.

PR #2052, opened 2019-01-06
make g_argvs a function static variable

### LINKED ISSUE #1075: g_argvs is a static object causing exit time failures
I build gtest-1.8.0 on OpenBSD 6.1-amd64. When I run protobuf-3.2.0 test suite relying on gtest, I get exit time failures complaining about double-freeing memory.  The object implicated is g_argvs. Quick code inspection reveals a typical pattern outlawed in [google style guide](https://google.github.io/styleguide/cppguide.html#Static_and_Global_Variables):

Objects with static storage duration, including global variables, static variables, static class member variables, and function static variables, must be Plain Old Data (POD): only ints, chars, floats, or pointers, or arrays/structs of POD.

The following patch makes this symptom disappear (though the next bug is still causing trouble and will require more work). Admittedly, this is c++11-only code and may require changes to work for the full set of supported compilers.

```patch
$OpenBSD$
--- googletest/src/gtest.cc.orig
+++ googletest/src/gtest.cc
@@ -384,13 +384,17 @@ void AssertHelper::operator=(const Message& message) c
 GTEST_API_ GTEST_DEFINE_STATIC_MUTEX_(g_linked_ptr_mutex);
 
 // A copy of all command line arguments.  Set by InitGoogleTest().
-::std::vector<testing::internal::string> g_argvs;
+::std::vector<testing::internal::string>& g_argvs() {
+  static ::std::vector<testing::internal::string>* const gargvs =
+      new ::std::vector<testing::internal::string>();
+  return *gargvs;
+}
 
 const ::std::vector<testing::internal::string>& GetArgvs() {
 #if defined(GTEST_CUSTOM_GET_ARGVS_)
   return GTEST_CUSTOM_GET_ARGVS_();
 #else  // defined(GTEST_CUSTOM_GET_ARGVS_)
-  return g_argvs;
+  return g_argvs();
 #endif  // defined(GTEST_CUSTOM_GET_ARGVS_)
 }
 
@@ -5347,9 +5351,9 @@ void InitGoogleTestImpl(int* argc, CharType** argv) {
 
   if (*argc <= 0) return;
 
-  g_argvs.clear();
+  g_argvs().clear();
   for (int i = 0; i != *argc; i++) {
-    g_argvs.push_back(StreamableToString(argv[i]));
+    g_argvs().push_back(StreamableToString(argv[i]));
   }
 
   ParseGoogleTestFlagsOnly(argc, argv);

```

blackgnezdo:
Fixes #1075

### DIFF
--- a/googletest/src/gtest.cc
+++ b/googletest/src/gtest.cc
@@ -411,7 +411,10 @@ void AssertHelper::operator=(const Message& message) const {
 }
 
 // A copy of all command line arguments.  Set by InitGoogleTest().
-static ::std::vector<std::string> g_argvs;
+static ::std::vector<std::string>& g_argvs() {
+  static auto& g_argvs = *new ::std::vector<std::string>;
+  return g_argvs;
+}
 
 ::std::vector<std::string> GetArgvs() {
 #if defined(GTEST_CUSTOM_GET_ARGVS_)
@@ -420,7 +423,7 @@ static ::std::vector<std::string> g_argvs;
   const auto& custom = GTEST_CUSTOM_GET_ARGVS_();
   return ::std::vector<std::string>(custom.begin(), custom.end());
 #else   // defined(GTEST_CUSTOM_GET_ARGVS_)
-  return g_argvs;
+  return g_argvs();
 #endif  // defined(GTEST_CUSTOM_GET_ARGVS_)
 }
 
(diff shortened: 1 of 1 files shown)


gennadiycivil:
Thank you for this contribution.

There are several issues with this PR
1) The issue states that "gtest-1.8.0" is the problem. Yet the PR is asking to merge into master. Which one is it?
2) Proper unit test would be helpful. Since this is specific for OpenBSD 6.1-amd64 (? is this right? hard to say from the issue) maybe create Travis build for that platform?

Overall it just requires a bit more due diligence - clearly stating what the issue appears to be, presenting a clear simple unit test, and fix.

gennadiycivil:
Closing for now, will come back as discussed

Issue #1637, opened 2018-06-21
compiler flags from CMAKE_CXX_FLAGS doubled by googletest


MirkoFrommberger:
In File googletest/cmake/internal_utils.cmake is a line (number 133 in release 1.8.0)
set(cxx_exception "${CMAKE_CXX_FLAGS} ${cxx_base_flags} ${cxx_exception_flags}")

This lets the CMake variable 'cxx_exception' include all compiler flags already defined in 'CMAKE_CXX_FLAGS'.
But CMake adds the CMAKE_CXX_FLAGS to the compiler call by itself, so if cxx_exception also contains the CMAKE_CXX_FLAGS and is then added to the compiler flags by function 'cxx_library_with_type' (defined in the same file) , they are two times in the compiler call.

This is OK for most use cases, but not if you are using compiler flags they may only be set once.

The same happens in the next line in internal_utils.cmake for 'cxx_no_exception'.

Solution: just don't add the CMAKE_CXX_FLAGS to cxx_exception / cxx_no_exception and let cmake add the CMAKE_CXX_FLAGS.

wkaluza:
Related issue #1611 and PR #1612 . On closer inspection of the log sample in the PR (VS2017 64 bit build), the only duplicates seem to be the UNICODE and _UNICODE preprocessor definitions; other options which cmake adds by default (such as /EHsc) seem to get successfully deduplicated somewhere along the way.

MirkoFrommberger:
I am not building with MSVC but with a the GNU Arm Embedded Toolchain for STM32F. This includes passing a .specs file to gcc, which only works once. The CMake toolchain file adds the --specs=nosys.specs part to the CMAKE_CXX_FLAGS. And the mentioned function doubles that, which does not work.
There is simply no need to add the CMAKE_CXX_FLAGS to the cxx_exception variable, since CMake passes both to the compiler.

gennadiycivil:
Looks stale, corresponding PR has been closed due to no CLA

PR #2057, opened 2019-01-08
Update internal_utils.cmake

MirkoFrommberger:
If CMAKE_CXX_FLAGS are part of cxx_no_exception, the compiler flags are doubled, because CMake adds the CMAKE_CXX_FLAGS to the compiler call by itself, so if cxx_no_exception also contains the CMAKE_CXX_FLAGS and is then added to the compiler flags by function 'cxx_library_with_type' (defined in the same file) , they are two times in the compiler call.

This is OK for most use cases, but not if you are using compiler flags that may only be set once.

See discussion in https://github.com/google/googletest/issues/1637

### DIFF
--- a/googletest/cmake/internal_utils.cmake
+++ b/googletest/cmake/internal_utils.cmake
@@ -121,8 +121,7 @@ macro(config_compiler_and_linker)
 
   # For building gtest's own tests and samples.
   set(cxx_exception "${cxx_base_flags} ${cxx_exception_flags}")
-  set(cxx_no_exception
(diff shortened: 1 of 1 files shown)


gennadiycivil:
NO CLA

MirkoFrommberger:
Sorry, took me a while to re-activate my google account. It's signed now.

gennadiycivil:
Sorry I still dont see a CLA for you. Make sure you use the same email / github username that you used to commit your changes

Issue #1419, opened 2018-01-23
INSTANTIATE_TEST_CASE_P generates warnings with clang++ -Wpedantic


llugin:
Reproduction:

foo.cc
```
#include "gtest/gtest.h"

class Foo : public ::testing::TestWithParam<int> {
public:
  virtual void SetUp() { return; }
  virtual void TearDown() { return; }
};

TEST_P(Foo, FOO_TEST) { ASSERT_TRUE(GetParam()); }

INSTANTIATE_TEST_CASE_P(InstantiationName, Foo, ::testing::Values(1, 2, 3));

int main(int argc, char **argv) {
  ::testing::InitGoogleTest(&argc, argv);
  return RUN_ALL_TESTS();
} 
```
```                                                          
$ clang++ -Wall -Wpedantic -Werror -isystem ${GTEST_DIR}/include -pthread \
foo.cc ${GTEST_DIR}/build/libgtest.a -o foo
```

Output:

> foo.cc:11:75: error: must specify at least one argument for '...' parameter of variadic macro [-Werror,-Wgnu-zero-variadic-macro-arguments]
> INSTANTIATE_TEST_CASE_P(InstantiationName, Foo, ::testing::Values(1, 2, 3));
>                                                                           ^
> /my/dir/googletest/googletest//include/gtest/gtest-param-test.h:1421:10: note: macro 'INSTANTIATE_TEST_CASE_P' defined here
> \# define INSTANTIATE_TEST_CASE_P(prefix, test_case_name, generator, ...) \
>          ^
> 1 error generated.

On the other hand, it compiles for g++ (checked on 7.2.0).

Tested on ubuntu 17.10, for clang versions:  
- 4.0.1-6 (tags/RELEASE_401/final)
- 5.0.0-3 (tags/RELEASE_500/final)

Found on: 3c5e064ca089cbe3e89278e16c2c6cc3a97dba1f

Vlad-ll:
I have the same with GCC 7.1.

improbable-nickkrempel:
Confirmed happening here too. Using the compiler option `-Wno-gnu-zero-variadic-macro-arguments` to suppress the warning as a workaround.

patrick96:
Adding a comma after the last argument also solves this for me:
```cpp
INSTANTIATE_TEST_CASE_P(InstantiationName, Foo, ::testing::Values(1, 2, 3),);
                                                                          ^
```

gennadiycivil:
[...]

PR #2063, opened 2019-01-15
Fix INSTANTIATE_TEST_CASE_P with zero variadic arguments

mathbunnyru:
To fix: https://github.com/google/googletest/issues/1419
INSTANTIATE_TEST_CASE_P generates warnings with clang++ -Wpedantic #1419

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,7 +43,7 @@ matrix:
       env: BUILD_TYPE=Release VERBOSE=1 CXX_FLAGS=-std=c++11
     - os: linux
       compiler: clang
-      env: BUILD_TYPE=Release VERBOSE=1 CXX_FLAGS=-std=c++11 NO_EXCEPTION=ON NO_RTTI=ON COMPILER_IS_GNUCXX=ON
+      env: BUILD_TYPE=Release VERBOSE=1 CXX_FLAGS="-std=c++11 -Wgnu-zero-variadic-macro-arguments" NO_EXCEPTION=ON NO_RTTI=ON COMPILER_IS_GNUCXX=ON
     - os: osx
       compiler: gcc
       env: BUILD_TYPE=Release VERBOSE=1 CXX_FLAGS=-std=c++11 HOMEBREW_LOGS=~/homebrew-logs HOMEBREW_TEMP=~/homebrew-temp
--- a/googletest/include/gtest/gtest-param-test.h
+++ b/googletest/include/gtest/gtest-param-test.h
@@ -556,15 +556,20 @@ internal::CartesianProductHolder10<Generator1, Generator2, Generator3,
 // alphanumeric characters or underscore. Because PrintToString adds quotes
 // to std::string and C strings, it won't work for these types.
 
+#define GTEST_GET_FIRST_(first, ...) first
+#define GTEST_GET_SECOND_(first, second, ...) second
+
 #define INSTANTIATE_TEST_SUITE_P(prefix, test_suite_name, ...)                \
   static ::testing::internal::ParamGenerator<test_suite_name::ParamType>      \
       gtest_##prefix##test_suite_name##_EvalGenerator_() {                    \
-    return VA_GETFIRST(__VA_ARGS__);                                          \
+    return GTEST_GET_FIRST_(__VA_ARGS__, DUMMY_PARAM_);                       \
   }                                                                           \
   static ::std::string gtest_##prefix##test_suite_name##_EvalGenerateName_(   \
       const ::testing::TestParamInfo<test_suite_name::ParamType>& info) {     \
-    return ::testing::internal::CreateParamGenerator<                         \
-        test_suite_name::ParamType>(VA_GETREST(__VA_ARGS__, 0))(info);        \
+    return GTEST_GET_SECOND_(                                                 \
+        __VA_ARGS__,                                                          \
+        ::testing::internal::DefaultParamName<test_suite_name::ParamType>,    \
+        DUMMY_PARAM_)(info);                                                  \
   }                                                                           \
   static int gtest_##prefix##test_suite_name##_dummy_                         \
       GTEST_ATTRIBUTE_UNUSED_ =                                               \
--- a/googletest/include/gtest/gtest-param-test.h
+++ b/googletest/include/gtest/gtest-param-test.h
@@ -557,20 +557,20 @@ internal::CartesianProductHolder10<Generator1, Generator2, Generator3,
 // to std::string and C strings, it won't work for these types.
 
 #define GTEST_EXPAND_(arg) arg
-#define GTEST_GET_FIRST_(first, ...) GTEST_EXPAND_(first)
-#define GTEST_GET_SECOND_(first, second, ...) GTEST_EXPAND_(second)
+#define GTEST_GET_FIRST_(first, ...) first
+#define GTEST_GET_SECOND_(first, second, ...) second
 
 #define INSTANTIATE_TEST_SUITE_P(prefix, test_suite_name, ...)                \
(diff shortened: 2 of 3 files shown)


gennadiycivil:
CI broken,

mathbunnyru:
Yes, sorry for that. I will fix this tomorrow, I hope.

mathbunnyru:
@VSol-ll do you know why windows build is broken? I don't know anything about msvc and I don't even have Windows to check it out.

mathbunnyru:
https://stackoverflow.com/questions/5134523/msvc-doesnt-expand-va-args-correctly

mathbunnyru:
Squashed, rebased on the latest master, cla signed, fixed CIs. Please, take a look.

VSol-ll:
Good finding with this VS preprocessor issue!
I didn't use VS for ages and don't have it. Was going to play with godbolt to see what's going wrong. But you've done things.

Minor stuff to mention: 
* To my opinion, location of these VA-related macros is slightly illogical. 
* Also quick search doesn't show too many places of using of GetParamNameGen(). Which means that there were no need introducing a wrapper for it (I'd probably choose to modify the GetParamNameGen() itself). 
* Finally I'd introduce a fake argument type instead of using int.
However I'd leave it for maintainers as I'm not familiar with the code-base.

Good job!

mathbunnyru:
Thanks a lot. Waiting for maintainers' opinion on this.

mathbunnyru:
@gennadiycivil please, take a look. This is preventing us from upgrading to new Ubuntu LTS.

gennadiycivil:
Let's discuss the purpose of this change. Are we only making these changes to placate -Wpedantic?
There are a lot of changes including changes to internals. I am skeptical ( but can be persuaded ) - please explain and prove that this is needed.

gennadiycivil:
In addition - please add a unit test that demonstrates :"the problem" and proves that the solution works

mathbunnyru:
Thank you for your response.

Yes, this change is only for those, who use Wpedantic.
I personally think that everyone should use this flag in their source code - at least a lot of people teach to use flags like Wall, Werror, Wpedantic and so on in your codebase.
Good thing about this change, is that it doesn't require to change any user code.

Do you have any way to force compiler flags for a given test? Actually, any INSTANTIATE_TEST_CASE_P usage without optional argument would work.
Or would you prefer to use something like `#pragma GCC diagnostic error`?

gennadiycivil:
I think the best way would be

* Add -Wgnu-zero-variadic-macro-arguments into one of the travis clang builds. This is more focused that blanket -Wpedantic

Could you please take a look

mathbunnyru:
Sure, will look at this tomorrow.

googletest/include/gtest/gtest-param-test.h
```diff
@@ -556,15 +556,15 @@ internal::CartesianProductHolder10<Generator1, Generator2, Generator3,
 // alphanumeric characters or underscore. Because PrintToString adds quotes
 // to std::string and C strings, it won't work for these types.
 
-#define INSTANTIATE_TEST_SUITE_P(prefix, test_suite_name, generator, ...)     \
+#define INSTANTIATE_TEST_SUITE_P(prefix, test_suite_name, ...)                \
```
gennadiycivil: This is following your direction but with more suitable names, etc

```suggestion
#define GTEST_GET_FIRST_(first, ...) first
#define GTEST_GET_SECOND_(first, second, ...) second

#define INSTANTIATE_TEST_SUITE_P(prefix, test_suite_name, ...)                \
  static ::testing::internal::ParamGenerator<test_suite_name::ParamType>      \
      gtest_##prefix##test_suite_name##_EvalGenerator_() {                    \
    return GTEST_GET_FIRST_(__VA_ARGS__, DUMMY_PARAM_);                                         \
  }                                                                           \
  static ::std::string gtest_##prefix##test_suite_name##_EvalGenerateName_(   \
      const ::testing::TestParamInfo<test_suite_name::ParamType>& info) {     \
    return GTEST_GET_SECOND_(                                                        \
        __VA_ARGS__,                                                          \
        ::testing::internal::DefaultParamName<test_suite_name::ParamType>,    \
        DUMMY_PARAM_)(info);                                                             \
  }                                                                           \

```

googletest/include/gtest/internal/gtest-param-util.h
```diff
@@ -396,6 +397,30 @@ typename ParamNameGenFunc<ParamType>::Type *GetParamNameGen() {
   return DefaultParamName;
 }
 
+// INTERNAL IMPLEMENTATION - DO NOT USE IN USER CODE.
```
gennadiycivil: Don't need any of those at all, see suggestion in the googletest/include/gtest/gtest-param-test.h

googletest/include/gtest/internal/gtest-param-util.h
```diff
@@ -396,6 +397,30 @@ typename ParamNameGenFunc<ParamType>::Type *GetParamNameGen() {
   return DefaultParamName;
 }
```
gennadiycivil: We can delete these now, all 3 
ParamNameGenFunctor GetParamNameGen...
struct ParamNameGenFunc...
typename ParamNameGenFunc<ParamType>....

googletest/include/gtest/internal/gtest-param-util.h
```diff
@@ -41,6 +41,7 @@
 #include <memory>
 #include <set>
 #include <tuple>
+#include <type_traits>
```
gennadiycivil: no need anymore

.travis.yml
```diff
@@ -43,7 +43,7 @@ matrix:
       env: BUILD_TYPE=Release VERBOSE=1 CXX_FLAGS=-std=c++11
     - os: linux
       compiler: clang
-      env: BUILD_TYPE=Release VERBOSE=1 CXX_FLAGS=-std=c++11 NO_EXCEPTION=ON NO_RTTI=ON COMPILER_IS_GNUCXX=ON
+      env: BUILD_TYPE=Release VERBOSE=1 CXX_FLAGS="-std=c++11 -Wgnu-zero-variadic-macro-arguments" NO_EXCEPTION=ON NO_RTTI=ON COMPILER_IS_GNUCXX=ON
```
mathbunnyru: I hope this is the correct way to add the flag.

googletest/include/gtest/gtest-param-test.h
```diff
@@ -556,15 +556,21 @@ internal::CartesianProductHolder10<Generator1, Generator2, Generator3,
 // alphanumeric characters or underscore. Because PrintToString adds quotes
 // to std::string and C strings, it won't work for these types.
 
-#define INSTANTIATE_TEST_SUITE_P(prefix, test_suite_name, generator, ...)     \
+#define GTEST_EXPAND_(arg) arg
```
mathbunnyru: Needed for msvc build.

gennadiycivil:
I merged the changes in http://go/gh/google/googletest/commit/1f42ae7353b3c52f0fe014da31e5210f266274cc. Due to the way we import/export the code I have to do the travis.yml change separarately

mathbunnyru:
Thanks.

mathbunnyru:
Yes, thank you.

Right now I get

```
foo.cc:11:1: error: 'InstantiateTestCase_P_IsDeprecated' is deprecated: INSTANTIATE_TEST_CASE_P is deprecated, please use INSTANTIATE_TEST_SUITE_P [-Werror,-Wdeprecated-declarations]
INSTANTIATE_TEST_CASE_P(InstantiationName, Foo, ::testing::Values(1, 2, 3));
```

And this is perfectly fine. When I change to INSTANTIATE_TEST_SUITE_P everything works just fine.

Issue #890, opened 2016-09-28
Generalize EXPECT_NEAR/ASSERT_NEAR to work with arbitrary floating-point like types


helloworld922:
The EXPECT_NEAR and ASSERT_NEAR macros expect double parameters, which potentially means loss of precision or for types with greater accuracy than double, or compile errors for types not implicitly convertible to double (gcc/intel quad precision long double, Boost.MultiPrecision, etc.).

I've attached a patch which demonstrates a working example fix which generalizes this to work with arbitrary types as long as the required basic operations are available (can find an abs function, subtraction operator, comparison operator).

[fp_fix.txt](https://github.com/google/googletest/files/498845/fp_fix.txt)

chiphogg:
So, was this closed because the fix was merged?  Closed because it won't be fixed?  @gennadiycivil -- can you please elaborate?

gennadiycivil:
I have to say that I dont know. I have been doing a lot of cleanup, this
was back in August and it is possible that I simply clicked the wrong
button.
Sorry about that
Thanks
G

chiphogg:
It seems there is an open pull request.  Perhaps this issue should be reopened?

gennadiycivil:
Re-Opening. Please read https://github.com/google/googletest/pull/2077#issuecomment-460729350.

chiphogg:
Even though it looks like I shouldn't expect this to close any time soon, I'm still happy the issue is in the correct state.  :slightly_smiling_face: Thanks!

chiphogg:
Perhaps a good way to move this forward would be:

- take one of Google's internal tests which this breaks
- create a test in the public repo which reproduces the failure

What do you think?

EvanHonnold:
I am encountering similar issues with `EXPECT_NEAR`. Looks like there was some recent discussion in https://github.com/google/googletest/pull/2077#issuecomment-783465831 about a possible fix -- might this be a way forward?

chiphogg:
Happy New Year 2025 to all of the various `EXPECT_NEAR` enthusiasts here!  I'm moving the conversation here, because an open issue seems like a better home than a closed PR.

In revisiting this familiar page, I realized I was taking the wrong approach pinging @gennadiycivil and @mattcalabrese-google on #2077: I don't even know if they still work on this project.  Instead, I went to the issues page, and took a peek to see who is actively responding to new issues.  By far, the most common name was @derekmauro.  So, Derek --- are you able to clear up the status here?

[...]

PR #2077, opened 2019-01-24
added general floating point near comparison

helloworld922:
The EXPECT_NEAR and ASSERT_NEAR macros expect double parameters, which potentially means loss of precision or for types with greater accuracy than double, or compile errors for types not implicitly convertible to double (gcc/intel quad precision long double, Boost.MultiPrecision, etc.).

This is essentially the same fix as the patch I submitted in issue #890. I fixed a problem where I called the wrong absolute value funciton (abs vs. fabs).

Side note: I don't know why the original bug report was closed without comment.

### DIFF
--- a/googletest/include/gtest/gtest.h
+++ b/googletest/include/gtest/gtest.h
@@ -1784,11 +1784,9 @@ AssertionResult CmpHelperFloatingPointEQ(const char* lhs_expression,
 //
 // INTERNAL IMPLEMENTATION - DO NOT USE IN A USER PROGRAM.
 template <class R1, class R2, class T>
-GTEST_API_ AssertionResult FPNearPredFormat(const char* expr1,
-                                            const char* expr2,
-                                            const char* abs_error_expr,
-                                            const R1& val1, const R2& val2,
-                                            const T& abs_error) {
+AssertionResult FPNearPredFormat(const char* expr1, const char* expr2,
+                                 const char* abs_error_expr, const R1& val1,
+                                 const R2& val2, const T& abs_error) {
   using namespace std;
   auto diff = fabs(val1 - val2);
   if (diff <= abs_error) {
--- a/googletest/include/gtest/gtest.h
+++ b/googletest/include/gtest/gtest.h
@@ -52,7 +52,6 @@
 #ifndef GTEST_INCLUDE_GTEST_GTEST_H_
 #define GTEST_INCLUDE_GTEST_GTEST_H_
 
-#include <cmath>
 #include <iomanip>
 #include <limits>
 #include <memory>
@@ -1787,8 +1786,7 @@ template <class R1, class R2, class T>
 AssertionResult FPNearPredFormat(const char* expr1, const char* expr2,
                                  const char* abs_error_expr, const R1& val1,
                                  const R2& val2, const T& abs_error) {
-  using std::fabs;
-  auto diff = fabs(val1 - val2);
+  auto diff = val1 > val2 ? val1 - val2 : val2 - val1;
   if (diff <= abs_error) {
     return AssertionSuccess();
   }
--- a/googletest/include/gtest/gtest.h
+++ b/googletest/include/gtest/gtest.h
@@ -52,7 +52,6 @@
 #ifndef GTEST_INCLUDE_GTEST_GTEST_H_
 #define GTEST_INCLUDE_GTEST_GTEST_H_
 
-#include <iomanip>
 #include <limits>
 #include <memory>
 #include <ostream>
--- a/googletest/include/gtest/gtest.h
+++ b/googletest/include/gtest/gtest.h
@@ -56,6 +56,8 @@
 #include <memory>
 #include <ostream>
 #include <vector>
+#include <iomanip>
+#include <cmath>
 
 #include "gtest/internal/gtest-internal.h"
 #include "gtest/internal/gtest-string.h"
@@ -1781,12 +1783,26 @@ AssertionResult CmpHelperFloatingPointEQ(const char* lhs_expression,
 // Helper function for implementing ASSERT_NEAR.
 //
 // INTERNAL IMPLEMENTATION - DO NOT USE IN A USER PROGRAM.
-GTEST_API_ AssertionResult DoubleNearPredFormat(const char* expr1,
-                                                const char* expr2,
-                                                const char* abs_error_expr,
-                                                double val1,
-                                                double val2,
-                                                double abs_error);
+template <class R1, class R2, class T>
+GTEST_API_ AssertionResult FPNearPredFormat(const char* expr1,
+                                            const char* expr2,
+                                            const char* abs_error_expr,
+                                            const R1& val1, const R2& val2,
+                                            const T& abs_error) {
+  using namespace std;
+  auto diff = fabs(val1 - val2);
+  if (diff <= abs_error) {
+    return AssertionSuccess();
+  }
+  return AssertionFailure()
+         << std::setprecision(
+                std::numeric_limits<typename std::decay<R1>::type>::max_digits10)
+         << "The difference between " << expr1 << " and " << expr2 << " is "
+         << diff << ", which exceeds " << abs_error_expr << ", where\n"
+         << expr1 << " evaluates to " << val1 << ",\n"
+         << expr2 << " evaluates to " << val2 << ", and\n"
+         << abs_error_expr << " evaluates to " << abs_error << ".";
+}
 
 // INTERNAL IMPLEMENTATION - DO NOT USE IN USER CODE.
 // A class that enables one to stream messages to assertion macros
@@ -2167,11 +2183,11 @@ class TestWithParam : public Test, public WithParamInterface<T> {
                       val1, val2)
 
 #define EXPECT_NEAR(val1, val2, abs_error)\
-  EXPECT_PRED_FORMAT3(::testing::internal::DoubleNearPredFormat, \
+  EXPECT_PRED_FORMAT3(::testing::internal::FPNearPredFormat, \
                       val1, val2, abs_error)
 
 #define ASSERT_NEAR(val1, val2, abs_error)\
-  ASSERT_PRED_FORMAT3(::testing::internal::DoubleNearPredFormat, \
+  ASSERT_PRED_FORMAT3(::testing::internal::FPNearPredFormat, \
                       val1, val2, abs_error)
 
 // These predicate format functions work on floating-point values, and
--- a/googletest/include/gtest/gtest.h
+++ b/googletest/include/gtest/gtest.h
@@ -52,12 +52,12 @@
 #ifndef GTEST_INCLUDE_GTEST_GTEST_H_
 #define GTEST_INCLUDE_GTEST_GTEST_H_
 
+#include <cmath>
+#include <iomanip>
 #include <limits>
 #include <memory>
 #include <ostream>
 #include <vector>
-#include <iomanip>
-#include <cmath>
 
 #include "gtest/internal/gtest-internal.h"
 #include "gtest/internal/gtest-string.h"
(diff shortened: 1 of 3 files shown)


googletest/include/gtest/gtest.h
```diff
@@ -1781,12 +1783,24 @@ AssertionResult CmpHelperFloatingPointEQ(const char* lhs_expression,
...
-                                                const char* expr2,
-                                                const char* abs_error_expr,
-                                                double val1,
-                                                double val2,
-                                                double abs_error);
+template <class R1, class R2, class T>
+AssertionResult FPNearPredFormat(const char* expr1, const char* expr2,
+                                 const char* abs_error_expr, const R1& val1,
+                                 const R2& val2, const T& abs_error) {
+  using namespace std;
```
gennadiycivil: Please follow the style guide. See https://google.github.io/styleguide/cppguide.html#Namespaces

googletest/include/gtest/gtest.h
```diff
@@ -56,6 +56,8 @@
 #include <memory>
 #include <ostream>
 #include <vector>
+#include <iomanip>
+#include <cmath>
```
gennadiycivil: Order of includes, see style guide

gennadiycivil:
Please also add unit test with clearly reproducible behaviour

googletest/include/gtest/gtest.h
```diff
@@ -1787,14 +1787,15 @@ template <class R1, class R2, class T>
 AssertionResult FPNearPredFormat(const char* expr1, const char* expr2,
                                  const char* abs_error_expr, const R1& val1,
                                  const R2& val2, const T& abs_error) {
-  using namespace std;
+  using std::fabs;
   auto diff = fabs(val1 - val2);
```
gennadiycivil: lets just spell it put, there is no sense for the using statement 1 line above and there is only 1 usage

googletest/test/gtest_unittest.cc
```diff
@@ -3103,6 +3103,101 @@ TEST_F(DoubleTest, DoubleLEFails) {
...
+  {
+    EXPECT_FATAL_FAILURE(long double a = -1.0; long double b = -1.5;
+                         ASSERT_NEAR(a, b, 0.25),  // NOLINT
+                         "The difference between a and b is 0.5, "
+                         "which exceeds 0.25, where\na evaluates to -1,\nb "
+                         "evaluates to -1.5, and\n0.25 evaluates to 0.25.");
+  }
+}
+
+// a custom floating-point like type which implements the minimal interface
```
gennadiycivil: lets put anonymous namespace around this custom class and associated test

googletest/test/gtest_unittest.cc
```diff
@@ -3103,6 +3103,103 @@ TEST_F(DoubleTest, DoubleLEFails) {
...
+// required to test EXPECT_NEAR on a custom type
+class custom_real {
+  // internally stores as a double to prevent having to implement an actual
+  // custom real-like type
+  double value_;
+
+ public:
+  custom_real(double value) : value_(value) {}
+  custom_real(const custom_real& o) : value_(o.value_) {}
```
gennadiycivil: No need for copy ctor

googletest/test/gtest_unittest.cc
```diff
@@ -3103,6 +3103,103 @@ TEST_F(DoubleTest, DoubleLEFails) {
...
+    EXPECT_FATAL_FAILURE(long double a = -1.0; long double b = -1.5;
+                         ASSERT_NEAR(a, b, 0.25),  // NOLINT
+                         "The difference between a and b is 0.5, "
+                         "which exceeds 0.25, where\na evaluates to -1,\nb "
+                         "evaluates to -1.5, and\n0.25 evaluates to 0.25.");
+  }
+}
+
+namespace {
+// a custom floating-point like type which implements the minimal interface
```
gennadiycivil: After internal discussion, we prefer the same code to look a bit different, please see below:

```suggestion
// a custom floating-point like type which implements the minimal interface
namespace custom_real_ns {

// a custom floating-point like type which implements the minimal interface
// required to test EXPECT_NEAR on a custom type
class custom_real {
  // internally stores as a double to prevent having to implement an actual
  // custom real-like type
  double value_;

 public:
  explicit custom_real(double value) : value_(value) {}

  friend custom_real operator-(custom_real l, custom_real r) {
    return custom_real(l.value_ - r.value_);
  }

  friend bool operator<(custom_real l, custom_real r) {
    return l.value_ < r.value_;
  }
  friend bool operator>(custom_real l, custom_real r) {
    return l.value_ > r.value_;
  }
  friend bool operator<=(custom_real l, custom_real r) {
    return l.value_ <= r.value_;
  }
  friend bool operator>=(custom_real l, custom_real r) {
    return l.value_ >= r.value_;
  }
  friend bool operator==(custom_real l, custom_real r) {
    return l.value_ == r.value_;
  }
  friend bool operator!=(custom_real l, custom_real r) {
    return l.value_ != r.value_;
  }
  friend std::ostream& operator<<(std::ostream& out, custom_real v) {
    out << v.value_;
    return out;
  }
};

}  // namespace custom_real_ns

```

googletest/test/gtest_unittest.cc
```diff
@@ -3103,6 +3103,103 @@ TEST_F(DoubleTest, DoubleLEFails) {
...
+};
+
+custom_real fabs(const custom_real& v) { return std::fabs(v.value_); }
+
+std::ostream& operator<<(std::ostream& out, const custom_real& v) {
+  out << v.value_;
+  return out;
+}
+
+TEST(CustomRealTest, EXPECT_NEAR) {
```
gennadiycivil: After internal discussion we prefer the same code to look a bit different, please see below
//After end of //namespace custom_real_ns
```suggestion
TEST(CustomRealTest, EXPECT_NEAR) {
  {
    custom_real_ns::custom_real a(-1.0);
    custom_real_ns::custom_real b(-1.1);
    EXPECT_NEAR(a, b, custom_real_ns::custom_real(0.2));
  }
  {
    custom_real_ns::custom_real a(-1.0);
    custom_real_ns::custom_real b(-1.5);
    EXPECT_NONFATAL_FAILURE(
        EXPECT_NEAR(a, b, custom_real_ns::custom_real(0.25)),  // NOLINT
        "The difference between a and b is 0.5, "
        "which exceeds custom_real_ns::custom_real(0.25), where\na evaluates "
        "to -1,\nb "
        "evaluates to -1.5, and\ncustom_real_ns::custom_real(0.25) evaluates "
        "to 0.25.");
  }
}

TEST(CustomRealTest, ASSERT_NEAR) {
  {
    custom_real_ns::custom_real a(-1.0);
    custom_real_ns::custom_real b(-1.1);
    ASSERT_NEAR(a, b, custom_real_ns::custom_real(0.2));
  }
  {
    EXPECT_FATAL_FAILURE(
        custom_real_ns::custom_real a(-1.0);
        custom_real_ns::custom_real b(-1.5);
        ASSERT_NEAR(a, b, custom_real_ns::custom_real(0.25)),  // NOLINT
        "The difference between a and b is 0.5, "
        "which exceeds custom_real_ns::custom_real(0.25), where\na evaluates "
        "to -1,\nb "
        "evaluates to -1.5, and\ncustom_real_ns::custom_real(0.25) evaluates "
        "to 0.25.");
  }
}
```

googletest/include/gtest/gtest.h
```diff
@@ -52,6 +52,8 @@
 #ifndef GTEST_INCLUDE_GTEST_GTEST_H_
 #define GTEST_INCLUDE_GTEST_GTEST_H_
 
+#include <cmath>
+#include <iomanip>
```
gennadiycivil: Not sure we need iomanip either

gennadiycivil:
Just to give you an update, during internal testing this change breaks a number of existing tests. it nay be a while before I can get to it

daravi:
Can we rename the functions and the commit to be more general? Maybe call it "ordered type near comparison". I would like to be able to compare types that define a comparison operator. For example:
```
using namespace std::chrono;
EXPECT_NEAR(my_obj->getTimeSinceLastUpdate(), system_clock::now(), seconds(1));
```

gennadiycivil:
Closing , looks abandoned

helloworld922:
Your last message to me was to wait on you to fix internal testing, specifically "it may be a while before I can get to it"

Has this issue with internal testing been resolved?

gennadiycivil:
Apologies, I did not make it clear. The internal testing with existing millions of tests blocks this change in this form.

helloworld922:
What needs to be modified with this commit so this type of change could be pulled?

chiphogg:
This is a great question.  How can this change be moved forward?  It would fix a pretty frustrating oversight in the design of `EXPECT_NEAR`.

skorokhod:
Any news on this?
I'm looking for some way to compare chrono::time_points with some precision bit it looks like gtest has nothing to deal with it.

chiphogg:
Let me suggest a way forward based on my admittedly imperfect information.  :slightly_smiling_face:

It seems very likely to me that if `EXPECT_NEAR` were being written today, and the author had these two implementations to choose from, they would choose the form introduced in this PR over the form now on `master`, every time.  That's my explicit assumption; please let me know if I'm mistaken!  But it seems clear to me that "compare types directly and natively" is strictly better than "implicitly convert everything to a `double`"... especially considering the generic name `EXPECT_NEAR`.

If I'm right, the blocking tests are the equivalent of "behaviour which depends on a bug".  If so, then to fix this situation, we need to remove the dependency on the bug.  One way to do this would be for Google to add another macro which preserves the current behaviour---say, `EXPECT_NEAR_DOUBLE`---internally in their monorepo.  You could then use [rosie](https://opensource.google/docs/glossary/#rosie) to migrate the offending tests to this alternative.  That should unblock this PR.

Other repositories "in the wild" may also depend on this bug, and this patch would break them.  I think this is OK because it would only break them when they consciously choose to upgrade their googletest version.  It's very common to need to deal with slight behaviour changes when upgrading to a new version of a dependency.

@gennadiycivil (as previous participant in this thread) and @mattcalabrese-google (as assignee of #890): WDYT about this analysis, and the approach I suggest?  Have I missed anything important?  Let's see if we can figure out a way forward.

daravi:
Failing that I suggest introducing a new macro such as EXPECT_APPROX or some other name with the more logical behaviour.

chiphogg:
I like that solution, especially if `EXPECT_NEAR` is also deprecated in future googletest versions.  The advantage is that it would also avoid breaking any existing users.  The downside is that `EXPECT_APPROX` is a slightly worse name, and I'd rather pair the best behaviour with the best name... but, very often, the perfect is the enemy of the good.  🙂 

@gennadiycivil, @mattcalabrese-google: we now have two viable solutions that can unblock this work.

1. Migrate Google's dubious usages of `EXPECT_NEAR` to a new, well-named macro that preserves the present behaviour.
2. Provide a new macro---probably `EXPECT_APPROX`---which does what users expect; probably deprecate `EXPECT_NEAR` also.

Are you folks able to weigh in on which of these approaches you would prefer?

chiphogg:
Hi!  Friendly ping. :slightly_smiling_face: I appreciate that folks are surely busy, but I thought I'd check in to see if we can move the conversation forward.

- Does Google consider the present, behaviour of `EXPECT_NEAR` (i.e., restricting to arguments that can be converted implicitly to `double`) to be desirable?
- If not: does Google plan to fix it?  And if so: which above approach would Google prefer, if either?

Thanks!

chiphogg:
Hi!  Happy 2024.  Can @gennadiycivil, @mattcalabrese-google, or anyone else at Google please respond to the discussion starting in [this comment](https://github.com/google/googletest/pull/2077#issuecomment-772788023)?

chiphogg:
For folks who have been following the conversation here, where it has mainly been taking place: I decided to put [my latest ping](https://github.com/google/googletest/issues/890#issuecomment-2568520838) on the issue page #890, because an open issue is a better place for conversation than a (temporarily? :pray:) closed PR.  :slightly_smiling_face:

Issue #2071, opened 2019-01-21
Added special catch for std::exception in GTEST_TEST_NO_THROW reverted?


Zitrax:
We noticed when using EXPECT_NO_THROW that we just see "Actual: it throws." when it throws, we would like to see that was actually thrown. (We are currently on 1.8.0).

Looking at the git log it seems that there has been a fix for this on master: 0354ccb049ea2b9e4696adb8b5a013287daf0c33

But in a later commit that is reverted again, however that commit message is "support printing std::reference_wrapper<T> in gUnit" which does not mention any changes to EXPECT_NO_THROW: 1b20bd176fb3bc6feef4ab17488a2ac8850f42dd

Based on the commit message the revert seem to be a mistake? If not, why was it reverted?

AmatanHead:
This issue is open, but it appears to be fixed in master.

PR #2084, opened 2019-01-26
Add More information to EXPECT_NO_THROW/ASSERT_NO_THROW output

Zitrax:
Cherry picked back commit 0354ccb that went missing in 1b20bd1 due to merge
issues as reported in #2071.

Changed from original commit:

* Use const char* in the constructor instead of const std::string& to avoid
  compile issue (error: conversion from ‘const char [2]’ to non-scalar type
  ‘testing::internal::AdditionalMessage’ requested). This follows the style of
  the existing ConstCharPtr.

* Applied clang-format on the changes to avoid exceeding the line length limit.

(cherry picked from commit 0354ccb049ea2b9e4696adb8b5a013287daf0c33)


Zitrax:
Since I just made minor modifications I let the original author stand from the cherry-pick.

@Jonny007-MKD does this seem fine with you?

Zitrax:
@Jonny007-MKD Thanks.

I see you seem to have reviewed your original commit 0354ccb though, see the "files changed" tab here for my commit.

Jonny007-MKD:
@Zitrax: I have compared both commits in 2 browser tabs and then commented my own one on the line lengths 🤦‍♂️

Zitrax:
I see 👍

gennadiycivil:

Zitrax:
The problem is that when using EXPECT_NO_THROW or ASSERT_NO_THROW and the code throws the output from googletest is:

Expected: <statement> doesn't throw an exception.
Actual: it throws.

which is not very useful. The patch changes this to:

Expected: <statement> doesn't throw an exception.
Actual: it throws <message>

where message is taken from std::exception::what() which is very helpful to get some information on what was thrown.

Prior to this fix one often has to launch the debugger, create a coredump or add debug print messages to find out what was thrown while with it the message might be enough to pinpoint what happened without that debug session.

googletest should give us this information when it is available.

gennadiycivil:
I see. So I changed the PR title to actually say what we are doing

googletest/include/gtest/internal/gtest-internal.h
```diff
@@ -1338,17 +1353,21 @@ class FlatTuple
 
 #define GTEST_TEST_NO_THROW_(statement, fail) \
   GTEST_AMBIGUOUS_ELSE_BLOCKER_ \
-  if (::testing::internal::AlwaysTrue()) { \
+  if (::testing::internal::AdditionalMessage message = ".") { \
```
gennadiycivil: I am missing why this change is beneficial? it appears that the change amounts to the same thing anyway. What is the advantage of adding oprator bool which evaluates to  return ::testing::internal::AlwaysTrue() and not comparing to the ::testing::internal::AlwaysTrue() directly as it was before?

googletest/include/gtest/internal/gtest-internal.h
```diff
@@ -1338,17 +1353,21 @@ class FlatTuple
...
+    } \
     catch (...) { \
       goto GTEST_CONCAT_TOKEN_(gtest_label_testnothrow_, __LINE__); \
     } \
   } else \
     GTEST_CONCAT_TOKEN_(gtest_label_testnothrow_, __LINE__): \
-      fail("Expected: " #statement " doesn't throw an exception.\n" \
-           "  Actual: it throws.")
+      fail(("Expected: " #statement " doesn't throw an exception.\n" \
+           "  Actual: it throws" + message.get()).c_str())
```
gennadiycivil:
```suggestion
           "  Actual: it throws." + message.get()).c_str())
```

googletest/test/gtest_unittest.cc
```diff
@@ -3803,6 +3804,10 @@ TEST(AssertionTest, ASSERT_NO_THROW) {
   EXPECT_FATAL_FAILURE(ASSERT_NO_THROW(ThrowAnInteger()),
                        "Expected: ThrowAnInteger() doesn't throw an exception."
                        "\n  Actual: it throws.");
+  EXPECT_FATAL_FAILURE(ASSERT_NO_THROW(ThrowAnException("blablubb")),
+                       "Expected: ThrowAnException(\"blablubb\") "
+                       "doesn't throw an exception.\n  "
+                       "Actual: it throws: blablubb");
```
gennadiycivil:
```suggestion
                       "Actual: it throws.: blablubb");
```

googletest/include/gtest/internal/gtest-internal.h
Jonny007-MKD: We need declare `message` in the scope of the `if`. It can't go into the body of the `if` because then we cannot access it in the `else`.

googletest/test/gtest_unittest.cc
Jonny007-MKD: In terms of punctuation this is wrong, isn't it? 😄 Why do you think it is better?

googletest/test/gtest_unittest.cc
```diff
@@ -4540,6 +4545,10 @@ TEST(ExpectTest, EXPECT_NO_THROW) {
   EXPECT_NONFATAL_FAILURE(EXPECT_NO_THROW(ThrowAnInteger()),
                           "Expected: ThrowAnInteger() doesn't throw an "
                           "exception.\n  Actual: it throws.");
+  EXPECT_NONFATAL_FAILURE(EXPECT_NO_THROW(ThrowAnException("blablubb")),
+                          "Expected: ThrowAnException(\"blablubb\") "
+                          "doesn't throw an exception.\n  "
+                          "Actual: it throws: blablubb");
```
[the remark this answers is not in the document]
Jonny007-MKD: That's reasonable :)

Zitrax:
One travis build failed early before even compiling, some known issue?

googletest/include/gtest/internal/gtest-internal.h
gennadiycivil: Lets add a comment to describe why this is here"
We need declare message in the scope of the if. It can't go into the body of the if because then we cannot access it in the else. It also can't go before the if block because than we have to make up some very unique name (line number is not enough)."

gennadiycivil:
> One travis build failed early before even compiling, some known issue?

Travis does this some times

googletest/include/gtest/internal/gtest-internal.h
```diff
@@ -1282,6 +1282,21 @@ class FlatTuple
   }
 };
 
+class AdditionalMessage {
```
gennadiycivil: Lets make this a struct with value being public, there is really no reason for access control.

gennadiycivil:
@Zitrax  My sincere apologies for the delay. (Got very busy for a few days ) . We had an internal discussion, like the functionality. 
Could you please take a look at my comment, merging ConstCharPtr with new AdditionalMessage.

googletest/include/gtest/internal/gtest-internal.h
Zitrax: I merged the functionality in AdditionalMessage into ConstCharPtr and used the same style for the new code. It could have been done the other way too I guess, not sure what makes most sense - but this way results in a smaller diff.

googletest/include/gtest/internal/gtest-internal.h
```diff
@@ -1282,21 +1286,6 @@ class FlatTuple
...
 };
 
-class AdditionalMessage {
- public:
-  AdditionalMessage(const char* message) : value(message) {}
-  AdditionalMessage& operator=(const std::string& message) {
-    value = message;
-    return *this;
-  }
-  operator bool() const { return ::testing::internal::AlwaysTrue(); }
```
Zitrax: Just a note that the previous code here used AlwaysTrue(), I don't know what significance that has so I left the ConstCharPtr implementation as it was with a simple "return true;"

Jonny007-MKD: I don't know either, I just kept the AlwaysTrue() from the previous if statement

Jonny007-MKD:
@gennadiycivil: you requested a change: ".". I suppose you didn't mean to do this :) How can this be resolved?

gennadiycivil:
Please take a look at the fail:
https://travis-ci.org/google/googletest/jobs/500044517


> @gennadiycivil: you requested a change: ".". I suppose you didn't mean to do this :) How can this be resolved?

You suppose correctly, thank you, my bad. Please feel free to ignore that request for

Zitrax:
> Please take a look at the fail:
> https://travis-ci.org/google/googletest/jobs/500044517

So it failed only osx-clang, while it passed in all other configurations. The output is:

```
[ RUN      ] ExpectTest.EXPECT_NO_THROW
/Users/travis/build/google/googletest/googletest/src/gtest.cc:707: Failure
Expected: 1 non-fatal failure containing "Expected: ThrowAnException("blablubb") doesn't throw an exception.
  Actual: it throws.: blablubb"
  Actual:
/Users/travis/build/google/googletest/googletest/test/gtest_unittest.cc:4548: Non-fatal failure:
Expected: ThrowAnException("blablubb") doesn't throw an exception.
  Actual: it throws.
```

I don't know why that would happen only there and I have no mac for further investigation.

Jonny007-MKD:
Weird. Can wie re-run the build?

Zitrax:
@gennadiycivil ping, can we trigger the build again to rule out some travis hickup?

gennadiycivil:
> @gennadiycivil ping, can we trigger the build again to rule out some travis hickup?

Running, 
https://travis-ci.org/google/googletest/jobs/500044517

Zitrax:
Seems consistent, it failed again.

Looking at the output as I understand it the problem is that for the EXPECT_NO_THROW test we except to see `"Actual: it throws.: blablubb"` but we get `"Actual: it throws."`. Looking at the code that could mean that `gtest_msg` was not set or was emptied.

Could this mean we have entered the `catch(...)` block instead of the `catch (const std::exception& e)` block?

But even if that happened shouldn't we have seen `"Actual: it throws.."`, since we initialize the value to "." ?

I also don't understand why EXPECT_NO_THROW fails and ASSERT_NO_THROW pass.

googletest/include/gtest/internal/gtest-internal.h
```diff
@@ -1363,19 +1367,27 @@ constexpr bool InstantiateTypedTestCase_P_IsDeprecated() { return true; }
...
+// some very unique name (line number is not enough).
 #define GTEST_TEST_NO_THROW_(statement, fail) \
   GTEST_AMBIGUOUS_ELSE_BLOCKER_ \
-  if (::testing::internal::AlwaysTrue()) { \
+  if (::testing::internal::ConstCharPtr gtest_msg = ".") { \
     try { \
       GTEST_SUPPRESS_UNREACHABLE_CODE_WARNING_BELOW_(statement); \
     } \
+    catch (const std::exception& e) { \
+      gtest_msg = std::string(": ") + e.what(); \
```
Youw: no woulder CI is failing:
`gtest_msg` holds a pointer to a temporary object that is destroyed immediately after this statement - 
hello usage of dangling pointer :)

Original implementation with `AdditionalMessage` didn't have this issue, since it owned a copy of the message, which is fine.

My opinion - we need `AdditionalMessage` class back, maybe named more consistently with `ConstCharPtr`.
And `ConstCharPtr& ConstCharPtr::operator=(const std::string& message)` shouldn't be allowed, by definition of `class ConstCharPtr`.

googletest/include/gtest/internal/gtest-internal.h
```diff
@@ -1363,19 +1367,27 @@ constexpr bool InstantiateTypedTestCase_P_IsDeprecated() { return true; }
     GTEST_CONCAT_TOKEN_(gtest_label_testthrow_, __LINE__): \
       fail(gtest_msg.value)
 
+// We need to declare the ConstCharPtr object in the scope of the if. It
+// can't go into the body of the if because then we cannot access it in the
+// else. It also can't go before the if block because than we have to make up
+// some very unique name (line number is not enough).
```
Youw: > It also can't go before the if block because than we have to make up some very unique name (line number is not enough).

This is not completely accurate - there is much deeper problem:
We cannot go before the if block because `ASSERT_NO_THROW(...)` should expand to a single statement, so it can be used as:
```cpp
if (some_condition)
    ASSORT_NO_THROW(some_expression) << "Additional error message";
```

gennadiycivil:
Closing, looks inactive

Issue #1786, opened 2018-08-28
CMake WRITE_BASIC_CONFIG_VERSION_FILE for CMake Versions < 3.0


athrun22:
Using CMake 2.8.9, building GoogleTest fails because the PROJECT_VERSION variable is required by the write_basic_config_version_file command on [line 96](https://github.com/google/googletest/blob/58f3f1005cffce8a9d005d0361d3471cd9947501/googletest/CMakeLists.txt#L96), but it is not set for CMake versions < 3.0 on [line 44 of the CMakeLists.txt](https://github.com/google/googletest/blob/58f3f1005cffce8a9d005d0361d3471cd9947501/googletest/CMakeLists.txt#L44). Because the version cannot be set in the project command for versions of CMake less than 3, it is possible to fix this by supplying the version explicitly to the write_basic_config_version_file command. This is shown in the following line, which allowed my project to build successfully.
```cmake
  write_basic_package_version_file(${version_file} VERSION ${GOOGLETEST_VERSION} COMPATIBILITY AnyNewerVersion)
```

gennadiycivil:
Ping

fidergo-stephane-gourichon:
> Thank you very much for this report. The best way to approach this would be to create a proper PR and submit it for consideration.

The fix by @athrun22  works here and allows CMake to complete.
Preparing a PR.

PR #2089, opened 2019-01-30
Fix issue #1786

fidergo-stephane-gourichon:
Proper PR for issue #1786 as requested on https://github.com/google/googletest/issues/1786 .

The fix by @athrun22 works here and allows CMake to complete.

### DIFF
--- a/googletest/CMakeLists.txt
+++ b/googletest/CMakeLists.txt
@@ -93,7 +93,7 @@ if (INSTALL_GTEST)
   set(generated_dir "${CMAKE_CURRENT_BINARY_DIR}/generated" CACHE INTERNAL "")
   set(cmake_files_install_dir "${CMAKE_INSTALL_LIBDIR}/cmake/${cmake_package_name}")
   set(version_file "${generated_dir}/${cmake_package_name}ConfigVersion.cmake")
-  write_basic_package_version_file(${version_file} COMPATIBILITY AnyNewerVersion)
+  write_basic_package_version_file(${version_file} VERSION ${GOOGLETEST_VERSION} COMPATIBILITY AnyNewerVersion)
   install(EXPORT ${targets_export_name}
     NAMESPACE ${cmake_package_name}::
     DESTINATION ${cmake_files_install_dir})


gennadiycivil:
No CLA

gennadiycivil:
This introduces dependency on GOOGLETEST_VERSION which is not documented anywhere. CMake support is community-only however it would be good to understand exactly **why* we are making these changes. 
Please explain in greater details

fidergo-stephane-gourichon:
> This introduces dependency on GOOGLETEST_VERSION

GOOGLETEST_VERSION is a simple and straightforward CMAKE variable already used in several places.

(1) It's set in main CMakeLists.txt, see https://github.com/google/googletest/blob/478a518590901cf3da0868ae553c4ed7f7cf54e1/CMakeLists.txt#L15-L16

(2) It's used in all CMakeList.txt in subdirectories:

https://github.com/google/googletest/blob/50f1a77955bde27cf49173c4133688d61f201e85/googlemock/CMakeLists.txt#L36-L42

https://github.com/google/googletest/blob/e9085769d4819f8491a5ab19e9e3813144e6aa13/googletest/CMakeLists.txt#L43-L49

> which is not documented anywhere.

As people say on WikiTravel, [No advice from Captain Obvious](https://wikitravel.org/en/Wikitravel:No_advice_from_Captain_Obvious "Wikitravel:No advice from Captain Obvious - Wikitravel").

No advice from Captain Obvious - Wikitravel

Seriously, though, this variable is not part of any API to be maintained. It is an implementation detail in the current implementation of the CMake scaffolding of the project.

> CMake support is community-only
> however it would be good to understand exactly *why we are making these changes.
> Please explain in greater details

In order to generate the `<PackageName>ConfigVersion.cmake` which makes Google Test available to downstreams packages, existing CMakeLists uses CMake's built-in `write_basic_package_version_file` as documented on https://cmake.org/cmake/help/latest/module/CMakePackageConfigHelpers.html#generating-a-package-version-file . 

## Search for alternative

The fact is, in the case of a CMake <3.0, the CMakeLists.txt does not include version information in project (see snippet above). This change is the simplest one that provides the version number to `write_basic_package_version_file`.

I tried to include `VERSION ${GOOGLETEST_VERSION}` in pre-3.0 path, and CMake 2.8 rejected that option because it does not support `VERSION` keyword in `project()`. This confirms that current `CMakeList.txt` structure is sound.

## Conclusion

I think we can confidently say the change in this PR is:

* simplest
* uses a simple and straightforward variable
* doesn't actually introduce any new dependency, because the files already depend on it
* doesn't change any public API or introduce any maintenance burden
* is very unlikely to cause any regression
* fixed the problem observed in the first place, as confirmed by two independent persons.

gennadiycivil:
Thank you for the explanation.
Is there anything that would need to happen when we cut the next googletest release?
Will we have to change this variable?
Note that it is already out of sync with the actual version , which is 1.8.x ( see the branches ) 

Thanks



> Seriously, though, this variable is not part of any API to be maintained. It is an implementation detail in the current implementation of the CMake scaffolding of the project.

> In order to generate the `<PackageName>ConfigVersion.cmake` which makes Google Test available to downstreams packages, existing CMakeLists uses CMake's built-in `write_basic_package_version_file` as documented on https://cmake.org/cmake/help/latest/module/CMakePackageConfigHelpers.html#generating-a-package-version-file .

fidergo-stephane-gourichon:
> Thank you for the explanation.

Thanks for merging the PR.

> Is there anything that would need to happen when we cut the next googletest release?
> Will we have to change this variable?

I'm surprised by your question. I would expect any maintainer looking working towards a newer release in such a context to look at the CMakeList.txt, see that variable and increment it when applicable.

> Note that it is already out of sync with the actual version , which is 1.8.x ( see the branches )

It might be or not at all, depending on project conventions.

This is master branch right after 1.8.x and the README.md announces 1.9.x. 

Bumping the version number in master branch just after a release to distinguish the code from the latest release is an existing practice (see https://stackoverflow.com/a/40884130/1429390 for the first example Google search suggested just now). So, seems sane to me.

Some project adopts different practices.

gennadiycivil:
> > Thank you for the explanation.
> 
> Thanks for merging the PR.
> 
> > Is there anything that would need to happen when we cut the next googletest release?
> > Will we have to change this variable?
> 
> I'm surprised by your question. I would expect any maintainer looking working towards a newer release in such a context to look at the CMakeList.txt, see that variable and increment it when applicable.
> 
> > Note that it is already out of sync with the actual version , which is 1.8.x ( see the branches )
> 
> It might be or not at all, depending on project conventions.
> 
> This is master branch right after 1.8.x and the README.md announces 1.9.x.
> 
> Bumping the version number in master branch just after a release to distinguish the code from the latest release is an existing practice (see https://stackoverflow.com/a/40884130/1429390 for the first example Google search suggested just now). So, seems sane to me.
> 
> Some project adopts different practices.

The reason I am asking is that we do not use CMake at all.  CMake is completely community supported for googletest . 
As you may see this project is supporting bazel ( main release mechanism ) , CMake, automake - community supported.

Therefore  if this variable  is expected ( please do be careful when making assumptions about releases, semantic versioning and such ) to play any role in the release process whatsoever there has to be *something* documented to this effect .

What is going to happen with the next release is that at some point we will cut the branch 1.9.x and tag it. That is it. , which would make anything based on this variable instantly wrong

fidergo-stephane-gourichon:
> (...) CMake is completely community supported for googletest . (...)

Thank you for those explanations. It makes much more sense now. 

> What is going to happen with the next release is that at some point we will cut the branch 1.9.x and tag it. That is it. , which would make anything based on this variable instantly wrong

Given the context, that's a totally valid point.

It is possible in CMake to automatically compute the value of the variable, based on `git describe`.
For example, this answer: https://stackoverflow.com/a/38836941

Is it a good idea some day to open an issue titled like "CMake variable GOOGLETEST_VERSION is fixed, not computed from non-CMake-specific information" and offer a PR? I'll also offer to add some comments on top of the CMakeLists.txt to make clear that bazel is primary build system and CMake is community supported.

Thanks for the explanation and keep up the good work! I guess you can close this issue now.

gennadiycivil:
@fidergo-stephane-gourichon Thank you very much. Yes I agree that 
>>"to open an issue titled like "CMake variable GOOGLETEST_VERSION is fixed, not computed from >>non-CMake-specific information" and offer a PR? I'll also offer to add some comments on top of the >>CMakeLists.txt to make clear that bazel is primary build system and CMake is community supported.

This would be a worthy thing to do and if you could take a look and propose a PR this would be welcomed.

In addition your idea to add some comments on top of the >>CMakeLists.txt to make clear that bazel is primary build system and CMake is community supported sounds great.. If you have some time a PR would be welcomed as well

Thanks again

gennadiycivil:
Closing for now, hoping to see PRs as described, thank you again

PR #2090, opened 2019-01-30
Update advanced.md casing in example

KellyWalker:
Updated the example that says not to use SetupTestSuite with a small 'u' to actually use a small 'u'


KellyWalker:
I signed it!

PR #2096, opened 2019-01-31
[Feature / Enhancement]: New Macros for checking exact exception type and what().

shahzadlone:
Implemented a ```EXPECT_WHAT_THROW``` and ```ASSERT_WHAT_THROW``` that
checks if the exact type of exception we expect is thrown and it is
thrown with the exact message we expect (i.e. it's err.what() is the
same.)

Works like a charm, maybe someone can help me clean this up by putting a
common part into the internal header.

Also I can unit test this if someone can guide me where I should place,
and where some similar unit tests are located.

### DIFF
--- a/googletest/include/gtest/gtest.h
+++ b/googletest/include/gtest/gtest.h
@@ -2187,6 +2187,49 @@ GTEST_API_ AssertionResult DoubleLE(const char* expr1, const char* expr2,
                                     double val1, double val2);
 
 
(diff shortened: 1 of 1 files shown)


gennadiycivil:
Thank you for this report. We are actively discouraging adding new macros

shahzadlone:
Would there be a better way to do this then?

PR #2034, opened 2018-12-27
Fix FreeBSD -Werror compilation issues

ngie-eign:
The attached pull request fixes 2 trivial issues when compiling the googletest tests on FreeBSD with WARNS >= 1, i.e., -Werror on, with clang.

This issue was seen when trying to integrate googletest 1.8.1 into the FreeBSD base system on my GitHub project branch.

This PR requires #2100.

PR #2100, opened 2019-02-04
Test out changes with clang/OSX each PR using Travis CI

ngie-eign:
This will help ensure that test coverage isn't missed when changes that
can impact OSX with clang are submitted. Although not perfect, testing
changes on OSX with clang is an ok proxy [for now] for testing changes on
FreeBSD with clang (testing on FreeBSD is non-trivial, as Travis CI doesn't
support operating systems other than Linux and OSX).

In order to support this, install the packages via homebrew using a
`addons::homebrew::packages` block (similar to `addons::apt::packages`
for Ubuntu), as documented in the Travis CI docs
( https://docs.travis-ci.com/user/installing-dependencies/ ). While
here, try pushing apt* calls into the Travis config for Ubuntu, instead
of delaying the equivalent calls in the `ci/*.sh` scripts. Keep the
`ci/*.sh` scripts for ease of testing locally (and extend the OSX one to
install gcc 4.9 and llvm 3.9, like the travis config does).

In order to accomodate this change (and because the homebrew package for
llvm@3.9 doesn't automatically add clang*-3.9 to `$PATH`), `$PATH` needs
to be adjusted to find the llvm@3.9 toolchain.

This PR was originally part of #2034.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,8 +48,8 @@ matrix:
       compiler: gcc
       env: BUILD_TYPE=Release VERBOSE=1 CXX_FLAGS=-std=c++11 HOMEBREW_LOGS=~/homebrew-logs HOMEBREW_TEMP=~/homebrew-temp
     - os: osx
+      compiler: clang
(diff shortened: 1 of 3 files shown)


ngie-eign:
This change was originally a part of https://github.com/google/googletest/pull/2034 .

CC: @EricWF, @gennadiycivil

PR #2101, opened 2019-02-04
Fix Cygwin CMake Build

MaEtUgR:
**Why?**
Based on the findings I wrote [here](https://github.com/google/googletest/pull/1890#issuecomment-460063010) I'm fixing the global CMakeLists file such that building under Cygwin also works out of the box e.g. when following [the readme and including the repo in your CMake project](https://github.com/google/googletest/blame/0ea2d8f8fa1601abb9ce713b7414e7b86f90bc61/googletest/README.md#L160
).

>On Cygwin specifying `set(CMAKE_CXX_EXTENSIONS OFF)` breaks the build exactly like described in https://stackoverflow.com/q/18784112/6326048
>
>- `add_definitions(-std=c++11)` also breaks it
>- Removing the whole block of compile std defines builds fine
>
>@peterjc123 Is `set(CMAKE_CXX_EXTENSIONS OFF)` really necessary for you? If yes I'll create a pr to only exclude the line for Cygwin to makes sure it builds out of the box on that platform as well.

**What?**
- **First commit** To detect Cygwin I had to move the project initialization to the beginning, see https://stackoverflow.com/a/35914521/6326048
- **Second commit** I'm not disabling C++ extensions when building the repository using the global CMake file in a Cygwin environment.

**Really?**
Here's the output after the following commands in the freshly cloned repo wit the latest Cygwin environment:
```
mkdir build
cd build
cmake ..
make
```
Before change:
gtest-fail
After change:
gtest-success

**Additional info**
Building gtest without any C++ standard definitions works fine on Cygwin, probably the defaults are just right. That's also why diectly building it using the `CMakeLists.txt` files in the `googletest` or `googlemock` folder is fine. I'm fixing the global one because I intend to use it and it's also frustrating to new users to first have a failing build.


MaEtUgR:
`I signed it!` 😃

CMakeLists.txt
```diff
@@ -1,20 +1,22 @@
 cmake_minimum_required(VERSION 2.8.8)
 
+project(googletest-distribution)
+set(GOOGLETEST_VERSION 1.9.0)
```
gennadiycivil: The PR description is specifc to "Fix Cygwin CMake Build". Why are we also inserting a version? Does Gygwin not work without it?

MaEtUgR: I'm not inserting a version. In the first commit I moved these two lines exactly like they were to the beginning of the file because cygwin gets detected by cmake when the project is defined and hence the project definition had to be before using the variable `CYGWIN`. I wrote a description including a link to more explanation under the title "What? - First commit" of my description.

gennadiycivil: My apologies I do not follow, what is the :"first file" . This PR shows me one files changed and the diff shows inserted line 4, +set(GOOGLETEST_VERSION 1.9.0). What I am missing?

gennadiycivil: Oh, nvm, you just moved the 2 lines to the beginning of the file

gennadiycivil:
merged now

MaEtUgR:
@gennadiycivil I just tested and it works perfect thanks! I could probably contribute CI on Cygwin using appveyor if there's any interest.

PR #2102, opened 2019-02-04
Add additional -W* flags in cxx_strict_flags for the tests with clang

ngie-eign:
This should help catch more issues with warnings upfront on platforms
that use clang, e.g., FreeBSD, Linux, OSX.

This PR was originally part of #2034 and requires #2034 and #2100 in order to not break the build.


ngie-eign:
This is also failing to build due to Travis CI issues with fetching security keys. I'm digging into this.

gennadiycivil:
Please see my comment in https://github.com/google/googletest/pull/2034

PR #2109, opened 2019-02-10
Minor enhancements and fixes

knuto:
These are a number of minor enhancements and bug fixes that we have been using for several years 
within Oracle, but not been able to contribute until now, as the licensing issues needed to be sorted out.
Hopefully the individual commit messages should mostly speak for themselves:

- **Report the number of actually executed assertions:**  I added this after I had been "running" 
  lots of tests once, just to discover that for configuration reasons, no actual asserts were executed.
- **Report number of filtered out tests:** When running complicated filter expressions, it's useful to know how many tests ended up being filtered out.
- **Remove / from parameterized test names if base test name is empty:**:  In KTF 
     (see github project oracle/ktf) we slightly misuse the parameterizable tests to create a namespace for executing tests inside the Linux kernel. This makes it possible to do this fully transparently.
- **Stop TestInfo::Run() calling a function through null pointer** : I ran into this one and discovered there were already a fix for it - why hasn't it been merged?
- **Set gtest version correctly for older cmake versions**: At Oracle, we rely on being able to run our tests against both new and old versions of Linux, so the ability to rely on version checking is important.
- **Generate a libgtest.la to help libtool managing dependencies**: I know running with a dynamically linked used to be deprecated by Googletest, but we have been using this mode for many years now and it serves our use and purpose well. Having libtools support allows Googletest to work seamlessly for developer builds and for installed builds against several projects we have that uses autotools.

**Changes in v2**:
Fixed the type conversion issue reported by appveyor on Windows.


knuto:
I'm not able to figure out what the Google bot reacts to with commit "Generate a libgtest.la..." - I am the sole author of it and there's no mentions of anyone else?

jwakely:
I'm happy for my commit to be contributed (I already tried to get it included once via #967 but it wasn't taken, for no good reason).

gennadiycivil:
@knuto Thank you so much for this contribution! There is a number of changes here and it would be much easier to review if they were split into individual focused PRs. Could you please do that.
Thank you again

knuto:
> @knuto Thank you so much for this contribution! There is a number of changes here and it would be much easier to review if they were split into individual focused PRs. Could you please do that.
> Thank you again

Ok, the two first commits are falsely dependent due to their effect on the "golden" test output, and also somewhat related, so I keep them together, I have cherry-picked the rest onto separate branches.

PR #2110, opened 2019-02-11
Report the number of actually executed assertions, and the number of filtered out tests

knuto:
First part of split from PR #2109

Since these are falsely dependent I'd prefer to rebase only if you decide to keep 
one but reject the other.

### DIFF
--- a/googletest/include/gtest/gtest.h
+++ b/googletest/include/gtest/gtest.h
@@ -596,6 +596,9 @@ class GTEST_API_ TestResult {
   // UNIX epoch.
   TimeInMillis start_timestamp() const { return start_timestamp_; }
 
+  // Return the number of assert executed with no errors in this test.
+  int success_assert_count() const { return success_assert_count_; }
+
   // Returns the i-th test part result among all the results. i can range from 0
   // to total_part_count() - 1. If i is not in that range, aborts the program.
   const TestPartResult& GetTestPartResult(int i) const;
@@ -656,6 +659,9 @@ class GTEST_API_ TestResult {
   // Increments the death test count, returning the new count.
   int increment_death_test_count() { return ++death_test_count_; }
 
+  // Increments the successful assert count, returning the new count.
+  int increment_success_assert_count(int v = 1) { return success_assert_count_ += v; }
+
   // Clears the test part results.
   void ClearTestPartResults();
 
@@ -674,6 +680,8 @@ class GTEST_API_ TestResult {
   int death_test_count_;
   // The start time, in milliseconds since UNIX Epoch.
   TimeInMillis start_timestamp_;
+  // Running count of number of successful asserts.
+  int success_assert_count_;
   // The elapsed time, in milliseconds.
   TimeInMillis elapsed_time_;
 
@@ -760,6 +768,13 @@ class GTEST_API_ TestInfo {
   // Returns the result of the test.
   const TestResult* result() const { return &result_; }
 
+  // Returns the number of executed assertions that has passed silently:
+  int success_assert_count() const { return result_.success_assert_count(); }
+
+  // Increment the number of assertions executed without errors:
+  int increment_success_assert_count(int v = 1) {
+    return result_.increment_success_assert_count(v);
+  }
  private:
 #if GTEST_HAS_DEATH_TEST
   friend class internal::DefaultDeathTestFactory;
@@ -1348,6 +1363,10 @@ class GTEST_API_ UnitTest {
 
   // Gets the elapsed time, in milliseconds.
   TimeInMillis elapsed_time() const;
+
+  // Increment the number of assertions executed without errors:
+  int increment_success_assert_count(int v = 1)
+      GTEST_LOCK_EXCLUDED_(mutex_);
 
   // Returns true if and only if the unit test passed (i.e. all test suites
   // passed).
--- a/googletest/include/gtest/gtest_pred_impl.h
+++ b/googletest/include/gtest/gtest_pred_impl.h
@@ -75,7 +75,7 @@ namespace testing {
 #define GTEST_ASSERT_(expression, on_failure) \
   GTEST_AMBIGUOUS_ELSE_BLOCKER_ \
   if (const ::testing::AssertionResult gtest_ar = (expression)) \
-    ; \
+    ::testing::UnitTest::GetInstance()->increment_success_assert_count(); \
   else \
     on_failure(gtest_ar.failure_message())
 
--- a/googletest/src/gtest.cc
+++ b/googletest/src/gtest.cc
@@ -2026,7 +2026,11 @@ std::string AppendUserMessage(const std::string& gtest_msg,
 
 // Creates an empty TestResult.
 TestResult::TestResult()
-    : death_test_count_(0), start_timestamp_(0), elapsed_time_(0) {}
+    : death_test_count_(0),
+      start_timestamp_(0),
+      success_assert_count_(0),
+      elapsed_time_(0) {
+}
 
 // D'tor.
 TestResult::~TestResult() {
@@ -2184,6 +2188,7 @@ void TestResult::Clear() {
   test_part_results_.clear();
   test_properties_.clear();
   death_test_count_ = 0;
+  success_assert_count_ = 0;
   elapsed_time_ = 0;
 }
 
@@ -3245,6 +3250,8 @@ void PrettyUnitTestResultPrinter::OnTestEnd(const TestInfo& test_info) {
   if (test_info.result()->Failed())
     PrintFullTestCommentIfPresent(test_info);
 
+  printf(", %d assertions",
+	 test_info.result()->total_part_count() + test_info.result()->success_assert_count());
   if (GTEST_FLAG(print_time)) {
     printf(" (%s ms)\n", internal::StreamableToString(
            test_info.result()->elapsed_time()).c_str());
@@ -4696,6 +4703,13 @@ int UnitTest::test_to_run_count() const { return impl()->test_to_run_count(); }
 // UNIX epoch.
 internal::TimeInMillis UnitTest::start_timestamp() const {
     return impl()->start_timestamp();
+}
+
+// Increment the number of assertions executed without errors:
+int UnitTest::increment_success_assert_count(int v)
+  GTEST_LOCK_EXCLUDED_(mutex_) {
+  internal::MutexLock lock(&mutex_);
+  return impl_->increment_success_assert_count(v);
 }
 
 // Gets the elapsed time, in milliseconds.
--- a/googlemock/test/gmock_output_test_golden.txt
+++ b/googlemock/test/gmock_output_test_golden.txt
@@ -7,7 +7,7 @@ FILE:#: Mock function call matches EXPECT_CALL(foo_, Bar2(0, _))...
     Function call: Bar2(0, 0)
           Returns: false
 Stack trace:
-[       OK ] GMockOutputTest.ExpectedCall
+[       OK ] GMockOutputTest.ExpectedCall, 0 assertions
 [ RUN      ] GMockOutputTest.ExpectedCallToVoidFunction
 
 FILE:#: EXPECT_CALL(foo_, Bar3(0, _)) invoked
@@ -16,7 +16,7 @@ Stack trace:
 FILE:#: Mock function call matches EXPECT_CALL(foo_, Bar3(0, _))...
     Function call: Bar3(0, 0)
 Stack trace:
-[       OK ] GMockOutputTest.ExpectedCallToVoidFunction
+[       OK ] GMockOutputTest.ExpectedCallToVoidFunction, 0 assertions
 [ RUN      ] GMockOutputTest.ExplicitActionsRunOut
 
 GMOCK WARNING:
@@ -26,7 +26,7 @@ GMOCK WARNING:
 FILE:#: Actions ran out in EXPECT_CALL(foo_, Bar2(_, _))...
 Called 2 times, but only 1 WillOnce() is specified - returning default value.
 Stack trace:
-[       OK ] GMockOutputTest.ExplicitActionsRunOut
+[       OK ] GMockOutputTest.ExplicitActionsRunOut, 0 assertions
 [ RUN      ] GMockOutputTest.UnexpectedCall
 unknown file: Failure
 
@@ -40,7 +40,7 @@ FILE:#: EXPECT_CALL(foo_, Bar2(0, _))...
            Actual: 1
          Expected: to be called once
            Actual: never called - unsatisfied and active
-[  FAILED  ] GMockOutputTest.UnexpectedCall
+[  FAILED  ] GMockOutputTest.UnexpectedCall, 1 assertions
 [ RUN      ] GMockOutputTest.UnexpectedCallToVoidFunction
 unknown file: Failure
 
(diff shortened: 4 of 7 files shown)


gennadiycivil:
@knuto Thank you very much for this. Could you please provide clear description what this PR is doing, why it is important.
The PR needs to stand on its own and be possible to review independently.  Provide more information on exactly the issue / problems this is trying to address. What is the before/after effect this would create and how is this beneficial?

knuto:
The idea is that the commit log messages and the effect on golden output would be sufficient, but I can add back in the overview I put in the original PR.

googletest/include/gtest/gtest.h
```diff
@@ -601,6 +601,9 @@ class GTEST_API_ TestResult {
   // Returns the elapsed time, in milliseconds.
   TimeInMillis elapsed_time() const { return elapsed_time_; }
 
+  // Return the number of assert executed with no errors in this test.
+  size_t success_assert_count() const { return success_assert_count_; }
```
gennadiycivil: even though sizet_t is correct-er we should be using int to prevent bugs. For example you have in gtest.cc this line " test_info.result()->total_part_count() + test_info.result()->success_assert_count());"  one is an int the other one is size_t, which will produce undesired behavior.  Please change everything to be int, to be consistent with the existing code and prevent these issues

gennadiycivil: This comment pertains to may places in this PR so please change to int where applicable

googletest/include/gtest/gtest.h
```diff
@@ -658,6 +661,13 @@ class GTEST_API_ TestResult {
   // Increments the death test count, returning the new count.
   int increment_death_test_count() { return ++death_test_count_; }
 
+  // Increments the successful assert count, returning the new count.
+  size_t increment_success_assert_count(size_t v = 1) { return success_assert_count_ += v; }
```
gennadiycivil: Should just return void, I dont see where the return is actually used

googletest/include/gtest/gtest.h
```diff
@@ -658,6 +661,13 @@ class GTEST_API_ TestResult {
...
+
+  // Set/get executed flag.
+  bool was_executed() const { return executed_; }
```
gennadiycivil: This data member is redundant, the value can be calculated run-time between Passed(), "Failed" , "Skipped". Please remove

googletest/include/gtest/gtest.h
```diff
@@ -760,6 +778,13 @@ class GTEST_API_ TestInfo {
   // Returns the result of the test.
   const TestResult* result() const { return &result_; }
 
+  // Returns the number of executed assertions that has passed silently:
+  size_t success_assert_count() const { return result_.success_assert_count(); }
+
+  // Increment the number of assertions executed without errors:
+  size_t increment_success_assert_count(size_t v = 1) {
```
gennadiycivil: return void as before

gennadiycivil: Please remove argument size_t v = 1, it is unused, therefore not needed for the internal-only API

gennadiycivil: Should be private, this is for internal use

googletest/include/gtest/gtest.h
knuto: I am not sure I understand how this can become a problem?
size_t can hold a larger positive integer than int, and even though total_part_count() returns 
an int, it should never become negative? 

I have no problem converting success_assert_count to int but thought size_t a better choice as the number of asserts for a test can easily become huge quickly,
possible wrapping to negative for an int.

googletest/include/gtest/gtest.h
knuto: Personally I think it makes a better API being able to read the old value while incrementing,
even though it is not used, and it is a common way to implement such functions. 
But I can remove it.

googletest/include/gtest/gtest.h
knuto: As far as I recall the problem is that Skipped will only count tests that were actually attempted executed. If let's say you have an environment issue which causes the tests never to be even considered, they are not counted as skipped. We had runs where no tests were being run because conditions checked by AddGlobalTestEnvironment callbacks caused them to not be attempted run at all.

googletest/include/gtest/gtest.h
knuto: We need the v argument in KTF, because the test has executed in the kernel, and the number of asserts in that test are reported together. Since we try to implement our KTF as a user of Googletest without hacks to get to internal APIs, we also need it to be public.

knuto:
> In addition to the review comments, please make sure that the new information is also included in XML and JSON reports

I agree that all the variants of reporting should be treated as "equal citizens", but it does not seem to be that way now? There's no "golden_output" file and just minimal tests for anything but the text output. We were also puzzled about the "errors" indication with seems to be hard coded to "0" ?

While looking into this, I did find a number of obvious merge mistakes in my patches, so I'll send an updated version once I get some feedback to my concerns to your changes.

googletest/include/gtest/gtest.h
knuto: Do we have an example of a Skipped test? There's none in the "golden" output?

knuto: There's also no JSON or XML output of DISABLED tests, which is of the same category as FILTERED tests, which is the natural pattern to follow IMHO.

googletest/include/gtest/gtest.h
gennadiycivil: It is better to be consistent with the existing code

knuto: ok - will change back to int then

knuto:
I have changed the type of the assertion counter from size_t to int as requested, and fixed a mistaken merge change. With my comments above I believe I have covered the concerns you had apart from inclusion into XML and JSON reports, where I am waiting for feedback to my above questions.

knuto:
After several rounds of compiler warnings from visual studio in Appveyor, I added an extra patch I think is desired, on top here, so that error/warnings for Linux/g++ resembles more the ones from Visual Studio - hope that's ok even in the same pull req since it is kind of related to the fix..

knuto:
@gennadiycivil From my perspective I have no outstanding todos on this one either so awaiting your feedback!

googletest/include/gtest/gtest_pred_impl.h
```diff
@@ -75,7 +75,7 @@ namespace testing {
 #define GTEST_ASSERT_(expression, on_failure) \
   GTEST_AMBIGUOUS_ELSE_BLOCKER_ \
   if (const ::testing::AssertionResult gtest_ar = (expression)) \
-    ; \
+    ::testing::UnitTest::GetInstance()->increment_success_assert_count(); \
```
gennadiycivil: This seems dangerous.
Is there any synchronization here?
If there is, it is going to make this slower.
If there isn't, it is going to make this thread-hostile.

gennadiycivil: (this is from internal review)

knuto: Good point,
 - the main purpose of the assert counting, and why I think it is important, is to be able to easily spot whether a test actually has tested something reasonably close to what the intentions are, so a 100% accuracy is not really critical, but it is of course better/easier if it is accurate, so this is a tradeoff.

Unless there's a way to detect whether a test is exposed to multithreading, I suppose there's then two main choices: 
a) Leave it as it is for performance reasons, and document that it is a hint.
b) Use an atomic operation for the counting, which at least is more light weight than a locking based scheme.

 It could possibly be made configurable to be able to disable it for particularly challenging tests.
I suggest to opt for b) - I would argue that getting this information is worth the extra cycle cost.

knuto: There's of course always the remaining question of whether using an atomic will have undesired effects on the parallelism in the test, possibly affect test outcomes, but that can probably be both positive and negative depending on the purpose of the test.

gennadiycivil:
Closing, looks inactive.

knuto:
> Closing, looks inactive.

Please don't! I have been waiting for your feedback on this since May 18!

knuto:
Please let me know what the problem is - my understanding is that I have addressed all issues

gennadiycivil:
Sorry for the disconnect, I did not read your response as an actual complete answer - My bad.

So you are proposing to " Use an atomic operation for the counting, which at least is more light weight than a locking based scheme." ?

Not sure how this wold work on all platforms. Please advise. Best would be to rebase this PR, resolve the conflicts and propose what it would look like.

gennadiycivil:
Also , could you please limit the changes to 1 PR. So for example this PR could be about "Add ability to count and report the number of actually executed assertions for a test".  and only that, lets move all other things ( report number of filtered tests ) someplace else

gennadiycivil:
Looking at this again, I does not seem worth to add atomic ops for the sake of counting assertions.
You are correct - using an atomic might have undesired effects on the parallelism in the test.
What exactly is the use case to count assertions? Lets start with that again

knuto:
The main reason why I think it is a good idea is that it gives easy visual info about whether that test actually did something or not, and what amount of testing actually went on. It helps people avoid running into situations where they think they have tested something, but for some reason something went wrong within the test that caused that all the assertions in the test did not end up being executed. Getting "0 assertions" (or lets say 5 assertions when ~50000 is expected) is a clear indication of that.

knuto:
>Also , could you please limit the changes to 1 PR. So for example this PR could be about "Add ability to count and report the number of actually executed assertions for a test". and only that, lets move all other things ( report number of filtered tests ) someplace else

Ok, I assume this means you are happy with the current version (lost updates may make the assertion count imprecise). [I have to justify the extra time spent here ;-) ]

Being used to having multiple patches in a PR, may I ask why you need this to go into two PRs? 
They both are simple report enhancing commits, they are already in separate commits.
Since they both change the golden output, it is some work to separate them, and they will create a conflict when pulled separately, since they will end up with a "false dependency" between them.
So it seems like a lot of work for no real benefit?

gennadiycivil:
> The main reason why I think it is a good idea is that it gives easy visual info about whether that test actually did something or not, and what amount of testing actually went on. It helps people avoid running into situations where they think they have tested something, but for some reason something went wrong within the test that caused that all the assertions in the test did not end up being executed. Getting "0 assertions" (or lets say 5 assertions when ~50000 is expected) is a clear indication of that.
>

 I am tempted to say (forgive me because I don't have the context here )  that the tests like that should have produced a FAIL and this would have been an indication that *something is wrong* which could have led to people making the tests in question more reliable.

* Of course I don't know the context. * There is however a nagging feeling that this assertion count is trying to cover up a smell in not-optimally-designed tests that should have failed.  

> Ok, I assume this means you are happy with the current version (lost updates may make the assertion count imprecise). [I have to justify the extra time spent here ;-) ]
> 

Well, not really happy unfortunately. Having a counter that is under certain conditions is incorrect is in my opinion worse than having no counter at all. 

> Being used to having multiple patches in a PR, may I ask why you need this to go into two PRs?
> They both are simple report enhancing commits, they are already in separate commits.
> Since they both change the golden output, it is some work to separate them, and they will create a conflict when pulled separately, since they will end up with a "false dependency" between them.
> So it seems like a lot of work for no real benefit?

Having just one thing to review makes it much easier to roll back should this be necessary. Since googletest is used internally by Google to run literally millions of tests this is a pretty strong requirement. 
Just to explain, when there is a change to googletest the maintainers often have to run this change against an entire existing body of code to make sure there are no negative side effects because given this large body of code the [Hyrum's Law]( http://www.hyrumslaw.com/) takes effect.
When there is only one logical change it makes this job a lot easier and lot safer.

Since there is no logical link between reporting of number of assertions and reporting number of filtered out tests they should be separate.  The corresponding changes in the golden file should be atomic and independent on one another, so it is possible to roll back if needed.

knuto:
>     * Of course I don't know the context. * There is however a nagging feeling that this assertion count is trying to cover up a smell in not-optimally-designed tests that should have failed.

So I'll try to construct a short example here:
Consider a test that is subject to ifdef's because it is run across multiple platforms.
By accident, on one of the platforms, the ifdefs were made wrong in a way that the body of the test did not actually get compiled in. So the test passes perfectly on all platforms, but on the faulty platform, no assertions were executed. On the other platforms, the test executes as it should, so 
there's no attention to any issue. Half a year later a regression slips through which would have been caught by the faulty ifdef'ed test. Now, looking at logs of older test results, nothing is 
unusual, it seems the test has been passing all along. With the assertion count, a count of 0 would be easy to spot, and the whole problem might not have slipped through. And even if it had, test logs might still have recorded the 0 assertions and made it easier to realize what went wrong.

> > Ok, I assume this means you are happy with the current version (lost updates may make the assertion count imprecise). [I have to justify the extra time spent here ;-) ]
> 
> Well, not really happy unfortunately. Having a counter that is under certain conditions is incorrect is in my opinion worse than having no counter at all.

Then atomic is the solution.

> > Being used to having multiple patches in a PR, may I ask why you need this to go into two PRs?
> > They both are simple report enhancing commits, they are already in separate commits.
> > Since they both change the golden output, it is some work to separate them, and they will create a conflict when pulled separately, since they will end up with a "false dependency" between them.
> > So it seems like a lot of work for no real benefit?
> 
> Having just one thing to review makes it much easier to roll back should this be necessary. Since googletest is used internally by Google to run literally millions of tests this is a pretty strong requirement.

The commit set must be (and is) bisectable of course, but that's not what I am referring to here.

> Just to explain, when there is a change to googletest the maintainers often have to run this change against an entire existing body of code to make sure there are no negative side effects because given this large body of code the [Hyrum's Law](http://www.hyrumslaw.com/) takes effect.
> When there is only one logical change it makes this job a lot easier and lot safer.

So I assume this must be some form of build technical issue then. It's always possible to pick the first out of a commit set of multiple commits.

> Since there is no logical link between reporting of number of assertions and reporting number of filtered out tests they should be separate. The corresponding changes in the golden file should be atomic and independent on one another, so it is possible to roll back if needed.

Well, since they both affect related context lines, that's unfortunately not the case, and it's a consequence of the way the "golden output" works, not the code changes. 
But I can split them up (or we can do one at a time) if that's what it takes.

knuto:
So what I propose I do to save work on both sides, if you want to go ahead (I presume then with the atomic based version?), is that we use this PR for the assertion commit. I'll take out the filtered test commit and submit that as a separate pull req after the assertion count has been decided upon and if accepted, also merged. Ok?

gennadiycivil:
In order to be able to property review this PR , yes lets limit the work in it to the assertion commit count. This way we can move forward. Please not that there is no guarantee that the PR will be accepted and merged, this is just so we can do the proper review. 
Thank you for looking into this!

CJ-Johnson:
@knuto I'd like to take another look at this PR. Are you able to resolve the merge conflicts?

Thanks!

knuto:
@CJ-Johnson thanks - it should be ok now

googletest/include/gtest/gtest_pred_impl.h
CJ-Johnson: Has this been updated to use atomics?

googletest/src/gtest.cc
```diff
@@ -4698,6 +4705,13 @@ internal::TimeInMillis UnitTest::start_timestamp() const {
     return impl()->start_timestamp();
 }
 
+// Increment the number of assertions executed without errors:
+int UnitTest::increment_success_assert_count(int v)
+  GTEST_LOCK_EXCLUDED_(mutex_) {
```
CJ-Johnson: This appears to still use a mutex, instead of atomics. Were atomics tried?

gennadiycivil:
@knuto @CJ-Johnson ping :)

CJ-Johnson:
After discussing this change and estimating the compute cost, this just isn't something we can adopt. The success path for test runs must be quick. Even with atomics, this will be too much of a bottle neck.

Thanks for the PR!

PR #2120, opened 2019-02-13
Compile clang with `-Wall -Wshadow -Werror`

ngie-eign:
It was not compiling any of the code with warnings prior to this.


ngie-eign:
This PR requires #2119 to land first.

PR #2123, opened 2019-02-13
Fix clang `-Winconsistent-missing-override` warnings

ngie-eign:
`MatchAndExplain(..)` in `gmock-matchers_test` overrides a virtual method.
Remove the `virtual` keyword and apply `override` to it instead.


ngie-eign:
@gennadiycivil: ping?

PR #2125, opened 2019-02-13
Fix clang `-Wunused-parameter` warnings

ngie-eign:
Comment out unused method parameters in their relevant definitions.

Signed-off-by: Enji Cooper <yaneurabeya@gmail.com>

PR #2126, opened 2019-02-13
Add `cxx_strict_flags` for clang to match FreeBSD's WARNS flags

ngie-eign:


ngie-eign:
This PR requires #2123 and #2125.

ngie-eign:
@gennadiycivil: ping?

gennadiycivil:
@ngie-eign There is a number of flags ... what would it take to "just fix the code?"
Thanks again for all this work!
G

gennadiycivil:
@ngie-eign also if you could please explain in greater details these flags ( for example I am not familiar with Wformat-y2k

ngie-eign:
> @ngie-eign There is a number of flags ... what would it take to "just fix the code?"
> Thanks again for all this work!
> G

The additional flags are to ensure that the code doesn't regress (as much) between FreeBSD and Linux/OSX.

> @ngie-eign also if you could please explain in greater details these flags ( for example I am not familiar with Wformat-y2k

These flags are probably set as a precursor of a larger -Wformat set. Per the `gcc` manpage:
```
           -Wformat=2
               Enable -Wformat plus additional format checks.  Currently
               equivalent to -Wformat -Wformat-nonliteral -Wformat-security
               -Wformat-y2k.
...
           -Wformat-y2k
               If -Wformat is specified, also warn about "strftime" formats
               that may yield only a two-digit year.
```

This is protecting against `%g`/`%y` [on FreeBSD at least]:
```
     %g    is replaced by the same year as in “%G”, but as a decimal number
           without century (00-99).
...
     %y    is replaced by the year without century as a decimal number
           (00-99).
```
They want to ensure that double digit years aren't provided, resulting in hard to decipher dates when centuries may vary, as some dates might be 20th century values and some might be 21st century values (or later). Example `00` can be 1800, 1900, 2000, 2100, etc.

gennadiycivil:
Hmm. So I just built googletest with bazel build --cxxopt=-Wformat googletest/... and it worked without any issues.
Then I tried bazel build --cxxopt=-Wformat-y2k googletest/... also worked just fine.
Then I did bazel build --cxxopt=-Wformat=2 googletest/... also worked just fine.

Not sure why we are doing -Wno-format-y2k when in fact it does not seem to create an issue.... could you please advise?

ngie-eign:
> Hmm. So I just built googletest with bazel build --cxxopt=-Wformat googletest/... and it worked without any issues.
> Then I tried bazel build --cxxopt=-Wformat-y2k googletest/... also worked just fine.
> Then I did bazel build --cxxopt=-Wformat=2 googletest/... also worked just fine.
> 
> Not sure why we are doing -Wno-format-y2k when in fact it does not seem to create an issue.... could you please advise?

I was mimicking FreeBSD's options, but (as you suggested), this flag isn't necessary. I'll remove it now and force push the change.

ngie-eign:
I just removed `-Wno-format-y2k`.

gennadiycivil:
Merged

PR #2127, opened 2019-02-13
Provide a complete example for `EXPECT_NONFATAL_FAILURE(..)`

ngie-eign:
The existing example is a bit lean on details, in particular, the
headers that are required in order to use the API and the method for
calling the API.

The new example provides a fully functional test program that
demonstrates how to implement the calls, in both the non-curly brace and
curly brace form. I originally got confused when trying to augment
existing testcases (which were expected to fail on FreeBSD) in
https://github.com/google/capsicum-test/pull/36 .

### DIFF
--- a/googletest/docs/advanced.md
+++ b/googletest/docs/advanced.md
@@ -1789,7 +1789,33 @@ current thread whose message contains the given `substring`, or use
   EXPECT_NONFATAL_FAILURE(statement, substring);
 ```
 
-if you are expecting a non-fatal (e.g. `EXPECT_*`) failure.
+if you are expecting a non-fatal (e.g. `EXPECT_*`) failure. A full example is
+as follows:
+
+```c++
+  #include <gtest/gtest.h>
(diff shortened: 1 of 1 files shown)


ngie-eign:
@gennadiycivil: ping?

gennadiycivil:
Thank you for this contribution! It appears to be redundant, you can see examples in the unit test 
https://github.com/google/googletest/blob/0599a7b8410dc5cfdb477900b280475ae775d7f9/googletest/test/gtest_unittest.cc#L1291

ngie-eign:
@gennadiycivil: for items like this, having clear documentation would be extremely helpful--otherwise it makes it difficult to build/use the APIs.

My stripped down example aimed to provide a simple, clear way to test with the API. My hope is that (with that information), one could make the next logical leap towards implementing something with `EXPECT_NONFATAL_FAILURE_ON_ALL_THREADS`.

PR #2132, opened 2019-02-18
Added new macros to specify override or noexcept specifiers #533 #1821

Julien-Blanc-tgcm:
* following the model of MOCK_CONST_METHOD, added MOCK_NOEXCEPT_METHOD
* also added the MOCK_VIRTUAL_METHOD, which adds the override keyword
* added all combination of CONST / NOEXCEPT / VIRTUAL where it made
sense.

### DIFF
--- a/googlemock/include/gmock/gmock-generated-function-mockers.h
+++ b/googlemock/include/gmock/gmock-generated-function-mockers.h
@@ -134,17 +134,18 @@ using internal::FunctionMocker;
     GTEST_CONCAT_TOKEN_(gmock##constness##arity##_##Method##_, __LINE__)
 
 // INTERNAL IMPLEMENTATION - DON'T USE IN USER CODE!!!
-#define GMOCK_METHOD0_(tn, constness, ct, Method, ...) \
+#define GMOCK_METHOD0_(tn, constness, noexceptness, overrideness, ct, Method, \
+    ...) \
   static_assert(0 == \
       ::testing::internal::Function<__VA_ARGS__>::ArgumentCount, \
       "MOCK_METHOD<N> must match argument count.");\
   GMOCK_RESULT_(tn, __VA_ARGS__) ct Method( \
-      ) constness { \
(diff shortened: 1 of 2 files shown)


AriehSchneier:
https://github.com/google/googletest/pull/704

AriehSchneier:
Note I already created a pull request for this a while back, which included heavily cleaning up the pump file to remove a lot of the duplication that shouldn't have been in there.
https://github.com/google/googletest/pull/704

Julien-Blanc-tgcm:
Didn't see your PR, apparently it was closed.

You did not handle noexcept, though, and i think MOCK_VIRTUAL is a better alternative than MOCK_NOOVERRIDE: it does not change the default behaviour, so it doesn't break any existing tests.

However, i could use something inspired by your logic to generate the combinatory.

gennadiycivil:
No CLA. 
In addition we are actively discouraging new macros.

AriehSchneier:
The difference is, I made it so by default every mock adds the override keyword, unless you specifically remove it (but added a flag to enable it for backwards compatibility).
Generally mocks are virtual, with only a few cases not being virtual.

Julien-Blanc-tgcm:
> Generally mocks are virtual, with only a few cases not being virtual.

While that is true, the side-effect of the change must be taken into account. Adding the override keyword by default breaks every test that is not virtual. This can means, breaking a lot of tests. Adding a new macro does not change anything for existing code, does not break anything.

> In addition we are actively discouraging new macros.

While i understand that, i doubt there can be a solution for noexcept that does not incurs the usage of new macros. IIRC exceptions are not in use at google, so this may not be an issue, but is for any user that want to mock noexcept functions. Override can be handled with pragmas, noexcept can't.

gennadiycivil:
No CLA means that we can not continue working on this, sorry.

AriehSchneier:
> Adding the override keyword by default breaks every test that is not virtual.

That is why I added a flag that you need to turn on to enable the change.

Issue #2131, opened 2019-02-18
Feature Request - Change the behaviour of a mock at runtime


Julien-Blanc-tgcm:
Currently mocks may either be Strict, Naggy or Nice. This change the outcome of the tests, regarding the presence or not of unexpected calls.

In some scenarios, you want to test an object in a particular state. During this initialization phase, the object may call a lot of methods on your mock objects, for which you don’t care (they’re part of other tests). So the mock should be nice at this point, to avoid polluting logs. However, when the actual tests run, you may want it to be strict (i do). And you may want it to be Nice again at cleanup time.

The ```::testing::XXXXUninterestingCalls``` methods could be used to switch the behaviour, but they are not exposed to the user.

I’m proposing to add a new mock type (FickleMock might be a good name), which is Naggy by default, and exposes three additional methods: ```Nice()```, ```Naggy()``` and ```Strict()``` which switches to the respective behaviour.

PR coming soon.

PR #2133, opened 2019-02-18
Added a fickle mock, whose behaviour can change at runtime #2131

Julien-Blanc-tgcm:
* can change from nice, strict or naggy, at any time, using the helper functions.

### DIFF
--- a/googlemock/include/gmock/gmock-nice-strict.h
+++ b/googlemock/include/gmock/gmock-nice-strict.h
@@ -182,6 +182,61 @@ class StrictMock : public MockClass {
   GTEST_DISALLOW_COPY_AND_ASSIGN_(StrictMock);
 };
 
+template <class MockClass>
+class FickleMock : public MockClass {
+ public:
+  FickleMock() : MockClass() {
+    ::testing::Mock::WarnUninterestingCalls(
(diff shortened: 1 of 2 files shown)


gennadiycivil:

Also, the CLA must be signed first

Julien-Blanc-tgcm:
I described it in issue #2131 .

The idea is that a test as a whole has three phases:
* initialization
* the actual test
* finalization

During initialization or finalization phase, i don't want to check anything. Usually, failure in those are detected by other tests. However, during the actual test, i want to check every side effect, so i want a strict mock (the same could be done with expect_call, but then you’d also get errors if they are called in the finalization phase).

So a test is usually written this way
* makes all mocks *nice*.
* do the initialization
* makes relevant mocks *strict*, add expect calls
* do the test
* makes all mocks nice again, so that finalization can be done correctly

Putting the object to test in the specific state i want to check can be an operation which involves a lot of operations (such as, connect to db, create some threads, etc.), i don't want a change of them to break all tests.


As for the CLA: i accept the CLA but can't sign it: it requires me to create a google account, which implies accepting a lot of conditions i don’t want to accept. If there is another way i’ll gladly sign it.

gennadiycivil:
No CLA means that we can not continue working on this, sorry.

PR #2137, opened 2019-02-19
Ignore `-Wsign-conversion` issues

ngie-eign:
clang++ compilation when `-Wsign-conversion` is currently broken and the
issues within the code are varied and widespread. For the time being
ignore `-Wsign-conversion` issues, even though some of them are valid
and bleed over into issues that would be found with
`-Wtautological-compare`, et al.

### DIFF
--- a/googletest/cmake/internal_utils.cmake
+++ b/googletest/cmake/internal_utils.cmake
@@ -78,7 +78,7 @@ macro(config_compiler_and_linker)
     # http://stackoverflow.com/questions/3232669 explains the issue.
     set(cxx_base_flags "${cxx_base_flags} -wd4702")
   elseif (CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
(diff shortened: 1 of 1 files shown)


ngie-eign:
@gennadiycivil: ping?

gennadiycivil:
@ngie-eign Feels that the "right thing to do" would be to fix the code rather than ignoring the warning.

gennadiycivil:
... But for now, yes,

ngie-eign:
> ... But for now, yes,

I'll open up an issue for this warning.

It's interesting how many downstream projects are impacted by adding `-Werror` with clang. This was one area that I definitely agree should be fixed, but it needs additional work to do so, since there are signed and sizing issues, and this will likely require making a number of macros more strict, and may require changing APIs; AFAICT they were all private APIs, but I didn't get too deep into resolving the underlying issues because of `-ferror-limit`.

Issue #2138, opened 2019-02-19
googlemock not compatible with clang -Wextra-semi


markmentovai:
Chromium is looking into building with clang’s -Wextra-semi flag. More information at [Chromium bug 926235](https://crbug.com/926235).

When attempting to build gmock tests with -Wextra-semi, two files fail to build:

googlemock/test/gmock-function-mocker_test.cc
googlemock/test/gmock-generated-function-mockers_test.cc

```
…/googletest/googlemock/test/gmock-function-mocker_test.cc:127:3: error: extra ';' inside a class [-Werror,-Wextra-semi]
  MOCK_METHOD(void, VoidReturning, (int n));  // NOLINT
  ^
…/googletest/googlemock/include/gmock/gmock-function-mocker.h:8:3: note: expanded from macro 'MOCK_METHOD'
  GMOCK_PP_VARIADIC_CALL(GMOCK_INTERNAL_MOCK_METHOD_ARG_, __VA_ARGS__)
  ^
…/googletest/googlemock/include/gmock/internal/gmock-pp.h:82:3: note: expanded from macro 'GMOCK_PP_VARIADIC_CALL'
  GMOCK_PP_CAT(_Macro, GMOCK_PP_NARG(__VA_ARGS__))(__VA_ARGS__)
  ^
…/googletest/googlemock/include/gmock/internal/gmock-pp.h:18:30: note: expanded from macro 'GMOCK_PP_CAT'
#define GMOCK_PP_CAT(_1, _2) GMOCK_PP_INTERNAL_CAT(_1, _2)
                             ^
…/googletest/googlemock/include/gmock/internal/gmock-pp.h:180:39: note: expanded from macro 'GMOCK_PP_INTERNAL_CAT'
#define GMOCK_PP_INTERNAL_CAT(_1, _2) _1##_2
                                      ^
<scratch space>:152:1: note: expanded from here
GMOCK_INTERNAL_MOCK_METHOD_ARG_3
^
…/googletest/googlemock/include/gmock/gmock-function-mocker.h:17:3: note: expanded from macro 'GMOCK_INTERNAL_MOCK_METHOD_ARG_3'
  GMOCK_INTERNAL_MOCK_METHOD_ARG_4(_Ret, _MethodName, _Args, ())
  ^
…/googletest/googlemock/include/gmock/gmock-function-mocker.h:24:42: note: expanded from macro 'GMOCK_INTERNAL_MOCK_METHOD_ARG_4'
[...]

PR #2139, opened 2019-02-19
Build cleanly with Clang -Wextra-semi

markmentovai:
For https://crbug.com/926235.
Fixes https://github.com/google/googletest/issues/2138.

### DIFF
--- a/googlemock/include/gmock/gmock-function-mocker.h
+++ b/googlemock/include/gmock/gmock-function-mocker.h
@@ -21,7 +21,7 @@
   GMOCK_INTERNAL_ASSERT_PARENTHESIS(_Spec);                                   \
   GMOCK_INTERNAL_ASSERT_VALID_SIGNATURE(                                      \
       GMOCK_PP_NARG0 _Args, GMOCK_INTERNAL_SIGNATURE(_Ret, _Args));           \
-  GMOCK_INTERNAL_ASSERT_VALID_SPEC(_Spec);                                    \
+  GMOCK_INTERNAL_ASSERT_VALID_SPEC(_Spec)                                     \
   GMOCK_INTERNAL_MOCK_METHOD_IMPL(                                            \
       GMOCK_PP_NARG0 _Args, _MethodName, GMOCK_INTERNAL_HAS_CONST(_Spec),     \
       GMOCK_INTERNAL_HAS_OVERRIDE(_Spec), GMOCK_INTERNAL_HAS_FINAL(_Spec),    \
--- a/googlemock/include/gmock/gmock-function-mocker.h
+++ b/googlemock/include/gmock/gmock-function-mocker.h
@@ -21,7 +21,7 @@
   GMOCK_INTERNAL_ASSERT_PARENTHESIS(_Spec);                                   \
   GMOCK_INTERNAL_ASSERT_VALID_SIGNATURE(                                      \
       GMOCK_PP_NARG0 _Args, GMOCK_INTERNAL_SIGNATURE(_Ret, _Args));           \
-  GMOCK_INTERNAL_ASSERT_VALID_SPEC(_Spec)                                     \
+  GMOCK_INTERNAL_ASSERT_VALID_SPEC(_Spec);                                    \
   GMOCK_INTERNAL_MOCK_METHOD_IMPL(                                            \
       GMOCK_PP_NARG0 _Args, _MethodName, GMOCK_INTERNAL_HAS_CONST(_Spec),     \
       GMOCK_INTERNAL_HAS_OVERRIDE(_Spec), GMOCK_INTERNAL_HAS_FINAL(_Spec),    \
@@ -62,7 +62,7 @@
           _N) " arguments. Parenthesize all types with unproctected commas.")
 
 #define GMOCK_INTERNAL_ASSERT_VALID_SPEC(_Spec) \
-  GMOCK_PP_FOR_EACH(GMOCK_INTERNAL_ASSERT_VALID_SPEC_ELEMENT, ~, _Spec)
+  GMOCK_PP_FOR_EACH_IN_CLASS(GMOCK_INTERNAL_ASSERT_VALID_SPEC_ELEMENT, ~, _Spec)
 
 #define GMOCK_INTERNAL_MOCK_METHOD_IMPL(_N, _MethodName, _Constness,           \
                                         _Override, _Final, _Noexcept,          \
@@ -127,7 +127,7 @@
        GMOCK_PP_HAS_COMMA(GMOCK_INTERNAL_DETECT_NOEXCEPT(_i, _, _elem)) + \
        GMOCK_INTERNAL_IS_CALLTYPE(_elem)) == 1,                           \
       GMOCK_PP_STRINGIZE(                                                 \
-          _elem) " cannot be recognized as a valid specification modifier.");
+          _elem) " cannot be recognized as a valid specification modifier.")
 
 // Modifiers implementation.
 #define GMOCK_INTERNAL_DETECT_CONST(_i, _, _elem) \
--- a/googlemock/include/gmock/internal/gmock-pp.h
+++ b/googlemock/include/gmock/internal/gmock-pp.h
@@ -160,6 +160,17 @@ static_assert(
   GMOCK_PP_CAT(GMOCK_PP_INTERNAL_FOR_EACH_IMPL_, GMOCK_PP_NARG0 _Tuple) \
   (0, _Macro, _Data, _Tuple)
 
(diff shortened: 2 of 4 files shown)


googlemock/include/gmock/gmock-function-mocker.h
```diff
@@ -21,7 +21,7 @@
   GMOCK_INTERNAL_ASSERT_PARENTHESIS(_Spec);                                   \
   GMOCK_INTERNAL_ASSERT_VALID_SIGNATURE(                                      \
       GMOCK_PP_NARG0 _Args, GMOCK_INTERNAL_SIGNATURE(_Ret, _Args));           \
-  GMOCK_INTERNAL_ASSERT_VALID_SPEC(_Spec);                                    \
+  GMOCK_INTERNAL_ASSERT_VALID_SPEC(_Spec)                                     \
```
ngie-eign: `GMOCK_INTERNAL_ASSERT_VALID_SPEC` should not have a semicolon at the end of the definition IMHO.

markmentovai: Removing the semicolon at the point of use was the “easy” way to get this to work. There’s also a “hard” way that involves duplicating some internal gtest macros in order to support this case, as distinct from other users of that macro machinery which are in different contexts.

Here’s what’s happening.

googlemock/include/gmock/gmock-function-mocker.h:64:
```
#define GMOCK_INTERNAL_ASSERT_VALID_SPEC(_Spec) \
  GMOCK_PP_FOR_EACH(GMOCK_INTERNAL_ASSERT_VALID_SPEC_ELEMENT, ~, _Spec)
```

googlemock/include/gmock/internal/gmock-pp.h:159:
```
#define GMOCK_PP_FOR_EACH(_Macro, _Data, _Tuple)                        \
  GMOCK_PP_CAT(GMOCK_PP_INTERNAL_FOR_EACH_IMPL_, GMOCK_PP_NARG0 _Tuple) \
  (0, _Macro, _Data, _Tuple)
```

For `GMOCK_PP_NARG0 _Tuple == 0`, googlemock/include/gmock/internal/gmock-pp.h:257:
```
#define GMOCK_PP_INTERNAL_FOR_EACH_IMPL_0(_i, _Macro, _Data, _Tuple)
```

So in that case, the entire `GMOCK_INTERNAL_ASSERT_VALID_SPEC` expands to nothing, and an extra dangling semicolon would be left behind.

For `GMOCK_PP_NARG0 _Tuple > 0`, the same problem would be experienced following the last element expansion. googlemock/include/gmock/internal/gmock-pp.h:258:
```
#define GMOCK_PP_INTERNAL_FOR_EACH_IMPL_1(_i, _Macro, _Data, _Tuple) \
  GMOCK_PP_INTERNAL_CALL_MACRO(_Macro, _i, _Data, GMOCK_PP_HEAD _Tuple)
#define GMOCK_PP_INTERNAL_FOR_EACH_IMPL_2(_i, _Macro, _Data, _Tuple)    \
  GMOCK_PP_INTERNAL_CALL_MACRO(_Macro, _i, _Data, GMOCK_PP_HEAD _Tuple) \
  GMOCK_PP_INTERNAL_FOR_EACH_IMPL_1(GMOCK_PP_INC(_i), _Macro, _Data,    \
                                    (GMOCK_PP_TAIL _Tuple))
```

`_Macro` here is `GMOCK_INTERNAL_ASSERT_VALID_SPEC_ELEMENT`, googlemock/include/gmock/gmock-function-mocker.h:122:
```
#define GMOCK_INTERNAL_ASSERT_VALID_SPEC_ELEMENT(_i, _, _elem)            \
  static_assert(                                                          \
      (GMOCK_PP_HAS_COMMA(GMOCK_INTERNAL_DETECT_CONST(_i, _, _elem)) +    \
       GMOCK_PP_HAS_COMMA(GMOCK_INTERNAL_DETECT_OVERRIDE(_i, _, _elem)) + \
       GMOCK_PP_HAS_COMMA(GMOCK_INTERNAL_DETECT_FINAL(_i, _, _elem)) +    \
       GMOCK_PP_HAS_COMMA(GMOCK_INTERNAL_DETECT_NOEXCEPT(_i, _, _elem)) + \
       GMOCK_INTERNAL_IS_CALLTYPE(_elem)) == 1,                           \
      GMOCK_PP_STRINGIZE(                                                 \
          _elem) " cannot be recognized as a valid specification modifier.");
```

There will always be a semicolon at the end of each element (from the text of `GMOCK_INTERNAL_ASSERT_VALID_SPEC_ELEMENT`), plus the trailing semicolon that followed `GMOCK_INTERNAL_ASSERT_VALID_SPEC` in the body of `GMOCK_INTERNAL_MOCK_METHOD_ARG_4`.

It sounds like you favor removing the reliance on the semicolon from the macro body (here, `GMOCK_INTERNAL_ASSERT_VALID_SPEC_ELEMENT`), and letting `GMOCK_PP_FOR_EACH` or a work-alike place them between expansions for each element, but not following the tail element. This is possible, but requires forking `GMOCK_PP_FOR_EACH` into another macro intended for use in class context. In between element expansions, it would insert a semicolon; for empty tuples, it’d place a harmless no-op that expects a semicolon terminator, such as `static_assert(true, "")`. `GMOCK_PP_FOR_EACH` itself can’t be modified because its other uses are in different contexts where these semicolons and static_asserts would be undesirable.

I’ve updated https://github.com/google/googletest/pull/2139 to this approach. Please let me know if you like it better.

+ @nico too.

googlemock/include/gmock/internal/gmock-pp.h
```diff
@@ -313,5 +324,65 @@ static_assert(
   GMOCK_PP_INTERNAL_CALL_MACRO(_Macro, _i, _Data, GMOCK_PP_HEAD _Tuple) \
   GMOCK_PP_INTERNAL_FOR_EACH_IMPL_14(GMOCK_PP_INC(_i), _Macro, _Data,   \
                                      (GMOCK_PP_TAIL _Tuple))
+#define GMOCK_PP_INTERNAL_FOR_EACH_IN_CLASS_IMPL_0(_i, _Macro, _Data, _Tuple) \
+  static_assert(true, "")
```
ngie-eign: Why add this definition?

markmentovai: That’s explained in [this comment](https://github.com/google/googletest/pull/2139/commits/bd1773ee7577b10a601b07f59625df36bd339e5e#diff-98584e9a9fab1f527c40de2900753506R163).

The `static_assert` is used to consume the semicolon properly. It’s conceptually similar to the `do {…} while (false)` macro construct, except intended to work in struct/class context when there’s nothing else to go in its place. `static_assert(true, "")` is simply a helpful no-op that consumes a semicolon.

@gennadiycivil, what do you think about this (bd1773ee7577) or its precursor (aef38af5348f)? If it’s easier, I can also propose it for review internally.

gennadiycivil: @markmentovai I like the "easy" ([aef38af](https://github.com/google/googletest/commit/aef38af5348f344d0778d233f9e36699265fdbce))  way but not sure if it would break existing tests. It is easier to review internally , lets go that route for now.

ngie-eign: > That’s explained in [this comment](https://github.com/google/googletest/pull/2139/commits/bd1773ee7577b10a601b07f59625df36bd339e5e#diff-98584e9a9fab1f527c40de2900753506R163).
> 
> The `static_assert` is used to consume the semicolon properly. It’s conceptually similar to the `do {…} while (false)` macro construct, except intended to work in struct/class context when there’s nothing else to go in its place. `static_assert(true, "")` is simply a helpful no-op that consumes a semicolon.
> 
> @gennadiycivil, what do you think about this ([bd1773e](https://github.com/google/googletest/commit/bd1773ee7577b10a601b07f59625df36bd339e5e)) or its precursor ([aef38af](https://github.com/google/googletest/commit/aef38af5348f344d0778d233f9e36699265fdbce))? If it’s easier, I can also propose it for review internally.

@markmentovai: Got it. Thanks for the description :).
@gennadiycivil: the catch with the easier way is that it's inconsistent with the convention of the other macro definitions; all the other GTEST macros require trailing semicolons, and many other macros I've dealt with in the open source community require trailing semicolons. Departing from this common convention results in a design wart with gmock, which will confuse implementors when they approach this API for the first time.

gennadiycivil:
I am closing this now because we accepted an expedient solution https://github.com/google/googletest/issues/2138. to help resolving an existing issue.

However @ngie-eign , your argument deserves consideration , if you feel strongly about it please open separate issue in terms of "macro definition is inconsistent with the convention of the other macro definitions" and propose a PR , we will take a look.
Thank you!

ngie-eign:
Will do @gennadiycivil!

PR #2148, opened 2019-02-21
clang: disable RTTI support with `cxx_no_rtti_flags` case for parity with g++

ngie-eign:
Add `cxx_no_rtti_flags` for clang in order to test the case with RTTI off,
similar to g++.

This increases test coverage with the non-RTTI case with clang. This something
I used in when investigating test failures on FreeBSD, as the tests that rely
are failing with googletest 1.8.1 on the OS/platform, as described in 
issue #2172.


gennadiycivil:
@ngie-eign could you please rebase/resolve conflicts

ngie-eign:
> @ngie-eign could you please rebase/resolve conflicts

@gennadiycivil: done!

EricWF:
How do we turn on C++11 today?

The fact we failed to set `cxx_no_rtti_flags` with Clang seems like a bug? Do we not have a test that catches it?

googletest/cmake/internal_utils.cmake
```diff
@@ -78,10 +78,11 @@ macro(config_compiler_and_linker)
     # http://stackoverflow.com/questions/3232669 explains the issue.
     set(cxx_base_flags "${cxx_base_flags} -wd4702")
   elseif (CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
-    set(cxx_base_flags "-Wall -Wshadow -Werror -Wno-error=sign-conversion")
+    set(cxx_base_flags "-Wall -Wshadow -Werror -Wno-error=sign-conversion -std=c++11 -frtti")
```
EricWF: Passing `-frtti` just reads weird.

RTTI is enabled by default. There is no need to specify it.

gennadiycivil:
@EricWF [C++ 11 set](https://github.com/google/googletest/search?q=-std%3Dc%2B%2B11&unscoped_q=-std%3Dc%2B%2B11). Some set in travis configs, some elsewhere, it is generally not well organized.
Given that CMake support is community-based this is probably an artifact of organic growth.

googletest/cmake/internal_utils.cmake
```diff
@@ -78,10 +78,11 @@ macro(config_compiler_and_linker)
...
     set(cxx_exception_flags "-fexceptions")
     set(cxx_no_exception_flags "-fno-exceptions")
     set(cxx_strict_flags "-W -Wpointer-arith -Wreturn-type -Wcast-qual -Wwrite-strings -Wswitch -Wunused-parameter -Wcast-align -Wchar-subscripts -Winline -Wredundant-decls")
+    set(cxx_no_rtti_flags "-fno-rtti")
```
EricWF: Why don't we need to pass `-DGTEST_HAS_RTTI=0`  like we do when using GCC?

EricWF:
@gennadiycivil It looks like we're using [`CMAKE_CXX_STANDARD`](https://github.com/google/googletest/blob/c4ccab33aaa0e86741db7c02a1d1040af76bd059/CMakeLists.txt#L9) to specify the dialect. I believe this is better practice than passing `-std=c++11` as a flag.

googletest/cmake/internal_utils.cmake
ngie-eign: > Passing `-frtti` just reads weird.
> 
> RTTI is enabled by default. There is no need to specify it.

What if a downstream project disables it, for some specific reason though? This is an attempt to be defensive by specifying the value explicitly.

googletest/cmake/internal_utils.cmake
ngie-eign: Because the compiler (clang) reports whether or not it's disabled and googletest tests for it: https://github.com/google/googletest/blob/master/googletest/include/gtest/internal/gtest-port.h#L519

ngie-eign:
> @gennadiycivil It looks like we're using [`CMAKE_CXX_STANDARD`](https://github.com/google/googletest/blob/c4ccab33aaa0e86741db7c02a1d1040af76bd059/CMakeLists.txt#L9) to specify the dialect. I believe this is better practice than passing `-std=c++11` as a flag.

Yes. I'll remove that explicit setting.

gennadiycivil:
Ping?

ngie-eign:
@gennadiycivil: pong :).

ngie-eign:
@gennadiycivil: ping?

ngie-eign:
@gennadiycivil: ping?

EricWF:
I still don't understand why we weed to pass `-frtti` explicitly. We should never have to do that. RTTI should already be enabled by default.

gennadiycivil:
I edited the comment a bit to just leave rtti , as C++11 is no longer in scope...
Could you please explain a bit more , isnt RTTI is enabled by default with clang? Why do we have to explicitly enable it.
I can see how we may want to have a disable flag, yes....

Please advise?

Thanks again
G

ngie-eign:
@EricWF: the concern I have is that downstream projects might disable RTTI, resulting in unexpected behavior when compiling with clang, or the default may change due to an upstream change or local modification. If either of these things happen, googletest may be compiled improperly, resulting in confusing behavior and test results, forcing one or more engineers to have to analyze why googletest isn’t functioning as expected.

gennadiycivil:
@ngie-eign I am not sure I agree with this premise. I don't believe googletest can presume to know what the downstream client means to when they choose disable RTTI . If the theoretical "they" chose to do this there must have been a reason.
Why do we want to force anything?

ngie-eign:
> @ngie-eign I am not sure I agree with this premise. I don't believe googletest can presume to know what the downstream client means to when they choose disable RTTI . If the theoretical "they" chose to do this there must have been a reason.
> Why do we want to force anything?

Ok, I've walked by the changes to assume the implicit `-frtti` being on behavior and have edited the commit log, etc to note this.

gennadiycivil:
Should be in now https://github.com/google/googletest/commit/5ba69d5cb93779fba14bf438dfdaf589e2b92071

ngie-eign:
> test

We should have a test that catches this. I'll post something soon to add it to Travis.

Issue #2146, opened 2019-02-21
Fix `-Wsign-conversion` issues with clang


ngie-eign:
Some downstream consumers of googletest compile with clang and use `-Wsign-conversion` (or in the case of https://github.com/jcarpent/pinocchio-ceres/issues/1, `-Wconversion`, which is a superset of `-W*conversion` warnings).

In order to unbreak their CI pipelines, `-Wsign-conversion` was explicitly ignored in #2137. It's in the project's best interest to not ignore these issues, however, so they should be fixed.

ngie-eign:
@gennadiycivil: I'll work on this.

CC: @FlashSheridan, @jcarpent

gennadiycivil:
@ngie-eign Thank you for your help with this. The -Wsign-conversion is a compiler warning that we should support.

I suggest the following:
1) Create a branch with CI configuration that includes this warning
2) Do a small change that just addresses a few cases, send this for review so we can look internally and make sure we are on the right path 
3) Once this small change has been reviewed, approved , do more changes. 

Let me know if this is OK
Thanks again
G

PR #2165, opened 2019-03-05
Proof-of-concept: fix `-Wsign-conversion` issues

ngie-eign:
This proof of concept replaces some int scalars used as loop counters and int
returning methods with size_t equivalents, as it's a viable type for
counting values and indexing arrays.

This is just a small amount of what needs to be done to make googletest
compile cleanly with `-Wsign-conversion`.

This starts to address #2146.

### DIFF
--- a/googlemock/test/gmock-spec-builders_test.cc
+++ b/googlemock/test/gmock-spec-builders_test.cc
@@ -1140,7 +1140,7 @@ TEST(UnexpectedCallTest, UnsatisifiedPrerequisites) {
   }
 
   // There should be one non-fatal failure.
-  ASSERT_EQ(1, failures.size());
+  ASSERT_EQ(1U, failures.size());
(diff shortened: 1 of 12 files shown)


EricWF:
Changing the return types of these API's from `int` to `size_t` violates the Google C++ style guide (https://google.github.io/styleguide/cppguide.html#Integer_Types).

Also, won't the `int` -> `size_t` change break `-Wsign-conversion` for users who expect it to be a signed type?

ngie-eign:
> Changing the return types of these API's from `int` to `size_t` violates the Google C++ style guide (https://google.github.io/styleguide/cppguide.html#Integer_Types).
> 
> Also, won't the `int` -> `size_t` change break `-Wsign-conversion` for users who expect it to be a signed type?

Yes.

I wasn't aware of the style guide. I will go ahead and unwind some of the proposed changes.

gennadiycivil:
I see (version 2 ) PR, should we close this one.

PR #2167, opened 2019-03-05
gcc: add braces around conditional statements

ngie-eign:
g++ 8.2 was citing issues with `-Wdangling-else`. Add additional curly
braces around conditional statements to avoid triggering these compiler
warnings.


EricWF:
I don't think this change is correct. If I understand the tests, they're trying to demonstrate that using `EXPECT_DEATH` after an `if` with no braces is allowed.

gennadiycivil:
@EricWF is 100% correct. This PR can not be considered. Thank you!

ngie-eign:
> @EricWF is 100% correct. This PR can not be considered. Thank you!

As it stands, gcc 8 warns on this issue, which will cause CI failures if/when using a newer version of g++ (I found this issue with g++ 8.2).

I'll look into fixing the macro so the compiler no longer finds this to be an issue.

Issue #2122, opened 2019-02-13
misspelling may happen when using “SetUpTestCase” or “TearDownTestCase”


rezaebrh:
If we misspell for example `SetUpTestcase `instead of `SetUpTestCase`, no error/warning will be shown.
But this mistake may happen frequently. Although for `SetUp `and `TearDown `(Because they are virtual) we can use `override `keyword and any misspell will cause error but for `SetUpTestCase `and `TearDownTestCase `there is no similar solution.

gennadiycivil:
Thank you very much for this report. The best way to approach this would be to create a proper PR and submit it for consideration.

gennadiycivil:
Closing

rezaebrh:
@gennadiycivil Thanks

PR #2175, opened 2019-03-14
misspelling may happen when using “SetUpTestCase” or “TearDownTestCase”

rezaebrh:
If we misspell for example `SetUpTestcase `instead of `SetUpTestCase`, no error/warning will be shown.
But this mistake may happen frequently. Although for `SetUp `and `TearDown `(Because they are virtual) we can use override keyword and any misspell will cause error but for `SetUpTestCase `and `TearDownTestCase `there is no similar solution.

I've created this PR based on the issue I had reported and its response:
[https://github.com/google/googletest/issues/2122](url)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,6 +35,7 @@ googletest/m4/ltoptions.m4
 googletest/m4/ltsugar.m4
 googletest/m4/ltversion.m4
 googletest/m4/lt~obsolete.m4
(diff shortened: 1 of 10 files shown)


gennadiycivil:
No CLA

gennadiycivil:
Content of this PR does not correspond to the description

Issue #247, opened 2015-07-28
Fatal failure in SetUpTestCase() function should skip running tests.


GoogleCodeExporter:
```
I think googletest should skip running tests (or automatically mark as
failed) in a particular fixture if we call SetUpTestCase() function
and there is fatal failure in it. Suppose we create some shared object
in SetUpTestCase() function and try to access that in TEST_F. But the
failure to create that object in SetUpTestCase() cause to access NULL
inside TEST_F.

So if we make fatal assertion in SetUpTestCase() itself, then above
problem can be avoided.


```

Original issue reported on code.google.com by `pvshew...@gmail.com` on 11 Jan 2010 at 6:39

hadrielk:
You can "fix" this yourself by using a test event listener, as described in the advanced user guide, by doing something like this in your derived test event listener class (which you would register in your main()):

```
class TestEventListener : public ::testing::EmptyTestEventListener
{
    // Fired before the test case starts and before SetUpTestCase() is invoked.
    virtual void OnTestCaseStart(const ::testing::TestCase& test_case) override
    {
        firstTestInTestCase = true;
        testCaseHadFailure = false;
    }

    // Fired before each individual test starts: before the test fixture is constructed
    // and SetUp() is invoked.
    virtual void OnTestStart(const ::testing::TestInfo& info) override
    {
        if (firstTestInTestCase) {
            size_t numFailures = numFatalFailures();

            if (numFailures > previousNumFatalFailures) {
                testCaseHadFailure = true;
                previousNumFatalFailures = numFailures;
            }

            firstTestInTestCase = false;
        }

        if (testCaseHadFailure) {
            // this will avoid running the test fixture's SetUp() and
            // TearDown() and TEST()/TEST_F(), but the test fixtures
            // constructor is still invoked (that's unavoidable)
            FAIL() << "The test case's SetUpTestCase() or executable had a fatal failure";
        }
    }

    // gets the number of fatal failures in the UnitTest's ad_hoc_test_result,
    // which is the number of fatal failures other than those inside of test
    // cases; fatal failures in SetUpTestCase() also count in this, but those
    // in SetUp(), TearDown(), and TEST()/TEST_F() do not.
    size_t numFatalFailures()
    {
        const auto& result = ::testing::UnitTest::GetInstance()->ad_hoc_test_result();
        const int numParts = result.total_part_count();
        size_t failures = 0;

        for (int i=0; i < numParts; ++i) {
[...]

PR #2177, opened 2019-03-15
when an error is thrown in SetUpTestCase() do not catch it

henriksjostrom:
This means that uncaught exceptions in SetUpTestCase() aborts the test execution to avoid running tests with faulty setup.

closes https://github.com/google/googletest/issues/247

### DIFF
--- a/googletest/src/gtest.cc
+++ b/googletest/src/gtest.cc
@@ -2454,7 +2454,7 @@ Result HandleExceptionsInMethodIfSupported(
   // throw statement in the code under test.  For this reason, we perform
   // the check early, sacrificing the ability to affect Google Test's
   // exception handling in the method where the exception is thrown.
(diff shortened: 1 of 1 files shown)


gennadiycivil:
Thank you for this. .How will this work when C++ Exceptions are no avaialble?

ngie-eign:
> Thank you for this. .How will this work when C++ Exceptions are no avaialble?

`HandleExceptionsInMethodIfSupported(..)` looks like it's specifically for exception handling..?

gennadiycivil:
So... *what* we are doing this is "Abort execution in case of unsuccessful SetUpTestCase "

And "how" we chose to implement it is by putting a special case into somewhat related function. 
NOTE: The function this is addressing SetUpTestCase is actually deprecated. There is a newer one called SetUpTestSuite.

There should be a better way to address the root cause. In this form this can not be accepted but I would welcome a discussion on how to do this right.

PR #2192, opened 2019-03-21
Printing nested Cxx Exceptions

### LINKED ISSUE #2186: the way (nested) exceptions are printed
Hi there,

`HandleExceptionsInMethodIfSupported` only prints the last thrown exception `e.what()` (formatted in `FormatCxxExceptionMessage`).

If the exception was nested (*i.e.* thrown with `std::throw_with_nested`) it would be great to unwrap and print all nested exceptions to get a useful message (using `std::rethrow_if_nested` like presented in https://en.cppreference.com/w/cpp/error/throw_with_nested).

Otherwise it would be great to be able to override `HandleExceptionsInMethodIfSupported` :D

matthieu-nesme:
Closes #2186 

- unwrap and print nested cxx exceptions

### DIFF
--- a/googletest/src/gtest.cc
+++ b/googletest/src/gtest.cc
@@ -2370,14 +2370,36 @@ namespace internal {
 
 #if GTEST_HAS_EXCEPTIONS
 
+// unwrap and print nested exceptions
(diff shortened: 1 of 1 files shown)


gennadiycivil:
No CLA
Builds fail

PR #2194, opened 2019-03-21
Add HWASan annotations.

eugenis:
These mirror existing ASan annotations.
This change is purely speculative, but the tools are very similar and we
expect to see the same issues with HWASan that these annotations solve
for ASan.

### DIFF
--- a/googletest/include/gtest/internal/gtest-port.h
+++ b/googletest/include/gtest/internal/gtest-port.h
@@ -819,6 +819,18 @@ typedef struct _RTL_CRITICAL_SECTION GTEST_CRITICAL_SECTION;
 # define GTEST_ATTRIBUTE_NO_SANITIZE_ADDRESS_
 #endif  // __clang__
 
+// A function level attribute to disable HWAddressSanitizer instrumentation.
+#if defined(__clang__)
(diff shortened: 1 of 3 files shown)


gennadiycivil:

eugenis:
Sure.
HWASan uses memory (address) tagging to detect memory errors:
https://clang.llvm.org/docs/HardwareAssistedAddressSanitizerDesign.html

It inserts a random tag in the MSB of heap and stack allocation addresses. This tag dominates pointer comparison in StackGrowsDown(), making the result non-deterministic, and entirely unrelated to the actual stack growth direction. The function attribute disables this behavior.

The annotations in gtest-printers are there because the printers are used to basically dump memory. The sanitizers may have ideas why this memory should not be accessed, and that is counter productive. In particular, the test may access only part of an array, but in case of a test failure gtest will dump the entire array which may contain uninitialized bytes - that's what SANITIZE_MEMORY annotation is for. There are similar reasons for ADDRESS and THREAD annotations. HWADDRESS in its current implementation can not cause issues there, I believe, but it falls under the same umbrella of tools whose checking should not apply to test printers because it is not the code under test.

gennadiycivil:
will work internally, closing for now

Issue #2195, opened 2019-03-25
Death tests fail in a macOS sandbox


wilhuff:
I'm working on porting a system that works on iOS and Linux to macOS (firebase/firebase-ios-sdk#434). We've written some death tests that work on Linux but are skipped on iOS because that platform doesn't support the mechanism. Now that we're trying macOS these tests are enabled, but failing.

Our system still has considerable Objective-C code, so on iOS and macOS we're running Googletest via XCTest, Apple's testing framework for Objective-C. We've written a [bridge](https://github.com/firebase/firebase-ios-sdk/blob/master/Firestore/core/test/firebase/firestore/FSTGoogleTestTests.mm) that allows our Objective-C and C++ tests to run together in one pass under Xcode.

This works swimmingly on iOS, but on macOS this ends up running XCTest-based tests in an app sandbox, where it fails because `/tmp` isn't writeable from a sandbox. In this configuration the `mkstemp` in [gtest-port.cc at line 1115](https://github.com/google/googletest/blob/master/googletest/src/gtest-port.cc#L1115), fails with `errno` = 1 (Operation not permitted).

Apparently the right thing to do is use the value of `TMPDIR` from the environment, which will point somewhere else. On my host on one run its value looked like `/var/folders/60/8r_t5wts5rn8bqz2l90c6ncm00380s/T/`. Adding some logic to compute the temporary directory prior to calling `mkstemp` allows the death tests to succeed.

I've cooked up a PR to fix this and will submit it shortly.

SoylentGraham:
I have just encountered a seemingly similar/same problem with a sandboxed app (crashing as `TEST` globals initialise)

I'm using 391ce627def20c1e8a54d10b12949b15086473dd and the macro has changed...
`#ifdef GTEST_HAS_DEATH_TEST` is now...
https://github.com/google/googletest/blob/main/googletest/src/gtest-internal-inl.h#L683

But reading the 2019 change above, and the current code... it feels like that the fix was maybe lost over the years?
Or is it a different problem

```
 if (original_working_dir_.IsEmpty()) {
      original_working_dir_.Set(FilePath::GetCurrentDir());
      GTEST_CHECK_(!original_working_dir_.IsEmpty())
          << "Failed to get the current working directory.";
    }
```
This GTEST_CHECK fails (Unfortuntely I can't see the contents of `original_working_dir`)

[...]

PR #2196, opened 2019-03-25
On macOS, use $TMPDIR for temporary files

wilhuff:
On macOS when running in an App Sandbox, /tmp isn't writeable, so ask
the system where to place temporary files.

Fixes google/googletest#2195.


gennadiycivil:
Closing here, working downstream

Issue #2189, opened 2019-03-20
GTEST_SKIP doesn't work in test environments


asomers:
GTEST_SKIP() works great when used in fixtures.  But it doesn't seem to have any effect when used in a test environment.  I would expect that every test case in the program gets skipped if an environment calls GTEST_SKIP().

Example program:
```c++
#include "gtest/gtest.h"

using ::testing::Test;

class SkipEnvironment : public testing::Environment {
 public:
  void SetUp() override {
    GTEST_SKIP() << "Skipping the entire environment";
  }
};

TEST(SkipTest, ShouldBeSkippedByEnvironment) {
  EXPECT_EQ(0, 1);
}

int main(int argc, char **argv) {
  testing::InitGoogleTest(&argc, argv);

  SkipEnvironment* const env = new SkipEnvironment;
  testing::AddGlobalTestEnvironment(env);

  return (RUN_ALL_TESTS());
```
Actual output:
```
Executing tests from //googletest/test:gtest_skip_environment_test
-----------------------------------------------------------------------------
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from SkipTest
[ RUN      ] SkipTest.ShouldBeSkippedByEnvironment
googletest/test/gtest_skip_environment_test.cc:16: Failure
Expected equality of these values:
  0
  1
[  FAILED  ] SkipTest.ShouldBeSkippedByEnvironment (0 ms)
[----------] 1 test from SkipTest (0 ms total)
  
[----------] Global test environment tear-down
[==========] 1 test from 1 test suite ran. (0 ms total)
[  PASSED  ] 0 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] SkipTest.ShouldBeSkippedByEnvironment

 1 FAILED TEST
```

cc @ngie-eign

ngie-eign:
This seems like a subset of the issue I reported in #2169.

ngie-eign:
Ok, interesting note. If we used GTEST_FAIL, it would bail out here. Let's give this a shot...

```
$ ./setup_environment_fatal_fail_demo 
[==========] Running 1 test from 1 test case.
[----------] Global test environment set-up.
setup_environment_fatal_fail_demo.cc:8: Failure
Failed
Calling a fatal failure
[----------] Global test environment tear-down
[==========] 1 test from 1 test case ran. (0 ms total)
[  PASSED  ] 1 test.
[  FAILED  ] 0 tests, listed below:

[...]

PR #2203, opened 2019-03-30
Handle GTEST_SKIP() when calling `Environment::SetUp()`

ngie-eign:
gtest prior to this change would completely ignore `GTEST_SKIP()` if
called in `Environment::SetUp()`, instead of bailing out early, unlike
`Test::SetUp()`, which would cause the tests themselves to be skipped.
The only way (prior to this change) to skip the tests would be to
trigger a fatal error via `GTEST_FAIL()`.

Desirable behavior, in this case, when dealing with
`Environment::SetUp()` is to check for prerequisites on a system
(example, kernel supports a particular featureset, e.g., capsicum), and
skip the tests. The alternatives prior to this change would be
undesirable:

- Failing sends the wrong message to the test user, as the result of the
  tests is indeterminate, not failed.
- Having to add per-test class abstractions that override `SetUp()` to
  test for the capsicum feature set, then skip all of the tests in their
  respective SetUp fixtures, would be a lot of human and computational
  work; checking for the feature would need to be done for all of the
  tests, instead of once for all of the tests.

For those reasons, making `Environment::SetUp()` handle `GTEST_SKIP()`,
by not executing the testcases, is the most desirable solution.

In order to properly diagnose what happened when running the tests if
they are skipped, print out the diagnostics in an ad hoc manner.

Update the documentation to note this change and integrate a new test,
gtest_skip_in_environment_setup_test, into the test suite.

This change addresses #2189.

### DIFF
--- a/googletest/CMakeLists.txt
+++ b/googletest/CMakeLists.txt
@@ -222,6 +222,7 @@ $env:Path = \"$project_bin;$env:Path\"
     test/gtest-typed-test2_test.cc)
   cxx_test(gtest_unittest gtest_main)
   cxx_test(gtest-unittest-api_test gtest)
+  cxx_test(gtest_skip_in_environment_setup_test gtest_main)
   cxx_test(gtest_skip_test gtest_main)
 
   ############################################################
--- a/googletest/Makefile.am
+++ b/googletest/Makefile.am
@@ -283,6 +283,12 @@ test_gtest_all_test_SOURCES = test/gtest_all_test.cc
 test_gtest_all_test_LDADD = lib/libgtest_main.la \
                             lib/libgtest.la
 
+TESTS += test/gtest_skip_in_environment_setup_test
+check_PROGRAMS += test/gtest_skip_in_environment_setup_test
+test_gtest_skip_in_environment_setup_test_SOURCES = test/gtest_skip_in_environment_setup_test.cc
+test_gtest_skip_in_environment_setup_test_LDADD= lib/libgtest_main.la \
+                                                 lib/libgtest.la
+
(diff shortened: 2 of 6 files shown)


ngie-eign:
> Works for me!

I'll proceed with http://reviews.freebsd.org/D19765, adjusting the code if need be later based on review from the maintainers.

emaste:
For reference, now committed to FreeBSD's in-tree googletest: https://svnweb.freebsd.org/changeset/base/345770

gennadiycivil:
Looking, may tweak the tests a little.

gennadiycivil:
@ngie-eign somehow he bot is not able to verify the CLA.  Could you please confirm that all commits are authored by you and you are ok with Google CLA?
Thanks

ngie-eign:
Hi @gennadiycivil! Yes, I have one commit attached to this PR, and it’s signed by me... hmm... is the bot working?

ngie-eign:
Oh, I read the second paragraph... it says that “it was authored by someone other than the PR submitter or a collaborator”. I think it’s getting cranky about your commit to merge the change to the branch...?

ngie-eign:
Just to rule out one possibility, I removed my middle name from my signed CLA (just to see if it’s doing strict author matches).

gennadiycivil:
Oh look at that, it found you now :-) Thank you

gennadiycivil:
Sorry for all back and forth, I had some issues on the back end... should be good now

gennadiycivil:
This code should be in now as of https://github.com/google/googletest/commit/1f3b098ef3afb1d186778d47ccecf9d188ef49b2.

PR #2206, opened 2019-03-31
clang: enable typed tests on non-GNU/Linux platforms

ngie-eign:
clang also supports variadic macros. Thus, the fact that `GTEST_HAS_TYPED_TEST`
and `GTEST_HAS_TYPED_TEST_P` are not defined in `gtest-port.h` results
in failures to compile tests which are typed using clang.

Check for clang and define the constants to 1 when using them, as the
compiler supports those tests.

This enables typed tests on FreeBSD and OSX (for instance) when using clang.

### DIFF
--- a/googletest/include/gtest/internal/gtest-port.h
+++ b/googletest/include/gtest/internal/gtest-port.h
@@ -619,10 +619,10 @@ typedef struct _RTL_CRITICAL_SECTION GTEST_CRITICAL_SECTION;
 
 // Determines whether to support type-driven tests.
 
-// Typed tests need <typeinfo> and variadic macros, which GCC, VC++ 8.0,
+// Typed tests need <typeinfo> and variadic macros, which clang, GCC, VC++ 8.0,
 // Sun Pro CC, IBM Visual Age, and HP aCC support.
-#if defined(__GNUC__) || defined(_MSC_VER) || defined(__SUNPRO_CC) || \
-    defined(__IBMCPP__) || defined(__HP_aCC)
+#if defined(__clang__) || defined(__GNUC__) || defined(_MSC_VER) || \
+    defined(__SUNPRO_CC) || defined(__IBMCPP__) || defined(__HP_aCC)
 # define GTEST_HAS_TYPED_TEST 1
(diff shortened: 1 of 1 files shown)


ngie-eign:
@gennadiycivil: please merge this change before the next release. I found it really annoying how clang didn't support these tests out of the box on FreeBSD; it works on Linux because it checks for `__GNUC__`, which is a misleading, at best.

ngie-eign:
CC: @asomers

ngie-eign:
Ok, I'm hilariously confused and amused at the same time.

Looking online, clang++ does also define `__GNUC__` as a preprocessor macro... so why isn't this working today with googletest on FreeBSD?

```
$ echo "" | clang++ -dM -E -xc++ -c - | grep GNU                                                                                                                
#define __GNUC_GNU_INLINE__ 1
#define __GNUC_MINOR__ 2
#define __GNUC_PATCHLEVEL__ 1
#define __GNUC__ 4
#define __GNUG__ 4
$ uname -a
FreeBSD pinklady-fbsd-current.local 13.0-CURRENT FreeBSD 13.0-CURRENT r344999+4cb64feca8a9(projects/import-googletest-1.8.1) GENERIC-NODEBUG  amd64
```

```
pinklady:~ ngie$ echo "" | clang++ -dM -E -xc++ -c - | grep GNU
#define __GNUC_GNU_INLINE__ 1
#define __GNUC_MINOR__ 2
#define __GNUC_PATCHLEVEL__ 1
#define __GNUC__ 4
#define __GNUG__ 4
pinklady:spark ngie$ uname -a
Darwin pinklady.local 18.2.0 Darwin Kernel Version 18.2.0: Thu Dec 20 20:46:53 PST 2018; root:xnu-4903.241.1~1/RELEASE_X86_64 x86_64
```

ngie-eign:
Oh.... I have an idea.

Clang does some fun stuff with macros if they aren't defined. I wonder if that's what happened with the port version of clang.

Let me futz around a bit more and see whether or not this occurs on master.

ngie-eign:
Ok, this is unfortunately annoying. It works on master, but not the 1.8.1 installed from ports :/.

Issue #296, opened 2015-07-28
[PATCH] Add Haiku support, adapted from llvm gtest


GoogleCodeExporter:
```
This patch adds support for Haiku.
It only adds the haiku detection preprocessor flag and adds it to the checks 
for certain features in gtest-port.h
I do not believe it touches anything that would require the contributor license 
agreement.

```

Original issue reported on code.google.com by `plmdvy@gmail.com` on 23 Jun 2010 at 8:07
- Merged into: #294

Attachments:
- [gtest.diff](https://storage.googleapis.com/google-code-attachments/googletest/issue-296/comment-0/gtest.diff)

Issue #294, opened 2015-07-28
patch to compile on Haiku


GoogleCodeExporter:
```
Patch to compile on Haiku operating system.
```

Original issue reported on code.google.com by `grzegorz...@gmail.com` on 17 Jun 2010 at 6:59

Attachments:
- [googletest-haiku.patch](https://storage.googleapis.com/google-code-attachments/googletest/issue-294/comment-0/googletest-haiku.patch)

PR #2229, opened 2019-04-11
Add Haiku platform support.

return:
This allows googletest to recognize the Haiku operating system when
running tests in other projects.

This patch also supersedes related patches #296 and #294 to enable death tests support which is required for building V8.


return:
I signed it!

Issue #2124, opened 2019-02-13
CMake variable GOOGLETEST_VERSION is fixed, not computed from non-CMake-specific information


fidergo-stephane-gourichon:
Opening issue as approved by maintainer on #2089 .

# Current state

CMake variable GOOGLETEST_VERSION is fixed, not computed from non-CMake-specific information, see line below:

https://github.com/google/googletest/blob/ea43be9d1a781b73dbd573954ee4a371c233a583/CMakeLists.txt#L4

# Consequence

As a consequence, whenever googletest version number changes, this file has to be changed, too.

This would be a non-issue if that line was the ["single, unambiguous, authoritative representation"](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself).

The reality is googletest maintainers do not use CMake at all. CMake is completely community supported for googletest. bazel is main release mechanism.

If we do nothing about this, people building using CMake will get wrong version information as soon as version changes, until someone fixes that each time.

# Solution

It is possible in CMake to automatically compute the value of the variable, based on `git describe`.
For example, this answer: https://stackoverflow.com/a/38836941

Best solutions are those that keep maximum portability (avoid making unnecessary assumptions about CMake-time platform).

As a side note, it would be best to add some comments on top of the CMakeLists.txt to make clear that bazel is primary build system and CMake is community supported.

I may do this some time in the future, but don't hold your breath. Anyone can offer a PR, too.

xilaraux:
Can I help with that?

ngie-eign:
This is true and gets more annoying when dealing with autotools-based builds as well.

It would be nice if these things were consistent.

gennadiycivil:
https://github.com/google/googletest/commit/703bd9caab50b139428cea1aaff9974ebee5742e

PR #2242, opened 2019-04-26
Fix compilation on CentOS 7

SylvestreG:
Hi here is a fix to allow google test cmake to works on CentOS 7. On this cmake VERSION was not mendatory.

user > uname -a 
Linux localhost.localdomain 3.10.0-957.10.1.el7.x86_64 #1 SMP Mon Mar 18 15:06:45 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
user > cmake -v
cmake version 2.8.12.2

Best Regards,


SylvestreG:
I signed it!

gennadiycivil:

SylvestreG:
HI Gennadiy,

I was trying to build googletest on CentOS 7 using cmake. But It does not work, it produce an error.

An issue seems to be created : https://github.com/google/googletest/issues/1786 but If I am right it was not fixed on the sourcetree.

My PR fix this issue and allow people to use googletest on CentOS7.

Best Regards,

gennadiycivil:
Also related, https://github.com/google/googletest/issues/2124

PR #2253, opened 2019-05-11
Feature: Add support for google unit tests to register tests dynamically

tapika:
This change allows simple dynamic test registration for google unit test.

Sample code how it can be used can be found here: 

https://github.com/tapika/cppscriptcore/blob/8d29136f86e3ff874d21cc76014c852f3f1c9fda/SolutionProjectModel/cppexec/cppexec.cpp#L144

This change:

https://github.com/tapika/googletest/blob/ae45849962424c0353cc01d0ea873ccff1dc4163/googletest/src/gtest.cc#L5563

Allows also for test application to provide source code position information back to caller, using special `<loc>` syntax. Currently Google test adapter does not supports it, my own fork does.

### DIFF
--- a/googletest/include/gtest/gtest.h
+++ b/googletest/include/gtest/gtest.h
@@ -485,7 +485,7 @@ class GTEST_API_ Test {
 
   // Deletes self.  We deliberately pick an unusual name for this
   // internal method to avoid clashing with names used in user TESTs.
-  void DeleteSelf_() { delete this; }
+  virtual void DeleteSelf_() { delete this; }
 
   const std::unique_ptr<GTEST_FLAG_SAVER_> gtest_flag_saver_;
 
@@ -683,7 +683,7 @@ class GTEST_API_ TestInfo {
  public:
   // Destructs a TestInfo object.  This function is not virtual, so
   // don't inherit from TestInfo.
-  ~TestInfo();
+  virtual ~TestInfo();
 
   // Returns the test suite name.
   const char* test_suite_name() const { return test_suite_name_.c_str(); }
@@ -747,7 +747,7 @@ class GTEST_API_ TestInfo {
   // Returns the result of the test.
   const TestResult* result() const { return &result_; }
 
- private:
+ protected:
 #if GTEST_HAS_DEATH_TEST
   friend class internal::DefaultDeathTestFactory;
 #endif  // GTEST_HAS_DEATH_TEST
@@ -801,8 +801,9 @@ class GTEST_API_ TestInfo {
   bool matches_filter_;             // True if this test matches the
                                     // user-specified filter.
   bool is_in_another_shard_;        // Will be run in another shard.
-  internal::TestFactoryBase* const factory_;  // The factory that creates
+  internal::TestFactoryBase* factory_;        // The factory that creates
                                               // the test object
+  bool ownFactory_;                 // delete factory_ on class termination
 
   // This field is mutable and needs to be reset before running the
   // test for the second time.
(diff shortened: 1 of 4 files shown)


tapika:
> please reply here (e.g. I signed it!) and we'll verify it.

Should be signed now.

tapika:
I have also more changes, but let's see how this pull will go inside.

See also https://stackoverflow.com/a/56097374/2338477
for new class `ScriptBasedTestInfo` (which is derived from `DynamicTestInfo`).

At the moment my changes to gtest resides in https://github.com/tapika/cppscriptcore repository.

gennadiycivil:
CI fails. 
Also googletest already provides a way to dynamically register a test.

gennadiycivil:
ping?

tapika:
> ping?

pong. 

Please let me know on which environment I need to test my changes. Currently I have tried to mimic build command lines for cmake and via my last commit compilation on my pc seems to be working, there is however linux side as well, but haven't tested that one as environment is not available to me.

tapika:
wow, I even managed to fix it. :)

gennadiycivil:
Thank you again for this contribution. Could you please elaborate and provide more information on exactly the issue / problems this is trying to address.
How is this change better than the existing functionality? 
The only functional change that is apparent to me is "ability to specify source code location" . If this is the case I would expect to see a much smaller tweak into the existing code base.

tapika:
> Thank you again for this contribution. Could you please elaborate and provide more information on 
> exactly the issue / problems this is trying to address.
> What is the before/after effect this would create and how is this beneficial?
> How is this change better than the existing functionality?
> The only functional change that is apparent to me is "ability to specify source code location" . If this is > the case I would expect to see a much smaller tweak into the existing code base.

There was two changes, maybe start from more breaking

1. When listing google tests - test application by itself specifies location in source code where particular test resides. This happens by providing additional "{space}{space}<loc>{path}({line})"
printing. This is breaking change, older application might detect this as additional test.

Test location resolving typically happened on application which tried to display tests - for example in Google Test Adapter (shortly as GTA, Visual Studio extension). GTA used debug information to extract location of test, which in a turn required project to have special compilation ("Generate Debug Information" > "Generate Debug Information optimized for sharing and publishing (/DEBUG:FULL)")

With my change "Genedate Debug Information" project change is not required anymore.

2. For dynamically added tests it was not possible to
2a. specify test location
2b. it was not easy to map dynamically executed test to which file it corresponded to.

2a. can specify test location to be any source code, even run-time compiled. (At test discovery phase we might not have c++ code compiled as .dll, application will compile it when test will be executed)

2b. It's now also possible to inherit from DynamicTestInfo and have more precise meta data associated with given test. For example, I can place full source code path and restore it back during testing.

https://stackoverflow.com/questions/19160244/create-tests-at-run-time-google-test/56097374#56097374

Example use here:
https://github.com/tapika/cppscriptcore/blob/master/SolutionProjectModel/cppexec/cppexec.cpp#L148

Getting class pointer back during testing.

gennadiycivil:
Unfortunately there is still not enough to convince us that this change is needed to be made in this form

I still only see one new functional change - "ability to specify source code location" . 

While we are not objecting to this change, the implementation should not require so many changes and should not be backwards-incompatible.

I am closing this PR, please feel free to open a new one that would propose the same change with smaller backward-compatible implementation

tapika:
> should not be backwards-incompatible

It's not possible to be backwards compatible - to be able to specify source code location, you need mechanism to tell about source code location to Google test adapter. Of course this could be made in backwards compatible manner, but I suspect that instead of using plain text to list tests you will need xml or json to pass through required information. Going in that direction might give more complex solution.

PR #2254, opened 2019-05-12
Use std::addressof to instead of plain '&'.

chaoran:
Otherwise the code won't compile if the '&' operater is overloaded and
return something that cannot be casted to void *.

### DIFF
--- a/googletest/include/gtest/gtest-printers.h
+++ b/googletest/include/gtest/gtest-printers.h
@@ -152,8 +152,8 @@ class TypeWithoutFormatter {
   // This default version is called when kTypeKind is kOtherType.
   static void PrintValue(const T& value, ::std::ostream* os) {
     PrintBytesInObjectTo(static_cast<const unsigned char*>(
-                             reinterpret_cast<const void*>(&value)),
-                         sizeof(value), os);
+                             reinterpret_cast<const void*>(
+                                 std::addressof(value))), sizeof(value), os);
   }
 };
 
--- a/googletest/test/googletest-overload-addressof-operator-test.cc
+++ b/googletest/test/googletest-overload-addressof-operator-test.cc
@@ -1,0 +1,91 @@
+// Copyright 2008, Google Inc.
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are
+// met:
+//
+//     * Redistributions of source code must retain the above copyright
+// notice, this list of conditions and the following disclaimer.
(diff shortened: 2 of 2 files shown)


chaoran:
I signed it!

gennadiycivil:
Thank your for this contribution. Could you please elaborate and provide more information , do you have a concrete example?
The theoretical issue is clear however we have many years of usage and I am not aware of any complaints.
Do you have a concrete example?

chaoran:
Here is a minimal example of what I'm trying to do. Basically I want to create a synthetic pointer class that can be used as an iterator over a buffer of fixed size strings. And I created a synthetic reference class, which has the overloaded `operator&` to create such pointers.
```c++
#include <gtest/gtest.h>

class string_ref;

/**
 * This is a synthetic pointer to a fixed size string.
 */
class string_ptr {
  public:
    string_ptr(const char *data, size_t size) : data_(data), size_(size) {}

    string_ptr &operator++() noexcept {
        data_ += size_;
        return *this;
    }

    string_ref operator*() const noexcept;

  private:
    const char *data_;
    size_t size_;
};

/**
 * This is a synthetic reference of a fixed size string.
 */
class string_ref {
  public:
    string_ref(const char *data, size_t size) : data_(data), size_(size) {}

    string_ptr operator&() const noexcept { return {data_, size_}; }

    bool operator==(const char *s) const noexcept {
        if (size_ > 0 && data_[size_ - 1] != 0) {
            return std::string_view(data_, size_) == std::string_view(s);
        } else {
            return std::string_view(data_) == std::string_view(s);
        }
    }

  private:
    const char *data_;
    size_t size_;
};

string_ref string_ptr::operator*() const noexcept {
    return {data_, size_};
}

TEST(string_ref, compare) {
    const char *s = "alex\0davidjohn\0";
    string_ptr ptr(s, 5);
    // EXPECT_EQ(*ptr, "alex"); // I wish I can do this, but it won't compile
    EXPECT_TRUE(*ptr == "alex");
    ++ptr;
    // EXPECT_EQ(*ptr, "david");
    EXPECT_TRUE(*ptr == "david");
    ++ptr;
    // EXPECT_EQ(*ptr, "john");
    EXPECT_TRUE(*ptr == "john");
}
```

gennadiycivil:
@chaoran Thank you for this example. Could you please also add a unit test and we can proceed after that.
Thanks again
Gennadiy

chaoran:
Unit test added.

gennadiycivil:
248165482

gennadiycivil:
note that we moved the code around a bit not to create entire new file.

PR #2260, opened 2019-05-22
unbreak windows build

741g:
windows msvc toolchain with werror and wconversion
will break if converting long to DWORD.


gennadiycivil:
@741g Thank you for this contribution. Could you also find the appropriate appveyor build with the  werror and wconversion so we can make sure this stays fixed.

gennadiycivil:
@ngie-eign any comments?

741g:
> @741g Thank you for this contribution. Could you also find the appropriate appveyor build with the werror and wconversion so we can make sure this stays fixed.

How do I find the particular build? It's possible our setup (that got broken) might not be part of the current set of build targets, or the configuration doesnt overlap.

ngie-eign:
@gennadiycivil, @741g: LGTM!

gennadiycivil:
249462159

PR #2266, opened 2019-05-30
Make gtest_main testonly in Bazel

m3rcuriel:
`gtest_main` should not be a non-test dependency of anyone. Even though googletest is just a library I feel like that's reasonable to enforce on customers.

Also it prevents non-testonly libraries from erroneously pulling it in and using the wrong main or causing link errors.


gennadiycivil:
Thank you for this. We disagree,

PR #2268, opened 2019-06-04
Fuse gtest-spi.h into the resulting header.

kalaxy:
`gtest-spi.h` is what contains the [testing utilities for expanding](https://github.com/google/googletest/blob/master/googletest/docs/advanced.md#catching-failures) googletest.  Currently it is fused into the source file, but this prevents users of the fused files from using those utilities directly.

I changed the fused script to put it in the header instead.  Let me know what you think!

### DIFF
--- a/googletest/scripts/fuse_gtest_files.py
+++ b/googletest/scripts/fuse_gtest_files.py
@@ -172,6 +172,7 @@ def FuseGTestH(gtest_root, output_dir):
         output_file.write(line)
 
   ProcessFile(GTEST_H_SEED)
+  ProcessFile(GTEST_SPI_H_SEED)
   output_file.close()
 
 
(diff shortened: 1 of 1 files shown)


kalaxy:
I signed it!

gennadiycivil:
What is the before/after effect this would create and how is this beneficial? I understand the general explanation however it would be helpful to understand your use case how it does not work 'before' and works 'afer'

kalaxy:
Yeah, I'll try to err on the side of too much information, but let me know if I've missed what you are looking for.

Our team uses the fused google test files in our project.  We've also developed some of our own extensions to google test by way of `{ASSERT_,EXPECT_}PRED_FORMAT2` as described in [Predicate Assertions for Better Error Messages](https://github.com/google/googletest/blob/master/googletest/docs/advanced.md#predicate-assertions-for-better-error-messages).  In order to test our extensions we want to write tests using the infrastructure described in the ["Catching" Failures](https://github.com/google/googletest/blob/master/googletest/docs/advanced.md#catching-failures) section.  

That test infrastructure is contained in `gunit-spi.h`.  However, the script for fusing the files together puts `gunit-spi.h` in the fused .cc file instead of the header file.  So using just the fused files as is, we don't have access to the ["Catching" Failures](https://github.com/google/googletest/blob/master/googletest/docs/advanced.md#catching-failures) testing utilities.

This change to the fuse script puts `gunit-spi.h` in the fused header instead of the cpp.  Then in our project we can continue to use just the fused files yet also have access to the extra testing utilities.  

Another option would be to use to pull `gunit-spi.h` into our project along with the fused files, but we figured doing that is sort of contrary to the purpose of a fused file.

Issue #2282, opened 2019-06-11
cmake policy CMP0054 hit when a parent project uses variable called SHARED


xammy:
When I include googletest as a subproject in cmake and use a variable $SHARED in the master project then the following warning appears:
```
CMake Warning (dev) at build/googletest-src/googletest/cmake/internal_utils.cmake:170 (if):
Policy CMP0054 is not set: Only interpret if() arguments as variables or
keywords when unquoted.  Run "cmake --help-policy CMP0054" for policy
details.  Use the cmake_policy command to set the policy and suppress this
warning.

Quoted variables like "SHARED" will no longer be dereferenced when the
policy is set to NEW.  Since the policy is not set the OLD behavior will be
used.
```
This can be avoided by explicitly activating the policy within googletest/CMakeLists.txt with a call to `  cmake_policy(SET CMP0054 NEW)`.

gennadiycivil:
Thank you very much for this report. The best way to approach this would be to create a proper PR and submit it for consideration.

xammy:
Alright - did so. See #2283.

PR #2283, opened 2019-06-12
Setting cmake CMP0054 policy to NEW

xammy:
This allows to use the string "SHARED" without interpreting it as a variable. Resolves issue #2282.


xammy:
I signed it!

PR #2286, opened 2019-06-13
Removing make and automake.

gennadiycivil:
The only supported build systems are Bazel internally and CMake community supported


googletest/docs/Pkgconfig.md
```diff
@@ -47,55 +47,6 @@ goes for using `_LDFLAGS` over the more commonplace `_LIBRARIES`, which
 happens to discard `-L` flags and `-pthread`.
 
 
-### Autotools ###
```
coryan: CMake can (and maybe does?) create pkg-config files. So you *could* (and I think you should) Autotools and Meson by just installing the right `.pc` files.

EricWF:
Does Googletest keep a change log? Should we put "Support for X and Y was removed in version Z" anywhere? I don't know it's needed, but I thought I would ask.

Otherwise, this LGTM.

Since I'm sure people are integrating googletest into their travis CI and using make, I imagine this will be noticed fairly quickly.

gennadiycivil:
Since http://go/gh/google/googletest/commit/f7c178ecb33c92763fa41d4999db30aaf43a6b30 did not cause any visible issues I am merging this PR.

PR #2303, opened 2019-06-26
Removes export of main function. Closes #891

### LINKED ISSUE #891: Conflicting export decorations on Windows

petersteneteg:


petersteneteg:
I have signed the CLA

gennadiycivil:
Thank you for this contribution. You can not just remove the export, see https://ci.appveyor.com/project/GoogleTestAppVeyor/googletest/builds/25550995/job/y1wsp442x72jb95f#L2271, it causes fail on Windows.

PR #2308, opened 2019-06-28
Fix bad advice in cook book

sam0109:
This line directly contradicts the warning that google mock spits out on unused mock calls:
```
NOTE: You can safely ignore the above warning unless this call should not happen.  Do not suppress it by blindly adding an EXPECT_CALL() if you don't mean to enforce the call.  See https://github.com/google/googletest/blob/master/googlemock/docs/CookBook.md#knowing-when-to-expect for details.
```
One or the other should be changed, and I believe the advice in this file is incorrect.

### DIFF
--- a/googlemock/docs/cook_book.md
+++ b/googlemock/docs/cook_book.md
@@ -229,7 +229,7 @@ If a mock method has no `EXPECT_CALL` spec but is called, Google Mock
 will print a warning about the "uninteresting call". The rationale is:
 
   * New methods may be added to an interface after a test is written. We shouldn't fail a test just because a method it doesn't know about is called.
-  * However, this may also mean there's a bug in the test, so Google Mock shouldn't be silent either. If the user believes these calls are harmless, they can add an `EXPECT_CALL()` to suppress the warning.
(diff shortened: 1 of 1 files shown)


asoffer:
Thanks Sam,
Colud you explain in a bit more detail what the contradiction is? My reading is:
* The warning suggests if you do not require the call to be made, it should be ignored rather than suppressed by adding `EXPECT_CALL`s for calls that may not be necessary.
* The cookbook suggests that having more expectations than are actually necessary is bad.
But I'm likely misunderstanding your concern.

sam0109:
Of course! My point is that 
- the warning says:
  - "Do not suppress it by blindly adding an EXPECT_CALL() if you don't mean to enforce the call", which is pretty self explanatory. Don't just silence this error with an EXPECT_CALL (as it will lead to brittle tests). If and only if you want to enforce this call, then add an EXPECT_CALL
- while the documentation currently says 
  - "If the user believes these calls are harmless, they can add an `EXPECT_CALL()` to suppress the warning", which is directly contradictory to the above. It is saying you *should* silence any random uninteresting calls by *enforcing* them with an EXPECT_CALL. This seems wrong as it will lead to brittle tests (as mentioned above) and is contradicting advice elsewhere.

PR #2312, opened 2019-07-08
Fix 'String Comparison' table

edhaz:
In preview some of the code sections had affected the 'String Comparison' table.
This is now fixed, the side effect being it doesn't look as pretty in markdown.


edhaz:
I signed it!

gennadiycivil:
@edhaz Thank you for taking the time. This is no longer needed (I think) due to other fixes that made it in

Issue #2277, opened 2019-06-09
Add MockFunction support for std::function


adambadura:
`MockFunction` requires function signature `R(Args...)` as its template argument. This makes it difficult to use it with `std::function` hidden under some type name. `std::function` doesn't provide any member type to allow to (safely and easily) reflect its own arguments (the signature).

The problem shows when we want to refer to `std::function` through its type name like:
```c++
using my_callback = std::function<bool(int)>;
```
Now, we cannot do `::testing::MockFunction<my_callback>`. We have to do `::testing::MockFunction<bool(int)>`, which is a maintenance issue requiring us to repeat the type definition.

Below is my proposal to add the support:
````diff
diff --git a/googlemock/include/gmock/gmock-spec-builders.h b/googlemock/include/gmock/gmock-spec-builders.h
index 1f261bd2..47c34980 100644
--- a/googlemock/include/gmock/gmock-spec-builders.h
+++ b/googlemock/include/gmock/gmock-spec-builders.h
@@ -1836,6 +1836,24 @@ void ReportUninterestingCall(CallReaction reaction, const std::string& msg);
 //   EXPECT_CALL(callback, Call("bar")).WillOnce(Return(1));
 //   Foo(callback.AsStdFunction());
 // }
+template <typename F>
+struct SignatureOf;
+
+template <typename R, typename... Args>
+struct SignatureOf<R(Args...)> {
+ using type = R(Args...);
+};
+
+template <typename R, typename... Args>
+struct SignatureOf<std::function<R(Args...)>> {
+ using type = R(Args...);
+};
+
+template <typename F>
+using SignatureOfT = typename SignatureOf<F>::type;
+
+namespace internal {
+
 template <typename F>
 class MockFunction;
 
@@ -1858,20 +1876,25 @@ class MockFunction<R(Args...)> {
     return mock_.Invoke(std::forward<Args>(args)...);
   }
 
-  internal::MockSpec<R(Args...)> gmock_Call(Matcher<Args>... m) {
+  MockSpec<R(Args...)> gmock_Call(Matcher<Args>... m) {
     mock_.RegisterOwner(this);
     return mock_.With(std::move(m)...);
   }
 
-  internal::MockSpec<R(Args...)> gmock_Call(const internal::WithoutMatchers&,
-                                            R (*)(Args...)) {
+  MockSpec<R(Args...)> gmock_Call(const WithoutMatchers&,
+                                  R (*)(Args...)) {
[...]

Issue #2265, opened 2019-05-29
MockFunction::AsStdFunction is not documented


adambadura:
The `AsStdFunction` method of `MockFunction` is not documented.

Here is my proposal to have it documented:
````diff
diff --git a/googlemock/docs/CheatSheet.md b/googlemock/docs/CheatSheet.md
index bc2af11f..44ecbc0d 100644
--- a/googlemock/docs/CheatSheet.md
+++ b/googlemock/docs/CheatSheet.md
@@ -577,10 +577,28 @@ Google Mock defines a convenient mock class template
 class MockFunction<R(A1, ..., An)> {
  public:
   MOCK_METHODn(Call, R(A1, ..., An));
+  std::function<R(A1, ..., An)> AsStdFunction();
 };
 ```
 See this [recipe](CookBook.md#using-check-points) for one application of it.
 
+The `AsStdFunction()` method returns an `std::function` with the same signature
+as the `MockFunction`. The returned functor calls the `Call` mock method of the
+`MockFunction` correctly forwarding any arguments and returning its  result.
+This allows using the `MockFunction` as a callback mock provided to the tested
+code:
+```cpp
+using predicate_signature = bool(int);
+using predicate = std::function<predicate_signature>;
+...
+void my_filter(predicate pred);
+...
+StrictMock<MockFunction<predicate_signature>> pred_mock;
+EXPECT_CALL(pred_mock, Call(_))
+    .WillRepeatedly(Return(true));
+my_filter(pred_mock.AsStdFunction());
+```
+
 # Flags #
 
 | Flag | Description |
````

gennadiycivil:
Thank you for this. I have it on my list to working on synchronizing internal docs with the OSS, which will add the documentation for AsStdFunction

gennadiycivil:
https://github.com/google/googletest/blob/6a2adc0eedd8ca47f1808dcd315b17666fb2fbe4/googlemock/docs/cook_book.md#mock-stdfunction-mockfunction

PR #2319, opened 2019-07-11
MockFunction improvements

adambadura:


adambadura:
It seems the change 3d70a6b (_Add MockFunction::AsLambda_) will not build since it contains:
```cpp
template <typename R, typename... Args>
class MockFunction<R(Args...)> {
  /* ... */

  auto AsLambda() {
    return [this](Args... args) -> R {
      return this->Call(std::forward<Args>(args)...);
    };
  }

  /* ... */
};
```
This form of `auto` return type deduction requires C++14. While apparently here we have only C++11.

If this cannot be changed (seems reasonable to not require too much) then the code either would have to use preprocessor for providing `AsLambda` with C++14 and higher only or would have to be dropped altogether. Which is fine as well, from my point of view the key parts are bb1c776 (_Document MockFunction::AsStdFunction (#2265)_) and cd3b9df (_Add support for std::function in MockFunction (#2277)_) anyway.

adambadura:
> I was using CMake while developing. (I tried the Bazel but didn't understand it at all...) This means that probably CMake is under-configured by not being strict on the language standard as use. Otherwise, I would see the error immediately. If C++11 is indeed the required standard I will try to update CMake correspondingly.

This seems more complicated. The main `CMakeLists.txt` does include
```cmake
if (CMAKE_VERSION VERSION_LESS "3.1")
  add_definitions(-std=c++11)
else()
  set(CMAKE_CXX_STANDARD 11)
  set(CMAKE_CXX_STANDARD_REQUIRED ON)
  if(NOT CYGWIN)
    set(CMAKE_CXX_EXTENSIONS OFF)
  endif()
endif()
```

It seems to be an issue between CMake and Visual C++, even though I have recent versions of both (3.14.0 and 19.21.27702.2 respectively) and this Visual C++ does support specifying language standard. Maybe the issue is that it does not support specifying C++11, only C++14 or higher.

The C++11 config does work with Cygwin.

adambadura:
I have dropped the problematic change 3d70a6b (_Add MockFunction::AsLambda_). Instead, I have added a change updating CMake support under Cygwin.

googlemock/include/gmock/gmock-spec-builders.h
```diff
@@ -1790,10 +1791,74 @@ void ReportUninterestingCall(CallReaction reaction, const std::string& msg);
...
+template <typename R, typename... Args>
+class MockFunction<R(Args...)> {
+ public:
+  MockFunction() = default;
+  MockFunction(const MockFunction&) = delete;
+  MockFunction& operator=(const MockFunction&) = delete;
+
+  std::function<R(Args...)> AsStdFunction() {
+    return [this](Args... args) -> R {
+      return this->Call(std::forward<Args>(args)...);
```
kuzkry: This and other std::forwards won't forward anything. `args` can never be variadic universal references here as they are not function template parameters. Imo, std::move works better here (as it will try to move, unless not possible (l-refs and nonmovable types)).

gennadiycivil:

adambadura:
> Btw. recently there's been some rename of GoogleMock docs but your change should easily rebase onto it.

Well, I was rebased over the `HEAD` until your changes got merged in. ;)

googlemock/include/gmock/gmock-spec-builders.h
adambadura: Firstly, note that I haven't written it so. I only moved it a bit. It was so before.

Secondly, I think you are not right about it anyway. This was originally `std::move` back in time when the class was still located in `gmock-generated-function-mockers.h` but was corrected to `std::forward` by 7796273 related to #1867.

adambadura:
> Thank your for this contribution. Could you please elaborate and provide more information on exactly the issue / problems this is trying to address.
> What is the before/after effect this would create and how is this beneficial?

Essentially there are three changes here:

1. Minor correction. (1st and 2nd commits)
2. Document currently undocumented `MockFunction::AsStdFunction`. (3rd commit)
3. Allow `MockFunction` to accept `std::function` (or other similar by user extension) as template argument and extract the function signature out of it. (4th commit)
4. Minor improvement. (5th commit)

googlemock/include/gmock/gmock-spec-builders.h
kuzkry: Sorry then, thanks for a good reference to another PR.

gennadiycivil:
@adambadura Thank you for this request. There are too many things going on at the same time here.
I see documentation changes, CMake changes, code changes, etc.
I recommend splitting this into separate small PRs so we can take a proper look.
Thanks again

PR #2335, opened 2019-07-16
Googletest export

gennadiycivil:
Googletest export


kuzkry:
Hi @gennadiycivil, could you please tell me what those "Google export" commits are for? I can see you're basically introducing bad stuff and then reverting it in subsequent commits. I don't quite understand.

gennadiycivil:
There are just some manual tests of the internal changes that sometimes is hard to test otherwise. You should pay zero attention to them. And they are not commits, just PRs.

kuzkry:
All right, I see. Thanks!

PR #2350, opened 2019-07-27
Add support for std::function in MockFunction (#2277)

adambadura:
`MocFunction` is slightly changed to transform its argument through (added in this change) `SignatureOf` meta-function. The meta-function extracts the function signature type from the provided type. Thanks to this approach `MockFunction<F>` can now be used with `F` being both function signature (like `bool(int)`) and `std::function` (like `std::function<bool(int)>`).

Besides the immediate benefit, extra extensibility is added. Library user can specialize `SignaturOf` for other types (like `boost::function`) to gain the same support level.

Note that this PR comes from #2319.

### DIFF
--- a/googlemock/include/gmock/gmock-spec-builders.h
+++ b/googlemock/include/gmock/gmock-spec-builders.h
@@ -1786,10 +1786,80 @@ void ReportUninterestingCall(CallReaction reaction, const std::string& msg);
 
 }  // namespace internal
 
-// A MockFunction<F> class has one mock method whose type is F.  It is
-// useful when you just want your test code to emit some messages and
-// have Google Mock verify the right messages are sent (and perhaps at
-// the right times).  For example, if you are exercising code:
+namespace internal {
+
+template <typename F>
+class MockFunction;
+
+template <typename R, typename... Args>
+class MockFunction<R(Args...)> {
+ public:
+  MockFunction(const MockFunction&) = delete;
+  MockFunction& operator=(const MockFunction&) = delete;
+
+  std::function<R(Args...)> AsStdFunction() {
+    return [this](Args... args) -> R {
+      return this->Call(std::forward<Args>(args)...);
+    };
+  }
+
+  // Implementation detail: the expansion of the MOCK_METHOD macro.
+  R Call(Args... args) {
+    mock_.SetOwnerAndName(this, "Call");
+    return mock_.Invoke(std::forward<Args>(args)...);
+  }
+
+  MockSpec<R(Args...)> gmock_Call(Matcher<Args>... m) {
+    mock_.RegisterOwner(this);
+    return mock_.With(std::move(m)...);
+  }
+
+  MockSpec<R(Args...)> gmock_Call(const WithoutMatchers&,
+                                  R (*)(Args...)) {
+    return this->gmock_Call(::testing::A<Args>()...);
+  }
+
+ protected:
+  MockFunction() = default;
+  ~MockFunction() = default;
+
+ private:
+  FunctionMocker<R(Args...)> mock_;
+};
+
+/*
+The SignatureOf<F> struct is a meta-function returning function signature
+corresponding to the provided F argument.
+
+It makes use of MockFunction easier by allowing it to accept more F arguments
+than just function signatures.
+
+Specializations provided here cover only a signature type itself and
+std::function. However, if need be it can be easily extended to cover also other
+types (like for example boost::function).
+*/
+
+template <typename F>
+struct SignatureOf;
+
+template <typename R, typename... Args>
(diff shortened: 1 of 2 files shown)


adambadura:
> One thing I forgot to ask you is whether the diff in _googlemock/include/gmock/gmock-spec-builders.h_ has to be so huge :)

I don't know how to make it smaller. The change is simile - added part around `SignatureOf` and moving of `MockFunction` to `internal` namespace. Yet Git somehow perceives it in a more difficult way. :(

kuzkry:
Yes, sometimes the diff algorithm gets crazy. Thanks and keep up the good work :)

CJ-Johnson:
@adambadura Thank you for the PR! After discussing this with the other owners, we're not interested in changing the "meaning" of the name MockFunction (from class template to alias template).

That said, we would be willing to consider a new name for this exact behavior. The new name would be an alias template out the gate. Is that a change you'd be open to?

adambadura:
> @adambadura Thank you for the PR! After discussing this with the other owners, we're not interested in changing the "meaning" of the name MockFunction (from class template to alias template).
> 
> That said, we would be willing to consider a new name for this exact behavior. The new name would be an alias template out the gate. Is that a change you'd be open to?

Sure. I prefer my approach, but I'm fine with your approach as well. And definitely either one is better than none. ;)

However, what precisely you want to get?

1. I will get the `MockFunction` back to the main namespace (from `internal` where I have put it in this change).
2. But what name to pick for the new entity that will handle `std::function`? `MockFunctor`?
3. Should the new entity work with both `std::function` and function signatures? I would recommend so to have a universal solution.
4. Should the `SignatureOf` remain part of the public interface or should it be handled by the new entity directly? Either way, I would strongly recommend an approach that allows users to extend the behavior for compatible function wrappers (like `boost::function` for example).

sbenzaquen:
Sorry for the delay.
I consulted with a coworker and my misgivings about this change have been solved.
Let's do this:
 1. `MockFunction<T>` supports `T=std::function<F>`. We talked about supporting other callable types, like function pointers, but decided that we don't need to do that now. If there is a need in the future, it could be easily added.
 2. SignatureOf would be an implementation detail, live in internal and not be mentioned in the .md documentation.

Thanks for the effort!

adambadura:
@sbenzaquen I have updated the pull request. Although, I'm still not sure if I got right what you meant. So, please, check now.

googlemock/include/gmock/internal/gmock-SignatureOf.h
```diff
@@ -0,0 +1,38 @@
+#ifndef THIRD_PARTY_GOOGLETEST_GOOGLEMOCK_INCLUDE_GMOCK_SIGNATUREOF_H_
```
sbenzaquen: We don't need or want one file for each little helper API.
Keep it in the same file, just have it in the internal namespace.

adambadura: Sure, I changed this to be your way!

sbenzaquen:
After testing this change, I found that this breaks valid use cases.
We can't make `testing::MockFunction<T>` be an alias with a dependent type on it. That breaks type deduction, specializations, etc.

Eg, this code:
```
template <typename T>
void Foo(testing::MockFunction<T>& t);
```
will not be able to deduce the T anymore when passing an instance of the type.

We have to keep `testing::MockFunction` as a class template.
An alternative solution is to make `testing::MockFunction` be like:
```
template <typename T>
class MockFunction : public internal::MockFunction<internal::SignatureOfT<T>> {
  using Base = internal::MockFunction<internal::SignatureOfT<T>>;
 public:
  using Base::Base;
};
```

adambadura:
@sbenzaquen, I have implemented it the way you suggested. Also, I have added tests that check the property you asked for so that no such error is made in the future.

However, I don't know why the verification failed. It did work on my (similar I think) configuration at my end. And I don't know how to get to exact logs of the failure. What can I do about it?

sbenzaquen:
The problem is that tests in the different _test.cc files have the same names and are colliding when running the gmock_all test. (they are all linked together and failing then).
Maybe rename the one in generated- to have Generated in its name.

adambadura:
> The problem is that tests in the different _test.cc files have the same names and are colliding when running the gmock_all test. (they are all linked together and failing then).
> Maybe rename the one in generated- to have Generated in its name.

It did help! Thanks! I just couldn't see what is the problem. I tried building on Linux subsystem on my Windows and even on Linux itself at work but got the same and didn't know how to proceed. Now it works.

Although I used a different naming approach, to mimic already existing one in those files.

However, I wonder why those tests are duplicated in the first place.

adambadura:
@sbenzaquen , can we finally proceed with this change?

PR #2344, opened 2019-07-25
Fixes in documentation

kuzkry:

PR #2356, opened 2019-07-30
Fix typos

kuzkry:
As requested in #2344.

Changes:
iff -> if
exihibit -> exhibit
plese -> please
are -> were (2nd conditional)


kuzkry:
Travis continues to fail on configuration steps, I'm trying to please it.

kuzkry:
Travis continues to fail on configuration steps, I'm trying to please it.

gennadiycivil:
260777334

adambadura:
Is this change correct? `iff` means `if and only if` (see [Wikipedia](https://en.wikipedia.org/wiki/If_and_only_if)).

So, if for example, we changed
```
returns true iff successful
```
into
```
returns true if successful
```
we changed biconditional into conditional. Formally the new meaning says only that `true` is returned on success. It doesn't say what happens on failure, it could return `true` as well. Nor that `true` is not returned in any other state (if there would be more than two).

I don't think anyone will seriously take this soooo formally. However, keeping it formally correct isn't an issue as well, right?

kuzkry:
> Is this change correct? `iff` means `if and only if` (see [Wikipedia](https://en.wikipedia.org/wiki/If_and_only_if)).
> 
> So, if for example, we changed
> 
> ```
> returns true iff successful
> ```
> 
> into
> 
> ```
> returns true if successful
> ```
> 
> we changed biconditional into conditional. Formally the new meaning says only that `true` is returned on success. It doesn't say what happens on failure, it could return `true` as well. Nor that `true` is not returned in any other state (if there would be more than two).
> 
> I don't think anyone will seriously take this soooo formally. However, keeping it formally correct isn't an issue as well, right?

@adambadura Wooow :O I had no idea about it being a thing. It looked like a misprint and a copy-paste mistake combined. My bad, thank you! You can revert if you want, though to be honest, it's clear enough I think.

adambadura:
Perhaps then it would be better to have `if and only if` stated directly to avoid possible future confusion while keeping formal correctness?

kuzkry:
Btw. I made a small questionnaire at my work and also asked one external programmer, and everyone didn't know that :D Therefore, I suggest we either keep it as it is or change to the explicit "if and only if" form. I'm totally indifferent, I might prepare another PR if you want to have it restored / changed.

adambadura:
@kuzkry, I made a poll at my work as well (if you are a Nokia employ [here is Yammer link](https://www.yammer.com/nokia.com/#/Threads/show?threadId=265963143880704)). The poll was created 8 days ago and last answers were provided 3 days ago, so I guess it is pretty stable now.

The poll had three answers and 37 votes:
1. 35% - _it's a typo - should be "if"_
2. 59% - _it is a shortcut for "if and only if"_
3. 5% - _something else_

PR #2357, opened 2019-07-30
Update pre-C++11 paragraphs

kuzkry:
As requested in #2344.

This removes the TODO-ish comments of:
> we'll consider supporting using ACTION*() inside a function.

and

> we'll consider supporting using MATCHER*() inside a function.

My reasoning is:
* Actions can be easily changed to lambdas (or even generic ones if someone uses C++14) so no need to put effort into legacy functionality
* Matchers have also alternatives, e.g. Truly(). Perhaps I should add something like:
> Users are free to provide their local functors and invoke Truly() on them

What do you think (especially about the latter)?

### DIFF
--- a/googlemock/include/gmock/gmock-generated-actions.h
+++ b/googlemock/include/gmock/gmock-generated-actions.h
@@ -259,11 +259,10 @@ class ActionHelper {
 //
 // CAVEAT:
 //
-// ACTION*() can only be used in a namespace scope.  The reason is
-// that C++ doesn't yet allow function-local types to be used to
-// instantiate templates.  The up-coming C++0x standard will fix this.
-// Once that's done, we'll consider supporting using ACTION*() inside
-// a function.
+// ACTION*() can only be used in a namespace scope as templates cannot be
+// declared inside of a local class.
+// Users can, however, define any local functors (e.g. a lambda) that
+// can be used as actions.
 //
 // MORE INFORMATION:
 //
--- a/googlemock/include/gmock/gmock-generated-actions.h.pump
+++ b/googlemock/include/gmock/gmock-generated-actions.h.pump
@@ -182,11 +182,10 @@ $template
 //
 // CAVEAT:
 //
-// ACTION*() can only be used in a namespace scope.  The reason is
-// that C++ doesn't yet allow function-local types to be used to
-// instantiate templates.  The up-coming C++0x standard will fix this.
-// Once that's done, we'll consider supporting using ACTION*() inside
-// a function.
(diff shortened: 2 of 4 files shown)


kuzkry:
Travis continues to fail on configuration steps, I'm trying to please it.

googlemock/include/gmock/gmock-generated-actions.h.pump
```diff
@@ -182,11 +182,11 @@ $template
...
-// ACTION*() can only be used in a namespace scope.  The reason is
-// that C++ doesn't yet allow function-local types to be used to
-// instantiate templates.  The up-coming C++0x standard will fix this.
-// Once that's done, we'll consider supporting using ACTION*() inside
-// a function.
+// ACTION*() can only be used in a namespace scope as templates cannot be
+// declared inside of a local class.
+// Users can, however, define any local functors (e.g. a lambda) that
+// can be used as actions. Since C++14, it is also possible to use
+// generic lambdas which overcome the mentioned limitation.
```
gennadiycivil: I would remove C++14 language completely.

googlemock/include/gmock/gmock-generated-actions.h
```diff
@@ -259,11 +259,11 @@ class ActionHelper {
...
 //
-// ACTION*() can only be used in a namespace scope.  The reason is
-// that C++ doesn't yet allow function-local types to be used to
-// instantiate templates.  The up-coming C++0x standard will fix this.
-// Once that's done, we'll consider supporting using ACTION*() inside
-// a function.
+// ACTION*() can only be used in a namespace scope as templates cannot be
+// declared inside of a local class.
+// Users can, however, define any local functors (e.g. a lambda) that
+// can be used as actions. Since C++14, it is also possible to use
```
gennadiycivil: I would remove C++14 language completely. Less is more here

gennadiycivil:
@kuzkry thank you, please take a look at my comment(s)

googlemock/include/gmock/gmock-generated-actions.h.pump
kuzkry: Ok.

googlemock/include/gmock/gmock-generated-actions.h
kuzkry: Good point, thanks :)

kuzkry:
> @kuzkry thank you, please take a look at my comment(s)

Thank you too! It's done now, you can recheck.

gennadiycivil:
261341748

PR #2316, opened 2019-07-09
Fix typed- and type-parameterized test clang warnings

kuzkry:
This refers to [issue #2271](https://github.com/google/googletest/issues/2271)

PR #2358, opened 2019-07-31
Adjustment of two similar comments

kuzkry:


kuzkry:
Done as requested in #2316.

googletest/docs/advanced.md
```diff
@@ -1558,7 +1558,7 @@ test suite. You can repeat this as many times as you want:
 ```c++
 TYPED_TEST(FooTest, DoesBlah) {
   // Inside a test, refer to the special name TypeParam to get the type
-  // parameter.  Since we are inside a derived class template, C++ requires
```
asoffer: There are other places in this file and in other files that use two spaces. Personally I don't care much one way or the other, but we should at least be consistent? Moreover, it is "technically correct" to have two spaces with monospace fonts, so I'd lean towards reverting this and keeping two spaces.

kuzkry: Hmm, I see. All right, all right, but I guess I need to put 2 spaces in the other file as well now.

PR #2359, opened 2019-08-01
Remove an excessive mutable type specifier

kuzkry:
The private mutable member of MockFunction cannot be reached from outside and all its
non-static member functions are not const. Alternatively, we could add const to every function of MockFunction, though it's a much worse idea I believe.

@adambadura - watch out, this will conflict with your #2350 if this gets merged first.

### DIFF
--- a/googlemock/include/gmock/gmock-spec-builders.h
+++ b/googlemock/include/gmock/gmock-spec-builders.h
@@ -1870,7 +1870,7 @@ class MockFunction<R(Args...)> {
   }
 
  private:
-  mutable internal::FunctionMocker<R(Args...)> mock_;
+  internal::FunctionMocker<R(Args...)> mock_;
 };
 
 // The style guide prohibits "using" statements in a namespace scope


gennadiycivil:

kuzkry:
Hmm, pardon me, actually nothing concrete comes to my mind. This is, I suppose, because of the pure nature of mutable which should be given only to make members of a const object modifiable. We shouldn't use it without a specific, exceptional need.
I've even digged into git history in order to find and ask an author but it turned out to be no-one specific, i.e. Abseil Team in [de5be0](https://github.com/google/googletest/commit/de5be0eb28b74ecd6335e3bd61d9dc8914ce0e57). If this is possible for you to find them, please do.
Last but not least, I could apply the reversed logic and ask if there is a scenario that requires having this mutable ;) Perhaps, it's the right way to think about it.

gennadiycivil:
Looking, stay tuned

gennadiycivil:
261151513

PR #2372, opened 2019-08-06
Fix #2371: Redirect Windows CRT assertions to stderr

### LINKED ISSUE #2371: Redirect Windows CRT assertions to stderr
By default, assertions in the Windows CRT in debug mode will pop up a dialog box and wait for user input (*). This is undesirable in automated testing setups (such as cloud-based CI services) as it then appears to hang without any useful indication of why the tests stopped running.

Instead, gtest could redirect those messages by default to stderr. See e.g. https://stackoverflow.com/questions/13943665/how-can-i-disable-the-debug-assertion-dialog-on-windows/31611163#31611163

(*) These can happen e.g. when passing an invalid file descriptor to a CRT function, or other kinds of invalid input

pitrou:


pitrou:
Right... Unfortunately I have technical problems with Google Accounts. I'll try to sign the CLA with the right e-mail address but no guarantee I'll succeed.

pitrou:
@googlebot I signed it!

pitrou:
There are download failures on Travis-CI.

googletest/src/gtest.cc
```diff
@@ -80,6 +80,11 @@
 
 #elif GTEST_OS_WINDOWS  // We are on Windows proper.
 
+# include <windows.h>  // NOLINT
```
gennadiycivil: Is this include absolutely necessary? windows.h has been known to cause various issues. ( min in particular)

pitrou: This is just moved around in the file (see below). It must be included before `debugapi.h`.

gennadiycivil:
261930228

pitrou:
Thanks for merging!

googletest/src/gtest.cc
```diff
@@ -4914,6 +4914,16 @@ int UnitTest::Run() {
           0x0,                                    // Clear the following flags:
           _WRITE_ABORT_MSG | _CALL_REPORTFAULT);  // pop-up window, core dump.
 # endif
+
+    // In debug mode, the Windows CRT can crash with an assertion over invalid
+    // input (e.g. passing an invalid file descriptor).  The default handling
+    // for these assertions is to pop up a dialog and wait for user input.
+    // Instead ask the CRT to dump such assertions to stderr non-interactively.
+    if (!IsDebuggerPresent()) {
+      (void)_CrtSetReportMode(_CRT_ASSERT,
```
devjgm, line 4923: Why does this set only `_CRT_ASSERT` and not `_CRT_WARN` and `_CRT_ERROR` as well? is there a reason why the latter two do not need to also be redirected to stderr like this one?

PR #2373, opened 2019-08-06
Added special catch for std::exception in GTEST_TEST_NO_THROW_

Youw:
Re-open #2084 with fixes.

### DIFF
--- a/googletest/include/gtest/internal/gtest-internal.h
+++ b/googletest/include/gtest/internal/gtest-internal.h
@@ -825,6 +825,16 @@ struct GTEST_API_ ConstCharPtr {
   const char* value;
 };
 
+// Helper for declaring std::string within 'if' statement
+// in pre C++17 build environment.
+struct GTEST_API_ TrueWithString {
+  TrueWithString() = default;
+  explicit TrueWithString(const char* str) : value(str) {}
+  explicit TrueWithString(const std::string& str) : value(str) {}
+  explicit operator bool() const { return true; }
+  std::string value;
+};
+
 // A simple Linear Congruential Generator for generating random
 // numbers with a uniform distribution.  Unlike rand() and srand(), it
 // doesn't use global state (and therefore can't interfere with user
@@ -1335,19 +1345,37 @@ constexpr bool InstantiateTypedTestCase_P_IsDeprecated() { return true; }
     GTEST_CONCAT_TOKEN_(gtest_label_testthrow_, __LINE__): \
       fail(gtest_msg.value)
 
+#if GTEST_HAS_EXCEPTIONS
+
+#define GTEST_TEST_NO_THROW_CATCH_STD_EXCEPTION_() \
+  catch (std::exception const& e) { \
+    gtest_msg.value = "it throws std::exception-derived exception with description: \""; \
+    gtest_msg.value += e.what(); \
+    gtest_msg.value += "\"."; \
+    goto GTEST_CONCAT_TOKEN_(gtest_label_testnothrow_, __LINE__); \
+  }
+
+#else // GTEST_HAS_EXCEPTIONS
+
+#define GTEST_TEST_NO_THROW_CATCH_STD_EXCEPTION_()
+
+#endif  // GTEST_HAS_EXCEPTIONS
+
 #define GTEST_TEST_NO_THROW_(statement, fail) \
   GTEST_AMBIGUOUS_ELSE_BLOCKER_ \
-  if (::testing::internal::AlwaysTrue()) { \
+  if (::testing::internal::TrueWithString gtest_msg{}) { \
     try { \
       GTEST_SUPPRESS_UNREACHABLE_CODE_WARNING_BELOW_(statement); \
     } \
+    GTEST_TEST_NO_THROW_CATCH_STD_EXCEPTION_() \
     catch (...) { \
+      gtest_msg.value = "it throws."; \
       goto GTEST_CONCAT_TOKEN_(gtest_label_testnothrow_, __LINE__); \
     } \
   } else \
     GTEST_CONCAT_TOKEN_(gtest_label_testnothrow_, __LINE__): \
-      fail("Expected: " #statement " doesn't throw an exception.\n" \
-           "  Actual: it throws.")
+      fail(("Expected: " #statement " doesn't throw an exception.\n" \
+            "  Actual: " + gtest_msg.value).c_str())
 
 #define GTEST_TEST_ANY_THROW_(statement, fail) \
   GTEST_AMBIGUOUS_ELSE_BLOCKER_ \
--- a/googletest/test/gtest_unittest.cc
+++ b/googletest/test/gtest_unittest.cc
@@ -3351,6 +3351,9 @@ TEST_F(SingleEvaluationTest, OtherCases) {
 void ThrowAnInteger() {
   throw 1;
 }
+void ThrowRuntimeError(const char* what) {
+  throw std::runtime_error(what);
+}
 
 // Tests that assertion arguments are evaluated exactly once.
 TEST_F(SingleEvaluationTest, ExceptionTests) {
@@ -3830,6 +3833,11 @@ TEST(AssertionTest, ASSERT_NO_THROW) {
   EXPECT_FATAL_FAILURE(ASSERT_NO_THROW(ThrowAnInteger()),
                        "Expected: ThrowAnInteger() doesn't throw an exception."
                        "\n  Actual: it throws.");
+  EXPECT_FATAL_FAILURE(ASSERT_NO_THROW(ThrowRuntimeError("A description")),
+                       "Expected: ThrowRuntimeError(\"A description\") "
+                       "doesn't throw an exception.\n  "
+                       "Actual: it throws std::exception-derived exception "
+                       "with description: \"A description\".");
 }
 
 // Tests ASSERT_ANY_THROW.
@@ -4567,6 +4575,11 @@ TEST(ExpectTest, EXPECT_NO_THROW) {
   EXPECT_NONFATAL_FAILURE(EXPECT_NO_THROW(ThrowAnInteger()),
                           "Expected: ThrowAnInteger() doesn't throw an "
                           "exception.\n  Actual: it throws.");
+  EXPECT_NONFATAL_FAILURE(EXPECT_NO_THROW(ThrowRuntimeError("A description")),
+                          "Expected: ThrowRuntimeError(\"A description\") "
+                          "doesn't throw an exception.\n  "
+                          "Actual: it throws std::exception-derived exception "
+                          "with description: \"A description\".");
 }
 
 // Tests EXPECT_ANY_THROW.


gennadiycivil:
@Youw Thank you for this. 
The older PR has a lot of conversation back . and forth and is hard to follow.

Youw:
_Why_:

`ASSERT/EXPECT_NO_THROW` is non-informative when it comes to investigating of "why a unit-test failed", when it failed.

_Example_:
```cpp
BusinessLogic logic;
ASSERT_NO_THROW(logic.perform());
```

_Before_:
```
Expected: logic.perform() doesn't throw an exception.
Actual: it throws.
```
It is unclear why the execution has failed, and manual investigation/debugging required.

_After_:
```
Expected: logic.perform() doesn't throw an exception.
Actual: it throws std::exception-derived exception with description: Startup error: configuration file /etc/project/main.conf is missing.
```
(The example is completely made up, but real-life cases sometime ridiculously close to this one.)

---
Some notes (case "before"):
Simplest way to get the exception message is to omit `ASSERT_NO_THROW` at all, and std::exception will be caught by GTest later, but:
1) The log message won't include failed statement and line;
2) Some developers omitting `ASSERT_NO_THROW` on purpose, to use 1) instead;
3) As a result of 2): no explicit highlight of "what is actually tested" by a particular test case;

---
Of course, not all projects follow "throw only std::exception-derived/std::runtime_error" rule, and this patch cannot cover all possible cases, but it greatly improve some part of it.

darionesi:
Is there any way to know if something can be printed as part of the message? Like “does it have a stream operator<<?” ? That may help with other objects being thrown that are not deriving from stream but can be printed as additional information

Youw:
There is no way to know what "that something" is.
Basically, what we have:
```cpp
try {
    // test check body
}
catch(const std::exception& e) {
     // print e.what()
}
catch(...) { // the problem in "..."
    // we know that there is an exception, but don't know of which type
    // maximum what we can get - std::current_exception(), but that is not helpful
}

**Comment by darionesi** [comment]
Can’t the detection pattern be used to check if there is an overload of operator<<? 
https://people.eecs.berkeley.edu/~brock/blog/detection_idiom.php 
https://www.bfilipek.com/2019/07/detect-overload-from-chars.html 
just asking

**Comment by Youw** [comment]
One can use detection pattern in case if have an object of unknown type, i.e. when using template function/class.

In our case we don't have an object at all.

File: googletest/include/gtest/internal/gtest-internal.h [hunk 1]
```diff
@@ -1337,17 +1346,24 @@ constexpr bool InstantiateTypedTestCase_P_IsDeprecated() { return true; }
 
 #define GTEST_TEST_NO_THROW_(statement, fail) \
   GTEST_AMBIGUOUS_ELSE_BLOCKER_ \
-  if (::testing::internal::AlwaysTrue()) { \
+  if (::testing::internal::TrueWithString gtest_msg = std::string()) { \
```
EricWF: Why can't this be declared on the next line?

googletest/include/gtest/internal/gtest-internal.h
```diff
@@ -1337,17 +1346,24 @@ constexpr bool InstantiateTypedTestCase_P_IsDeprecated() { return true; }
...
     try { \
       GTEST_SUPPRESS_UNREACHABLE_CODE_WARNING_BELOW_(statement); \
     } \
+    catch (std::exception const& e) { \
+      gtest_msg.value = " std::exception-derived exception with description: "; \
+      gtest_msg.value += e.what(); \
+      gtest_msg.value += "."; \
```
EricWF: The value of `what()` is implementation defined. Why does adding a dot make sense?

googletest/include/gtest/internal/gtest-internal.h
```diff
@@ -1337,17 +1346,24 @@ constexpr bool InstantiateTypedTestCase_P_IsDeprecated() { return true; }
...
+    } \
     catch (...) { \
+      gtest_msg.value = "."; \
       goto GTEST_CONCAT_TOKEN_(gtest_label_testnothrow_, __LINE__); \
     } \
   } else \
     GTEST_CONCAT_TOKEN_(gtest_label_testnothrow_, __LINE__): \
-      fail("Expected: " #statement " doesn't throw an exception.\n" \
-           "  Actual: it throws.")
+      fail(("Expected: " #statement " doesn't throw an exception.\n" \
```
EricWF: `fail` won't take a `std::string`?

darionesi:
> In our case we don't have an object at all.

right! that makes my idea tricky to implement 🤣

googletest/include/gtest/internal/gtest-internal.h
Youw: Same reason as in `GTEST_TEST_THROW_` - `gtest_msg` is used in both `if` and `else`

googletest/include/gtest/internal/gtest-internal.h
Youw: No, it doesn't. It expands to 4-th parameter of `AssertHelper` c-tor: https://github.com/google/googletest/blob/90a443f9c2437ca8a682a1ac625eba64e1d74a8a/googletest/include/gtest/gtest.h#L1772

googletest/include/gtest/internal/gtest-internal.h
Youw: To make sure we do have a dot at the end of the sentence, as we have in case of any other exception.

Youw:
Do we have explicit change requests on this one?

googletest/include/gtest/internal/gtest-internal.h
CJ-Johnson: Instead of adding a period at the end, are you able to change this such that the output is like the following?

```
gtest_msg.value = "`std::exception`-derived exception thrown with description \""
                                + e.what()
                                "\"";
```

googletest/include/gtest/internal/gtest-internal.h
```diff
@@ -1337,17 +1346,24 @@ constexpr bool InstantiateTypedTestCase_P_IsDeprecated() { return true; }
...
       GTEST_SUPPRESS_UNREACHABLE_CODE_WARNING_BELOW_(statement); \
     } \
+    catch (std::exception const& e) { \
+      gtest_msg.value = " std::exception-derived exception with description: "; \
+      gtest_msg.value += e.what(); \
+      gtest_msg.value += "."; \
+      goto GTEST_CONCAT_TOKEN_(gtest_label_testnothrow_, __LINE__); \
+    } \
     catch (...) { \
+      gtest_msg.value = "."; \
```
CJ-Johnson: What is the purpose of this?

googletest/include/gtest/internal/gtest-internal.h
CJ-Johnson: Can this be written with `::testing::internal::ConstCharPtr` instead of introducing a new type?

googletest/include/gtest/internal/gtest-internal.h
Youw: To match previous behavior. See unit-tests

googletest/include/gtest/internal/gtest-internal.h
Youw: It cannot: we need to own a copy of a string. ConstCharPtr is non-owning a wrapper over `const char*`

googletest/include/gtest/internal/gtest-internal.h
CJ-Johnson: That works, for me! Add the quotes around the error message and then keep the dot as well :)

googletest/include/gtest/internal/gtest-internal.h
Youw: updated to be more explicit

googletest/include/gtest/internal/gtest-internal.h
Youw: updated

CJ-Johnson:
@Youw I apologize for not making this change request earlier. Can you mark the `TrueWithString`'s constructors and conversion operator as explicit? I'll try to merge after that's done.

Youw:
Sure, I will.

Except I'm not sure whether explicit `operator bool` would allow using `TrueWithString`'s variable declaration within `if` statement

I'll try and update, it that is possible.

Youw:
Explicit c-tor -  done.

Explicit `operator bool` isn't possible in given use-case.
The very purpose of the class is to be able to declare it in `if` statement, _always (implicitly) evaluate to `true`_ and to be able to store a `string`, so it can be modified inside `if` and available inside `else`.
With C++17 this would simply be like:
```cpp
if (string msg; AlwaysTrue()) {
// set msg here
}
else {
// using msg here
}
```

CJ-Johnson:
This appears to work https://godbolt.org/z/Cfwuzc

Can you help me understand why it isn't able to be explicit?

CJ-Johnson:
Also it appears this change it out of date. It looks like similar behavior was already added.

https://github.com/google/googletest/blob/ba513d2c9525a7c986c115ed5d603f2cf17c6016/googletest/include/gtest/internal/gtest-internal.h#L822-L826

https://github.com/google/googletest/blob/ba513d2c9525a7c986c115ed5d603f2cf17c6016/googletest/include/gtest/internal/gtest-internal.h#L1269-L1282

Youw:
1) hmm, I checked it locally on my PC, and it didn't compile, I'll double check: maybe a specific compiler, or a bit different implementation

2) the change you mentioned is in function GTEST_TEST_THROW_. This PR is about GTEST_TEST_NO_THROW_

Youw:
> This appears to work https://godbolt.org/z/Cfwuzc

> hmm, I checked it locally on my PC, and it didn't compile, I'll double check: maybe a specific compiler, or a bit different implementation

I'm not sure why, but I couldn't reproduce the issue with `explicit operator bool` I had locally. Maybe I was confused by a different issue I had, which I probably fixed later.

As of now - updated accordingly.

CJ-Johnson:
I will now try again to land this change.

CJ-Johnson:
I'm performing a global test run. This may take up to 24 hours at which point I will be able to submit it.

CJ-Johnson:
The test run has completed and some breakages were found. Essentially, some users depend on GCC headers that, under `-fno-exceptions`, expose something to the effect of:

```
#define catch(e) \
  if(0)
```

This causes compilation failure for your change. Inside the `std::exception` catch block, the body references an undeclared symbol `e`. I think the best solution at the moment is to conditionally add this catch block based on whether exceptions are enabled or disabled. Something like:

```
#if GTEST_HAS_EXCEPTIONS

#define GTEST_TESTNOTHROW_CATCH_STD_EXCEPTION_()  \
  catch (std::exception const& e) {  \
    gtest_msg.value = "it throws std::exception-derived exception with description: \"";  \
    gtest_msg.value += e.what();  \
    gtest_msg.value += "\".";  \
    goto GTEST_CONCAT_TOKEN_(gtest_label_testnothrow_, __LINE__);  \
  }

#else  // GTEST_HAS_EXCEPTIONS

#define GTEST_TESTNOTHROW_CATCH_STD_EXCEPTION_()

#endif  // GTEST_HAS_EXCEPTIONS
```

That way, when exceptions are disabled, `e` will not be present in source and will thus not fail to compile. Is this an acceptable change?

Youw:
updated as suggested, but I've failed adding a test for this case, so left it as is

CJ-Johnson:
Fantastic! Thank you :)

I just kicked off the global test run again. Fingers crossed!

PR #2374, opened 2019-08-06
Rename equality macro arguments to (expected, actual)

### LINKED ISSUE #1416: Name macros variables (expected,actual) instead of (val1,val2)
At the moment you can't differentiate them

https://github.com/google/googletest/blob/83fa0cb17dad47a1d905526dcdddb5b96ed189d2/googletest/include/gtest/gtest.h#L1957

This is actually useful in error messages.

We have this code `ASSERT_EQ(reply, nullptr);` and it generates this error message:

```
Failure
Value of: nullptr
  Actual: 8-byte object <00-00 00-00 00-00 00-00>
Expected: reply
Which is: 8-byte object <C0-08 00-24 9D-7F 00-00>
```

As you can see it's confusing: `nullptr` is expected to be `reply`.
Correct is `ASSERT_EQ(nullptr, reply);`: `reply` is expected to be `nullptr`.

```
Failure
Value of: reply
  Actual: 8-byte object <C0-08 00-24 9D-7F 00-00>
Expected: nullptr
Which is: 8-byte object <00-00 00-00 00-00 00-00>
```

So the macros should be

```cpp
#define ASSERT_EQ(expected, actual)
```

EmilWojak:
Fixes #1416

### DIFF
--- a/googletest/docs/advanced.md
+++ b/googletest/docs/advanced.md
@@ -333,10 +333,10 @@ want to learn more, see
 
 <!-- mdformat off(github rendering does not support multiline tables) -->
 
-| Fatal assertion                 | Nonfatal assertion              | Verifies                                 |
-| ------------------------------- | ------------------------------- | ---------------------------------------- |
-| `ASSERT_FLOAT_EQ(val1, val2);`  | `EXPECT_FLOAT_EQ(val1, val2);`  | the two `float` values are almost equal  |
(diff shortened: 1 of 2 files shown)


EmilWojak:
@googlebot I signed it!

EmilWojak:
I just came across the followin note in [Googletest Primer](../blob/master/googletest/docs/primer.md):
> **Historical note**: Before February 2016 `*_EQ` had a convention of calling it
> as `ASSERT_EQ(expected, actual)`, so lots of existing code uses this order. Now
> `*_EQ` treats both parameters in the same way.

PR #2364, opened 2019-08-02
[bazel] Split gtest and gmock and become compatible with Bazel incompatible changes

Yannic:
Fixes googletest for upcoming `--incompatible_load_cc_rules_from_bzl` (https://github.com/bazelbuild/bazel/issues/8743) and `--incompatible_load_python_rules_from_bzl` (https://github.com/bazelbuild/bazel/issues/9006).

This change also makes the build more hermetic.

PR #2377, opened 2019-08-06
Fix C++ includes in preparation for #2364

Yannic:
#2364 will make the Bazel build more hermetic, e.g. we no longer leak `src/gtest-internal-inl.h` to tests.

@gennadiycivil Can you take a look whether this breaks your import?


gennadiycivil:
As discussed in https://github.com/google/googletest/pull/2364#issuecomment-518826169. I ll close this one for now

PR #2381, opened 2019-08-07
Prepare for Bazel incompatible changes

Yannic:
Fixes googletest for upcoming `--incompatible_load_cc_rules_from_bzl` (https://github.com/bazelbuild/bazel/issues/8743) and `--incompatible_load_python_rules_from_bzl` (https://github.com/bazelbuild/bazel/issues/9006).

This change was automatically generated with
`buildifier -lint=fix -warnings=all $(find . -name "BUILD" -o -name "BUILD.bazel" -o -name "*.bzl")`.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -30,6 +30,8 @@
 #
 #   Bazel Build for Google C++ Testing Framework(Google Test)
 
+load("@rules_cc//cc:defs.bzl", "cc_library", "cc_test")
+
 package(default_visibility = ["//visibility:public"])
 
 licenses(["notice"])
@@ -81,6 +83,10 @@ cc_library(
         ":has_absl": ["GTEST_HAS_ABSL=1"],
         "//conditions:default": [],
     }),
+    features = select({
+        ":windows": ["windows_export_all_symbols"],
+        "//conditions:default": [],
+    }),
     includes = [
         "googlemock",
         "googlemock/include",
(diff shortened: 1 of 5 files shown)


BUILD.bazel
```diff
@@ -155,11 +157,11 @@ cc_test(
         "googletest/samples/sample7_unittest.cc",
         "googletest/samples/sample8_unittest.cc",
     ],
+    linkstatic = 0,
```
gennadiycivil: Looks like there is no good reason to move this line from previous location.Could you please limit the changes to the actual factual changes to remove the noise.

BUILD.bazel
```diff
@@ -81,6 +83,10 @@ cc_library(
         ":has_absl": ["GTEST_HAS_ABSL=1"],
         "//conditions:default": [],
     }),
+    features = select({
```
gennadiycivil: Looks like there is no good reason to move this line from previous location.Could you please limit the changes to the actual factual changes to remove the noise.

BUILD.bazel
```diff
@@ -102,20 +108,16 @@ cc_library(
...
 cc_library(
     name = "gtest_main",
     srcs = ["googlemock/src/gmock_main.cc"],
-    deps = [":gtest"],
     features = select({
         ":windows": ["windows_export_all_symbols"],
         "//conditions:default": [],
-    })
+    }),
+    deps = [":gtest"],
```
gennadiycivil: Looks like there is no good reason to move this line from previous location.Could you please limit the changes to the actual factual changes to remove the noise.

BUILD.bazel
Yannic: [buildifier](https://github.com/bazelbuild/buildtools/tree/master/buildifier) automatically formats `BUILD` files in the canonical way when it applies fixes, so the new location is the correct one. I can revert to the old format manually, but I don't think it adds a lot of noise, and the next person editing the `BUILD` file will likely do the same changes again.

(Dito other comments)

Issue #2337, opened 2019-07-18
Question about docs of gtest & gmock


ant35rookie:
Hello, thank you for this great project

I have two questions!

1. Do you think that gtest / gmock need translated docs in other languages in main branch?
(like googlemock/docs/some_language/cook_book.md)

2. Do I need to make an issue and get agreement before I submit PR? (if it fixes typo in [docs](https://github.com/google/googletest/tree/master/googlemock/docs) only)

Thank you ~!

EricWF:
Reopening. This is a great question.
I would like to see this issue discussed further.

gennadiycivil:
It would be great to have documentation in other languages, yes.
We would have to develop an on-going process so once the docs are there they stay in sync with the changes though.

ant35rookie:
Thank you for the comment!
I'm actually translating docs into Korean and I think that it would be helpful for other Koreans too

I will try to check it regularly and sync up the changes

And I believe that engineers in my country would help to update it if initial version is there ^^;

gennadiycivil:
This is really awesome, thank you so much.

hyukmyeong:
Hello!

This issue is for making googletest, gMock docs in Korean version^^

I think that my work is almost done !
And while I'm doing this I found below as awell

It must be grateful if you review these things !

1. I found typos and so make https://github.com/google/googletest/pull/2384
2. I think markdown link syntax `{#where}` is not working in github
3. I found term "parent class", "base class" are mingled, Do you think they need to be unified?
4. In advanced.md, it describes "pimpl" as "private Implementation", It is correct cuz the context is saying about testing private code but still "pointer to Implementation" looks more general term

Please kindly review https://github.com/google/googletest/pull/2384 !
and if it needs to be revised I think I could make PR on 2~4 as well

PR #2384, opened 2019-08-09
Let tables use the same syntax and fix some typos in docs

hyukmyeong:
This PR is trying to let tables use the same syntax and fix some typos in documents!

I submit this PR while I'm working on issue https://github.com/google/googletest/issues/2337

### DIFF
--- a/googlemock/docs/cheat_sheet.md
+++ b/googlemock/docs/cheat_sheet.md
@@ -235,10 +235,10 @@ several categories:
 
 #### Wildcard
 
-Matcher                     | Description
-:-------------------------- | :-----------------------------------------------
-`_`                         | `argument` can be any value of the correct type.
-`A<type>()` or `An<type>()` | `argument` can be any value of type `type`.
+| Matcher                     | Description                                      |
+| :-------------------------- | :----------------------------------------------- |
+| `_`                         | `argument` can be any value of the correct type. |
+| `A<type>()` or `An<type>()` | `argument` can be any value of type `type`.      |
 
 #### Generic Comparison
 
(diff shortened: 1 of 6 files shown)


hyukmyeong:
@googlebot I signed it!

googlemock/docs/cheat_sheet.md
```diff
@@ -235,10 +235,10 @@ several categories:
 
 #### Wildcard
 
-Matcher                     | Description
-:-------------------------- | :-----------------------------------------------
-`_`                         | `argument` can be any value of the correct type.
-`A<type>()` or `An<type>()` | `argument` can be any value of type `type`.
+| Matcher                     | Description                                      |
```
kuzkry: You're second commit called "Fix typo in tables in docs" does some changes which don't change anything. In Markdown you may build up a table row both with or without enclosing pipes, i.e. ` ABC | XYZ ` and ` | ABC | XYZ |` are synonyms. I certainly wouldn't call it a typo.
Note: this relates to other places in this review as well.

hyukmyeong: Thank you for your comment !

Yeap, you are right, They both work well !
I just did it to let them use the same syntax as the others in the document!

kuzkry: Cool! You're welcome. Could you possibly rename the commit as I mentioned?

hyukmyeong: I'm sorry for the late response!
Could you please review again whether commit's name is proper or not~?

googlemock/docs/cheat_sheet.md
```diff
@@ -490,12 +490,12 @@ which must be a permanent callback.
 #### Returning a Value
 
 <!-- mdformat off(no multiline tables) -->
-|                             |                                               |
+| Action                      | Description                                   |
 | :-------------------------- | :-------------------------------------------- |
 | `Return()`                  | Return from a `void` mock function.           |
-| `Return(value)`             | Return `value`. If the type of `value` is     different to the mock function's return type, `value` is converted to the latter type <i>at the time the expectation is set</i>, not when the action is executed. |
+| `Return(value)`             | Return `value`. If the type of `value` is different to the mock function's return type, `value` is converted to the latter type *at the time the expectation is set*, not when the action is executed. |
```
kuzkry: I can see you tried to keep the ending `|` aligned to adjacent ones, but here (and in some other places too) there is a small inconsistency.

On the other hand, this line didn't fit because of how long the description is. Maybe keeping everything aligned would make all other short descriptions unreadable or unnecessarily long.

What do you think? Do you have any other ideas?

hyukmyeong: Thank you for your feedback!

I tried to align all the tables in all the documents once!
But as you know, some of them were still impossible and even they were less readable
For that reason, I gave up and just aligned a few tables which have other modifications as well
It is also possible to put them back if you give me further feedback!

And I think using html <table> might be an option to describe table neatly cuz github does not support multiline table!

kuzkry: And so I thought ;) Btw. I like your idea with http tables.

PR #2387, opened 2019-08-12
Revert mistakenly removed iffs

kuzkry:
This **partially** reverts #2356 (it's not 100% revert, I made some modifications to prevent too long lines).

TL;DR: Wary as usual @adambadura (thanks mate, all credit to you ;) ) found out that I made a mistake in #2356 by thinking iff was a typo of "if", whereas it really stands for "if and only if".

For more, please follow the discussion started in #2356.


adambadura:
@kuzkry, I think it would be better to use `if and only if` instead of `iff`. This will prevent the same misunderstanding in future.

However, if comment lines are broken at some column (which seems to be a somewhat flexible rule) changing `iff` to `if and only if` cannot be done by simple find-and-replace and requires a manual check of each case to ensure that line length limits are observed.

kuzkry:
> @kuzkry, I think it would be better to use `if and only if` instead of `iff`. This will prevent the same misunderstanding in future.
> 
> However, if comment lines are broken at some column (which seems to be a somewhat flexible rule) changing `iff` to `if and only if` cannot be done by simple find-and-replace and requires a manual check of each case to ensure that line length limits are observed.

Working on it. Stay tuned.

kuzkry:
Ok, done. I'm keeping this WIP, as I would like you to make sure it's like you wanted it to be. @adambadura, please take some time and be very precise then pointing me to lines that you still consider too long.

Commits and their logic:
1a6d736 is find and replace
2f39f43 is what clang formatting tool thinks about it
f7b850e is a bunch of my propositions

When we're done, I will simply fix the last two and withdraw WIP from the PR title.

adambadura:
Seems OK to me. I have no particular needs about the line breaking. I just noticed that in general line breaking is used. Although the length limit didn't seem consistent.

kuzkry:
Thank you @adambadura!

kuzkry:
Temporarily closing this to avoid a merge conflict with another pull request.

PR #2391, opened 2019-08-13
Use -Wa,-mbig-obj for Cygwin/MinGW always

adambadura:
Another target fails on Cygwin with the too big file issue. This time it is `gmock-generated-matchers_test`. To get rid of the problem I just used `-Wa,-mbig-obj` for all Google Mock test targets instead of only the failing ones.


gennadiycivil:
263572442

Issue #2392, opened 2019-08-14
Unneeded reimplementations of C++11 type_trait utilities


kuzkry:
GTest and GMock provide a lot of code that has been included in C++11.

kuzkry:
There will be merge conflicts because they depend on each other.

PR #2395, opened 2019-08-14
Remove custom implementations of std::remove_reference

kuzkry:
#2392


asoffer:
263417291

asoffer:
It looks like we need to do some internal cleanup to get this working. I'll get that started and update when we're ready to merge this.

kuzkry:
Ok, thanks! The problem here is all these type traits either depend on each other or were defined in similar places.

kuzkry:
Retriggering CI

PR #2400, opened 2019-08-14
Remove a custom implementation of std::enable_if

kuzkry:
#2392


asoffer:
263426328

EricWF:
This LGTM. 

Although, it's fun to point out that this could change the behavior of code, because it's possible to observe the difference between `enable_if<false>` being an incomplete type as opposed to a complete type with an empty definition.
But anybody clever enough to do that is clever enough to fix it.

kuzkry:
Heh, right! But look, no fixes needed; incompilable code is still incompilable :)

kuzkry:
Merged by 9bf34ace.

PR #2402, opened 2019-08-14
Remove tests for signed wchar_t and unsigned wchar_t.

pcc:
Clang now errors out on these types as well by default. Rather than making the
condition for these types even more complicated, just remove the tests covering
these types since they don't seem to justify the maintenance burden. We
can reasonably expect these types to work in compilers that support them
without needing specific tests for them since they are treated as standard
integral types.


gennadiycivil:
Thank you very much for this . 

Please follow https://github.com/google/googletest/blob/master/CONTRIBUTING.md#are-you-a-googler

pcc:
cl/263406158

PR #2403, opened 2019-08-14
Remove unnecessary CompileAssertTypesEqual.

IYP-Programer-Yeah:


kuzkry:
Hahaha, I was about to submit it XD

googlemock/include/gmock/gmock-actions.h
```diff
@@ -842,7 +841,7 @@ class IgnoreResultAction {
     typedef typename internal::Function<F>::Result Result;
 
     // Asserts at compile time that F returns void.
-    CompileAssertTypesEqual<void, Result>();
+    static_assert(std::is_same<void, Result>::value, "Result type should be void.");
```
kuzkry: It's better if you changed it to:
```suggestion
    static_assert(std::is_void<Result>::value, "Result type should be void.");
```

googletest/include/gtest/gtest.h
```diff
@@ -1965,13 +1965,13 @@ class TestWithParam : public Test, public WithParamInterface<T> {
 // AssertionResult. For more information on how to use AssertionResult with
 // these macros see comments on that class.
 #define EXPECT_TRUE(condition) \
-  GTEST_TEST_BOOLEAN_(condition, #condition, false, true, \
+  GTEST_TEST_BOOLEAN_((condition), #condition, false, true, \
```
kuzkry: Why did you add parenthesis?

googlemock/include/gmock/gmock-actions.h
```diff
@@ -617,9 +617,8 @@ class ReturnNullAction {
 class ReturnVoidAction {
  public:
   // Allows Return() to be used in any void-returning function.
-  template <typename Result, typename ArgumentTuple>
+  template <typename Result, typename ArgumentTuple, typename = std::enable_if<std::is_same<Result, void>::value>>
```
kuzkry: I cannot make a suggestion here as it's multiple-liner, sorry.
This has the following implications:
1) Compiles longer (SFINAE is a performance killer)
2) Instead of giving a clear compile time error message, it will provide "couldn't find a function"
3) It is a bit harder to maintain than a plain static assert
4) There is no reason to SFINAE since you've got no additional overload of the Perform() function in this structure.
Please use a static assertion. In addition, std::is_same could be turned into std::is_void.

googletest/test/gtest_unittest.cc
```diff
@@ -233,7 +233,6 @@ using testing::internal::AppendUserMessage;
 using testing::internal::ArrayAwareFind;
```
kuzkry: #include <type_traits> is missing.

googletest/include/gtest/gtest.h
IYP-Programer-Yeah: Wasnt meant to be pushed :D thanks

kuzkry: Sure, you're welcome :) And carry on continuing my work :muscle:

googlemock/test/gmock-internal-utils_test.cc
```diff
@@ -39,6 +39,7 @@
 #include <string>
 #include <sstream>
 #include <vector>
+#include <type_traits>
```
kuzkry: Can you keep the alphabetical order please?

googletest/test/gtest_unittest.cc
```diff
@@ -64,6 +64,7 @@ TEST(CommandLineFlagsTest, CanBeAccessedInCodeOnceGTestHIsIncluded) {
 #include <vector>
 #include <ostream>
 #include <unordered_set>
+#include <type_traits>
```
kuzkry: Can you possibly make a change so they are sorted alphabetically?

googlemock/test/gmock-internal-utils_test.cc
IYP-Programer-Yeah: I tried to fix this, but it leads to changes in the whole order, since the previous code wasn't ordered. If necessary I can order them all.

googletest/test/gtest_unittest.cc
IYP-Programer-Yeah: Same as the above.

googlemock/test/gmock-internal-utils_test.cc
kuzkry: I've just reminded myself I do it in my pull requests :D

googletest/test/gtest_unittest.cc
kuzkry: No, I was more of a "nice to have" thing.

IYP-Programer-Yeah:
A gentle ping!

kuzkry:
I can see you've restored the stuff I already asked you to change, in particular:
* `std::is_void<T>` instead of `std::is_same<void, T>`
* unnecessary addition of parenthesis around `condition` in macro GTEST_TEST_BOOLEAN_
* use of static assertion instead of std::enable_if

IYP-Programer-Yeah:
@kuzkry Sorry for the inconvenience, fixed.

IYP-Programer-Yeah:
Also any idea why the Mac OS build is failing? Doesn't seem to be anything on my part.

kuzkry:
No worries, it's not yours. Everything fails now (even commits that were okay a few days ago). This is quite likely a bug in libc++, which is the default implementation of the standard library on Mac OS. Related issue is #2428

IYP-Programer-Yeah:
@kuzkry I agree with your point, but the main reason I used the run time test wast to avoid the confusion between a compilation error and a behavioral error. If you think the old way was better, I can switch back to it.

kuzkry:
Both ways are fine. This subtlety can be opinioned by maintainers.
And thanks for resolving other things I suggested.

IYP-Programer-Yeah:
A gentle ping

asoffer:
Apologies, I need to get better at communicating here. I have looked at this PR and it looks good to me. I needed to do a bit of an internal cleanup internal code outside of googletest has come to use these assertions over the years. They should all be fixed now, so I can pull this change in.

PR #2405, opened 2019-08-14
Remove CopyArray and use std::copy_n instead.

IYP-Programer-Yeah:
This copy is unnecessary since std::copy_n does the job, and doesn't require maintenance.

### DIFF
--- a/googletest/include/gtest/internal/gtest-internal.h
+++ b/googletest/include/gtest/internal/gtest-internal.h
@@ -60,6 +60,7 @@
 #include <string>
 #include <type_traits>
 #include <vector>
+#include <algorithm>
 
 #include "gtest/gtest-message.h"
 #include "gtest/internal/gtest-filepath.h"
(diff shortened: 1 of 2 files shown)


googletest/include/gtest/internal/gtest-internal.h
```diff
@@ -1040,33 +1041,6 @@ Iter ArrayAwareFind(Iter begin, Iter end, const Element& elem) {
...
-template <typename T, typename U>
-void CopyArray(const T* from, size_t size, U* to);
-
-// This generic version is used when k is 0.
-template <typename T, typename U>
-inline void CopyArray(const T& from, U* to) { *to = from; }
-
-// This overload is used when k >= 1.
-template <typename T, typename U, size_t N>
-inline void CopyArray(const T(&from)[N], U(*to)[N]) {
```
kuzkry: This is why it worked. You begin with char[2][3] and through the call chain you are given char*[3]. So std::copy_n will dereference it, leaving it with char[3], hence the error of unassignability.

I like your idea with std::copy_n but that won't help recursion. If I'm not mistaken with C++17 it could just be:
```c++
template <typename T, typename U>
void CopyArray(const T* from, size_t size, U* to) {
    if constexpr (std::is_array<T>::value) // we cannot use if constexpr yet
        CopyArray(*from, std::extent<T>::value, *to);
    else
        std::copy_n(from, size, to);
}
```
I think you can change the existing code only a bit :<

googletest/include/gtest/internal/gtest-internal.h
```diff
@@ -60,6 +60,7 @@
 #include <string>
 #include <type_traits>
 #include <vector>
+#include <algorithm>
```
kuzkry: IYP-Programer-Yeah 1:0 alphabetic order :d

googletest/include/gtest/internal/gtest-internal.h
IYP-Programer-Yeah: Thanks havent made time to look at it yet. Ill fix it soon

kuzkry:
Perhaps there is still a room for small improvement. For example, maybe you could turn those 3 functions into 2, couldn't you?

PR #2420, opened 2019-08-22
Remove references to autotools and combine gTest & gMock sections

kuzkry:
This is based on 6b8c1381.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -90,32 +90,40 @@ and their own tests from a git checkout, which has further requirements:
 *   [Python](https://www.python.org/) v2.3 or newer (for running some of the
     tests and re-generating certain source files from templates)
 *   [CMake](https://cmake.org/) v2.6.4 or newer
-*   [GNU Build System](https://en.wikipedia.org/wiki/GNU_Build_System) including
-    automake (>= 1.9), autoconf (>= 2.59), and libtool / libtoolize.
 
-## Developing Google Test
+## Developing Google Test and Google Mock
 
-This section discusses how to make your own changes to Google Test.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -90,32 +90,40 @@ and their own tests from a git checkout, which has further requirements:
...
+    cmake -Dgtest_build_tests=ON ${GTEST_DIR} # sets up Google Test tests
+    cmake -Dgmock_build_tests=ON ${GMOCK_DIR} # sets up Google Mock tests
 
 Make sure you have Python installed, as some of Google Test's tests are written
 in Python. If the cmake command complains about not being able to find Python
 (`Could NOT find PythonInterp (missing: PYTHON_EXECUTABLE)`), try telling it
 explicitly where your Python executable can be found:
 
-    cmake -DPYTHON_EXECUTABLE=path/to/python -Dgtest_build_tests=ON ${GTEST_DIR}
+    cmake -DPYTHON_EXECUTABLE=path/to/python ...
```
EricWF: Is this bit about finding python still necessary?

I've hit a lot of CMake mis-configurations before, but surprisingly failing to find python isn't one of them.

kuzkry: CMake uses [this](https://github.com/google/googletest/blob/fb49e6c164490a227bbb7cf5223b846c836a0305/googletest/cmake/internal_utils.cmake#L243) to find Python. I'm happy that it's not some strange workaround in the build system but maybe we save someone else's from headache with this small hint. Frankly, I'm indifferent when it comes to removing it, albeit I wouldn't like to mix it with my changes in this pull request if you don't mind me. :)
Would you like me to prepare another pull request targeting this?

PR #2421, opened 2019-08-22
Clean-up for regex

kuzkry:
Small findings that I prefer to keep in a separate pull request.


kuzkry:
Retriggering CI.

googlemock/docs/cheat_sheet.md
```diff
@@ -326,8 +326,7 @@ The `argument` can be either a C string or a C++ string object:
 `ContainsRegex()` and `MatchesRegex()` take ownership of the `RE` object. They
 use the regular expression syntax defined
 [here](../../googletest/docs/advanced.md#regular-expression-syntax).
-`StrCaseEq()`, `StrCaseNe()`, `StrEq()`, and `StrNe()` work for wide strings as
-well.
+All but them work for wide strings.
```
mplatings: This sentence isn't grammatically correct, and I'm not sure what it's intended to mean.

kuzkry: Thanks for the review.
Let me explain what I meant. [This source](https://www.collinsdictionary.com/dictionary/english/all-but) says "all but" means "all except". If you take a look at the [table](https://github.com/google/googletest/blob/1f2e2e8059540247537e18cdab33bc77259273af/googlemock/docs/cheat_sheet.md#string-matchers), you will find my text in close adjacency, so my references:
* All = all these matches from the table (i.e. ContainsRegex(string), EndsWith(suffix), HasSubstr(string), MatchesRegex(string), StartsWith(prefix), StrCaseEq(string), StrCaseNe(string), StrEq(string) and StrNe(string))
* them = specific matches from the preceding sentence

So by "All but them work for wide strings" I meant "Take every string, remove ContainsRegex and MatchesRegex and Bob's your uncle; you're left with the ones that work with wide strings".

I'm not a native, but hopefully now you have a clear understanding of what I wanted to express. If I'm still wrong, please tell me :) You're welcome to provide me with a suggestion too.

asoffer: I am trying to come up with the grammatical rule that explains why this is incorrect. Even as a native speaker I don't know why this is incorrect... but it does sound strange. The best I can do is say that "them" sounds awkward here. The word "these" would be okay, but I'd prefer being explicit.

Do `StartsWith` and `EndsWith` work with wide strings too? If so, then perhaps

"All of these matchers, except `ContainsRegex` and `MatchesRegex` work for wide strings as well."

If they don't, then I think it's fine as is.

googletest/docs/advanced.md
```diff
@@ -686,7 +686,7 @@ Note that a death test only cares about three things:
 2.  (in the case of `ASSERT_EXIT` and `EXPECT_EXIT`) does the exit status
     satisfy `predicate`? Or (in the case of `ASSERT_DEATH` and `EXPECT_DEATH`)
     is the exit status non-zero? And
-3.  does the stderr output match `regex`?
+3.  does the stderr output match `matcher`?
```
asoffer: This has to be a regex, right? Not any matcher?

kuzkry: Ok, pardon me ;)

googlemock/docs/cheat_sheet.md
kuzkry: They do, I'll take your version.
Hmm, quite interesting. Thanks for taking the time ;)

asoffer:
Apologies for taking so long on this. Looks good and I'm importing it into an internal changelist.

PR #2422, opened 2019-08-22
Add multiplatform support for regular expressions

### LINKED ISSUE #2404: Support regular expressions on non-POSIX platforms
In code there exists a POSIX detection mechanism that conditionally enables POSIX regexes. I'm working on creating a portable solution with C++11. This is similar to #2380.

kuzkry:
This is based on #2421 and it resolves #2404.

Possible future work is:
1. adding support for wide strings for `MatchesRegex()` and `ContainsRegex()` so the support of Google Mock for wide strings is complete.
2. adding support for wide strings for death tests (though it's less likely to occur and it's more optional).

I'll happily take care of it, just let me know what your opinion is.


googletest/include/gtest/internal/gtest-port.h
```diff
@@ -202,12 +194,6 @@
 //   Mutex, MutexLock, ThreadLocal, GetThreadCount()
 //                            - synchronization primitives.
 //
-// Regular expressions:
-//   RE             - a simple regular expression class using the POSIX
-//                    Extended Regular Expression syntax on UNIX-like platforms
-//                    GOOGLETEST_CM0008 DO NOT DELETE
```
kuzkry: What do I do this this marker or the one in line 170?

asoffer:
Thanks for working diligently on this. I really like the idea at its core, but Google does not use std::regex internally, so I don't think this approach is going to work for us. Could this be done with RE2?

kuzkry:
Hello Andy! I don't understand your reasoning since Gtest requires C++11 and my approach wouldn't make a new dependency. Anyway, if that's what you want, I might give it a try, though it might take me some time.

asoffer:
It's not a new external dependency, but rather a dependency on std::regex. It's not an issue of availability, but rather library choice. Internally Google does not use std::regex for a handful of reasons (both technical and historical). We do not want even tests to pull in std::regex.

Yes, if you got this working with RE2, we would consider it

asoffer:
@kuzkry is this still under active development? I'm happy to keep it open if this is something you're pursuing. Otherwise I'd like to close the issue.

kuzkry:
Please do not close it, I'm working on that. I've had personal things that kept me away from GitHub.

kuzkry:
@asoffer, I'm thinking about closing this pull request.
Right now those are our regex implementations:
- user injected one in custom/gtest-port.h when they specify GTEST_USES_PCRE=1
- GoogleTest's class `RE` in two variants:
a) if GTEST_USES_POSIX_RE=1, then we use POSIX's `regex_t`
b) if GTEST_USES_SIMPLE_RE=1, then we define custom regex support and it is "so-so"

I came to the conclusion that adding `RE2` makes little sense because that could be merely another alternative protected by some GTEST_USES_RE2, that would still, on failure, fall back to SIMPLE_RE (option b). I find adding `RE2` neither multiplatform nor actually needed. The whole point of this pull request was to simplify things by making them common. So what do you think?

Btw. may I kindly ask for explanation why is `std::regex` so unacceptable? ;) I understand it's slow and compiles painfully long, though we're in the GoogleTest context and not like we're doing optimisations for the Google search engine. :)

bgaifullin:
There is any chance that this PR would be commited?

kuzkry:
How about this approach? GoogleTest could by default attempt to find & use RE2 and if not found, then falls back to std::regex? Would that be ok, Googlers?

asoffer:
Sorry, I've been slow to respond.

We do not want a dependency on std::regex whatsoever. No this isn't in the guts of Google Search, but performance still matters. For running a test once in a while, performance isn't critical, but Google has and frequently runs lots of tests. Increasing the build and run time of these tests has impact that we care about.

I'm not super familiar, but... do RE2 and std::regex use the same syntax? I believe not so I'm skeptical of the fallback approach. If I move my code from one installation to another and it switches regex engines, it's likely that I also need to change the regular expressions somewhat defeating the purpose of making this multiplatform in the first place.

Sorry, I know this is frustrating. This might be best off as a MatchesRegex matcher in a local library that's not upstreamed into GoogleTest.

kuzkry:
I understand @asoffer. No, it's not a problem, I respect technical arguments.

So what do you want to do? Because the most clean way for Googletest would be to have only one implementation of regex, but taking a dependency on `RE2` would break build for current users of Googletest.

Would it be OK to do that and release GoogleTest 2.0? Or do you have another idea?

kuzkry:
Rebased due to a merge conflict so it's easier to work on that later on. I'm waiting for a reply from the Google team.

bgaifullin:
gentle ping, I really need this patch, because part of my project uses pcre_regex (for historical reasons), part of project uses std::regex.  So I cannot use gtest together with code, which includes pcre_regex, due to naming clash. And It would be really helpful if there will be one option to use std::regex or not instead of current 3 options.

kuzkry:
Another rebase due to merge conflict.

asoffer:
Sorry again for the delay. I'm going to close this. I think the best approach here is a regex Matcher specific to the regex type of choice (be it std::regex, RE2, or pcre_regex) and that these matchers should live in an external library not a part of GoogleTest.

PR #2427, opened 2019-08-24
DO NOT MERGE! triggering Travis

kuzkry:


kuzkry:
I tried to find a broken commit but checking out older commits doesn't help and every commit fails.
This is most likely a bug in libc++. Related issue: #2428.

PR #2438, opened 2019-09-01
Print C++17 types (any, optional, string_view and variant) and absl::any

### LINKED ISSUE #2437: GTest is unable to print C++17 types
```c++
#include <optional>
#include <string>

#include <gmock/gmock.h>
#include <gtest/gtest.h>

using my_type = std::optional<std::string>;

class Mock {
public:
  MOCK_METHOD(my_type, foo, (), ());
};

TEST(Foo, Test)
{
  my_type const lol;
  Mock mock;

  ON_CALL(mock, foo()).WillByDefault(::testing::Return(lol));

  mock.foo();
}
```

When my_type is:
* `std::any`, `16-byte object <B2-AA 19-00 00-00 00-00 51-58 1F-00 00-00 00-00>` is printed and there are no Valgrind complains
* `std::optional<std::string>`, `40-byte object <D0-90 DA-05 00-00 00-00 13-00 00-00 00-00 00-00 13-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 01-00 00-00 00-00 00-00>` is printed and Valgrind reports problems
* `std::string_view`, `{ 'n' (110, 0x6E), 'o' (111, 0x6F), ' ' (32, 0x20), 'e' (101, 0x65), 'a' (97, 0x61), 's' (115, 0x73), 't' (116, 0x74), 'e' (101, 0x65), 'r' (114, 0x72), ' ' (32, 0x20), 'e' (101, 0x65), 'g' (103, 0x67), 'g' (103, 0x67), 's' (115, 0x73), ' ' (32, 0x20), 'h' (104, 0x68), 'e' (101, 0x65), 'r' (114, 0x72), 'e' (101, 0x65) }` is printed and there are no Valgrind complains
* `std::variant<std::string>`,  `40-byte object <D0-90 DA-05 00-00 00-00 13-00 00-00 00-00 00-00 13-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00>` is printed and Valgrind reports problems

These Valgrind complains for optional and variant are:
> Use of uninitialised value of size 8
> Conditional jump or move depends on uninitialised value(s)

kuzkry:
Resolves #2437.

Future work:
Matchers could now support wide string_views. This is something new; Abseil doesn't support this one.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -103,6 +103,7 @@ cc_library(
             "@com_google_absl//absl/debugging:stacktrace",
             "@com_google_absl//absl/debugging:symbolize",
             "@com_google_absl//absl/strings",
+            "@com_google_absl//absl/types:any",
             "@com_google_absl//absl/types:optional",
             "@com_google_absl//absl/types:variant",
         ],
--- a/googlemock/include/gmock/gmock-matchers.h
+++ b/googlemock/include/gmock/gmock-matchers.h
@@ -659,16 +659,15 @@ class StrEqualityMatcher {
   StrEqualityMatcher(const StringType& str, bool expect_eq,
                      bool case_sensitive)
       : string_(str), expect_eq_(expect_eq), case_sensitive_(case_sensitive) {}
-
-#if GTEST_HAS_ABSL
-  bool MatchAndExplain(const absl::string_view& s,
+#if GTEST_HAS_CPP17_TYPES
+  bool MatchAndExplain(const string_view& s,
                        MatchResultListener* listener) const {
-    // This should fail to compile if absl::string_view is used with wide
-    // strings.
+    // This should fail to compile if std::string_view / absl::string_view
+    // is used with wide strings.
     const StringType& str = std::string(s);
     return MatchAndExplain(str, listener);
   }
-#endif  // GTEST_HAS_ABSL
+#endif  // GTEST_HAS_CPP17_TYPES
 
   // Accepts pointer types, particularly:
   //   const char*
@@ -685,8 +684,9 @@ class StrEqualityMatcher {
 
   // Matches anything that can convert to StringType.
   //
-  // This is a template, not just a plain function with const StringType&,
-  // because absl::string_view has some interfering non-explicit constructors.
+  // This is a template, not just a plain function with const std::string&,
+  // because std::string_view and absl::string_view have some interfering
+  // non-explicit constructors.
   template <typename MatcheeStringType>
   bool MatchAndExplain(const MatcheeStringType& s,
                        MatchResultListener* /* listener */) const {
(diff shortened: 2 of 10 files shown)


asoffer:
I like the idea here, but I think we need to have a bit of internal discussion first about how we want to support this.

asoffer:
Thanks for this PR. After internal discussion, we are going to have googletest take a hard dependency on Abseil very soon. Which means that compiling with C++17, all the types will be the standard types and have correct printing, and without C++17, the Abseil backports will have correct printing and the standard ones won't be available. So I think this PR won't be necessary. Thanks again though, for this effort as well as all the other PRs you've sent. They're super useful and very much appreciated.

kuzkry:
Thank you for recognition!
But I regret your team decision as C++17 is quite popular these days. What about non-Abseil users such as myself? Will users be forced to reinvent the wheel because of this? Because having Abseil just for that is a huge no-go for my project (additional documentation and stuff...).

kuzkry:
@asoffer ping

kuzkry:
@asoffer, please note that my code uses string_views and I get bad formatting. Even my custom PrintTo() function doesn't work, check out this:
```c++
#include <string_view>

#include <gtest/gtest.h>

void PrintTo(std::string_view const s, std::ostream* const str)
{
  *str << "view";
}

TEST(FailAndPrint, MyTest)
{
  using namespace std::string_view_literals;
  EXPECT_EQ("this is"sv, "not equal"sv);
}
```

As a user, I'm not supposed to alter ::testing::internal and std namespaces - however in order to make my string_views work I would need to inject my function into either of these places to make ADL work.

asoffer:
If you're using C++17, absl::{optional, variant, any, string_view} are all the same type as the std equivalents, so unless I'm missing something this should all "just work" fine. You do not need to spell it with the absl:: namespace qualification.

If any of this doesn't work, please open up issues.

The googletest dependency on Abseil is more about using Abseil internally and should not impose any requirements on how users write tests.

kuzkry:
But I've already pinned a related issue :) It's #2437 and it basically tells you that not only is GTest unable to print those types, but also there are some Valgrind issues.

kuzkry:
Ok, @asoffer. I didn't know what you meant was the latest "Coming soon":
> We are also planning to take a dependency on Abseil.

which appeared a few days ago. Yes, it will work too for std::string_view I think.

Please mind the other types!

What I wrote here:
> Future work:
> Matchers could now support wide string_views. This is something new; Abseil doesn't support this one.

effecitively means that [this particular function](https://github.com/google/googletest/blob/3f5a8c6ee6ceea308d07def0f8430045f2edbc8e/googlemock/include/gmock/gmock-matchers.h#L664) will not work for `std::wstring_view`, `std::u8string_view`, `std::u16string_view`, `std::u32string_view`.
With my pull request, you could change `std::string_view` into `std::basic_string_view<T>`. Besides, it doesn't have to have the base being of the char type family to be considered as a string, as `std::basic_string_view<int>` is also a valid class here. Last but not least, I've already proved the users cannot provide their own PrintTo() for std types without either extending std (which is undefined behaviour btw.) or ::testing::internal (which is also against the rules).

kuzkry:
@asoffer ping

PR #2443, opened 2019-09-04
fix comment mistake in gtest.h

hhforgithub:
at line 248 , the wrong number 8 is changed to 6


hhforgithub:
@googlebot I signed it!

asoffer:
Thanks for the keen eye. I believe this is correct as written because 8 is the 6th fibonacci number. That is, `8 == Fib(6)`. in this example.

PR #2444, opened 2019-09-04
Remove GTEST_ARRAY_SIZE_

kuzkry:
This macro didn't work when an array was passed to a function by pointer,
in which case the information about its size was lost.
Better alternatives are:
* std::extent<T>::value (compile-time)
* std::array<T, N>::size() (compile-time)
* std::distance(std::begin(array), std::end(array)) (run-time)

### DIFF
--- a/googlemock/test/gmock-generated-matchers_test.cc
+++ b/googlemock/test/gmock-generated-matchers_test.cc
@@ -41,6 +41,8 @@
 
 #include "gmock/gmock-generated-matchers.h"
 
+#include <array>
+#include <iterator>
 #include <list>
 #include <map>
 #include <memory>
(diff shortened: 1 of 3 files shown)


kuzkry:
Retriggering CI.

kuzkry:
There is a merge conflict so I'm rebasing this commit. I hope it's okay with you (@asoffer).

gennadiycivil:
273569389

kuzkry:
In my eyes, what disqualifies GTEST_ARRAY_SIZE_ is the fact that it will silently fail in the following scenarios:
```c++
void callee1(int arr[10]) // Even if I explicitly say 10 here
{
  assert(GTEST_ARRAY_SIZE_(arr) == 10); // FAIL
}

void callee2(int* arr)
{
  assert(GTEST_ARRAY_SIZE_(arr) == 10); // FAIL
}

void caller()
{
  int a[10];

  callee1(a);
  callee2(a);
}
```
It's very easy to abuse this macro and it makes users remember about the danger.
What I think we should use instead is (let me quote myself):
> This macro didn't work when an array was passed to a function by pointer,
> in which case the information about its size was lost.
> Better alternatives are:
> 
>     * std::extent::value (compile-time)
> 
>     * std::array<T, N>::size() (compile-time)
> 
>     * std::distance(std::begin(array), std::end(array)) (run-time)

In your case, I would suggest using `constexpr std::array<DvTestCase, 6> kDvCases = {/*..*/}` and then calling `kDvCases.size()`.

gennadiycivil:
@kuzkry just an update, I am still trying to clean up usages before this can go in. Thanks

kuzkry:
No worries @gennadiycivil

PR #2448, opened 2019-09-07
Merge stuff that has been given up by manual merges

kuzkry:
Leftovers from improperly closed manual reviews.


gennadiycivil:
273570181

This happens sometimes because the SoT for the development is internal and the process is not ideal. 
We are incrementally working to make sure this does not happen again

kuzkry:
@gennadiycivil, I realised that some time ago :) But it's nice you started warning people about it.

gennadiycivil:
> @gennadiycivil, I realised that some time ago :) But it's nice you started warning people about it.
Ah, this would be the "incrementally working" part :-)

PR #2452, opened 2019-09-10
Move bool_constant from gtest-port.h to gtest-internal.h

kuzkry:
Because `bool_constant` is portable, it should not belong to *gtest-port.h*.


asoffer:
Perhaps we can remove bool_constant altogether. There's one non-trivial use of it that I'm working on internally, so I may just remove it entirely.

PR #2453, opened 2019-09-10
Remove BiggestInt

### LINKED ISSUE #2451: CppCheck error in gtest-port.h
The following error was found by CppCheck:
[/usr/local/include/gtest/internal/gtest-port.h:2126]: (error) Shifting signed 64-bit value by 63 bits is undefined behaviour

The code in question is:
```
const BiggestInt kMaxBiggestInt =
    ~(static_cast<BiggestInt>(1) << (8*sizeof(BiggestInt) - 1));
```

My command line:
cppcheck -I /usr/local/include -D__GNUC__ *.cpp

kuzkry:
Resolves #2451.

Additionally I've moved it to gtest-internal.h because it requires no porting to specific platforms (long long int is a part of C++11).

### DIFF
--- a/googletest/include/gtest/gtest-printers.h
+++ b/googletest/include/gtest/gtest-printers.h
@@ -133,7 +133,7 @@ GTEST_API_ void PrintBytesInObjectTo(const unsigned char* obj_bytes,
 // nor PrintTo().
 enum TypeKind {
   kProtobuf,              // a protobuf type
-  kConvertibleToInteger,  // a type implicitly convertible to BiggestInt
+  kConvertibleToInteger,  // a type implicitly convertible to std::intmax_t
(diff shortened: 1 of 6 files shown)


vslashg:
This is not a very widely used type, and it's in an internal namespace.  I question whether we need it at all?

On first glance, it seems preferable to use `std::intmax_t` and `std::numeric_limits<intmax_t>::max()` for these purposes — both are available in C++11.

vslashg:
We've begun the internal review process of this PR.  Please don't many any further changes here, or they may be lost.

gennadiycivil:
@kuzkry Looks like we had to roll back this PR because of internal breakage. 
@vslashg

kuzkry:
But this pull request seems ok, doesn't it? Do you think it would be remerged?

PR #2455, opened 2019-09-10
Move char utilities from gtest-port.h to gtest-internal.h

kuzkry:
Some portable functions were a part of gtest-port.h.

Additionally, I've done a small refactor of StripTrailingSpaces. There is also a [benchmark](http://quick-bench.com/2_fS9fbJSGz-nYiVAe4LD-J-yNI), but it doesn't affect much to be honest so I'd rather consider code expressibility.

### DIFF
--- a/googletest/include/gtest/internal/gtest-internal.h
+++ b/googletest/include/gtest/internal/gtest-internal.h
@@ -52,6 +52,7 @@
 
 #include <float.h>
 #include <string.h>
+#include <algorithm>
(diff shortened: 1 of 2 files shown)


asoffer:
Thanks, for this. I think a better long-term solution is to rely on Abseil Strings for this, so rather than merging this and then later deleting these facilities, it'd be simpler to just get what we need directly from Abseil.

PR #2456, opened 2019-09-10
Remove GTEST_HAS_STD_STRING

kuzkry:
1st commit removes GTEST_HAS_STD_STRING. I think we can drop it, after all it's 2019 :)
2nd commit simply removes usings from the header, which is how all guidelines say

Both changes are backward incompatible. I would merge them though because README.md says the following about [post 1.8.x releases](https://github.com/google/googletest#post-18x):
> On-going work to improve/cleanup/pay technical debt

and we don't quite need it, right?

### DIFF
--- a/googletest/include/gtest/internal/gtest-port.h
+++ b/googletest/include/gtest/internal/gtest-port.h
@@ -439,15 +439,6 @@ typedef struct _RTL_CRITICAL_SECTION GTEST_CRITICAL_SECTION;
 # endif  // defined(_MSC_VER) || defined(__BORLANDC__)
 #endif  // GTEST_HAS_EXCEPTIONS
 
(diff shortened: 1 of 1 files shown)


gennadiycivil:
273567737

gennadiycivil:
@kuzkry 
So I run some internal tests, GTEST_HAS_STD_STRING looks OK ( I think but not 100% sure because of all of the noise from the other change ). 

There is another change in this PR, ::testing::tuple_size -> std::tuple_size. This one looks harder because there is internal code accumulated over the years that is still using ::testing::tuple_size.
May I suggest splitting these two changes into two separate PRs?
Thank you!
G

gennadiycivil:
Trying again
273993693

PR #2470, opened 2019-09-21
gmock_gen.py discards argument specifiers when they are defaulted

hermas55:
Example : 
```cpp
class Foo {
public:
  virtual void Bar( const int a = 42 ) ;
};
```
Expected output : 
```cpp
class MockFoo : public Foo {
public:
    MOCK_METHOD1(Bar, void(const int a));
};
```
Actual output (const and parameter name discarded) : 
```cpp
class MockFoo : public Foo {
public:
    MOCK_METHOD1(Bar, void(int));
};
```


hermas55:
@googlebot I signed it!

gennadiycivil:
@hermas55 The scripts in the /script directory are out of date and  not well supported anymore . I would be glad to merge this PR if you explain your usage.
Thank you
G

gennadiycivil:
@hermas55 ping?

hermas55:
Hello @gennadiycivil ,

Sure. There is as bug in the gmock_gen.py script. 
It's supposed to take in input a header file of the class you want to mock, and print the mocked class in output.

It works fine as long as the class in input doesn't have methods with default parameters. When there are default parameters, the ```const``` qualifier gets discarded.

Example of Input class : 
```cpp
class Foo {
public:
  virtual void Bar( const int a = 42 ) ;
};
```
Output before the fix **(const and parameter name discarded)** : 
```cpp
class MockFoo : public Foo {
public:
    MOCK_METHOD1(Bar, void(int));
};
```

The PR contains 3 commits : 
- PEP8 Automatic reformatting --> Only reformatting to comply with PEP8 standard
- Fix for compatibility with Python 3.7 --> Make the script work with both Python 2.x and Python 3.x
- Fix default parameters with specifiers issue --> Fix for the issue explained above + Unittests

Please let me know if you have any concerns.

Kind regards,
//HM

gennadiycivil:
@hermas55 Headsup, we are going to include the files under /scrips with the sync process that syncs the files internally and on GitHub. ( We will also include a disclaimer that these files are unsupported. )
All this means that there may be changes in the files you are working with. You may have to re-base this PR . we shall see

gennadiycivil:
@hermas55 please resolve conflicts

hermas55:
conflicts resolved

gennadiycivil:
(276062823)

Issue #2482, opened 2019-09-28
-pthread should probably be used instead of -lpthread


ienorand:
According to
https://stackoverflow.com/questions/23250863/difference-between-pthread-and-lpthread-while-compiling
https://stackoverflow.com/questions/2127797/significance-of-pthread-flag-when-compiling
-pthread should be used instead of -lpthread in order to get the correct preprocessor defines set.

This also affects pkg-config, which currently has -lpthread in Cflags (in additon to Libs), which is invalid (an error according to lintian on Debian).

This seems somewhat related to the closed issue #391.

https://stackoverflow.com/questions/1620918/cmake-and-libpthread contains some hints on how to achieve this in cmake.

gennadiycivil:
Corresponding PR has been closed. Can we close this issue?

ienorand:
The issue is still present but as discussed in #2483 I/we have given up trying to fix the issue with the current version of cmake.

I'm not sure what the policy regarding "wontfix-but-maybe-in-the-future" issues are? Closing to avoid clutter or keeping open to avoid duplicates?

Tachi107:
Isn't it possible to enable `THREADS_PREFER_PTHREAD_FLAG` only if the compiler is not nvcc?

higher-performance:
Closing per #2483. It's not clear what a proper fix would be and we don't have plans to work on this.

Tachi107:
On Tue Jul 21, 2026 at 4:06 PM -03, higher-performance wrote:
> Closing per #2483. It's not clear what a proper fix would be and we 
> don't have plans to work on this.

The pull request you linked mentions that `THREADS_PREFER_PTHREAD_FLAG` 
is the proper fix for CMake 3.13 and above. As GoogleTest requires 
CMake version 3.16, we could just set that and be done with this, no?

higher-performance:
The PR was closed so I assumed it doesn't work as written? If you know the correct solution (are you saying it's only the 1 line `set(THREADS_PREFER_PTHREAD_FLAG ON)`?), perhaps send another PR and notify me on the PR?

PR #2483, opened 2019-09-29
Use -pthread over -lpthread

ienorand:
Based on
https://gcc.gnu.org/onlinedocs/gcc/Preprocessor-Options.html#index-pthread
-pthread should be used both for compilation and linking, since -pthread
defines needed macros.

Update the cmake configuration to hard-code "-pthread" as a compile
option on cmake prior to 3.1.0 and to set THREADS_PREFER_PTHREAD_FLAG in
order to use -pthread automatically on cmake 3.1.0 and above.

Remove the shared thread_spec variable and set the link libraries and
compiler options directly, since now there is less commonality between
the pre-3.1.0 and post-3.1.0 cmake handling.

close: #2482


SoapGentoo:
@ienorand We've been down this road before, and I generally agree with your reasoning, unfortunately CMake is too braindamaged to stratify on this, and using `-pthread` during compilation breaks CUDA. See also https://github.com/google/googletest/pull/1977 and https://github.com/google/googletest/pull/2006

gennadiycivil:
Anything related to CMake is community supported. We do not use it internally and we are not in position to check. In general the maintainers would look for community consensus. Someone needs to review such PRs and comment on them

ienorand:
> @ienorand We've been down this road before, and I generally agree with your reasoning, unfortunately CMake is too braindamaged to stratify on this, and using `-pthread` during compilation breaks CUDA. See also #1977 and #2006

Sorry, I did not find those when searching.

So reading up on CUDA a bit it seems like -pthreads should be passed using -Xcompiler for nvcc as a special case, which seems to have been implemented only relatively recently in CMake according to https://gitlab.kitware.com/cmake/cmake/merge_requests/2512. Which seems to have ended up in Cmake 3.13.0.

I'm not sure if it's possible or desirable to do some nvcc workaround for older CMake versions in googletest as well?

In order to address -lpthreads being part of the compile flags. Could we remove 
Sorry, I did not find those when searching.

So reading up on CUDA a bit it seems like -pthreads should be passed using -Xcompiler for nvcc as a special case, which seems to have been implemented only relatively recently in CMake according to https://gitlab.kitware.com/cmake/cmake/merge_requests/2512. Which seems to have ended up in Cmake 3.13.0.

I'm not sure if it's possible or desirable to do some nvcc workaround for older CMake versions in googletest as well?

I've looked into how to shove TARGET Threads::Threads PROPERTY INTERFACE_COMPILE_OPTIONS into the generated pkg-config file instead, but I've not really understood how to evaluate the generator expression properly when generating the file...

In order to at least address -lpthreads being part of the compile flags. Could we for now remove @CMAKE_THREAD_LIBS_INIT@ from Cflags in the *.pc.in files. Or is there some case which does need this as well?

ienorand:
Ok, I've looked into using something like:

```
set(THREADS_PREFER_PTHREAD_FLAG ON)        
find_package(Threads)             
(...)
get_target_property(THREADS_COMPILE_OPTIONS Threads::Threads INTERFACE_COMPILE_OPTIONS)
get_target_property(THREADS_LINK_LIBS Threads::Threads INTERFACE_LINK_LIBRARIES)
set(PKGCONFIG_THREADS_COMPILE_OPTIONS "$<JOIN:${THREADS_COMPILE_OPTIONS}, >")
set(PKGCONFIG_THREADS_LINK_LIBS "$<JOIN:${THREADS_LINK_LIBS}, >")
configure_file("${CMAKE_CURRENT_BINARY_DIR}/foo.in" "${CMAKE_CURRENT_BINARY_DIR}/foo.cfg")
file(GENERATE OUTPUT "${CMAKE_CURRENT_BINARY_DIR}/foo.pc" INPUT "${CMAKE_CURRENT_BINARY_DIR}/foo.cfg")
```

But this still ends up with the expanded CUDA string as  `SHELL:-Xcompiler -pthread` where the `SHELL:` prefix is a no-go. So there would still be an added step needed to get rid of that.

For now I think I'll give up on this venture to do it properly and I'll open a new pull request to just remove `-lpthread` from `Cflags:`. Or do you think it would be better to "hijack" this pull request for that change?

gennadiycivil:
@ienorand @SoapGentoo any updates on this one? Thank you

SoapGentoo:
@gennadiycivil I'm good with it.

gennadiycivil:
> For now I think I'll give up on this venture to do it properly and I'll open a new pull request to just For now I think I'll give up on this venture to do it properly and I'll open a new pull request to just remove remove `-lpthread` from `Cflags:`. Or do you think it would be better to "hijack" this pull request for that change?

The above leads me to believe that this PR is to be closed. Closing. Let me know if I misunderstood.
Thanks

ienorand:
For posterity, some of my issues/attempts with getting thread flags from cmake to the generated pkg-config files are referenced in https://gitlab.kitware.com/cmake/cmake/issues/19848 .

Indeed, cmake is currently not particularly helpful when it comes to generating pkg-config files...

SoapGentoo:
On a related note, I just tried `nvcc` again and can confirm that it fails on `-pthread`.

PR #2491, opened 2019-10-05
Fix gnu install dirs pkg config

SoapGentoo:
Hi @gennadiycivil 
the pkg-config change from last october breaks valid `GNUInstallDirs` uses as discussed in https://github.com/google/googletest/commit/095b3113e7438fd0be901e20b40d376bcd8ef860#r35373291
This reverts the change and adds documentation how to use pkg-config in a cross-compilation setting.


gennadiycivil:
Anything related to CMake is community supported. We do not use it internally and we are not in position to check. In general the maintainers would look for community consensus. Someone needs to review such PRs and comment on them

gennadiycivil:
273339550

gennadiycivil:

PR #2495, opened 2019-10-07
Fix typo in documents

hyukmyeong:
Fix typo in documents


gennadiycivil:
273343132

PR #2497, opened 2019-10-07
Use FormatFileLocation for streaming file and line

### LINKED ISSUE #819: Test Event: OnTestPartResult() gives C++ exception upon not finding file_name

thejcannon:
This will fix null file names potentially crashing the program.

Fixes #819 as well.

### DIFF
--- a/googlemock/src/gmock-spec-builders.cc
+++ b/googlemock/src/gmock-spec-builders.cc
@@ -34,6 +34,7 @@
 // EXPECT_CALL).
 
 #include "gmock/gmock-spec-builders.h"
+#include "gtest/internal/gtest-port.h"
 
 #include <stdlib.h>
 #include <iostream>  // NOLINT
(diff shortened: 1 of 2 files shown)


gennadiycivil:
273565699

Issue #2490, opened 2019-10-04
MOCK_METHOD macro not working with clang-cl


Corristo:
## My environment
 - Windows 10 Pro version 1809 OS build 17763.775
 - cmake 3.12.18081601-MSVC_2
 - cl version 19.16.27034
 - clang-cl versions 7.0.0 (tags/RELEASE_700/final),  8.0.0 (tags/RELEASE_800/final) and 9.0.0 (https://github.com/llvm/llvm-project.git 0399d5a9682b3cef71c653373e38890c63c4c365) (self-built clang from 9.0.0 tag)
  - googletest 1.10.0  
## Steps to reproduce
 - Start the x64 Native Tools Command Prompt for VS 2017
 - Ensure clang-cl can be found in the PATH
 - While in the top-level googletest directory, try to build using
    ```
    mkdir build
    cd build
    cmake -Dgmock_build_tests=ON -DCMAKE_C_COMPILER=clang-cl -DCMAKE_CXX_COMPILER=clang-cl -G "NMake Makefiles" ..
    cmake --build .
    ``` 

## Error output (truncated)
```
[ 43%] Building CXX object googlemock/CMakeFiles/gmock-function-mocker_test.dir/test/gmock-function-mocker_test.cc.obj
C:\TEMP\googletest\googlemock\test\gmock-function-mocker_test.cc(127,3): error: static_assert failed
      "(int n) should be enclosed in parentheses."
  MOCK_METHOD(void, VoidReturning, (int n));  // NOLINT
  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
C:\TEMP\googletest\googlemock\include\gmock/gmock-function-mocker.h(43,3): note: expanded from macro
      'MOCK_METHOD'
  GMOCK_PP_VARIADIC_CALL(GMOCK_INTERNAL_MOCK_METHOD_ARG_, __VA_ARGS__)
  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
C:\TEMP\googletest\googlemock\include\gmock/internal/gmock-pp.h(82,3): note: expanded from macro
      'GMOCK_PP_VARIADIC_CALL'
  GMOCK_PP_CAT(_Macro, GMOCK_PP_NARG(__VA_ARGS__))(__VA_ARGS__)
[...]

PR #2498, opened 2019-10-07
Fix MOCK_METHOD to handle noexcept correctly

### LINKED ISSUE #2472: GMock doesn't support noexcept functions
Hi, 

What is the current plan for supporting mocking `noexcept` functions? From searching StackOverflow, I understand that it's currently not possible to mock `noexcept` functions. Is there a workaround? 

Thanks

thejcannon:
Godbolt link to the preprocessed output of using the macro:
https://godbolt.org/z/rptA9J
(Scroll down to the bottom of the preprocessed output to see it parrotting the `noexcept` specifier verbatim)

Fixes #2472 

Let me know if y'all can think of any other test cases to throw at it.


googlemock/test/gmock-function-mocker_test.cc
```diff
@@ -656,5 +657,27 @@ TEST(MockMethodMockFunctionTest, MockMethodSizeOverhead) {
...
+  MOCK_METHOD(void, func3, (), (noexcept(false)));
+  MOCK_METHOD(void, func4, (), (noexcept(noexcept(1+1))));
+  MOCK_METHOD(void, func5, (), (const, noexcept(noexcept(1+1))));
+  MOCK_METHOD(void, func6, (), (noexcept(noexcept(1+1)), const));
+  // Put commas in the noexcept expression
+  MOCK_METHOD(void, func7, (), (noexcept(noexcept(hasTwoParams(1,2))), const));
+};
+
+TEST(MockMethodMockFunctionTest, NoexceptSpecifierPreserved) {
+  EXPECT_TRUE(noexcept(std::declval<MockMethodNoexceptSpecifier>().func1()));
```
ugoren: Is there a reason to use declval here? I'd simply define a MockMethodNoexceptSpecifier object.
In case the compiler warns that it's unused, it can be suppressed by cast to void.

googlemock/test/gmock-function-mocker_test.cc
```diff
@@ -656,5 +657,27 @@ TEST(MockMethodMockFunctionTest, MockMethodSizeOverhead) {
...
+  EXPECT_TRUE(noexcept(std::declval<MockMethodNoexceptSpecifier>().func2()));
+  EXPECT_FALSE(noexcept(std::declval<MockMethodNoexceptSpecifier>().func3()));
+  EXPECT_EQ(noexcept(std::declval<MockMethodNoexceptSpecifier>().func4()), noexcept(1+1));
+  EXPECT_EQ(noexcept(std::declval<MockMethodNoexceptSpecifier>().func5()), noexcept(1+1));
+  EXPECT_EQ(noexcept(std::declval<MockMethodNoexceptSpecifier>().func6()), noexcept(1+1));
+  EXPECT_EQ(noexcept(std::declval<MockMethodNoexceptSpecifier>().func7()), noexcept(hasTwoParams(1,2)));
+}
```
ugoren: I see no point in comparing with `noexcept(1+1)` - it simply evaluates to true. I suggest:

```
void Throws();
void DoesntThrow() noexcept;
struct MockMethodNoexceptSpecifier {
  MOCK_METHOD(void, func4, (), (noexcept(Throws())));
  MOCK_METHOD(void, func5, (), (noexcept(DoesntThrow())));
};
TEST(...) {
  EXPECT_FALSE(noexcept(mock.func4()));
  EXPECT_TRUE(noexcept(mock.func5()));
}
```

googlemock/test/gmock-function-mocker_test.cc
thejcannon: I had that previously, but the compiler would always say the result of the expression is `noexcept(false)`.
https://travis-ci.org/google/googletest/jobs/594800623#L2313
My guess would be it has to do with the constructor not being marked `noexcept`.
But `declval` really is the "correct" solution as we just want to test the `noexcept` specifier for the function (and not the default constructor as well).

Another solution would be to use `std::is_nothrow_invocable`, but I don't think there's a technical reason to prefer one over another.

thejcannon:
> Another small issue is behavior in case noexcept is used multiple times (which is invalid usage). I think it will lead to cryptic error messages, and better validation would be nice. But it isn't critical.

Seemingly it currently (before my change) leads to no error messages: https://godbolt.org/z/0TtYoG
So, some would argue that _any_ error message is better than none. I'll swing back around and work on a follow-up PR that makes the error message nice. It should be easy.

gennadiycivil:
@thejcannon I am going to import this into internal system so we can have an internal review.
273951497

googlemock/test/gmock-function-mocker_test.cc
ugoren: My intention was to define one MockMethodNoexceptSpecifier object in the test, and use it in all expectations. Then it wouldn't matter if the constructor is noexcept.

thejcannon: That's definitely another solution. Personally, I wouldn't like having to declare an instance of an object (which is a runtime construct) just so syntactically (since `noexcept` expressions are evaluated at compile-time) I can reason about the methods. 
But I also can't argue it would be less code.
If this comes up in the "internal review" I'd be alright changing it.

gennadiycivil:
@thejcannon Looks like this broke Windows builds.
See https://ci.appveyor.com/project/GoogleTestAppVeyor/googletest/builds/28041555

thejcannon:
@gennadiycivil 
So the good news is that the bug was there before my change: https://godbolt.org/z/WbouCI
Looks like my tests just exposed them.

I'll take a look at it when I get to work today.

Might be related to #2490

thejcannon:
#2514 has been created to fix the issue.

googlemock/test/gmock-function-mocker_test.cc
thejcannon: Yeah, it's all about tradeoffs if you want to use `static_assert`s in your test code to test some logic. 
If you move it into a compile-time check, you know it'll always be true. The tests wouldn't compile otherwise. On the other hand, normally one test failing doesn't bar other tests from running and reporting their successes/failures. Since the `noexcept` spec wouldn't be absolutely fatal if it was wrong, I opted to implement them as runtime checks so they can report successes/failures just like other tests.

PR #2509, opened 2019-10-10
Do not pollute the "testing" namespace with std::tuple and related names

kuzkry:
See #2456.


asoffer:
Definitely a welcome change, but as usual, this one will take a while to clean up all the internal `::testing::get` calls. I'm working on it, but to temper expectations, there looks to be a lot of cleanup and it's somewhat back-burner for me. I'll try to update regularly.

kuzkry:
@asoffer - do not mind the latest commit, it merely solves the conflict with headers.

asoffer:
As an update, I've totally let this fall through the cracks. I'm going to find a way to reinvest time into getting this done.

kuzkry:
Let's just say there's been a ton of other work for us both. Welcome back :)

asoffer:
Two years later, we still haven't made a push for this. It's unfortunate, but I just don't think we're going to be able to prioritize this effort. Thanks for the effort, on this PR and in general, you've done a lot for the health of the googletest codebase. Your work is invaluable and very much appreciated, even if we cannot always find a way to accept the PRs you send.

SaaadSaadAl:
G

PR #2510, opened 2019-10-10
Update PlatformIO Version

ciband:
The PlatformIO library version needs to be updated to match the latest release.


ciband:
@kuzkry Sorry I was late on this.  Thanks for taking care of it.  I will review.

asoffer:
Pulling into internal changelist

PR #2512, opened 2019-10-11
Fix some broken links in docs

hyukmyeong:
Fixed some broken links in docs
Because it seems like below markdown syntax is not working on github

(e.g.)
```
< temp1.md >
#### Complex Subject {#Summary}

<temp2.md>
make a [link](temp1.md#Summary) to temp1.md
```


hyukmyeong:
@googlebot

asoffer:
These links are used in an internal flavor of markdown and it would be nice to keep the same anchors. Can you modify the headers to use these anchors in a way that works both internally and externally? For example,

### [My header](#MyTag)

hyukmyeong:
I can modify but I don't know how to let them work in both ways yet
Do you mean below ?

```
< temp1.md >
### [My header](#MyTag)

<temp2.md>
[link](temp1.md#MyTag)
```

asoffer:
Yes, this will work

Before:
`### [My header] {#MyTag}`

After:
`### [My header](#MyTag)`

hyukmyeong:
There are 2 things I should check before I make a new commit
Could you please confirm below things are OK after applying your solution?

1. **Making fake link** : because headers will be looked like links , even though they don't link to anywhere
2. **Unity problem** : headers will have different colors (links are blue and others are black) as you know some of headers have tags but most of them don't at the time

asoffer:
Internally we use a home-rolled markdown tool.

You're correct that they would look like links but not jump anywhere.

I'm going to look into an alternate solution where we transform the source files before pushing to github.

hyukmyeong:
Hello~

Do you have any update for this pull request?

PR #2515, opened 2019-10-14
Add support for esp8266

ciband:
* Added support for ESP8266
* Removed previous Arduino defines in favor
  of GTEST_OS_* syntax
* Improved PlatformIO library file to be more
  stable when consumed in client applications.


googletest/src/gtest_main.cc
```diff
@@ -30,13 +30,20 @@
 #include <cstdio>
 #include "gtest/gtest.h"
 
-#ifdef ARDUINO
+#if GTEST_OS_ESP8266 || GTEST_OS_ESP32
+# if GTEST_OS_ESP8266
+extern "C" {
```
asoffer: I'm a bit unsure about this.
Could you explain why it is necessary?

asoffer:
Thank you for this PR. While I don't expect any significant technical problems here, I'm worried about adding support from a maintenance perspective. This is not a commonly used platform at Google, and we don't have these devices to test on. Do you have suggestions for testing infrastructure (see or AppVeyor and Travis CI).

googletest/src/gtest_main.cc
ciband: This was a weird thing.  The linked on esp8266 want these to be extern C. Other platforms I’ve used do not required for Arduino setup/loop functions. 

I am definitely open to a better way.

ciband:
The PR should add the esp8266 to CI in Travis. It just extends the esp32 support. 

As far as actual running any unit tests, that is harder to automate  on embedded hardware. I assumed unit tests are only ran on supported CI platforms and not for more embedded platforms like Windows CE. 

Is building the code for the platform during CI good enough or is there another way to verify that I am not thing of?

Thanks for the consideration.

googletest/include/gtest/internal/gtest-port.h
```diff
@@ -2036,7 +2052,7 @@ inline int Close(int fd) { return close(fd); }
 inline const char* StrError(int errnum) { return strerror(errnum); }
 #endif
 inline const char* GetEnv(const char* name) {
-#if GTEST_OS_WINDOWS_MOBILE || GTEST_OS_WINDOWS_PHONE || GTEST_OS_WINDOWS_RT
+#if GTEST_OS_WINDOWS_MOBILE || GTEST_OS_WINDOWS_PHONE || GTEST_OS_WINDOWS_RT || GTEST_OS_ESP8266
   // We are on Windows CE, which has no environment variables.
```
vslashg: This comment is stale, could you please update it?

googletest/src/gtest-filepath.cc
```diff
@@ -236,7 +236,7 @@ bool FilePath::DirectoryExists() const {
     result = true;
   }
 #else
-  posix::StatStruct file_stat;
+  posix::StatStruct file_stat = {};
```
vslashg: This change feels unrelated to the rest of what's going on here.  Is this needed to make esp8266 work, or was this just a drive-by cleanup?

ciband: I think this was drive by clean up.  I can revert the line if desired.

googletest/include/gtest/internal/gtest-port.h
ciband: will do

vslashg:
Thanks.  We've started the internal review of this change; please don't make further changes to this PR, or they may be lost.

ciband:
Will do.  Thank you.

vslashg:
Merged into mainline.

PR #2527, opened 2019-10-22
Prevent using ReturnRef on reference to temporary

PiotrNycz:
Fixed issue: 2471

### DIFF
--- a/googlemock/test/gmock-actions_test.cc
+++ b/googlemock/test/gmock-actions_test.cc
@@ -46,6 +46,7 @@
 #include <iterator>
 #include <memory>
 #include <string>
+#include <type_traits>
 #include "gmock/gmock.h"
 #include "gmock/internal/gmock-port.h"
 #include "gtest/gtest.h"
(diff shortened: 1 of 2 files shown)


sbenzaquen:
Please include a test that triggers the bug and gets fixed with this change.

kuzkry:
How is he going to test non-compiling code? It isn't run-time.

sbenzaquen:
You can use `decltype` testing to evaluate the validity of an expression.
Similar to how `std::is_detected` is meant to work (but we can't use that because it is not available in C++11).

Example:

```
template <typename T, typename = decltype(ReturnRef(std::declval<T&&>()))>
bool CanCallReturnRef(T&&) { return true; }
bool CanCallReturnRef(Unused) { return false; }

...
int i = 0;
EXPECT_TRUE(CanCallReturnRef(i));
EXPECT_FALSE(CanCallReturnRef(i + 1));

```

kuzkry:
Ah, ok. I know this trick. Though I think the standard way of doing this is by using `void_t`, isn't it?

sbenzaquen:
> Ah, ok. I know this trick. Though I think the standard way of doing this is by using `void_t`, isn't it?

void_t is not available in C++11, but you can do without it too.

googlemock/test/gmock-actions_test.cc
```diff
@@ -646,6 +646,30 @@ TEST(ReturnRefTest, IsCovariant) {
   EXPECT_EQ(&derived, &a.Perform(std::make_tuple()));
 }
 
+namespace
```
sbenzaquen: We don't need an extra anonymous namespace.
The whole file is in one already.

googlemock/test/gmock-actions_test.cc
```diff
@@ -646,6 +646,30 @@ TEST(ReturnRefTest, IsCovariant) {
...
+  EXPECT_FALSE(CanCallReturnRef(std::move(value)));
+  EXPECT_FALSE(CanCallReturnRef(value + 1));
+  EXPECT_FALSE(CanCallReturnRef(123));
+}
+
+// Tests that ReturnRef(v) is not working with const temporaries (const T&&)
+TEST(ReturnRefTest, WillNotAcceptConstTemporaryAkaContRValueRef) {
+  const int value = 42;
+  EXPECT_TRUE(CanCallReturnRef(value));
+  EXPECT_FALSE(CanCallReturnRef(std::move(value)));
```
sbenzaquen: Is this the case that fails to compile now that compiled before?

PiotrNycz: Actually almost yes. 

The case for non-const case were "covered" by C++ language rule that r-value ref `T&&` does not bind to `T&` - so it would not compile even without the current change. 

For const  case, however, binding is allowed - BUT - basic types  - as int - are not good for this tests - because `const int foo();` it treated by C++ as `int foo();` - so better I will use non-POD type. See simplified example (and more in issue description): https://wandbox.org/permlink/MjjbW2G7KfqclRSX

```
#include <type_traits>
#include <utility>
#include <string>

template <typename T>
struct ActionRef {
    T& ref;
};
template <typename T>
ActionRef<T> make_action_ref(T& a)
{
    return {a};
}

template <typename T>
ActionRef<T> make_action_ref(T&&) = delete;


std::string non_const_func() { return "42"; }
const std::string const_func() { return "42"; }

int main() {
    // make_action_ref(non_const_func());    
    // make_action_ref(const_func());    // w/o ActionRef<T> make_action_ref(T&&) = delete;
                                                            // this line will compile

    std::string a = "1";
    const std::string b = "2";
    make_action_ref(a);
    make_action_ref(b);
}

```

PiotrNycz:
> "POD" is not really a relevant concept since C++11.
> Also, this bug was not related to pod-ness afaik. POD structs would still have triggered the bug.

You are right. The difference is between basic (integral) types and more complex classes. I'll change that (the names).

PiotrNycz:
> This is redundant with `CanCallReturnRef`.
> We should only have one.
> Is `HasReturnRefAction` giving something that `CanCallReturnRef` wasn't?

IMHO `HasReturnRefAction` adding more information to these tests - it is somehow easier to get the difference. But if you think it is redundant I can remove it, no problem at all.

sbenzaquen:
> > This is redundant with `CanCallReturnRef`.
> > We should only have one.
> > Is `HasReturnRefAction` giving something that `CanCallReturnRef` wasn't?
> 
> IMHO `HasReturnRefAction` adding more information to these tests - it is somehow easier to get the difference. But if you think it is redundant I can remove it, no problem at all.

I see it as a different way of asking the same question.

googlemock/test/gmock-actions_test.cc
```diff
@@ -646,6 +647,71 @@ TEST(ReturnRefTest, IsCovariant) {
...
+template<typename... Ts> using precpp17_void_t = typename precpp17_make_void<Ts...>::type;
+
+template <typename T, typename = void>
+struct HasReturnRefAction : std::false_type {};
+template <typename T>
+struct HasReturnRefAction<T, precpp17_void_t<decltype(ReturnRef(std::declval<T>()))>>
+    : std::true_type {};
+
+// Just an example of non-POD type
+class MyNonPodType {
```
sbenzaquen: Please move these types into the body of the test that uses them to avoid polluting the namespace scope.

PiotrNycz:
> > > This is redundant with `CanCallReturnRef`.
> > > We should only have one.
> > > Is `HasReturnRefAction` giving something that `CanCallReturnRef` wasn't?
> > 
> > 
> > IMHO `HasReturnRefAction` adding more information to these tests - it is somehow easier to get the difference. But if you think it is redundant I can remove it, no problem at all.
> 
> I see it as a different way of asking the same question.

Ok, I'll remove HasReturnRefAction with next commit.

googlemock/test/gmock-actions_test.cc
PiotrNycz: Ok. To avoid duplicated definitions (repeating the same types in two tests) I just use `int` for scalar types and `std::string` for non-scalar types. Is that OK?

googlemock/test/gmock-actions_test.cc
```diff
@@ -646,6 +647,40 @@ TEST(ReturnRefTest, IsCovariant) {
   EXPECT_EQ(&derived, &a.Perform(std::make_tuple()));
 }
 
+template <typename T, typename = decltype(ReturnRef(std::declval<T&&>()))>
+bool CanCallReturnRef(T&&) { return true; }
+bool CanCallReturnRef(Unused) { return false; }
+
+// Tests that ReturnRef(v) is working with non-temporaries (T&)
+TEST(ReturnRefTest, WorksForNonTemporary) {
+  int scalarValue = 123;
```
sbenzaquen: Please use snake_case syntax for variable names, as per the style guide for GoogleTest

googlemock/test/gmock-actions_test.cc
```diff
@@ -646,6 +647,40 @@ TEST(ReturnRefTest, IsCovariant) {
...
+
+// Tests that ReturnRef(v) is not working with temporaries (T&&)
+TEST(ReturnRefTest, DoesNotWorkForTemporary) {
+  auto scalarValue = []()  -> int { return 123; };
+  EXPECT_FALSE(CanCallReturnRef(scalarValue()));
+
+  auto nonScalarValue = []() -> std::string { return "ABC"; };
+  EXPECT_FALSE(CanCallReturnRef(nonScalarValue()));
+
+  // cannot use here callable returning "const scalar type" because C++ ignores such const for scalar return type, so the static_cast
```
sbenzaquen: Please wrap line to 80 chars.

kuzkry:
Do we really need 6 commits for that? :D I suggest squashing them all into one.

gennadiycivil:
(276737278)

PR #2530, opened 2019-10-23
Added CPack functionality for generating RPM or DEB packages

bf-bryants:
ie: run `cpack -G RPM` (or "DEB") after building with CMake.

The current configuration produces _only_ a development package with header, static library, and config files for CMake and pkgconfig.  There is no package with a shared library.

Testing status:
- RPM tested on OpenSUSE Leap 15.1
- DEB tested on Ubuntu Core 19.04
- NSIS (Windows/setup.exe) not tested, but might work anyway.


gennadiycivil:
@bf-bryants Thank you for your contribution. Please note that the CLA is required

PR #2534, opened 2019-10-23
TYPED_TEST_SUITE - outdated documentation

fletcher-blight:
The sample6_unittest.cc differs from the documentation here. The documentation has 'TYPED_TEST_SUITE' (which fails to build in my own projects) where the sample unit test has 'TYPED_TEST_CASE' (and this builds and runs fine for me).

I'm not sure if this is a legacy thing or not but I always forget this every time I come to the documentation, and then wonder why unit test doesn't compile.

### DIFF
--- a/googletest/docs/advanced.md
+++ b/googletest/docs/advanced.md
@@ -1531,10 +1531,10 @@ each type in the list:
 
 ```c++
 using MyTypes = ::testing::Types<char, int, unsigned int>;
-TYPED_TEST_SUITE(FooTest, MyTypes);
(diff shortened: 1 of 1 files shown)


fletcher-blight:
@googlebot I signed it!

gennadiycivil:
@fletcher-blight TYPED_TEST_SUITE is [correct](https://github.com/google/googletest/blob/bbe4b7363bbaa7c9e07dbacde11602a4c69504d3/googletest/include/gtest/gtest-typed-test.h#L59). TYPED_TEST_CASE is old and deprecated API.
You are probably building against older version of googletest.

PR #2542, opened 2019-10-29
Remove inclusion of deprecated C++ headers

alex1989hu:
Tested **(gmock_build_tests and gtest_build_tests)** within the following environments:
 - macOS: Apple clang version 11.0.0 (clang-1100.0.33.8)
 - SL7.6: gcc (GCC) 4.8.5 20150623 (Red Hat 4.8.5-36)
 - Ubuntu 18.04: gcc (Ubuntu 7.4.0-1ubuntu1~18.04.1) 7.4.0
 - CentOS8: gcc (GCC) 8.2.1 20180905 (Red Hat 8.2.1-3)

### DIFF
--- a/googlemock/include/gmock/gmock-actions.h
+++ b/googlemock/include/gmock/gmock-actions.h
@@ -38,7 +38,7 @@
 #define GMOCK_INCLUDE_GMOCK_GMOCK_ACTIONS_H_
 
 #ifndef _WIN32_WCE
-# include <errno.h>
+# include <cerrno>
 #endif
 
 #include <algorithm>
--- a/googlemock/include/gmock/gmock-cardinalities.h
+++ b/googlemock/include/gmock/gmock-cardinalities.h
@@ -39,7 +39,7 @@
 #ifndef GMOCK_INCLUDE_GMOCK_GMOCK_CARDINALITIES_H_
 #define GMOCK_INCLUDE_GMOCK_GMOCK_CARDINALITIES_H_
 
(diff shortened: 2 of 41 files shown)


vslashg:
Is there a motive behind this change other than purity?

I am concerned that this change will make googletest less portable.  In particular, including <foobar.h> puts the C library symbols in the global namespace, but including <cfoobar> only guarantees that they will be in the std namespace.  googletest isn't fully qualifying access to things like memcpy, for example.

alex1989hu:
@vslashg: thanks for review.

First of all there is no special case behind this PR. It is just a self motivation to update googletest - kinda follow up changes happened after pre C+11 compilers. I met more than enough legacy code (internal proprietary projects) where the original maintainer completely rejects to update (refresh) the code in the past years. Once we had to switch a newer operating system / new distribution I had to apply many changes - most of them were remove deprecated headers and include new ones; not a real code change. Since `v1.10` we have the opportunity to apply changes which reduces technical debt thanks to the minimum compiler C++11 compliance. To cover your portability concern I would say we can not prepare to unseen thing. I think there is a great community which greets changes related to making googletest compatible with a new target architecture. Furthermore, users who are stuck on old system can continue to use `v1.8` To sum up, I am happy to update the code to in another PR - if you specially listed those calls you mentioned.

kuzkry:
It wouldn't be _less portable_ if you added the missing **std::** qualifier to every function but this is likely rejectable because it'd have to blur our git history a bit.

PR #2548, opened 2019-11-02
Update pump_manual.md

### LINKED ISSUE #2321: pump_manual.md is out of date
> Variadic templates and variadic macros can alleviate the problem. However, while both are being considered by the C++ committee, neither is in the standard yet or widely supported by compilers.

Not only have variadic macros and templates been added to the standard, support for them is widespread among compilers at this point.

kuzkry:
Closes #2321.


zhangxy988:
Thank you, we have started internal review. Please don’t push any more changes into this PR as they might be overwritten. (278391763)

PR #2549, opened 2019-11-02
pump.py: add support for Python 3

### LINKED ISSUE #2365: pump.py doesn't run with Python 3
Running [`pump.py`](../blob/master/googletest/scripts/pump.py) with Python 3 (`3.7.4` to be precise) results in the following error:
```
  File "pump.py", line 249
    print 'ERROR: %s expected at %s.' % (token_type, pos)
                                    ^
SyntaxError: invalid syntax
```

Since [Python 2.7 will not be maintained past 2020](https://pythonclock.org/) perhaps it is worth to rewrite the script for Python 3. (And possibly other scripts judging on comments at #2075 and version declared in [CONTRIBUTING.md](../blob/master/CONTRIBUTING.md#requirements-for-contributors).)

Perhaps it would be fairly easy to make it compatible with both versions?

What are the plans here considering the end of Python 2.7 maintenance?

kuzkry:
Closes #2365.

Use of _io_ module makes it work with Python 2 and 3. If it wasn't for Python 2, _io_ wouldn't be needed.


zhangxy988:
Thank you, we have started internal review. Please don’t push any more changes into this PR as they might be overwritten. (278391918)

PR #2550, opened 2019-11-02
pump.py: fix CRLF generation

### LINKED ISSUE #2366: pump.py issues with line end style
Running [`pump.py`](../blob/master/googletest/scripts/pump.py) on Windows generates files with Windows line end style (`<CR><LF>` instead of `<LF>`). In effect the generated file has every line changed.

Perhaps it would be better if the [`pump.py`](../blob/master/googletest/scripts/pump.py) would respect the line end style of the source `.pump` file? However, I'm not sure how it would interact with line ends which are consumed and not emitted.

kuzkry:
Closes #2366.


zhangxy988:
This seems duplicate of https://github.com/google/googletest/pull/2549.

kuzkry:
@zhangxy988, this isn't duplicate. This is just based on the other change, so it has a dependency to that, but this pull request solves a different issue (please check out the commit 6a1a919).

kuzkry:
@zhangxy988, please reopen this issue. This is a one-liner change (containing just 6a1a919), even though GitHub shows something else (especially now, when it's closed).

zhangxy988:
Sorry about that. I missed 6a1a919.
I will leave it to Andy who is the release manager this week.

kuzkry:
Huh :) Now I can rebase on master so Github should detect and remove this one commit that's been already merged.

kuzkry:
I want to take a step back here. I will continue the discussion in the issue.

PR #2554, opened 2019-11-05
Update documentation from "iff" to "if and only if" in gen_gtest_pred_impl.py

kuzkry:
After 7bd4a7f3 [gtest_pred_impl_unittest.cc](https://github.com/google/googletest/blob/8697709e0308af4cd5b09dc108480804e5447cf0/googletest/test/gtest_pred_impl_unittest.cc) used to be newer than its template.

Sorry, I haven't checked it out before :)

### DIFF
--- a/googletest/scripts/gen_gtest_pred_impl.py
+++ b/googletest/scripts/gen_gtest_pred_impl.py
@@ -544,10 +544,10 @@ class Predicate%(n)sTest : public testing::Test {
     }
   }
 
-  // true iff the test function is expected to run to finish.
+  // true if and only if the test function is expected to run to finish.
   static bool expected_to_finish_;
 
-  // true iff the test function did run to finish.
+  // true if and only if the test function did run to finish.
   static bool finished_;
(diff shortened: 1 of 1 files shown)


zhangxy988:
Could you please explain why you change "iff" to "if and only if"? Isn't "iff" just a shorted form of "if and only if"?

kuzkry:
Ah, my favourite "iff"s xd

In short: seriously, not that many people know the meaning of "iff", and it even used to pass the review process in googletest...

Long story: It all started with me wrongfully thinking that "iff" is just a funny misprint of "if", so I changed them in commit bf6df7ea. Someone eventually pointed out that I was wrong, hence after a long discussion we decided that "if and only if" is the best option and 7bd4a7f3 came about.

Recently, I've noticed one place wasn't patched so this is to have it done once and for all.

zhangxy988:
Thanks for explaining.
We have started internal review. Please don’t push any more changes into this PR as they might be overwritten. (279381441)

PR #2556, opened 2019-11-05
pkg-config: Remove pthread link flag from Cflags

ienorand:
Remove the threads link library variable references from the pkg-config
Cflags: field, removing -lpthread(s) from the compile flags.

"-l*" linker flags should only be part of the Libs: section and should
not be part of the Cflags: section in pkg-config files.

This was first suggested in
https://github.com/google/googletest/pull/2006 and further discussed in
https://github.com/google/googletest/pull/2483 .


asoffer:
Google does not use CMake internally and we are not in the best position to vet these changes. Generally CMake changes are considered to be "community supported." This change looks reasonable, so I am accepting it as-is.

PR #2557, opened 2019-11-06
Use .has_value() instead of operator bool

initcrash:
We think that implicit conversions and especially to
bool are a terrible idea.
Google style guide agrees:
https://google.github.io/styleguide/cppguide.html#Implicit_Conversions

For this reason we have our own optional<T> that does not have
the implicit conversion to bool and would still like to use this matcher.
As std::optional also has .has_value() this should still work.

### DIFF
--- a/googlemock/include/gmock/gmock-matchers.h
+++ b/googlemock/include/gmock/gmock-matchers.h
@@ -3283,7 +3283,7 @@ class OptionalMatcher {
 
     bool MatchAndExplain(Optional optional,
                          MatchResultListener* listener) const override {
-      if (!optional) {
+      if (!optional.has_value()) {
         *listener << "which is not engaged";
         return false;
       }


initcrash:
@googlebot I signed it!

sbenzaquen:
You are right that the style guide bans implicit conversions, but these are not implicit conversions.
`optional` uses an explicit conversion to `bool`, which does not have the same issues and it is not banned by the style guide.

Please note that the change as is will break other users. Not all implementations of the concept use the same spelling. For example, folly's and llvm's use `hasValue()` instead, but they both have the explicit conversion to bool.

asoffer:
Thanks for bringing this up. It's definitely a subtle point that this is a "contextual conversion" rather than an implicit one (it really looks implicit, but the pedantry is important here). As Sam mentioned, this will break folly and llvm's usage, so I don't think it would be beneficial to accept this change.

That being said, if code-cleanup is something you enjoy, Googletest could use a bunch of it and we do generally welcome PRs along these lines. If you have questions about where to start along these lines, please feel free to reach out and we can point you in the right direction.

PR #2569, opened 2019-11-16
Fix FlatTuple compilation on older msvc.

bgianfo:
googletest 1.10.0 fails to compile on msvc version 19.00.23917
with one compilation error:

src\googletest\include\gtest\internal\gtest-internal.h(1188) : error C2039:
'FlatTupleBase<testing::internal::FlatTuple<bool,bool>,testing::internal::IndexSequence<0,1> >':
is not a member of 'testing::internal::FlatTuple<bool,bool>'

This PR fixes the compilation error by explicitly specifying the full type that Indices is
located in the base type.

### DIFF
--- a/googletest/include/gtest/internal/gtest-internal.h
+++ b/googletest/include/gtest/internal/gtest-internal.h
@@ -1185,7 +1185,7 @@ struct FlatTupleBase<FlatTuple<T...>, IndexSequence<Idx...>>
 
 // Analog to std::tuple but with different tradeoffs.
 // This class minimizes the template instantiation depth, thus allowing more
-// elements that std::tuple would. std::tuple has been seen to require an
+// elements than std::tuple would. std::tuple has been seen to require an
 // instantiation depth of more than 10x the number of elements in some
 // implementations.
 // FlatTuple and ElemFromList are not recursive and have a fixed depth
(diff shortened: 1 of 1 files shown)


googletest/include/gtest/internal/gtest-internal.h
```diff
@@ -1185,7 +1185,7 @@ struct FlatTupleBase<FlatTuple<T...>, IndexSequence<Idx...>>
 
 // Analog to std::tuple but with different tradeoffs.
 // This class minimizes the template instantiation depth, thus allowing more
-// elements that std::tuple would. std::tuple has been seen to require an
+// elements than std::tuple would. std::tuple has been seen to require an
 // instantiation depth of more than 10x the number of elements in some
 // implementations.
 // FlatTuple and ElemFromList are not recursive and have a fixed depth
```
jaykrell: "it is" a few lines below should be "is".
"ln" after that should probably be "lg". i.e. log base 2 instead of log base e. Just guessing.

gennadiycivil:
@bgianfo Thank you for the patch. Could you please explain exactly what version of Visual Studio, etc this is?
Thanks

bgianfo:
@gennadiycivil The MSVC compiler I was using is from Visual Studio 2015, version 14.0. https://en.wikipedia.org/wiki/Microsoft_Visual_C%2B%2B

I didn't do an exhaustive test to see at which MSVC version master starts building clean.

gennadiycivil:

PR #2570, opened 2019-11-16
Fix internal memory leak in Windows _Crt report.

xieyubo:
We use "MemoryIsNotDeallocated" to aovid internal expected leak reported
in Windows _Crt report, like:

```cpp
    {
    #ifdef _MSC_VER
    	MemoryIsNotDeallocated memory_is_not_deeallocated;
    #endif

    	static ThreadIdToThreadLocals* map = new ThreadIdToThreadLocals();
	return map;
    }
```

But int the above code, only "new ThreadIdToThreadLocals()" is
protected, if we invoke "insert()" function of the return value,
the memory allocated in "insert()" will be reported to _Crt report
also. This change try to fix this issue.

### DIFF
--- a/googletest/src/gtest-port.cc
+++ b/googletest/src/gtest-port.cc
@@ -536,6 +536,9 @@ class ThreadLocalRegistryImpl {
   // Returns a value that can be used to identify the thread from other threads.
   static ThreadLocalValueHolderBase* GetValueOnCurrentThread(
       const ThreadLocalBase* thread_local_instance) {
+#ifdef _MSC_VER
+    MemoryIsNotDeallocated memory_is_not_deallocated;
+#endif  // _MSC_VER
     DWORD current_thread = ::GetCurrentThreadId();
     MutexLock lock(&mutex_);
     ThreadIdToThreadLocals* const thread_to_thread_locals =


xieyubo:
@googlebot I signed it!

gennadiycivil:
@xieyubo Thank you for this patch. There is not enough information here, could you please provide detailed explanation , when does the error get reported, by what tools, etc.

xieyubo:
It is found in our test. In our test, we enable the windows crt memory leak detector and get a report there are some memory leaks after we upgrated gtest from 1.7 to 1.10. No code change in our side at all, so I believe this leak is in gtest 1.10.  After some investigation, I found the leak came from `ThreadLocalRegistryImpl::GetValueOnCurrentThread()`. 

In `ThreadLocalRegistryImpl::GetValueOnCurrentThread()` function, it invokes `GetThreadLocalsMapLocked` to get an `std::map (ThreadIdToThreadLocals)`, the memory allocation for the map itself is protected under `MemoryIsNotDeallocated`, so this memory leak would be ignored:

```cpp
    mutex_.AssertHeld();
#ifdef _MSC_VER
    MemoryIsNotDeallocated memory_is_not_deallocated;
#endif  // _MSC_VER
    static ThreadIdToThreadLocals* map = new ThreadIdToThreadLocals();
    return map;
  }
```

But after this function return, `GetValueOnCurrentThread()` actually invokes `insert()` on the map, and the memory allocated by `insert()` are not protected under `MemoryIsNotDeallocated`. And because the map won't be released, so the memory used inside in the map won't be released also, so finally, we get a memory leak in report at the text exiting.

```cpp
if (thread_local_pos == thread_to_thread_locals->end()) {

      //-------------------------------------------------
      // Because thread_to_thread_locals won't be released, so the memory
      // allocated from `thread_to_thread_locals->insert` won't be released also.
      thread_local_pos = thread_to_thread_locals->insert(
      //------------------------------------------^^^^^^

          std::make_pair(current_thread, ThreadLocalValues())).first;
      StartWatcherThreadFor(current_thread);
    }
```

gennadiycivil:
I need to trigger CI, so I will close this and re-open

gennadiycivil:
Merged via https://github.com/google/googletest/commit/af71b6607f9e63a0c70b59dcafe92a42a57e3a96

PR #2571, opened 2019-11-16
Fix internal memory leak in Windows _Crt report.

xieyubo:
We use "MemoryIsNotDeallocated" to aovid internal expected leak reported
in Windows _Crt report, like:

```cpp
    {
    #ifdef _MSC_VER
    	MemoryIsNotDeallocated memory_is_not_deeallocated;
    #endif

    	static ThreadIdToThreadLocals* map = new ThreadIdToThreadLocals();
	return map;
    }
```

But int the above code, only "new ThreadIdToThreadLocals()" is
protected, if we invoke "insert()" function of the return value,
the memory allocated in "insert()" will be reported to _Crt report
also. This change try to fix this issue.


gennadiycivil:
Duplicate of #2570

kuzkry:
@gennadiycivil - this pull is aimed to be delived to _master_. The other one (#2570) is targeted to _v1.10.x_.

PR #2580, opened 2019-11-20
Allow using other frameworks supported by PlatformIO

valeros:
There is no reason to bound googletest only to the Arduino framework. This will allow using googletest for testing on a native machine as well.

### DIFF
--- a/library.json
+++ b/library.json
@@ -9,7 +9,7 @@
     "url": "https://github.com/google/googletest.git"
   },
   "version": "1.10.0",
-  "frameworks": "arduino",
+  "frameworks": "*",
   "platforms": [
         "espressif32",
         "espressif8266"


valeros:
@googlebot I signed it!

gennadiycivil:
@valeros Thank you for this patch.
This functionality is community supported. We do not use it internally and we are not in position to check. In general the maintainers would look for community consensus. Someone needs to review such PRs and comment on them

gennadiycivil:
@ciband ping?

gennadiycivil:
(281975856)

ciband:
I'm all for the feature.  I limited the original set to arduino just because that was all I tested and was setup for.  My vote would be to add supported platforms explicitly rather than a wildcard.

Can you add test cases to the CI to show that esp8266 and/or esp32 work under your proposed framework?

valeros:
Hi @ciband ! 
> My vote would be to add supported platforms explicitly rather than a wildcard.

The reason why I added that wildcard is because it's the easiest way to allow using this library for testing on a native machine **without** any framework specified. Why force users to compile this library for the target device if they may want to run unit tests for platform-independent code only on the host machine?

ciband:
@valeros 
I am still relatively new to PlatformIO so I went to the docs to brush up.  
The platform field is where I was thinking about adding the native support, so just adding the "native" tag.
The frameworks is a bit of a confusion to me on desktop OS.  It doesn't seem like it is used.  If that is the case then I would rather see the "native" tag be added to the supported platforms (with CI test cases) and then leave the frameworks as is.

My main concern is just not advertising support that we don't have.

If I've misunderstood something about how the system works, please correct me.  

Thank you.

gennadiycivil:
@valeros ping?

valeros:
@ciband  Sorry for the late reply. The `native` tag won't help as there is no such thing as Arduino framework for the `native` platform. I do understand your concern, but IMO, I don't think that specifying `*` in framework field is advertising anything, it's just a convenient way for users to run GoogleTest on native platform without `framework` field in `platformio.ini` file.

ciband:
@valeros 
Looking at the PIO example for native (granted a pretty thin example) they don't use framework at all to denote a "native" platform target.  

https://github.com/platformio/platform-native/blob/master/examples/hello-world/platformio.ini

They use platform.  I'm all for supporting the native environment in PIO but just wildcard-ing the frameworks field doesn't do this.

Even if we let the '*' in frameworks in, this change should still add 'native' to platforms and provide CI modifications to show that it will at least build.  We don't really have anything in place to run the googletest unit tests on the embedded platforms but I would also assume you would test this new functionality on your own project before signing off on the PR.

If we can get these features added to the PR, we can talk about the frameworks tag some more.

These are my thoughts.  @gennadiycivil , please feel free to chime in since you seem to be the product owner here (certainly more than me :) ).

valeros:
> Even if we let the '*' in frameworks in, this change should still add 'native' to platforms and provide CI modifications to show that it will at least build. 

Actually, the default `value for lib_compat_mode` is `soft` which means that PlatformIO will check only framework compatibility, so the `platform` field makes no difference.

gennadiycivil:
@valeros @ciband I have no opinion on PlatformIO, we dont use it and we do not test on it - this is community ( seem to be just the two of you ) supported.

I don't see a consensus between two of you and I am leaning to just close this PR. Please let me know otherwise.

ciband:
@valeros I am pretty firm on my stance about proving compatibility with CI builds.  To keep this PR going, we need at least that.  I'm all for having the support for we need some minimum level of proof that it actually works.

@gennadiycivil I vote to close this PR unless @valeros agrees to make the requested changes.

gennadiycivil:
Closing

puzrin:
@ciband IMO fields `platforms` & `frameworks` should be completely removed. Those add nothing except problems :). I can make PR if you have no objections.

ciband:
We use these on the ark.io project to run unit tests on IoT hardware. 

Can you share the motivation for removal?

Is there some specific problem they are causing?

Thanks.

puzrin:
Reason is very simple - gtest has no dependencies on platform and framework. IMO goal of "library.json" should be to whitelist "exported" files, and nothing else.

In this scope `platforms` & `frameworks` do not add values (those do not protect from anything). But those cause error when someone tries build native / ststm32, ... or tries different (non-aruino) framework or framework-less mode.

PS. My projects uses stm32 MCUs, and i use 'native' for testing and fast GUI development.

ciband:
It does not add any dependency.  It only adds support for additional platforms using the existing methods. 

For desktop users of GTest, there is no issue. 

For supported embedded platforms (currently Arduino esp32 and Arduino esp8266), there is no issue. 

It is no different than the existing support for windows CE or any other platform/operating system. 

If you try to build googletest on an unsupported platform, then it will fail. The goal should be to add support for as many platforms as possible. 

If you want stm32 support then why not add it?

For our use case, we have cross platform code that runs on Windows, Linux, MacOS, and Arduino so a single testing solution is desirable.

puzrin:
Probably we missunderstand each other due language barrier. Let me explain again.

> If you try to build googletest on an unsupported platform, then it will fail. The goal should be to add support for as many platforms as possible. If you want stm32 support then why not add it?

IMO this statemenet confuse "op-in" with "opt-out" principle. You suggest to "solve problem which not exists".

Current approach drops support of all existing PIO combination and then add back some. Why? That just adds difficulties [for others] without any benefits.

> If you want stm32 support then why not add it?

What do you mean with word "support"? By default gtest source code support all possible combination of frameworks/platforms in PIO. My question is, why `library.json` restricts that? This resriction looks artificial:

- it does not improve anything (does not add feature, does not fix bug)
- it ad difficulties to everyone, who wish use package with "missed" platform/framework.

Or do i miss something? Does gtest contains some specific code for esp & arduino?

ciband:
Ah, ok. I might understand. 

Googletest has only been validated on esp32 and esp8266. This is why there is a restriction. There are some conditional compile flags in the code for these platforms, just like all the other platforms like Windows and Linux. 

Googletest might work on other Arduino MCUs, either just by compiling or with just a few modifications. Since these other platforms have not been validated, it does not officially support them. I would like googletest to support as many as possible but I don’t want to advertise support for an untested platform. 

I would vote that if you know of a platform that would be officially supported and added to the JSON file then please test and submit a PR for this. 

The only reason for the restriction is due to limited availability to test the many available platforms.

puzrin:
> There are some conditional compile flags in the code for these platforms, just like all the other platforms like Windows and Linux.

Could you explain details? As far as i understand, the only real difference is Arduino, because it doesn't has `main()` entry https://github.com/google/googletest/blob/master/googletest/src/gtest_main.cc (use setup/loop instead).

All other cases have just something gcc-based (or clang), without deviations. In theory, it would be nice to test those compilers, but i think such requirement is a bit excessive. It would be more comfortable to solve such issues in lazy way - when those happen.

ciband:
Every platform’s compiler is a bit different so there are many unknowns. 

I don’t control googletest so I can’t say how things should or should not be. 

My vote is to keep opt-in support for the platforms. 

The other issue is that there are CI builds and tests to verify the platform. 

For any platform we want to support I would assume the googletest maintainers would require those basic tests.

puzrin:
> My vote is to keep opt-in support for the platforms.

I can understand your reasons, but we have very uncomforable side effect. Everyone why tries use gtest with "uncovered" pio env will get guaranteed build error.

> For any platform we want to support I would assume the googletest maintainers would require those basic tests.

I do no agree. PIO uses gcc/clang. That's good enougth. Gtest can live without support/guarantees on some envs. But strict disable of use without maintainers is strange.

Note, i don't ask you to care about problems of other platforms. You can say "only esp/arduino supported officially". I just ask to not create additional difficulties.

ciband:
I get that.  You have a fair point. 

Probably my biggest concern is someone wanting to use googletest, finding in PIO, seeing that it supports all platforms (or implies that it does) then being disappointed when it fails to compile on their platform. 

I would support a PR I f you could maybe also put in the Description field that only esp32 and esp8266 are officially supported and that we welcome patches to add other platforms. 

Sorry for the long conversation and I appreciate the different view point. 

Hopefully we can get to a common solution together. 

Thanks.

PR #2587, opened 2019-11-22
Remove g++ 3.4.2 workaround: change access specifiers

kuzkry:


kuzkry:
This bug is still present in gcc 5.5.

PR #2588, opened 2019-11-22
Change incorrect comments

kuzkry:
This isn't a bug in gcc. Compiler has to know which version of a template function to instantiate.


rogeeff:
Thank you, we have started internal review. Please don’t push any more changes into this PR as they might be overwritten. (282416437)

PR #2590, opened 2019-11-22
Remove stale comments about older GCC versions

kuzkry:
```c++
  // To work around a bug in gcc 2.95.0, there is intentionally no
  // space after the first comma in the previous statement.
```
Because in the past someone forgot to remove these.

```c++
  // Some older GCC versions may issue a spurious warning in this or the next
  // assertion statement. This warning should not be suppressed with
  // static_cast since the test verifies the ability to use bare NULL as the
  // expected parameter to the macro.
```
No problems found. I assume this isn't a problem any longer.


rogeeff:
Thank you, we have started internal review. Please don’t push any more changes into this PR as they might be overwritten. (282455859)

PR #2591, opened 2019-11-22
Remove g++ 3.3 workaround: using on operator<<

kuzkry:


rogeeff:
Thank you, we have started internal review. Please don’t push any more changes into this PR as they might be overwritten. (282451653)

PR #2592, opened 2019-11-22
Remove MSVC workaround: error C2665

kuzkry:


rogeeff:
Why did we have a workaround before? Why is it not relevant anymore?

kuzkry:
Because it probably is an unneeded workaround. Since I've got the latest MSVC, I trust AppVeyor to check me on this. Fortunately, the mere fact of using C++11 cuts down a lot of old MSVCs :)

mbxx:
Agreed, this is an old enough workaround that I believe googletest's C++11 requirement makes it obsolete.

Thanks!  I've started the internal review for this.

PR #2593, opened 2019-11-22
Remove MSVC workaround: accessing namespace scope from within nested class

kuzkry:


rogeeff:
Why do you believe we do not need this workaround anymore?

kuzkry:
Based on [Microsoft C++ language conformance table](https://docs.microsoft.com/en-us/cpp/overview/visual-cpp-language-conformance?view=vs-2019) you can check that the earliest implementation of C++11 is in VS 2015.
The comment in code says:
> VC++ prior to version 8.0 SP1

But VC++ 8.0 = Visual C++ 2005

rogeeff:
Thank you, we have started internal review. Please don’t push any more changes into this PR as they might be overwritten. (282590818)

PR #2595, opened 2019-11-22
Remove MSVC workaround: warning 4355

kuzkry:
I didn't have a problem with that even after adding /we4355 which would turn all 4355 warnings into errors.

### DIFF
--- a/googlemock/include/gmock/gmock-spec-builders.h
+++ b/googlemock/include/gmock/gmock-spec-builders.h
@@ -1350,12 +1350,6 @@ class ReferenceOrValueWrapper<T&> {
   T* value_ptr_;
 };
 
-// MSVC warns about using 'this' in base member initializer list, so
-// we need to temporarily disable the warning.  We have to do it for
-// the entire class to suppress the warning, even though it's about
-// the constructor only.
-GTEST_DISABLE_MSC_WARNINGS_PUSH_(4355)
-
 // C++ treats the void type specially.  For example, you cannot define
 // a void-typed variable or pass a void value to a function.
 // ActionResultHolder<T> holds a value of type T, where T must be a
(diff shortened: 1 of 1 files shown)


rogeeff:
Why should we do this? Is this warning is not relevant anymore?

rogeeff:
What MSVC version did you test it on?

kuzkry:
Sorry for taking so long to reply :)

> What MSVC version did you test it on?

Unfortunately, mine is close to the latest because I used Visual Studio 2019.
Sorry, it's not easy to get older versions to test it, therefore I believe our AppVeyor job suite running different MSVCs should be enough.

> Why should we do this? Is this warning is not relevant anymore?

Because it probably is an unneeded workaround. Since I've got the latest MSVC, I trust AppVeyor to check me on this. Fortunately, the mere fact of using C++11 cuts down a lot of old MSVCs :)

mbxx:
I believe that this warning is indeed obsolete because the code that had been generating the warning was changed.  The warning is specifically about using `this` in a constructor's initializer list (https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-c4355?view=vs-2019).  Back in 2011, this file contained the snippet:

```
FunctionMockerBase() : current_spec_(this) {}
```

Which is exactly the situation addressed by this warning.  In 2015, this snippet was refactored but the warning suppression wasn't removed, probably because it covers such a broad swath of code.

I suspect the MSVC version doesn't actually enter into the picture, but if you wanted to verify it, you could try to remove the other place in googletest that this warning is suppressed: https://github.com/google/googletest/blob/master/googletest/src/gtest.cc#L5024  In that location, the code is still using `this` to initialize a member variable.

Anyway, that was a long-winded way of saying: "Thanks for the clean up!  I'll start the internal review now."

PR #2596, opened 2019-11-22
Remove MSVC workaround: wmain link error

kuzkry:
Two commits here:
298a40f0 - removing workaround to the bug that has been patched
a909becd - just me secretly trying to smuggle some unification to both main files

### DIFF
--- a/googlemock/src/gmock_main.cc
+++ b/googlemock/src/gmock_main.cc
@@ -48,20 +48,13 @@ void loop() { RUN_ALL_TESTS(); }
 #endif
 
 #else
-
-// MS C++ compiler/linker has a bug on Windows (not on Windows CE), which
-// causes a link error when _tmain is defined in a static library and UNICODE
-// is enabled. For this reason instead of _tmain, main function is used on
-// Windows. See the following link to track the current status of this bug:
-// https://web.archive.org/web/20170912203238/connect.microsoft.com/VisualStudio/feedback/details/394464/wmain-link-error-in-the-static-library
-// // NOLINT
-#if GTEST_OS_WINDOWS_MOBILE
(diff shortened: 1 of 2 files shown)


rogeeff:
Why we do not need this workaround anymore?

kuzkry:
Based on [Microsoft C++ language conformance table](https://docs.microsoft.com/en-us/cpp/overview/visual-cpp-language-conformance?view=vs-2019) you can check that the earliest implementation of C++11 is in VS 2015.
The comment in code has a link to a bug that has been confirmed as fixed in VS 2010.

rogeeff:
For copy-pasting: “Thank you, we have started internal review. Please don’t push any more changes into this PR as they might be overwritten. (282585468)

googlemock/src/gmock_main.cc
```diff
@@ -48,21 +48,14 @@ void loop() { RUN_ALL_TESTS(); }
...
 #else
-
-// MS C++ compiler/linker has a bug on Windows (not on Windows CE), which
-// causes a link error when _tmain is defined in a static library and UNICODE
-// is enabled. For this reason instead of _tmain, main function is used on
-// Windows. See the following link to track the current status of this bug:
-// https://web.archive.org/web/20170912203238/connect.microsoft.com/VisualStudio/feedback/details/394464/wmain-link-error-in-the-static-library
-// // NOLINT
-#if GTEST_OS_WINDOWS_MOBILE
+#if __MSC_VER
```
chfast: I think it should have been `_MSC_VER`.

kuzkry: I don't know how you managed to find it :astonished:

PR #2597, opened 2019-11-22
Remove Nokia's Symbian compiler workaround: SafeMatcherCastImpl

kuzkry:
I strongly advise to enable some sort of "skip whitespace changes" when reviewing this diff.


rogeeff:
Thank you, we have started internal review. Please don’t push any more changes into this PR as they might be overwritten. (282454113)

PR #2607, opened 2019-11-27
Rename test case to test suite in docs/primer.md

oyefremov:


oyefremov:
@googlebot I signed it!

rogeeff:
Thank you, we have started internal review. Please don’t push any more changes into this PR as they might be overwritten. (282798019)

PR #2608, opened 2019-11-27
Fix incorrect use of __ in __MSC_VER

kuzkry:
Small fix for #2596. Thanks @chfast!


kuzkry:
Broken build means that the removal of the workaround was incorrect. Roll-back is needed.

PR #2609, opened 2019-11-27
Revert #2596

kuzkry:
Turns out that the bug fixed in Visual Studio 2010 is still present in Visual Studio 2014. This was found in #2608.


rogeeff:
Thank you, we have started internal review. Please don’t push any more changes into this PR as they might be overwritten. (282799204)

PR #2624, opened 2019-12-10
Revert change that disallows empty test prefixes

ShabbyX:
Reverted commits:

- 20b5b8ecc7a81d23b4716e22a2b35fd53379c8c6: "Googletest export": disallow empty prefix
- 4c25e2b87fcb78abcfdee2739db9ed7a19754cd3: "Googletest export": Remove test for empty prefix

Reason for revert: Breaks existing applications, such as ANGLE (angleproject.org), requires adding an extra prefix that needs to be typed for no reason (when testing from command line) and increases the logs' sizes by a non-trivial amount due to the very large number of tests ANGLE runs.

The original commit that disallowed empty test prefixes says:

> Add a compile time check to ensure that the parameters to TEST_P and INSTANTIATE_TEST_SUITE_P are not empty. Some compilers already fail in that case and, even where it works, it's likely to result in technically invalid code by virtue of creating reserved identifiers:
https://en.cppreference.com/w/cpp/language/identifiers

First, every project is perfectly capable of adding a prefix if they want to support such a compiler. This change penalizes every project.

Second, using a prefix such as `_p` also results in reserved identifiers, so this change is not really solving the problem.  For that matter, instead of generating `gtest_##prefix##...`, you can generate `gtest_x##prefix##...` to correctly fix the issue, including when empty prefixes are used.

### DIFF
--- a/googletest/include/gtest/gtest-param-test.h
+++ b/googletest/include/gtest/gtest-param-test.h
@@ -58,9 +58,7 @@ class FooTest : public ::testing::TestWithParam<const char*> {
 
 // Then, use the TEST_P macro to define as many parameterized tests
 // for this fixture as you want. The _P suffix is for "parameterized"
-// or "pattern", whichever you prefer to think. The arguments to the
-// TEST_P macro are the test_suite_name and test_case (both which must be
-// non-empty) that will form the test name.
+// or "pattern", whichever you prefer to think.
 
 TEST_P(FooTest, DoesBlah) {
   // Inside a test, access the test parameter with the GetParam() method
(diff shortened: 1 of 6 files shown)


gennadiycivil:
@ShabbyX thank you for the PR.
Could you please reference the actual commit - in the PR description that this is trying to revert so it is easy to trace and understand. As of now I spend a little time searching and could not find the actual commit this is trying to revert.

asoffer:
Thanks for the PR, and we apologize for the breakage. We'd like to get this rolled back. If you can get the AppVeyor CI passing, we'll pull this in as soon as possible.

ShabbyX:
@gennadiycivil Sure I can. If you look at the individual commits in the PR, they each say what commit it is reverting.

@asoffer, glad to hear you agree with this. I'll see what I can do about the CI. @null77 the failures are from visual studio, I may WFH the rest of the week (so no windows machine around). If you have some spare time, do you think you could take a quick look?

ShabbyX:
@asoffer, all green now. I had to work around a bug in Visual Studio with macro expansions which is the first commit in this change. That was necessary for a4a5a7c768ceaf8fdc79ccf1d9e3bfb9f481efa6 which landed after 20b5b8ecc7a81d23b4716e22a2b35fd53379c8c6 (which is being reverted). Curiously, the former contains the commit message of the latter, which makes me suspect the "compile failures" alluded to in the latter is indeed the one I fixed in the first commit of this PR (and was a necessary change for the former to land).

bcsgh:
@ShabbyX regarding you speculation around "compile failures"; you are correct, that MSVC difference is what prompted the change.

Issue #2628, opened 2019-12-12
Visual Studio 2019 16.4 GoogleMock SEH Exception.


alec500oo:
## Problem
The newest update of Microsoft Visual Studio 2019 (16.4) is causing all googlemock mock methods to fail with an SEH Exception 0xc0000005 thrown. The mock methods are throwing this exception at mock object destruction this leads to leaked mock objects, and consequently failed tests. This test was built from the `release-1.10.0` tag of this git repository. Mock methods are working correctly with compiled with GCC on Ubuntu Linux. 

## Minimum to Reproduce
The minimum code to reproduce is listed below. This example was build with `Microsoft (R) C/C++ Optimizing Compiler Version 19.24.28314 for x86`. CMake was used to generate the build files. gmock_main was used as the program's entry point.

### TestClass.h
```c++
#pragma once

class TestClass {
public:
	virtual void TestMethod() = 0;

	virtual ~TestClass() = default;
};
```

### MockTest.h
```c++

#pragma once

#include "gtest/gtest.h"
#include "gmock/gmock.h"

#include "TestClass.h"

class MockTest final : public TestClass {
public:
	MOCK_METHOD(void, TestMethod, (), (override));
};
```

### Test.cpp
```c++
#include "gtest/gtest.h"
#include "gmock/gmock.h"

#include "MockTest.h"

using namespace testing;

TEST(Test, t) {
	MockTest t;
	EXPECT_CALL(t, TestMethod()).Times(1);

	t.TestMethod();
}
```

### Test Output
```
Running main() from gmock_main.cc
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from Test
[ RUN      ] Test.t
unknown file: error: SEH exception with code 0xc0000005 thrown in the test body.
[  FAILED  ] Test.t (1 ms)
[----------] 1 test from Test (2 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test suite ran. (5 ms total)
[  PASSED  ] 0 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] Test.t

 1 FAILED TEST

..\..\..\test\Test.cpp(11): ERROR: this mock object (used in test Test.t) should be deleted but never is. Its address is @000000FB9D96F490.
[...]

PR #2635, opened 2019-12-14
Iterate untyped expectations vector back without reverse_iterator to workaround a bug  in VS 2019 16.4

macqm:
Workaround fix for #2628 
Using range based for-loop generates code that doesn't crash and the bug does not reproduce.

### DIFF
--- a/googlemock/include/gmock/gmock-spec-builders.h
+++ b/googlemock/include/gmock/gmock-spec-builders.h
@@ -1731,9 +1731,13 @@ class FunctionMocker<R(Args...)> final : public UntypedFunctionMockerBase {
     g_gmock_mutex.AssertHeld();
     // See the definition of untyped_expectations_ for why access to
     // it is unprotected here.
-    for (auto& untyped_expectations : untyped_expectations_) {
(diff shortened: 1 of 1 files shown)


ThadHouse:
This is failing testing because the original iterator is a reverse iterator, whereas range for goes forward. Below is code using forward iterators, but going backward.


```
    for (typename UntypedExpectations::const_iterator it =
             untyped_expectations_.end();
         it != untyped_expectations_.begin(); --it) {
      typename UntypedExpectations::const_iterator tmp = it;
      --tmp;
      TypedExpectation<F>* const exp =
          static_cast<TypedExpectation<F>*>(tmp->get());
      if (exp->ShouldHandleArguments(args)) {
        return exp;
      }
    }
```

macqm:
Thanks @ThadHouse, I pushed a commit with your suggestion.

gennadiycivil:
@macqm Thank you for this. I am not sure if we should be changing code to work around various bugs that various compilers may have.

gennadiycivil:
The codebase is correct C++ . If a compiler has an issue we should we working with the compiler to fix it.

PR #2639, opened 2019-12-17
Disable move constructor and assignment operator for test classes.

trzecieu:
Modernize generated test classes to follow CppCoreGuidelines, and fix common static code analyzers problems:

1. Define move operations (deleted) that are noexcept.
2. Define default destructor
3. Move deleted operations (copy constructors, assignment operations) to public interface

Disable move operations for `TEST()` and `TEST_F()` macros.
That is, a generated class will be decorated (addition to current) with:
```c++
class Foo {
  public:
  ~Foo() = default;
  Foo(Foo&&) noexcept = delete;
  Foo& operator=(Foo &&) = delete;
}
```
Previous implementation disabled only copy ctor and assing operator, but this was violating [rule of 5], which was captured by static code analysis tools like clang-tidy `cppcoreguidelines-special-member-functions`.

Check-list:
- [ ] compiled and tested according to: https://github.com/google/googletest/blob/master/CONTRIBUTING.md#testing-google-test-and-google-mock-themselves
- [x] tested solution with affected internal project

[rule of 5]: https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#c21-if-you-define-or-delete-any-default-operation-define-or-delete-them-all

### DIFF
--- a/googletest/include/gtest/internal/gtest-port.h
+++ b/googletest/include/gtest/internal/gtest-port.h
@@ -682,7 +682,7 @@ typedef struct _RTL_CRITICAL_SECTION GTEST_CRITICAL_SECTION;
 // A macro to disallow move operator=
 // This should be used in the private: declarations for a class.
 #define GTEST_DISALLOW_MOVE_ASSIGN_(type) \
-  type& operator=(type &&) noexcept = delete
+  type& operator=(type &&) = delete
 
 // A macro to disallow move constructor and operator=
 // This should be used in the private: declarations for a class.
--- a/googletest/include/gtest/internal/gtest-port.h
+++ b/googletest/include/gtest/internal/gtest-port.h
@@ -671,7 +671,7 @@ typedef struct _RTL_CRITICAL_SECTION GTEST_CRITICAL_SECTION;
 // A macro to disallow copy operator=
 // This should be used in the private: declarations for a class.
 #define GTEST_DISALLOW_ASSIGN_(type) \
-  void operator=(type const &) = delete
+  type& operator=(type const &) = delete
 
 // A macro to disallow copy constructor and operator=
 // This should be used in the private: declarations for a class.
@@ -682,7 +682,7 @@ typedef struct _RTL_CRITICAL_SECTION GTEST_CRITICAL_SECTION;
 // A macro to disallow move operator=
 // This should be used in the private: declarations for a class.
 #define GTEST_DISALLOW_MOVE_ASSIGN_(type) \
-  void operator=(type &&) = delete
+  type& operator=(type &&) = delete
 
 // A macro to disallow move constructor and operator=
 // This should be used in the private: declarations for a class.
--- a/googletest/include/gtest/internal/gtest-port.h
+++ b/googletest/include/gtest/internal/gtest-port.h
@@ -190,8 +190,10 @@
 //   GTEST_AMBIGUOUS_ELSE_BLOCKER_ - for disabling a gcc warning.
 //   GTEST_ATTRIBUTE_UNUSED_  - declares that a class' instances or a
 //                              variable don't have to be used.
-//   GTEST_DISALLOW_ASSIGN_   - disables operator=.
+//   GTEST_DISALLOW_ASSIGN_   - disables copy operator=.
 //   GTEST_DISALLOW_COPY_AND_ASSIGN_ - disables copy ctor and operator=.
+//   GTEST_DISALLOW_MOVE_ASSIGN_   - disables move operator=.
+//   GTEST_DISALLOW_MOVE_AND_ASSIGN_ - disables move ctor and operator=.
 //   GTEST_MUST_USE_RESULT_   - declares that a function's result must be used.
 //   GTEST_INTENTIONAL_CONST_COND_PUSH_ - start code section where MSVC C4127 is
 //                                        suppressed (constant conditional).
@@ -666,7 +668,7 @@ typedef struct _RTL_CRITICAL_SECTION GTEST_CRITICAL_SECTION;
 #endif
 
 
-// A macro to disallow operator=
+// A macro to disallow copy operator=
 // This should be used in the private: declarations for a class.
 #define GTEST_DISALLOW_ASSIGN_(type) \
   void operator=(type const &) = delete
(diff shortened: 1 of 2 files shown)


trzecieu:
@googlebot I signed it!

googletest/include/gtest/internal/gtest-port.h
```diff
@@ -677,6 +679,17 @@ typedef struct _RTL_CRITICAL_SECTION GTEST_CRITICAL_SECTION;
   type(type const &) = delete; \
   GTEST_DISALLOW_ASSIGN_(type)
 
+// A macro to disallow move operator=
+// This should be used in the private: declarations for a class.
```
N-Dekker: @trzecieu Hi Piotr, for me as a GTest user, I think your pull request could certainly be useful to get rid of those clang-tidy warnings. I would just like to mention that the current [Google Style Guide](https://google.github.io/styleguide/cppguide.html#Copyable_Movable_Types) says that those deleted member functions should be in the `public` section, instead of being declared `private`! Furthermore I noted that the code examples suggested that `operator=` should return a reference to the type, rather than `void`  I leave it up to you if you want to address these things, these are just my 2 cents. Kind regards, Niels

trzecieu: Thx, it's a good suggestion that I'll follow. Also I've found 2 possible smells in this area. Will come back with further improvements

trzecieu: Should be fixed now. With some asterisks, that I'm unable to test code as I'm in travel at the moment. Will continue verification once I settle somewhere for a while :) Added also 2 more modernizations, explained in the main description. 
Thanks!

asoffer:
I'm generally really happy with this change. I'm going to try pulling it in and seeing how much it breaks.
The possible next steps are:
A: Nothing breaks (unlikely) -- We'll accept the PR
B: Very little breaks (likely) -- We'll clean up the internals and then accept the PR
C: A lot breaks, but in ways that are easy to clean up -- We'll take some time but eventually accept the PR
D: A lot breaks, and there's no good way to automate the clean up -- We'll reasses the situation, but might come to the conclusion that it's not worth the effort.

googletest/include/gtest/internal/gtest-port.h
```diff
@@ -682,12 +682,12 @@ typedef struct _RTL_CRITICAL_SECTION GTEST_CRITICAL_SECTION;
 // A macro to disallow move operator=
 // This should be used in the private: declarations for a class.
 #define GTEST_DISALLOW_MOVE_ASSIGN_(type) \
-  type& operator=(type &&) = delete
+  type& operator=(type &&) noexcept = delete
```
N-Dekker: No show-stopper, but in which case could it useful to declare a deleted function `noexcept`? Just wondering...

trzecieu: Well spotted, just move constructor should be noexcept.

N-Dekker: Thanks @trzecieu But the example at https://google.github.io/styleguide/cppguide.html#Copyable_Movable_Types does not even have its deleted move constructor (`NotCopyableOrMovable`) marked `noexcept`. Of course, the example could be out-of-date. But I'm just wondering, does `noexcept` have any effect on a deleted function. Anyway, no showstopper to me.

googletest/include/gtest/internal/gtest-port.h
```diff
@@ -682,12 +682,12 @@ typedef struct _RTL_CRITICAL_SECTION GTEST_CRITICAL_SECTION;
...
```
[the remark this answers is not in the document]
trzecieu: Hi @asoffer, @N-Dekker.
Main motivation is to fix one sonarqube smell/bug in this area. At the moment I don't have an access to verify which exactly rule wad triggered, but after searching in SonarQube rules that one https://rules.sonarsource.com/cpp/RSPEC-5018?search=Noexcept looks looks the most familiar. 
It's supported by https://github.com/isocpp/CppCoreGuidelines/blob/036324/CppCoreGuidelines.md#c66-make-move-operations-noexcept

Indeed it looks odd to have deleted function noexcept, but still it's a part of API.

I'm not strongly pushing to have this in. Just wanted to clean all smells from generated code.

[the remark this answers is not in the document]
N-Dekker: @trzecieu I'm not _opposed_ to declaring deleted move-member-functions `noexcept`, it's just that I had not seen it before! While it's a common practice to declare a deleted assignment operator to return `type&`, instead of `void`, even though it will never return anything at all!

So you could argue that adding `noexcept` improves consistency between deleted and non-deleted move-member-functions. I leave it up you. Either way is acceptable to me, at the moment.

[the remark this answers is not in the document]
asoffer: I'm accepting the PR as is, because this is not a particularly important style point. But I've summarized my thoughts below, because I do think the broader issue of style-guide conformance and consistency is an important discussion to have.

The google style guide also suggests marking move-constructors as noexcept. What I think is different here is that all of these guidelines are about move constructors that are actually present in code. The reasoning is typically "moves are about performance" which doesn't apply for deleted move ctors/assignments.

The return value should be `T&` for move-assignment, because that's part of how people syntactically recognize that it is a move-assignment operator. The noexcept specification is not part of that recognition, which makes me think it's less important. I don't think consistency here is important, but rather being able to quickly identify the two important pieces: "move operator" and "deleted."

Anyway, as I said, I'm accepting the PR as is, so please don't make any further commits. Otherwise our automated machinery will get confused.

asoffer:
Surprisingly, it looks like the answer was (A): Not much breaks. Thanks for this change!

trzecieu:
Hi @asoffer, thank you for your support. I'm happy that the change can be merged. 
I might rushed with removal of `noexcept` keyword in my original change for move assignment operator (e3018bd)
Do you think this can be put back, as the Core Guideline covers it? As you've asked, I'm not gonna commit anything more to the PR.

asoffer:
This change got merged as fbe34cecf4ff34e0bed76aefb2b4d31f7b1f6819. I'm not sure why that did not close this PR, but yes, this has been merged. Thank you for the contribution!

Issue #2630, opened 2019-12-12
Compilation error on master using MinGW 8.2.0 and MSYS


Kravlalala:
I've configured project with the following command:
`cmake.exe -G "MSYS Makefiles" -B build .`
And then make gives me the following result:
```
$ make.exe 
Scanning dependencies of target gtest
[ 12%] Building CXX object googletest/CMakeFiles/gtest.dir/src/gtest-all.cc.obj
In file included from C:/Users/ikravchu/Documents/Templates/Tests/googletest/googletest/include/gtest/internal/gtest-internal.h:40,
                 from C:/Users/ikravchu/Documents/Templates/Tests/googletest/googletest/include/gtest/gtest.h:62,
                 from c:/Users/ikravchu/Documents/Templates/Tests/googletest/googletest/src/gtest-all.cc:38:
C:/Users/ikravchu/Documents/Templates/Tests/googletest/googletest/include/gtest/internal/gtest-port.h: In function 'int testing::internal::posix::StrCaseCmp(const char*, const char*)':
C:/Users/ikravchu/Documents/Templates/Tests/googletest/googletest/include/gtest/internal/gtest-port.h:1959:10: error: '_stricmp' was not declared in this scope
   return _stricmp(s1, s2);
          ^~~~~~~~
C:/Users/ikravchu/Documents/Templates/Tests/googletest/googletest/include/gtest/internal/gtest-port.h:1959:10: note: suggested alternative: 'strncmp'      
   return _stricmp(s1, s2);
          ^~~~~~~~
          strncmp
C:/Users/ikravchu/Documents/Templates/Tests/googletest/googletest/include/gtest/internal/gtest-port.h: In function 'char* testing::internal::posix::StrDup(const char*)':
C:/Users/ikravchu/Documents/Templates/Tests/googletest/googletest/include/gtest/internal/gtest-port.h:1961:47: error: '_strdup' was not declared in this scope
 inline char* StrDup(const char* src) { return _strdup(src); }
                                               ^~~~~~~
C:/Users/ikravchu/Documents/Templates/Tests/googletest/googletest/include/gtest/internal/gtest-port.h:1961:47: note: suggested alternative: 'StrDup'       
 inline char* StrDup(const char* src) { return _strdup(src); }
                                               ^~~~~~~
                                               StrDup
[...]

PR #2653, opened 2020-01-02
Adds POSIX define to allow use of POSIX extensions

mjvk:
Should fix things like #2632 #2101 #2630 in a platform independent manner.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,14 +10,14 @@ endif (POLICY CMP0048)
 project(googletest-distribution)
 set(GOOGLETEST_VERSION 1.10.0)
 
+
 if (CMAKE_VERSION VERSION_LESS "3.1")
-  add_definitions(-std=c++11)
+  add_definitions(-std=c++11 -D_POSIX_C_SOURCE=200809L)
 else()
   set(CMAKE_CXX_STANDARD 11)
   set(CMAKE_CXX_STANDARD_REQUIRED ON)
-  if(NOT CYGWIN)
-    set(CMAKE_CXX_EXTENSIONS OFF)
-  endif()
+  set(CMAKE_CXX_EXTENSIONS OFF)
+  add_compile_definitions(_POSIX_C_SOURCE=200809L)
 endif()
 
 enable_testing()


CMakeLists.txt
```diff
@@ -10,14 +10,14 @@ endif (POLICY CMP0048)
...
 if (CMAKE_VERSION VERSION_LESS "3.1")
-  add_definitions(-std=c++11)
+  add_definitions(-std=c++11 -D_POSIX_C_SOURCE=200809L)
 else()
   set(CMAKE_CXX_STANDARD 11)
   set(CMAKE_CXX_STANDARD_REQUIRED ON)
-  if(NOT CYGWIN AND NOT MSYS)
-    set(CMAKE_CXX_EXTENSIONS OFF)
-  endif()
+  set(CMAKE_CXX_EXTENSIONS OFF)
```
asoffer: Is this a behavior change for CYGWIN and MSYS?

mjvk: Yes, I am actually unsure how this would compile outside MSYS and CYGWIN as well (or are posix extensions not used on other systems?). Compiling with extensions disabled also disables posix functionality. This commit enables the extensions that you need (posix) and not more. This makes it all as compiler agnostic as possible. Pure C++ is of course even nicer but not very practical.

Even nicer would be to put a define in the cpp files where you use them, but this is not really feasible without a biggish refactor (and provides limited advantage).

asoffer: Thanks for this. I think this has follow-on effects for a large enough collection of our user base that we're unable to accept this change.

mjvk: I understand a bit. I don't understand why gtest compiles without extensions at all under Linux. For example fork() and isatty() are not part of the standard but are POSIX extensions. Ref http://man7.org/linux/man-pages/man3/isatty.3.html and http://man7.org/linux/man-pages/man2/fork.2.html . As these and other calls are used throughout the project you should define somewhere which POSIX standard you are using. 

I do understand as I did not test it on every system, but still this change should have no impact as it only enables some features. It does not disable them such as the change adding ```set(CMAKE_CXX_EXTENSIONS OFF)``` which broke the build on at least CYGWIN, MINGW and QNX.

I will make a PR to fix QNX.

PR #2670, opened 2020-01-15
Add additional valid characters to parameter names

DerekTBrown:
I feel as though the alphanumeric + `_` charset is too restrictive, especially when formatting complex objects into names.  Added a few extra characters which will help with this.

### DIFF
--- a/googletest/include/gtest/internal/gtest-param-util.h
+++ b/googletest/include/gtest/internal/gtest-param-util.h
@@ -633,6 +633,8 @@ class ParameterizedTestSuiteInfo : public ParameterizedTestSuiteInfoBase {
       int line;
   };
   typedef ::std::vector<InstantiationInfo> InstantiationContainer;
+  
+  static std::string validParamChars = "_-:/";
 
   static bool IsValidParamName(const std::string& name) {
     // Check for empty string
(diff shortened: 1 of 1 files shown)


vslashg:
I think the current restriction is probably wise.  I imagine allowing "/" in parameter names would break user code that expects "/" to have special meaning in a parameterized test instance name in test output.

Most names in googletest are restricted to something close to C++ identifier spelling, and I'm not sure we'd want to deviate here.

stonigrl (approved):

DerekTBrown:
@stonigrl can you approve running the remaining workflows?

krishnacx (commented):
Can we declare as const at line no: 655

krishnacx (approved):
Any reason why we need space in begining of line 656?

Bat3an:
No pacific reason ,I'm sorry I didn't see that
    On Saturday, June 4, 2022, 07:00:05 AM PDT, Krishna Vivek Vitta ***@***.***> wrote:  
 
 


@krishnacx approved this pull request.

Any reason why we need space in begining of line 656?

DerekTBrown:
Nope. Fixed.

kangzixiang (approved):

PR #2677, opened 2020-01-16
Fix always false condition and clean function body

### LINKED ISSUE #2647: Bug in gtest-filepath
gtest-filepath.cc	355

Expression:
"if (pathname_.c_str() == nullptr) { "

is always false.

IYP-Programer-Yeah:
An always false condition was remove from Normalize member function of FilePath and the body of the function which was poorly written, is improved.

Closes #2647

### DIFF
--- a/googletest/src/gtest-filepath.cc
+++ b/googletest/src/gtest-filepath.cc
@@ -349,33 +349,19 @@ FilePath FilePath::RemoveTrailingPathSeparator() const {
 // For example, "bar///foo" becomes "bar/foo". Does not eliminate other
 // redundancies that might be in a pathname involving "." or "..".
 void FilePath::Normalize() {
-  if (pathname_.c_str() == nullptr) {
-    pathname_ = "";
-    return;
-  }
-  const char* src = pathname_.c_str();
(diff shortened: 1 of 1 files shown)


IYP-Programer-Yeah:
> I am not sure whether the reimplementation is 100% equivalent. Specifically, if the first character is a path separator, it seems to be UB. Not sure if that's a considered a valid input or not, but the change might have some impact.
> If the concern is the always false condition, can we just remove the if block?
Thanks for pointing out the issue, it's fixed. The main point of this merge request was to refactor the function, but again if the maintainers think it's too much of a change I can split the merge request into two.

IYP-Programer-Yeah:
Can I get a review please?

zhangxy988:
Sorry for the delay, I lost track of this PR.
The change looks good to me.
I will import it for internal testing.
Thanks!

zhangxy988:
Thank you, we have started internal review. Please don’t push any more changes into this PR as they might be overwritten. (312291968)

IYP-Programer-Yeah:
What is the issue with the CLA? I believe my CLA is okay, is there something I should do?

PR #2683, opened 2020-01-24
Set kErrorOnUninstantiatedParameterizedTest et al. to true

bcsgh:
This will start reporting missing instantiations of parameterized tests as a failing test under the GoogleTestVerification test suite (they are currently reported, but just as a notice from a passing test).

NOTE: It is expected that this will surface existing issues in downstream consumers.

### DIFF
--- a/googletest/src/gtest.cc
+++ b/googletest/src/gtest.cc
@@ -414,8 +414,8 @@ namespace {
 // inserted to report ether an error or a log message.
 //
 // This configuration bit will likely be removed at some point.
-constexpr bool kErrorOnUninstantiatedParameterizedTest = false;
-constexpr bool kErrorOnUninstantiatedTypeParameterizedTest = false;
+constexpr bool kErrorOnUninstantiatedParameterizedTest = true;
+constexpr bool kErrorOnUninstantiatedTypeParameterizedTest = true;
 
 // A test that fails at a given file/line location with a given message.
 class FailureTest : public Test {
--- a/googletest/test/googletest-output-test-golden-lin.txt
+++ b/googletest/test/googletest-output-test-golden-lin.txt
@@ -984,6 +984,8 @@ Stack trace: (omitted)
 [0;31m[  FAILED  ] [mPrintingStrings/ParamTest.Failure/a, where GetParam() = "a"
 [0;32m[----------] [m3 tests from GoogleTestVerification
 [0;32m[ RUN      ] [mGoogleTestVerification.UninstantiatedParamaterizedTestSuite<NoTests>
+googletest-output-test_.cc:#: Failure
+
 Paramaterized test suite NoTests is instantiated via INSTANTIATE_TEST_SUITE_P, but no tests are defined via TEST_P . No test cases will run.
 
 Ideally, INSTANTIATE_TEST_SUITE_P should only ever be invoked from code that always depend on code that provides TEST_P. Failing to do so is often an indication of dead code, e.g. the last TEST_P was removed but the rest got left behind.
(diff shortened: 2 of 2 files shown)


asoffer:
As discussed, this may take some time to fix internal users, but we do intend to accept this eventually.

bcsgh:
Yup. That was what I was expecting. Thanks.

bcsgh:
BTW: seems this fails presubmit by catching an extra INSTANTIATE_TEST_SUITE_P in the tests. It seems the issue is a orphaned .cc file and another that an overly wide glob() is picking up:

The relevant bit of the log:

> [----------] 2 tests from GoogleTestVerification
[ RUN      ] GoogleTestVerification.UninstantiatedParamaterizedTestSuite<ExternalInstantiationTest>
googletest/test/googletest-param-test2-test.cc:51: Failure
>
> Paramaterized test suite ExternalInstantiationTest is instantiated via INSTANTIATE_TEST_SUITE_P, but no tests are defined via TEST_P . No test cases will run.
>
> Ideally, INSTANTIATE_TEST_SUITE_P should only ever be invoked from code that always depend on code that provides TEST_P. Failing to do so is often an indication of dead code, e.g. the last TEST_P was removed but the rest got left behind.
>
> To suppress this error for this test suite, insert the following line (in a non-header) in the namespace it is defined in:
>
>GTEST_ALLOW_UNINSTANTIATED_PARAMETERIZED_TEST(ExternalInstantiationTest);
> Stack trace: ...
>
> [  FAILED  ] GoogleTestVerification.UninstantiatedParamaterizedTestSuite<ExternalInstantiationTest> (1 ms)
> [ RUN      ] GoogleTestVerification.UninstantiatedParamaterizedTestSuite<InstantiationInMultipleTranslationUnitsTest>
> googletest/test/googletest-param-test2-test.cc:60: Failure
>
> Paramaterized test suite InstantiationInMultipleTranslationUnitsTest is instantiated via INSTANTIATE_TEST_SUITE_P, but no tests are defined via TEST_P . No test cases will run.
>
> Ideally, INSTANTIATE_TEST_SUITE_P should only ever be invoked from code that always depend on code that provides TEST_P. Failing to do so is often an indication of dead code, e.g. the last TEST_P was removed but the rest got left behind.
>
> To suppress this error for this test suite, insert the following line (in a non-header) in the namespace it is defined in:
>
> GTEST_ALLOW_UNINSTANTIATED_PARAMETERIZED_TEST(InstantiationInMultipleTranslationUnitsTest);
Stack trace: ...
> [  FAILED  ] GoogleTestVerification.UninstantiatedParamaterizedTestSuite<InstantiationInMultipleTranslationUnitsTest> (0 ms)

bcsgh:
Bump? Any Progress on this?

antoniovicente:
To allow gradual detection and fix: would it be possible to add a compile time option to enable this mode?

bcsgh:
IIRC that was actually proposed (by me, internally to Google) at one point. I don't remember the details, but it got turned down. (I think it ended up with me reluctantly agreeing with the argument?)

antoniovicente:
> IIRC that was actually proposed (by me, internally to Google) at one point. I don't remember the details, but it got turned down. (I think it ended up with me reluctantly agreeing with the argument?)

So it goes.  I was fortunate enough to be cc'ed on an google-internal report about some tests in an open-source project that I'm a part of which are not running.  I'm sad that I can't protect the open source project from regressing once the relevant tests are fixed.

yurykats:
https://github.com/google/googletest/commit/ec94d9f24c92a5090fda5567156d6dde99cdbf31

mr-c:
Hello, has googletest switched to "live at HEAD" instead of regular releases? If so, has a refactoring tool been issued for this change, as per the [policy](https://abseil.io/about/philosophy#upgrade-support) ?

Is there a changelog or other announcement mechanism that developers should be subscribing to?

Debian [packaged a recent git snapshot](https://tracker.debian.org/news/1186727/accepted-googletest-110020201025-1-source-into-unstable/) of googletest, and now other Debian packages that build and run googletest tests at packaging building time [are broken](https://bugs.debian.org/973001) :-(

Issue #2646, opened 2019-12-24
Can't attach extra information to EXPECT_CALL


MovEaxEsp:
All the standard gtest EXPECT and ASSERT macros support at least 2 ways (that I know of) for attaching arbitrary information to their output to provide context in case of failure (<< at the end, or ScopedTrace).

EXPECT_CALL, however, doesn't support either of these mechanisms.  I haven't found any way of including any extra information in an EXPECT_CALL (or ON_CALL I suppose), which makes it impossible to debug EXPECT_CALLs in subroutines as described in https://github.com/google/googletest/blob/master/googletest/docs/advanced.md#using-assertions-in-sub-routines

I hacked something for myself quite easily here, for the interested: https://github.com/MovEaxEsp/googletest/commit/a32b47d7467110a7fa6a5064ca339381ffc7bcdc

Ideally, EXPECT_CALL would include any active ScopedTraces at the time it's invoked, and include them in any failure output, and/or support << streaming so you could do:

EXPECT_CALL(...).WillOnce(...).RetireOnSaturation() << "my extra info";

but both of these might involve some work.

asoffer:
Apologies for the delay.

SCOPED_TRACE doesn't make sense because EXPECT_CALL expectations live beyond the scope until the mock is destroyed. With this in mind, two options are:

1. Have your mocks destructor print the information you care about
2. Put the SCOPED_TRACE immediately after the construction of the mock?

But I think knowing the type of information you'd like to include would be useful here.

pgrzech1:
Thanks for taking a look.  Currently, the only thing I use this for is to stick the line number passed in from the caller of the subroutine that actually does the EXPECT_CALL, exactly as was described in #2732, but I'd hate for whatever solution is implemented for this to be limited to line numbers, and not be as general as whatever's available for other EXPECT_* or ASSERT_* macros.

With respect to the two options, they both seem to suggest sticking more information in the mock itself, which isn't what's being asked for here.  What I'm looking for is extra context information to be stored with each individual expectation, so it can be printed whenever gmock has to print something about the expectation.

As an example, with #2689 , I modified the gmock output to look like:

```
../../src/groups/rgw/rgwr/rgwr_router.g.cpp:282: EXPECT_CALL(*d_terminals[terminal], sendMessageToTerminal(toExpect, BlobEq(payload))) with [line: 448]...
  Expected arg #0: is equal to [ type = TMGR sessionNumber = 0 window = 0 ]
[...]

PR #2689, opened 2020-01-28
Allow EXPECT_CALL to have extra context information streamed to it

pgrzech1:
This PR adds stream support to ExpectationBase, which would allow one to write something like:
EXPECT_CALL(...).Clause().Clause() << "some extra context", and have 'some extra context' printed whenever any error involving this expectation is printed, as described in https://github.com/google/googletest/issues/2646


asoffer:
See #2646.

PR #2701, opened 2020-02-09
Fix unit test failure on a compiler where internal::GetTypeName<T> returns different string than in unit test

bysreg:
This fixes unit test failure in gmock-matchers_test.cc on a compiler where internal::GetTypeName<int>() returns "i" instead of "int" that is specified in the unit test. This is kinda similar to the merged pull request that I did in https://github.com/google/googletest/pull/2665 though with a different compiler. 

One of the tests error:
```
[ RUN ] MatcherAssertionTest.WorksForByRefArguments
googletest\src/gtest.cc:701: Failure
Expected: 1 fatal failure containing "Actual: 0 (of type int), which is located @"
Actual:
googlemock\test/gmock-matchers_test.cc:2908: Fatal failure:
Value of: n
Expected: does not reference the variable @82a6244 0
Actual: 0 (of type i), which is located @82a6244


[ FAILED ] MatcherAssertionTest.WorksForByRefArguments (0 ms)
```


mattcalabrese-google:
Thanks and sorry for our delay in processing this pull request -- I believe that in this case we'd rather have `OfType` become a template that takes the type as an explicit template parameter, as-in:

`OfType<int>()` rather than `OfType("int")`

And have `GetTypeName` be invoked internal to the `OfType` function template. Are you willing to update your PR for such an approach?

Also, what is your compiler/configuration?

bysreg:
No worries. Yes, I can update this to use template parameter. 

Unfortunately, I'm not allowed to say the compiler/configuration.

bysreg:
@mattcalabrese-google I've updated the pull request

bysreg:
@mattcalabrese-google any update on this?

PR #2709, opened 2020-02-13
Add std::complex<> matcher for equality.

### LINKED ISSUE #2708: std::complex comparison
As a user of GTest I would like to be able to compare complex numbers defines with std::complex the same way I compare floating point number.
I would like to have corresponding 'std::complex' matcher for
  - `FloatEq`
  - `FloatNear`
  - `NanSensitiveComplexFloatEq`
and their double equivalents:
   - `ComplexEq`
   - `ComplexNear`
   - `NanSensitiveComplexDoubleEq`
#1432 is a similar issue.

lkoenig:
To ease matching std::complex<T> where T
is either float or double this change
introduces equality and neighbourhood comparison.

That PR resolves #2708


asoffer:
Thank you for this pull request. See comments on #2708 for why we do not believe this is a good fit for GoogleTest. In short, we understand this the benefits of this matcher, but there are circumstances in which we believe it to be misleading and therefore prefer to not include it in the core of GoogleTest

PR #2720, opened 2020-02-21
Remove deprecated SetArgumentPointee

kuzkry:


asoffer:
Unfortunately there are still thousands of uses inside Google. We'll see how quickly I can get rid of them.

asoffer:
We're almost done cleaning up Google's codebase. Maybe another week or so before we can pull this in. Thanks for your patience.

kuzkry:
Rebased to resolve the merge conflict.

asoffer:
After some discussion internally, there is not consensus on removing this. The consensus for the time being is that it would break external users without providing significant benefit. We may revisit this in the future.

PR #2722, opened 2020-02-21
Fixed warnings

### LINKED ISSUE #2721: Build fails with Gcc >= 8 due to implicit conversion
When building with gcc 8 and above, I have the following error:

```bash
/home/travis/build/xtensor-stack/xtensor/build/test/googletest-src/googletest/src/gtest.cc:2112:72: error: conversion to ‘long unsigned int’ from ‘int’ may change the sign of the result [-Werror=sign-conversion]
 std::vector<std::string> ArrayAsVector(const char* const (&array)[kSize]) {
```

JohanMabille:
Fixes #2721


JohanMabille:
@googlebot I signed it!

asoffer:
Looks good. Pulling internally.

PR #2723, opened 2020-02-21
Fixed warnings

JohanMabille:
Backport of #2722

### DIFF
--- a/googletest/src/gtest.cc
+++ b/googletest/src/gtest.cc
@@ -2108,7 +2108,7 @@ static const char* const kReservedOutputTestCaseAttributes[] = {
     "classname",   "name", "status", "time",   "type_param",
     "value_param", "file", "line",   "result", "timestamp"};
 
-template <int kSize>
+template <size_t kSize>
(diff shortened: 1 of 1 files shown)


kuzkry:
Hi! On [releases](https://github.com/google/googletest/releases) webpage you will find release 1.10.0 and a note:
> The 1.10.x will not consider any requests for any new features, The bug fix requests will only be considered if proven "critical"

Warning isn't critical, so this pull request targeting [v.1.10.x](https://github.com/google/googletest/tree/v1.10.x) branch should be abandoned.

JohanMabille:
This warning prevents to build gtest 1.10 with recent versions of gcc and you consider it as non critical?

kuzkry:
It's only a warning after all. What prevents build is `-Werror` applied onto `-Wsign-conversion`.
Who knows, maybe GTest maintainers will accept it.

JohanMabille:
> What prevents build is `-Werror` applied onto `-Wsign-conversion`.

Indeed, but I guess they set `-Werror`  on purpose, so fixing the warning seems to be the thing to do rather than unsetting `-Werror` (or unsetting it for particular warnings).

PR #2725, opened 2020-02-23
Fix GetFailureReporter() memory leak

kuzkry:
Failing code:
```c++
#include "gmock/gmock.h"
#include "gtest/gtest.h"

class MyMock {
public:
    MOCK_METHOD(void, my_function, (char const*), ());
};

TEST(MyTest, Leak) {
    MyMock my_mock;

    EXPECT_CALL(my_mock, my_function("LEAK"));
    // unsatisfied expectation will cause a memory leak
}
```
---
I changed the implementation to Meyer's singleton so it's better than it was, though I'm not very happy with it. Feel free to propose sth.

### DIFF
--- a/googlemock/src/gmock-internal-utils.cc
+++ b/googlemock/src/gmock-internal-utils.cc
@@ -115,9 +115,8 @@ GTEST_API_ FailureReporterInterface* GetFailureReporter() {
   // thread-safe.  We may need to add additional synchronization to
   // protect failure_reporter if we port Google Mock to other
   // compilers.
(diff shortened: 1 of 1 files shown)


asoffer:
This isn't really a leak in the sense that the memory is always accessible. The Meyer's singleton gives us safe initialization, but the object's destruction order can still be a source of bugs, as it's destruction order is not statically knowable. For this reason, a common pattern is to heap allocate and "leak" the global. There's no real need to clean up this type, so we just don't bother.

PR #2733, opened 2020-02-27
Remove public buganizer reference.

taylorconor:
It also seems that this bug is obsolete.


asoffer:
Hi Conor, for Google employees we require that all submissions be made internally through CLs. Send me a CL and I'll approve it.

PR #2736, opened 2020-03-04
add CMAKE_CXX_STANDARD to CMakeLists.txt

o-micron:
this fixes errors related to enabling C++11 like `no member named 'make_tuple' in namespace 'std'` and like `deleted function definitions are a C++11 extension` in
googletest/googletest/include/gtest/internal/gtest-port.h

I can see there was a modification in #2706 which probably broke things lately

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,18 +11,13 @@ project(googletest-distribution)
 set(GOOGLETEST_VERSION 1.10.0)
 
 if (CMAKE_VERSION VERSION_GREATER_EQUAL "3.1")
-    set(CMAKE_CXX_STANDARD 11)
-    set(CMAKE_CXX_STANDARD_REQUIRED OFF)
-endif()
-
-if(WIN32)
-  if(MSVC)
-    if(MSVC_TOOLSET_VERSION EQUAL 140 OR MSVC_TOOLSET_VERSION EQUAL 141)
-      # there is a bug reported on https://developercommunity.visualstudio.com/content/problem/9696/c17-not-available-by-default.html
-      # string_view is not available unless we use the /std:c++latest switch
-      set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /std:c++latest")
+    if(NOT CYGWIN AND NOT MSYS)
+      set(CMAKE_CXX_EXTENSIONS OFF)
+      if(APPLE)
+        set(CMAKE_CXX_STANDARD 11)
+        set(CMAKE_CXX_STANDARD_REQUIRED OFF)
+      endif()
     endif()
-  endif()
 endif()
 
 enable_testing()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,8 +11,17 @@ project(googletest-distribution)
 set(GOOGLETEST_VERSION 1.10.0)
 
 if (CMAKE_VERSION VERSION_GREATER_EQUAL "3.1")
-  if(NOT CYGWIN AND NOT MSYS)
-    set(CMAKE_CXX_EXTENSIONS OFF)
+    set(CMAKE_CXX_STANDARD 11)
+    set(CMAKE_CXX_STANDARD_REQUIRED OFF)
+endif()
+
+if(WIN32)
+  if(MSVC)
+    if(MSVC_TOOLSET_VERSION EQUAL 140 OR MSVC_TOOLSET_VERSION EQUAL 141)
+      # there is a bug reported on https://developercommunity.visualstudio.com/content/problem/9696/c17-not-available-by-default.html
+      # string_view is not available unless we use the /std:c++latest switch
+      set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /std:c++latest")
+    endif()
   endif()
 endif()
 
--- a/googletest/include/gtest/internal/gtest-port.h
+++ b/googletest/include/gtest/internal/gtest-port.h
@@ -2234,21 +2234,38 @@ using StringView = ::absl::string_view;
 }  // namespace internal
 }  // namespace testing
 #else
-# ifdef __has_include
-#   if __has_include(<string_view>) && __cplusplus >= 201703L
-// Otherwise for C++17 and higher use std::string_view for Matcher<>
-// specializations.
-#   define GTEST_INTERNAL_HAS_STRING_VIEW 1
-#include <string_view>
-namespace testing {
-namespace internal {
-using StringView = ::std::string_view;
-}  // namespace internal
-}  // namespace testing
-// The case where absl is configured NOT to alias std::string_view is not
-// supported.
-#  endif  // __has_include(<string_view>) && __cplusplus >= 201703L
-# endif  // __has_include
+  #ifdef __has_include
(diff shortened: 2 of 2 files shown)


o-micron:
@googlebot I signed it!

kuzkry:
Hi! Commit 23b2a3b (not #2706) introduced this change and it was done to avoid mixed builds. Please think about C++17 users who want to use e.g. `Matcher<std::string_view>`. With this change, even setting `CMAKE_CXX_STANDARD=17` is going to be ignored.

o-micron:
> Hi! Commit [23b2a3b](https://github.com/google/googletest/commit/23b2a3b1cf803999fb38175f6e9e038a4495c8a5) (not #2706) introduced this change and it was done to avoid mixed builds. Please think about C++17 users who want to use e.g. `Matcher<std::string_view>`. With this change, even setting `CMAKE_CXX_STANDARD=17` is going to be ignored.

Hi! Yep you're right, so the solution to satisfy all cases is to supply cmake with -DCMAKE_CXX_STANDARD=17 .. 

Maybe setting `set(CMAKE_CXX_STANDARD_REQUIRED OFF)` would allow them so add the extra argument with cmake ?

kuzkry:
I think we're talking at cross purposes or maybe I didn't understand. :)

The problem with using `#if __cplusplus` in code is that if this is used in GoogleTest headers we don't know whether they are placed in context of either GoogleTest libraries or user code. Imagine we have a C++17 user that writes this:
```c++
// test; requires C++17

#include <gmock/gmock.h>
#include <gtest/gtest.h>
// there exists conditional enabling of Matcher<std::string_view>
// based on #__cplusplus

// from this file point of view there exist Matcher<std::string_view>

// other includes ...
using ::testing::StartsWith;

TEST(MyTest, TestFoo) {
  std::string_view sv = "something";
  EXPECT_THAT(Foo(), StartsWith(sv));
}
```

Unless GoogleTest libraries are also compiled in C++17, this user will get undefined reference linker errors because the header declared something that those libs don't have. So that's the mix of different contexts (i.e. mixed builds).

Last but not least, this user could either:
- use GoogleTest as subproject of their own one and let GoogleTest inherit CMAKE_CXX_STANDARD from a parent project
- they could build GoogleTest libs stand-alone (no inheritance from a parent project, it just exists somewhere on hard drive) and later point a linker to look for GoogleTest libraries there

Anyway, I hope I made it easy to digest. Enjoy the rest of the weekend :D

o-micron:
> 
> 
> I think we're talking at cross purposes or maybe I didn't understand. :)
> 
> The problem with using `#if __cplusplus` in code is that if this is used in GoogleTest headers we don't know whether they are placed in context of either GoogleTest libraries or user code. Imagine we have a C++17 user that writes this:
> 
> ```c++
> // test; requires C++17
> 
> #include <gmock/gmock.h>
> #include <gtest/gtest.h>
> // there exists conditional enabling of Matcher<std::string_view>
> // based on #__cplusplus
> 
> // from this file point of view there exist Matcher<std::string_view>
> 
> // other includes ...
> using ::testing::StartsWith;
> 
> TEST(MyTest, TestFoo) {
>   std::string_view sv = "something";
>   EXPECT_THAT(Foo(), StartsWith(sv));
> }
> ```
> 
> Unless GoogleTest libraries are also compiled in C++17, this user will get undefined reference linker errors because the header declared something that those libs don't have. So that's the mix of different contexts (i.e. mixed builds).
> 
> Last but not least, this user could either:
> 
>     * use GoogleTest as subproject of their own one and let GoogleTest inherit CMAKE_CXX_STANDARD from a parent project
> 
>     * they could build GoogleTest libs stand-alone (no inheritance from a parent project, it just exists somewhere on hard drive) and later point a linker to look for GoogleTest libraries there
> 
> 
> Anyway, I hope I made it easy to digest. Enjoy the rest of the weekend :D

That's exactly what I imagine should happen

The (C++17/20 or ..) user will generally add googletest as a sub project which will notify googletest's cmake with the desired `CMAKE_CXX_STANDARD` ... which is fine in this case I see no problem with that :)

and in case the user compiled googletest somewhere without actually specifying C++17/20 and some features (that his/her own code is using) are missing, an undefined reference is totally fine in that case, that's because the user did not actually tell googletest to compile with the newer features from C++17/20 ...

What I mean by this pull request is to keep C++11 the default and allow extra features in C++17/20 only available if the user passes `-DCMAKE_CXX_STANDARD=17` to cmake ...

I apologize if I didn't explain it properly first time :)

kuzkry:
Specifying C++11 as default would give us a certain level of control while leaving C++17 users with a possibility to enable additional features. Sounds reasonable, I like it.

Please carry on with https://github.com/google/googletest/pull/2736#pullrequestreview-370861650
[suggestion] May I kindly suggest to decrease commit count here? Could you squash them together?

kuzkry:
If done for all platforms, it'd resolve #2740.

o-micron:
@kuzkry Ok, done

o-micron:
now, done

derekmauro:
Hi,

I removed `CMAKE_CXX_STANDARD` from CMakeLists.txt for exactly the reasons @kuzkry has explained.

It seems like the problem you are trying to fix is that you want the build to work by default with compilers that by default use C++ < 11?

Does something like this work? (I don't know, I am asking.)
```
if(NOT "${CMAKE_CXX_STANDARD}")
  set(CMAKE_CXX_STANDARD 11)
endif()
```

I have to say that I'm not sure this is worth fixing. Mixed mode builds are dangerous, and I fear the by fixing this, we are only hiding it from the user. Of course it is still possible that the user is getting a mixed-mode build if the user's compiler defaults to C++ >= 11 when the user doesn't realize that both the user code and GoogleTest must be built with the same compile options if the options affect ABI.

kuzkry:
@derekmauro, probably yes but I'm no cmake expert. There's a cleaner way if we used cmake 3.8:
```cmake
# marks gtest as requiring at least C++11
target_compile_features(gtest PRIVATE cxx_std_11)
```
The problem with this is that GoogleTest uses some really really old cmake (3.0 or something) and I don't think it's nice to upgrade by that many versions.

And again, you're right, it will still be possible to mix up builds :<

o-micron:
Hi @derekmauro and @kuzkry thanks for your assistance, the problem originally was me building googletest seperately from my projects, keeping a "libs" folder with googletest and other libraries compiled and ready.
Recently that got broken because of the removed cmake `set(CMAKE_CXX_STANDARD 11)`
the compiler was complaining 
`no member named 'make_tuple' in namespace 'std' and like deleted function definitions are a C++11 extension`

After a while I understood that you were introducing some new C++17 features like string_view ?
My point was using something like this to keep the new features working while leaving C++11 as the default standard
```cpp
#if __cplusplus >= 201500
.. ADD_EXTRA_TEST_FOR_CPP17_AND_UP (Matcher<std::string_view>) ..
#endif
```

I know that makes the code even more complex and the templates may get bloated just because of that ... I really don't know I just didn't want to pass `-DCMAKE_CXX_STANDARD=11`
I now think it's not worth introducing more complexity to the template and just pass that and move on .. 😆

derekmauro:
It looks like we are going with #2808 to solve this problem.

PR #2742, opened 2020-03-07
Print C++17 types (any, optional and variant) and absl::any

kuzkry:
Rework of rejected #2438 due to [the plan on taking hard dependency on Abseil](https://github.com/google/googletest/pull/2438#issuecomment-534136820). I'm resubmitting because recent commit 6f5fd0d7 seems to suggest abandonment of Abseil incorporation like it was planned half a year ago.

### DIFF
--- a/googletest/include/gtest/gtest-printers.h
+++ b/googletest/include/gtest/gtest-printers.h
@@ -688,13 +688,20 @@ class UniversalPrinter<Any> {
  public:
   static void Print(const Any& value, ::std::ostream* os) {
     if (value.has_value())
-      *os << "'any' type with value of type " << GetTypeName();
+      *os << "'any' type with value of type " << GetTypeName(value);
     else
       *os << "'any' type with no value";
   }
 
  private:
-  static std::string GetTypeName() { return "the element type"; }
+  static std::string GetTypeName(const Any& value) {
+#if GTEST_HAS_RTTI
+    return internal::GetTypeName(value.type());
+#else
+    static_cast<void>(value); // possibly unused
+    return "the element type";
+#endif  // GTEST_HAS_RTTI
+  }
 };
 
 #endif  // GTEST_INTERNAL_HAS_ANY
--- a/googletest/include/gtest/gtest-printers.h
+++ b/googletest/include/gtest/gtest-printers.h
@@ -679,6 +679,26 @@ class UniversalPrinter {
   GTEST_DISABLE_MSC_WARNINGS_POP_()
 };
 
+#if GTEST_INTERNAL_HAS_ANY
+
+// Printer for std::any / absl::any
+
+template <>
+class UniversalPrinter<Any> {
+ public:
+  static void Print(const Any& value, ::std::ostream* os) {
+    if (value.has_value())
+      *os << "'any' type with value of type " << GetTypeName();
+    else
+      *os << "'any' type with no value";
+  }
+
+ private:
+  static std::string GetTypeName() { return "the element type"; }
+};
+
+#endif  // GTEST_INTERNAL_HAS_ANY
+
 #if GTEST_INTERNAL_HAS_OPTIONAL
 
 // Printer for std::optional / absl::optional
--- a/googletest/include/gtest/gtest-printers.h
+++ b/googletest/include/gtest/gtest-printers.h
@@ -114,7 +114,6 @@
 #if GTEST_HAS_ABSL
 #include "absl/strings/string_view.h"
 #include "absl/types/optional.h"
-#include "absl/types/variant.h"
 #endif  // GTEST_HAS_ABSL
 
 namespace testing {
@@ -699,14 +698,22 @@ class UniversalPrinter<::absl::optional<T>> {
   }
 };
 
-// Printer for absl::variant
+#endif  // GTEST_HAS_ABSL
+
+#if GTEST_INTERNAL_HAS_VARIANT
+
+// Printer for std::variant / absl::variant
 
 template <typename... T>
-class UniversalPrinter<::absl::variant<T...>> {
+class UniversalPrinter<Variant<T...>> {
  public:
-  static void Print(const ::absl::variant<T...>& value, ::std::ostream* os) {
+  static void Print(const Variant<T...>& value, ::std::ostream* os) {
     *os << '(';
-    absl::visit(Visitor{os}, value);
+#if GTEST_HAS_ABSL
+    absl::visit(Visitor{os, value.index()}, value);
+#else
+    std::visit(Visitor{os, value.index()}, value);
+#endif  // GTEST_HAS_ABSL
     *os << ')';
   }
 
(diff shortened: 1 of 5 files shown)


o-micron:
@kuzkry Ok, here is what the summary of what I understood so far ..
- Bazel builds are just fine, only cmake on [msvc 15 2017 Win64, msvc 14 2015 Win64] are failing to build so far and the error is at [gmock-matchers.h:434](https://github.com/google/googletest/blob/master/googlemock/include/gmock/gmock-matchers.h#L434)
- we replaced `#if GTEST_HAS_ABSL` by `#if GTEST_INTERNAL_HAS_ANY` since we now also defined that and `internal::StringView` is used as `::std::string_view` in case of no ABSL ..
- I kept jumping around the different templates and I don't know at this point where the missing part is, what I don't understand is that it compiles on all platforms and even on different msvc versions so I don't know if there is something in these templates that needs to be tweaked or the problem comes from cmake params that we I pass ... since basel builds are just fine .. same code :D
- Does these template errors have something to do with `string_view` and `explicit conversion` ?

kuzkry:
Hey @o-micron.

First things first, `#if GTEST_INTERNAL_HAS_ANY` has nothing to do with `internal::StringView`. The latter is guarded by `#if GTEST_INTERNAL_HAS_STRING_VIEW` ;)

To reply all your other points: I didn't investigate it but you don't need to care about these two failing AppVeyor builds as master is unstable now. I don't know what's the reason but I can tell it's been failing since commit fd538161 so either it's that commit fault or AppVeyor VMs were externally modified.

o-micron:
Yeah `#if GTEST_INTERNAL_HAS_STRING_VIEW` not `#if GTEST_INTERNAL_HAS_ANY` 🤓
So should I just squash the commits and just leave the 2 appveyor builds failing at the moment ?

kuzkry:
In case of your PR #2736, yes :) Just do your work there and ignore these 2 failing jobs for now. If you wanna fix the failing master, you could create another PR by branching out from master.

mattcalabrese-google:
Sorry for our delays in getting to this, we are looking at it again now. One quick suggestion upon skimming -- the variant output should likely also stream the index of the alternative that is mentioned (either in addition to or in place of the type info). This is because a variant can have multiple alternatives with the same type, so the type and value alone is not necessarily enough to determine which alternative was actually active.

googletest/include/gtest/internal/gtest-port.h
```diff
@@ -2223,6 +2229,64 @@ const char* StringFromGTestEnv(const char* flag, const char* default_val);
...
+// is built with absl support.
+# define GTEST_INTERNAL_HAS_ANY 1
+#include "absl/types/any.h"
+namespace testing {
+namespace internal {
+using Any = ::absl::any;
+}  // namespace internal
+}  // namespace testing
+#else
+# ifdef __has_include
```
zhangxy988, line 2243: I think this detection logic doesn't work for MSVC.
See https://github.com/abseil/abseil-cpp/blob/7853a7586c492ce8905c9e49f8679dada6354f2c/absl/base/config.h#L543
We can leave a TODO comment.

googletest/include/gtest/gtest-printers.h
```diff
@@ -681,14 +679,41 @@ class UniversalPrinter {
...
+      *os << "'any' type with no value";
+  }
+
+ private:
+  static std::string GetTypeName(const Any& value) {
+#if GTEST_HAS_RTTI
+    return internal::GetTypeName(value.type());
+#else
+    static_cast<void>(value); // possibly unused
+    return "the element type";
```
zhangxy988, line 702: I don't think "the element type" is more informative than just "<type>". Maybe just call `internal::GetTypeName`?

googletest/test/googletest-printers-test.cc
```diff
@@ -1531,23 +1531,57 @@ TEST(UniversalTersePrintTupleFieldsToStringsTestWithStd, PrintsTersely) {
...
-#if GTEST_HAS_ABSL
+#if GTEST_INTERNAL_HAS_ANY
+class PrintAnyTest : public ::testing::Test {
+ protected:
+  template <typename T>
+  static std::string ExpectedTypeName() {
+#if GTEST_HAS_RTTI
+    return internal::GetTypeName<T>();
+#else
+    return "the element type";
```
zhangxy988, line 1542: Same here, we do not need this function any more if we just use `internal::GetTypeName`.

zhangxy988:
> Sorry for our delays in getting to this, we are looking at it again now. One quick suggestion upon skimming -- the variant output should likely also stream the index of the alternative that is mentioned (either in addition to or in place of the type info). This is because a variant can have multiple alternatives with the same type, so the type and value alone is not necessarily enough to determine which alternative was actually active.

I think this is a good suggestion for improvement but doesn't need to be part of this change. This change aims at supporting users who uses C++17 std types without the dependency on Abseil.

I would even suggest separating the `(std|absl)::any` support into a separate PR but that is probably going too far.

googletest/include/gtest/gtest-printers.h
```diff
@@ -699,14 +724,22 @@ class UniversalPrinter<::absl::optional<T>> {
...
+// Printer for std::variant / absl::variant
 
 template <typename... T>
-class UniversalPrinter<::absl::variant<T...>> {
+class UniversalPrinter<Variant<T...>> {
  public:
-  static void Print(const ::absl::variant<T...>& value, ::std::ostream* os) {
+  static void Print(const Variant<T...>& value, ::std::ostream* os) {
     *os << '(';
+#if __cplusplus >= 201703L
```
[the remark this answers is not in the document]
kuzkry: Done

googletest/include/gtest/internal/gtest-port.h
kuzkry: I based that on the existing StringView detection as delivered in 6f5fd0d. If we wanna change this PR, then we should adjust the existing code.

googletest/include/gtest/gtest-printers.h
kuzkry: I based that on [VariantMatcher<T>::GetTypeName()](https://github.com/google/googletest/blob/6f5fd0d7199b9a19faa9f499ecc266e6ae0329e7/googlemock/include/gmock/gmock-matchers.h#L3605) and [AnyCastMatcher<T>::GetTypeName()](https://github.com/google/googletest/blob/6f5fd0d7199b9a19faa9f499ecc266e6ae0329e7/googlemock/include/gmock/gmock-matchers.h#L3666). It's the same on the current master too. Funny, I had the same impression as you ;)

googletest/test/googletest-printers-test.cc
kuzkry: Yeah, it's the same as above.

kuzkry:
@mattcalabrese-google, thanks, done :)
@zhangxy988, thanks too. I didn't want to implement it all as it doesn't match the existing style, though I agree with you. I can create another small PR and have it all done there if you want.

mattcalabrese-google:
LGTM. I think the `<index>` way of including the alternative index may end up being confusing since it makes it look like a template being instantiated with an integral argument, but if we decide that's a problem we can change it in a different PR.

zhangxy988:
> LGTM. I think the `<index>` way of including the alternative index may end up being confusing since it makes it look like a template being instantiated with an integral argument, but if we decide that's a problem we can change it in a different PR.

I agree "int<0>" is a bit confusing. How about "int(0)" or "int@0"?

kuzkry:
> > LGTM. I think the `<index>` way of including the alternative index may end up being confusing since it makes it look like a template being instantiated with an integral argument, but if we decide that's a problem we can change it in a different PR.
> 
> I agree "int<0>" is a bit confusing. How about "int(0)" or "int@0"?

Sure, I've chosen int(0).

kuzkry:
I've lowered the amount of commits so it's neat and concise now.

zhangxy988:
push any more changes into this PR as they might be overwritten. (314204435)

PR #2754, opened 2020-03-21
Delete .clang-format

metro44060:


zhangxy988:
Sorry, we do want clang format and since there is no good reason stated above to delete it, I am closing this PR.

PR #2756, opened 2020-03-21
Ensure that gtest/gmock pkgconfig requirements specify version

Conan-Kudo:
Google Test and Google Mock require matching versions to work,
so this requirement should be described in the pkgconfig files.

This change is derived from the one used for the Fedora gtest package.

Signed-off-by: Neal Gompa <ngompa13@gmail.com>


zhangxy988:
(303413368)

kloczek:
Any plans to make new trelease?
It is already year since last releaese ..

PR #2761, opened 2020-03-23
Add TOC to googletest and googlemoc docs markdowns

### LINKED ISSUE #2696: Markdown documentation table of contents
Hi,

First off thank you for this fantastic project.

I often find myself going through the gtest or gmock markdowns and wishing there was a TOC at the beginning.

Is there any reason this is not the case?

I could PR some in if not?

bricoletc:
This fixes #2696 by adding a table of contents to the markdown docs.

I have a simple script to produce it in-place: https://gist.github.com/bricoletc/79b3cd9efb862c55f48cac13cbf86ed9

This PR would personally help me navigate the docs, I hope it would for others too :-)


bricoletc:
@googlebot I signed it!

zhangxy988:
Could you please also include the script that generates the toc in this PR?

bricoletc:
Sure, should I just copy it in the root of the project? Or paste it here?

zhangxy988:
@bricoletc I think it should probably live with the docs in googletest/docs/.

bricoletc:
@zhangxy988 just did that :+1: 

You can feed the script 1+ .md in python3 and it will 
1. do nothing if finds existing TOC 
2. delete existing TOC and regenerate one, if run with `--force`, for when the documentation is updated. 
To work, this relies on an empty new line after the TOC section (needs not to be deleted)

zhangxy988:
(307051077)

kuzkry:
Hey @zhangxy988, please notice that this PR is a bit outdated. For example, I've recently removed some chapters in #2765. You might probably want to regenerate TOCs.

zhangxy988:
@bricoletc Could you please sync to master and regenerate the TOCs? Sorry for the inconvenience, I will be quick to sync once you upload. Thanks!

bricoletc:
@kuzkry  thanks for the spot, @zhangxy988 I rebased the work onto master and regenerated TOCs !

bricoletc:
Just to repeat what I put in docstring in the generating script: some `.md`s have only `#` and `###` headers, with no `##` in between. I'd recommend not doing those jumps in the future, because then you have to manually dedent the generated TOC.

zhangxy988:
> Just to repeat what I put in docstring in the generating script: some `.md`s have only `#` and `###` headers, with no `##` in between. I'd recommend not doing those jumps in the future, because then you have to manually dedent the generated TOC.

I see. I don't think there is a good reason for them to stay the way they are. I'd encourage you to send PR to us to fix them if you like. Thanks!

bricoletc:
Sorry @zhangxy988 , getting a bit swamped in my PhD at the moment- can't do that at the moment. The TOCs I made are valid as is though!

zhangxy988:
During the internal review, we found out that our internal documentation viewer automatically generates the table of contents as a side bar so the added TOC here will be redundant. Since we try to maintain the same "source code" internal and external, adding such disparity introduces costly maintenance burden.
That said, I don't think there is a way to take this PR as is and we probably need to seek a different approach that can be easily maintained. I will close this PR for now. Feel free to discuss further about ideas on the issue #2696 and we will also think about how to provide TOC for github users. It's very sad that github doesn't support a simple [TOC] via extension.

zhangxy988:
Forgot to say that we really appreciate your efforts on this and we do want to find a way to solve this problem. Sorry for the inconvenience.

PR #2764, opened 2020-03-24
Remove dead code in googletest-output-test

kuzkry:
It's been dead for 10 years.


zhangxy988:
(303410134)

PR #2765, opened 2020-03-24
Remove chapters on unsupported build systems (Autotools, Meson and plain Makefiles)

kuzkry:


zhangxy988:
(303409432)

PR #2768, opened 2020-03-26
Added junit-like xml output

Butterneck:
Roughly added new output option to output an XML formatted as a junit output XML in order to be accepted by services like sonarqube


Butterneck:
@googlebot I signed it!

zhangxy988:
The tests are failing. Please take a look and make sure the tests pass.

derekmauro:
We talked about this today, and while this in itself isn't something we want to accept and maintain, we would accept some mechanism that made test output more extensible so that users could use their own custom formats.

PR #2772, opened 2020-03-28
Fix a typo in .travis.yml

Quuxplusone:
The old code was trying to pass -Wgnu-zero-variadic-macro-arguments
as part of CXXFLAGS, but it forgot the quotation marks needed around
whitespace. This meant that option was ignored:
https://travis-ci.org/github/google/googletest/jobs/666534177#L760

Unfortunately, the codebase is not remotely clean with respect to that
warning option. It fails like this:
https://travis-ci.org/github/Quuxplusone/googletest/jobs/668118135

So, remove that failing configuration from the test matrix until
someone has time to look at it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,9 +28,6 @@ matrix:
       env: BUILD_TYPE=Debug VERBOSE=1 CXX_FLAGS=-std=c++11
     - os: linux
       compiler: clang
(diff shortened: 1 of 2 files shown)


Quuxplusone:
@zhangxy988

zhangxy988:
Created https://github.com/google/googletest/issues/2806 to track the issue with building with  -Wgnu-zero-variadic-macro-arguments.

zhangxy988:
Sorry for the late reply, and thank you for your effort. We have started internal review. Please don’t
push any more changes into this PR as they might be overwritten. (305706291)

Issue #2777, opened 2020-03-29
First dash of --gtest_print_time is not colored


lvjp:
The first dash of `--gtest_print_time` is not colored when you use the `--help` option.

CJ-Johnson:
The PR is being merged. Thanks!

PR #2778, opened 2020-03-29
Fix --gtest_print_time coloring

lvjp:
Fix the #2777 coloring problem :
The first dash of --gtest_print_time is not colored when you use the --help option.


CJ-Johnson:
Thank you for this contribution! I'm importing this PR now. Please refrain from making any further updates to the code, since they will be overwritten. 😄

PR #2785, opened 2020-03-31
Remove unneeded sys/ includes

kuzkry:
Small clean-up, which I didn't want to make a part of another pull request.


kuzkry:
It's wrong.

PR #2787, opened 2020-03-31
Support multithreading on non-POSIX platforms and require threading support

### LINKED ISSUE #2380: Support multithreading on non-POSIX platforms
I've got an idea to leverage the fact that we use C++11.

In primer.md there is a [known limitation](https://github.com/google/googletest/blob/master/googletest/docs/primer.md#known-limitations) that has been around since 642acb (more than 4 years old) and it suggests to write custom platform-specific implementations for concurrency support.
Quote:
> Google Test is designed to be thread-safe. The implementation is thread-safe on systems where the pthreads library is available. It is currently unsafe to use Google Test assertions from two threads concurrently on other systems (e.g. Windows). In most tests this is not an issue as usually the assertions are done in the main thread. If you want to help, you can volunteer to implement the necessary synchronization primitives in gtest-port.h for your platform.

Now, after we've finally switched to C++11, I suppose it's high time to do it right and have one, portable implementation which will work everywhere. I've started working on a new pull request that will deal with it, though it's not that easy and it's presumably going to take some longer time.

@gennadiycivil - please let me know if you have insight against it.

kuzkry:
Resolves #2380. @EricWF, may I kindly ask you to have a look into this?

1. This is (almost) a minimal set of changes to utilise C++11 concurrency in GoogleTest. Commits are dependent on each other so it's impossible to divide it into smaller pull requests. 
2. Until now, GoogleTest has had 3 models of concurrency: POSIX, Windows and "unsupported", where only POSIX was thread-safe. This PR is inspired by POSIX implementation, hence thread-safety has been preserved.
3. Speaking of "unsupported" model, this PR removes dummy implementation for them as it unconditionally uses C++11 threading tools. Until now, [unsupported platforms have been silently compiling GTest synchronisation primitives](https://github.com/google/googletest/blob/67cc66080d64e3fa5124fe57ed0cf15e2cecfdeb/googletest/include/gtest/internal/gtest-port.h#L1846). I disliked the fact that users weren't even given a warning, so my implementation simply requires threading support (e.g. linking with pthread on Linux, Win32 on Windows etc.); otherwise it won't link. Options we have are:
a) embrace KISS, accept this breaking change, prepare GoogleTest for release 2.0 according to [Live at Head](https://youtu.be/tISy7EJQPzI?t=1127) philosophy. This yields crystal clear message that there is major incompatibility and allow us to pay off a bit more of technical debt. I guess this breaking change wouldn't be that severe because if someone's platform is _old and unsupported_ or _too bare-metal_, they probably don't even bother compiling tests on it.
b) not make any breaking changes by having proxy types such as `::testing::Mutex`, `::testing::MutexLock` and proxy macro `GTEST_THREAD_LOCAL`. Optionally, we could add an extra `#warning` directive for users of unsupported platforms that they shouldn't use threads.
4. Code is equivalent in its meaning to the previous POSIX model. I didn't even try to introduce atomics because my goal was to port things, not to optimise.
5. I've had problems with Windows DLLs - formally commits 784d0a6 and 142925c break the build and they are patched by 2c74ab5 and 455eb9d respectively. I extracted "fix" commits just in case you got some better idea how to handle that - but if you don't, then they should be squashed together so no commit breaks the build.

You'll find the detailed description in commit messages. Happy reviewing and please ask me questions.

### DIFF
--- a/googletest/src/gtest-port.cc
+++ b/googletest/src/gtest-port.cc
@@ -34,9 +34,11 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <chrono>
 #include <cstdint>
 #include <fstream>
 #include <memory>
+#include <thread>
 
 #if GTEST_OS_WINDOWS
 # include <windows.h>
@@ -277,11 +279,11 @@ size_t GetThreadCount() {
 
 #endif  // GTEST_OS_LINUX
 
-#if GTEST_IS_THREADSAFE && GTEST_OS_WINDOWS
-
 void SleepMilliseconds(int n) {
-  ::Sleep(static_cast<DWORD>(n));
+  std::this_thread::sleep_for(std::chrono::nanoseconds(n));
 }
+
+#if GTEST_IS_THREADSAFE && GTEST_OS_WINDOWS
 
 AutoHandle::AutoHandle()
     : handle_(INVALID_HANDLE_VALUE) {}
--- a/googletest/include/gtest/internal/gtest-port.h
+++ b/googletest/include/gtest/internal/gtest-port.h
@@ -540,10 +540,7 @@ typedef struct _RTL_CRITICAL_SECTION GTEST_CRITICAL_SECTION;
 // gtest-port.h guarantees to #include <pthread.h> when GTEST_HAS_PTHREAD is
 // true.
 # include <pthread.h>  // NOLINT
-
-// For timespec and nanosleep, used below.
-# include <time.h>  // NOLINT
-#endif
+#endif  // GTEST_HAS_PTHREAD
 
 // Determines whether clone(2) is supported.
 // Usually it will only be available on Linux, excluding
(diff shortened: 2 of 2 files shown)


kuzkry:
Looks like esp8266 doesn't support C++11 threads. Please let me know if this breakage is a no-go.

googletest/src/gtest-port.cc
```diff
@@ -277,12 +279,12 @@ size_t GetThreadCount() {
 
 #endif  // GTEST_OS_LINUX
 
-#if GTEST_IS_THREADSAFE && GTEST_OS_WINDOWS
-
 void SleepMilliseconds(int n) {
-  ::Sleep(static_cast<DWORD>(n));
+  std::this_thread::sleep_for(std::chrono::nanoseconds(n));
```
ojeda, line 279: This sleeps for nanoseconds, not milliseconds.

kuzkry: Thanks :)

ojeda: You're welcome! :)

kuzkry:
Rebased due to a merge conflict.

kuzkry:
Rebased due to a merge conflict.

kuzkry:
Rebased due to a merge conflict.

kuzkry:
Rebased due to a merge conflict.

PR #2795, opened 2020-04-04
Remove duplicate codes existed in get-nprocessors.sh

Jyun-Neng:
Remove the duplicate code in `travis.sh`, which already exists in `get-nprocessors.sh`.


Jyun-Neng:
@googlebot I fixed

Issue #2779, opened 2020-03-29
The complexity of the function IsValidUTF8 in the file gtest-printers.cc


desosa-9:
The complexity of the function IsValidUTF8 can be decreased, by distilling the conditions of the if-statements into separate functions.

asoffer:
Thanks! Moving conversation to the PR.

PR #2803, opened 2020-04-07
Decrease complexity of the function IsValidUTF8 #2779

desosa-9:
This change addresses issue #2779. The complexity of the function IsValidUTF8 is decreased, by distilling the conditions of the if-statements into separate functions.


vslashg:
We don't think refactoring this code improves readability enough to justify the cost.  Please reopen this PR if this change uncovers a bug in the current UTF parsing code.

Issue #1519, opened 2018-03-27
no member named 'tuple' in namespace 'testing'


lawben:
On my OSX 10.13.3 with AppleClang 9.0.0.9000039 I cannot build the current master. I get the following error:

```
Users/law/repos/googletest/googlemock/include/gmock/internal/gmock-generated-internal-utils.h:72:33: error:
      no member named 'tuple' in namespace 'testing'
struct MatcherTuple< ::testing::tuple<> > {
                     ~~~~~~~~~~~^
/Users/law/repos/googletest/googlemock/include/gmock/internal/gmock-generated-internal-utils.h:72:39: error:
      expected expression
struct MatcherTuple< ::testing::tuple<> > {
                                      ^
/Users/law/repos/googletest/googlemock/include/gmock/internal/gmock-generated-internal-utils.h:77:33: error:
      no member named 'tuple' in namespace 'testing'
struct MatcherTuple< ::testing::tuple<A1> > {
                     ~~~~~~~~~~~^
/Users/law/repos/googletest/googlemock/include/gmock/internal/gmock-generated-internal-utils.h:77:39: error:
      'A1' does not refer to a value
struct MatcherTuple< ::testing::tuple<A1> > {
                                      ^
/Users/law/repos/googletest/googlemock/include/gmock/internal/gmock-generated-internal-utils.h:76:20: note:
      declared here
template <typename A1>
                   ^
/Users/law/repos/googletest/googlemock/include/gmock/internal/gmock-generated-internal-utils.h:82:33: error:
      no member named 'tuple' in namespace 'testing'
struct MatcherTuple< ::testing::tuple<A1, A2> > {
                     ~~~~~~~~~~~^
/Users/law/repos/googletest/googlemock/include/gmock/internal/gmock-generated-internal-utils.h:82:39: error:
      'A1' does not refer to a value
struct MatcherTuple< ::testing::tuple<A1, A2> > {
                                      ^
/Users/law/repos/googletest/googlemock/include/gmock/internal/gmock-generated-internal-utils.h:81:20: note:
      declared here
template <typename A1, typename A2>
                   ^
/Users/law/repos/googletest/googlemock/include/gmock/internal/gmock-generated-internal-utils.h:87:33: error:
      no member named 'tuple' in namespace 'testing'
struct MatcherTuple< ::testing::tuple<A1, A2, A3> > {
                     ~~~~~~~~~~~^
[...]

PR #2808, opened 2020-04-11
Set cpp11 compile feature with target_compile_features if available.

OlivierLDff:
Use `target_compile_features` to use c++11 if target_compile_features is available. 
This fix compilation with clang and gcc when c++11 isn't specified by user. Otherwise user need to set `-DCMAKE_CXX_STANDARD=11`.

The fix the issue that have been reopened there: https://github.com/google/googletest/issues/1519

This only fix the issue for user using cmake version that support `target_compile_features`, ie CMake greater than 3.1.

This solution is better than `set(CMAKE_CXX_STANDARD 11)` because in superbuild, it won't override any user `CMAKE_CXX_STANDARD` already set variables.
This c++11 requirement will be propagated to any target that `target_link_libraries` as a minimum requirement and won't downgrade any user c++ standard requirement.

And maybe to test if it work as expected, i suggest that `CXX_FLAGS=-std=c++11` gets removed from `.travis.yml` and `-DCMAKE_CXX_FLAGS=$CXX_FLAGS ` from `ci/travis.sh` ? Min version for ci seems to be CMake 3.12 so everything would work the same.

### DIFF
--- a/googletest/cmake/internal_utils.cmake
+++ b/googletest/cmake/internal_utils.cmake
@@ -188,6 +188,10 @@ function(cxx_library_with_type name type cxx_flags)
     endif()
     target_link_libraries(${name} PUBLIC ${threads_spec})
   endif()
+
+  if (NOT "${CMAKE_VERSION}" VERSION_LESS "3.8")
+    target_compile_features(${name} PUBLIC cxx_std_11)
+  endif()
 endfunction()
 
 ########################################################################


kuzkry:
Hey @OlivierLDff.
Indeed `target_compile_features` will work for users of cmake >= 3.1 but cxx_std_11 was introduced in cmake 3.8 (see [release notes](https://cmake.org/cmake/help/v3.8/release/3.8.html?highlight=cxx_std_11#c-c)). I think this pull request will break build for users of cmake in versions [3.1, 3.8).

OlivierLDff:
Ok i see, would this PR be considered ok if i'm checking that cxx_std_11 is available?
Either with hard code version check
``` cmake
if (NOT "${CMAKE_VERSION}" VERSION_LESS "3.8")
  target_compile_features(${name} PUBLIC cxx_std_11)
endif()
```

Or by checking in `CMAKE_CXX_KNOWN_FEATURES` that `cxx_std_11` is present
``` cmake
list (FIND ${CMAKE_CXX_KNOWN_FEATURES} "cxx_std_11" _index)
if (${_index} GREATER -1)
  if(COMMAND target_compile_features)
    target_compile_features(${name} PUBLIC cxx_std_11)
  endif()
endif()
```

kuzkry:
Sorry, I don't know what are good CMake practices. I only noticed what I noticed :)

OlivierLDff:
Yes i don't know what would be the best practice too.
I think `if (NOT "${CMAKE_VERSION}" VERSION_LESS "3.8")` is the way to go since in is used in the same file `googletest/cmake/internal_utils.cmake`

seze4251:
I put up another pull request that does essentially the same thing: https://github.com/google/googletest/pull/2873

OlivierLDff:
Any news on this PR?
I see some discussion here https://github.com/google/googletest/pull/2736 about the same issue.
I think we should at least merge this PR for user using CMake v3.8 or greater.

googletest/cmake/internal_utils.cmake
```diff
@@ -188,6 +188,10 @@ function(cxx_library_with_type name type cxx_flags)
     endif()
     target_link_libraries(${name} PUBLIC ${threads_spec})
   endif()
+
+  if (NOT "${CMAKE_VERSION}" VERSION_LESS "3.8")
```
derekmauro, line 192: I'm wondering if we can help out users using CMake < 3.8 with an else branch. Maybe something like `cxx_auto_type`, `cxx_nullptr`, and `cxx_static_assert` to effective say C++11 for those users?

OlivierLDff: Or check that user defined `CMAKE_CXX_STANDARD` greater than 11 ? If not issue a warning saying build might break

derekmauro: I'm happy accepting this as is. We can iterate if people complain about older cmake versions.

PR #2774, opened 2020-03-28
Simple

Quuxplusone:

PR #2758, opened 2020-03-22
Fix Clang's `-Wdeprecated-copy` warnings in C++20

Quuxplusone:
Fixes #1305.
(I think. Hopefully you've got a buildbot that will tell me if this breaks something.)

PR #2815, opened 2020-04-16
Fix all -Wdeprecated warnings; eliminate GTEST_DISALLOW_ASSIGN_

### LINKED ISSUE #1305: Clang warning with -Wdeprecated in MATCHER_P generated matchers
As reported by @francoechea in google/googlemock#211 , in the expansion of `MATCHER_P` there's a -Wdeprecated warning by Clang because the class defines a copy assignment operator (to make it private), but does not define a copy constructor, and implicit copy constructor generation is deprecated in such cases in C++11:

     definition of implicit copy constructor for 'FooMatcherP2<double, double>' is deprecated because it has a user-declared copy assignment operator [-Werror,-Wdeprecated]

When I bumped into this myself, I tried simply adding a copy constructor like:

    name##MatcherP(const name##MatcherP &other) : p0(other.p0) {\
    }\

to the `MATCHER_P` macro I was using, and it seemed to work.

Should perhaps whatever code generates these `MATCHER_P` macros make sure they have explicitly defined copy constructors like this?

Quuxplusone:
This is the spiritual successor to #2774 (which I tried and failed to reopen) and #2758.
Fixes #1305.

Remove all uses of `GTEST_DISALLOW_{MOVE_,}ASSIGN_`.
    
    None of these are strictly needed for correctness.
    A large number of them (maybe all of them?) trigger `-Wdeprecated`
    warnings on Clang trunk as soon as you try to use the implicitly
    defaulted (but deprecated) copy constructor of a class that has
    deleted its copy assignment operator.
    
    By declaring a deleted copy assignment operator, the old code
    also caused the move constructor and move assignment operator
    to be non-declared. This means that the old code never got move
    semantics -- "move-construction" would simply call the defaulted
    (but deprecated) copy constructor instead. With the new code,
    "move-construction" calls the defaulted move constructor, which
    I believe is what we want to happen. So this is a runtime
    performance optimization.

Add `-Wdeprecated` to the build configuration.

The change in `VariadicMatcher` looks deceptively simple, but I actually still don't really understand what's going on with it; feel free to dig deeper if you can.

@mbxx @zhangxy988 please take a look!

### DIFF
--- a/googlemock/include/gmock/gmock-actions.h
+++ b/googlemock/include/gmock/gmock-actions.h
@@ -570,13 +570,9 @@ class PolymorphicAction {
 
    private:
     Impl impl_;
-
-    GTEST_DISALLOW_ASSIGN_(MonomorphicImpl);
   };
 
   Impl impl_;
-
-  GTEST_DISALLOW_ASSIGN_(PolymorphicAction);
 };
 
 // Creates an Action from its implementation and returns it.  The
@@ -717,13 +713,9 @@ class ReturnAction {
    private:
     bool performed_;
     const std::shared_ptr<R> wrapper_;
-
-    GTEST_DISALLOW_ASSIGN_(Impl);
   };
 
   const std::shared_ptr<R> value_;
-
-  GTEST_DISALLOW_ASSIGN_(ReturnAction);
 };
 
 // Implements the ReturnNull() action.
@@ -784,13 +776,9 @@ class ReturnRefAction {
 
    private:
     T& ref_;
-
-    GTEST_DISALLOW_ASSIGN_(Impl);
   };
 
   T& ref_;
-
-  GTEST_DISALLOW_ASSIGN_(ReturnRefAction);
 };
 
 // Implements the polymorphic ReturnRefOfCopy(x) action, which can be
@@ -831,13 +819,9 @@ class ReturnRefOfCopyAction {
 
    private:
     T value_;
-
-    GTEST_DISALLOW_ASSIGN_(Impl);
   };
 
   const T value_;
-
-  GTEST_DISALLOW_ASSIGN_(ReturnRefOfCopyAction);
 };
 
 // Implements the polymorphic ReturnRoundRobin(v) action, which can be
@@ -894,8 +878,6 @@ class AssignAction {
  private:
   T1* const ptr_;
   const T2 value_;
-
-  GTEST_DISALLOW_ASSIGN_(AssignAction);
 };
 
 #if !GTEST_OS_WINDOWS_MOBILE
@@ -917,8 +899,6 @@ class SetErrnoAndReturnAction {
  private:
   const int errno_;
   const T result_;
-
-  GTEST_DISALLOW_ASSIGN_(SetErrnoAndReturnAction);
 };
 
 #endif  // !GTEST_OS_WINDOWS_MOBILE
@@ -1024,13 +1004,9 @@ class IgnoreResultAction {
         OriginalFunction;
 
     const Action<OriginalFunction> action_;
-
-    GTEST_DISALLOW_ASSIGN_(Impl);
   };
 
   const A action_;
-
-  GTEST_DISALLOW_ASSIGN_(IgnoreResultAction);
 };
 
 template <typename InnerAction, size_t... I>
@@ -1467,13 +1443,7 @@ auto InvokeArgumentAdl(AdlTag, F f, Args... args) -> decltype(f(args...)) {
       template <GMOCK_ACTION_TEMPLATE_ARGS_NAMES_>                            \
       return_type gmock_PerformImpl(GMOCK_ACTION_ARG_TYPES_AND_NAMES_) const; \
       GMOCK_ACTION_FIELD_PARAMS_(params)                                      \
-                                                                              \
-     private:                                                                 \
-      GTEST_DISALLOW_ASSIGN_(gmock_Impl);                                     \
     };                                                                        \
-                                                                              \
-   private:                                                                   \
-    GTEST_DISALLOW_ASSIGN_(full_name);                                        \
   };                                                                          \
   template <GMOCK_ACTION_TYPENAME_PARAMS_(params)>                            \
   inline full_name<GMOCK_ACTION_TYPE_PARAMS_(params)> name(                   \
@@ -1512,13 +1482,7 @@ auto InvokeArgumentAdl(AdlTag, F f, Args... args) -> decltype(f(args...)) {
       }                                                                       \
       template <GMOCK_ACTION_TEMPLATE_ARGS_NAMES_>                            \
       return_type gmock_PerformImpl(GMOCK_ACTION_ARG_TYPES_AND_NAMES_) const; \
-                                                                              \
-     private:                                                                 \
-      GTEST_DISALLOW_ASSIGN_(gmock_Impl);                                     \
     };                                                                        \
-                                                                              \
-   private:                                                                   \
-    GTEST_DISALLOW_ASSIGN_(name##Action);                                     \
   };                                                                          \
   inline name##Action name() { return name##Action(); }                       \
   template <typename F>                                                       \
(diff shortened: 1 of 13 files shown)


Quuxplusone:
@zhangxy988: Sure, I think I'm amenable to `=default`'ed members. But can you show what you mean via a comment on a specific line of code, because I can't tell what exactly you mean by "these types." All types? Just the Matchers? Other? Does that include making changes to 0d2269c b543eb7 40827d9 as well?

Also, for types with `const` members, their move-constructor will actually copy-construct the `const` member (that's one reason I personally say "member data should never be const-qualified," which I guess I forgot to mention in [my blog post on `const`](https://quuxplusone.github.io/blog/2019/01/03/const-is-a-contract/)). [Godbolt.](https://godbolt.org/z/K9pcn4)  So when you say "these types should still be move constructible," do you mean that they should be _cheaply_ move-constructible, i.e., I should remove the `const` from data members? (I expect you don't mean that, but if you did, I'd like to do it in yet another separate PR.)

googlemock/include/gmock/gmock-actions.h
```diff
@@ -570,13 +570,9 @@ class PolymorphicAction {
 
    private:
     Impl impl_;
-
```
zhangxy988, line 573: The original intent is to make `MonomorphicImpl` unassignable, so I think we should explicitly write:
`MonomorphicImpl& operator=(const MonomorphicImpl&) = delete;`
This should match the old semantics.

googlemock/include/gmock/gmock-actions.h
```diff
@@ -717,13 +713,9 @@ class ReturnAction {
    private:
     bool performed_;
     const std::shared_ptr<R> wrapper_;
-
-    GTEST_DISALLOW_ASSIGN_(Impl);
```
zhangxy988, line 721: Since the original intent is to make `Impl` not copy assignable. We should explicitly write
`Impl& operator=(const Impl&) = delete;`
not relying on the `const` member to disable the assignment operator.
You are right that `const` member makes move ctor call copy ctor even we declare move ctor `=default`, but I think it's better to make the intention clear if we want this type to be move constructible (even though under the hood it invokes copy constructor). If the `const` somehow is removed later, we will automatically get the optimization.
However, feel free to NOT declare the move ctor here. It's technically a separate thing that could be handled separately.

googletest/test/googletest-port-test.cc
```diff
@@ -90,10 +90,10 @@ TEST(IsXDigitTest, ReturnsFalseForWideNonAscii) {
 
 class Base {
  public:
-  // Copy constructor and assignment operator do exactly what we need, so we
-  // use them.
   Base() : member_(0) {}
   explicit Base(int n) : member_(n) {}
+  Base(const Base&) = default;
```
zhangxy988, line 95: Are these two lines of change related to the removal of `GTEST_DISALLOW_ASSIGN_`?

googlemock/include/gmock/gmock-actions.h
```diff
@@ -570,13 +570,9 @@ class PolymorphicAction {
 
    private:
     Impl impl_;
-
-    GTEST_DISALLOW_ASSIGN_(MonomorphicImpl);
```
zhangxy988, line 574: `MonomorphicImpl& operator=(const MonomorphicImpl&) = delete;`
The move ctor doesn't seem to matter for this class.

googlemock/include/gmock/gmock-actions.h
```diff
@@ -717,13 +713,9 @@ class ReturnAction {
    private:
     bool performed_;
     const std::shared_ptr<R> wrapper_;
-
-    GTEST_DISALLOW_ASSIGN_(Impl);
   };
 
   const std::shared_ptr<R> value_;
-
-  GTEST_DISALLOW_ASSIGN_(ReturnAction);
```
zhangxy988, line 726: If there is a const member, we should still explicitly write
`Impl& operator=(const Impl&) = delete;`
Regarding move ctor, even though it invokes the copy ctor, I think it's more clear to write
`Impl(Impl&&) = default;`, otherwise it would not be implicitly declared due to the user-declared copy assignment operator.
If the `const` is removed some day, we will automatically get the optimization.

googlemock/include/gmock/gmock-spec-builders.h
```diff
@@ -499,7 +499,8 @@ class GTEST_API_ Expectation {
  public:
   // Constructs a null object that doesn't reference any expectation.
   Expectation();
-
+  Expectation(const Expectation&) = default;
```
zhangxy988, line 502: Are these 2 lines of change related to the removal `GTEST_DISALLOW_ASSIGN_(ExpectationBase)`?
What about the move ctor and assignment operator?

googletest/test/googletest-param-test-test.cc
```diff
@@ -490,16 +490,15 @@ TEST(CombineTest, CombineWithMaxNumberOfParameters) {
 class NonDefaultConstructAssignString {
  public:
   NonDefaultConstructAssignString(const std::string& s) : str_(s) {}
+  NonDefaultConstructAssignString() = delete;
+  NonDefaultConstructAssignString(const NonDefaultConstructAssignString&) = default;
+  NonDefaultConstructAssignString& operator=(const NonDefaultConstructAssignString&) = delete;
+  ~NonDefaultConstructAssignString() = default;
```
zhangxy988, line 496: What is the reason for explicitly declaring the destructor?

googlemock/include/gmock/gmock-spec-builders.h
Quuxplusone: These are in a separate commit (look at the "Commits" tab). It fixes a -Wdeprecated warning.
I guess it would make sense to default the move operations too, yeah. Will do.

googletest/test/googletest-param-test-test.cc
Quuxplusone: No reason. I'm just in the habit of declaring all the special members — at least the Big 3 or the Big 5, but not, like, 2 of the 3, or 4 of the 5. I can remove the destructor declaration here if you want.

googletest/test/googletest-port-test.cc
Quuxplusone: They're in a separate commit, and related to -Wdeprecated.

googlemock/include/gmock/gmock-actions.h
Quuxplusone: If we `=delete` the copy assignment operator, then (to avoid relying on deprecated behavior) we must also put something for the copy constructor (e.g. `=default` it). Clang doesn't actually warn on this specific deprecated behavior yet, but I just filed https://bugs.llvm.org/show_bug.cgi?id=45634 about it.

Anyway, I contend that in all these places where I removed `GTEST_DISALLOW_ASSIGN_ `, there _was_ no discernable intent. There's no obvious reason that these types _should_ be non-assignable; and on the flip side, most of them are accidentally non-assignable anyway even if you remove the deleted assignment operator. I think it's better to follow the Rule of Zero unless we have a specific metaprogramming reason that we _need_ `std::is_assignable<T&, const T&>::value` to be `false` or else code would break. (And in that case, there should be unit tests to regression-test that code that would have broken.)

googlemock/include/gmock/gmock-spec-builders.h
zhangxy988: Do you want to add the move ctor and assignment before I import it?

Quuxplusone: Done!

zhangxy988:
Though this macro is supposed to be internal, it's unfortunately used inside Google and removing it would cause breakages. I am going to clean them up, but you can choose to wait for my cleanup to complete (which delays this PR) or keep the macro for now. Let me know how you want to proceed.

Quuxplusone:
@zhangxy988 wrote:
> Though this macro is supposed to be internal, it's unfortunately used inside Google and removing it would cause breakages. I am going to clean them up, but you can choose to wait for my cleanup to complete (which delays this PR) or keep the macro for now.

No problem. I've modified the first commit in this PR so that it no longer removes the macro definitions.

PR #2819, opened 2020-04-20
Correctly name the extended header

tamird:


vslashg:
This is fixed by a conflicting change.

tamird:
Which change? This is still incorrect on master at far as I can tell.

PR #2824, opened 2020-04-24
Handle c++17 nested namespaces in gmock generation

xwl:


vslashg:
We are not opposed to investigating this change, but you will need to sign the CLA.

xwl:
@googlebot I signed it!

Issue #2650, opened 2019-12-31
Compiler emits warning on MOCK_METHOD missing argument to '...' parameter of variadic macro


keithmendozasr:
When compiling with the -pedantic switch the compiler clang++ 11.0.0 emits the following warning on all calls to MOCK_METHOD:
```
warning: must specify at least one argument for '...' parameter of variadic macro [-Wgnu-zero-variadic-macro-arguments]
    MOCK_METHOD(int, select,
    ^
/Users/kmendoza/rv6_mount/tlslookieloo/googletest/googlemock/include/gmock/gmock-function-mocker.h:50:3: note: expanded from macro 'MOCK_METHOD'
  GMOCK_PP_VARIADIC_CALL(GMOCK_INTERNAL_MOCK_METHOD_ARG_, __VA_ARGS__)
  ^
/Users/kmendoza/rv6_mount/tlslookieloo/googletest/googlemock/include/gmock/internal/gmock-pp.h:49:7: note: expanded from macro 'GMOCK_PP_VARIADIC_CALL'
      GMOCK_PP_CAT(_Macro, GMOCK_PP_NARG(__VA_ARGS__))(__VA_ARGS__))
      ^
/Users/kmendoza/rv6_mount/tlslookieloo/googletest/googlemock/include/gmock/internal/gmock-pp.h:5:30: note: expanded from macro 'GMOCK_PP_CAT'
#define GMOCK_PP_CAT(_1, _2) GMOCK_PP_INTERNAL_CAT(_1, _2)
                             ^
note: (skipping 5 expansions in backtrace; use -fmacro-backtrace-limit=0 to see all)
/Users/kmendoza/rv6_mount/tlslookieloo/googletest/googlemock/include/gmock/internal/gmock-pp.h:103:15: note: expanded from macro 'GMOCK_PP_NARG0'
  GMOCK_PP_IF(GMOCK_PP_IS_EMPTY(__VA_ARGS__), 0, GMOCK_PP_NARG(__VA_ARGS__))
              ^
/Users/kmendoza/rv6_mount/tlslookieloo/googletest/googlemock/include/gmock/internal/gmock-pp.h:80:30: note: expanded from macro 'GMOCK_PP_IS_EMPTY'
  GMOCK_PP_INTERNAL_IS_EMPTY(GMOCK_PP_HAS_COMMA(__VA_ARGS__),                \
                             ^
[...]

PR #2826, opened 2020-04-27
Fix `-Wgnu-zero-variadic-macro-arguments`

Kangz:
Passing zero arguments to the variadic part of a macro is a GNU
extension and triggers warnings when build projects using GTest and
GMock with `-pedantic`.

 - Fix uses of `GMOCK_PP_INTERNAL_16TH` to always receive at least 17
arguments. (this was triggered when `GMOCK_PP_NARG` or `GMOCK_PP_HAS_COMMA`
were used with an argument containing no commas).
 - Fix `GMOCK_PP_HEAD` to append a dummy unused argument so that
`GMOCK_PP_INTERNAL_HEAD` always has two arguments.
 - Fix `TYPED_TEST_SUITE` and `INSTANTIATE_TYPED_TEST_SUITE_P` when called
without a name generator by making the argument default to no token
before proxying to an internal implementation.

### DIFF
--- a/googlemock/include/gmock/internal/gmock-pp.h
+++ b/googlemock/include/gmock/internal/gmock-pp.h
@@ -28,16 +28,16 @@
 // Requires: the number of arguments after expansion is at most 15.
 #define GMOCK_PP_NARG(...) \
   GMOCK_PP_INTERNAL_16TH(  \
-      (__VA_ARGS__, 15, 14, 13, 12, 11, 10, 9, 8, 7, 6, 5, 4, 3, 2, 1))
+      (__VA_ARGS__, 15, 14, 13, 12, 11, 10, 9, 8, 7, 6, 5, 4, 3, 2, 1, 0))
 
 // Returns 1 if the expansion of arguments has an unprotected comma. Otherwise
 // returns 0. Requires no more than 15 unprotected commas.
 #define GMOCK_PP_HAS_COMMA(...) \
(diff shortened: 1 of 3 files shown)


Bagira80:
@Kangz Is there a specific reason why you closed this pull-request before it got reviewed/merged?

It looks like a very useful addition to GoogleTest and should be able to fix #2650.

With C++20 this fix will probably be obsolete because calling variadic macros without arguments should be valid then.  
However, for someone not using C++20 this is a very useful fix. So I would recommend you re-open this pull-request.

Besides, I think it would be great if this got backport to the [v1.10.x](https://github.com/google/googletest/tree/v1.10.x) branch, too, as it prevents compilation if someone wants to compile his/her unit-tests with GoogleMock and compiler-flags `-Werror -pedantic`.

Kangz:
I closed it because I work at Google and the contribution guidelines say googlers should contribute internally when possible. Hopefully it is in an export very soon!

Kangz:
Landed as https://github.com/google/googletest/commit/a09ea700d32bab83325aff9ff34d0582e50e3997.

Bagira80:
It seems this will not be backported into the [v1.10.x](https://github.com/google/googletest/tree/v1.10.x) branch!?

Kangz:
> It seems this will not be backported into the [v1.10.x](https://github.com/google/googletest/tree/v1.10.x) branch!?

No idea how this usually works. Maybe you can make a merge PR on it? IDK.

PR #2837, opened 2020-05-05
Detect proto messages based on presence of DebugString.

### LINKED ISSUE #2817: Proto messages from open source protobufs are printed as bytes.
Open source protobuf messages derive from `google::protobuf::MessageLite`.

GTest enables ProtbufPrinter based on:

```
template <typename T>
struct IsAProtocolMessage
   : public std::is_convertible<const T*, const ::proto2::MessageLite*> {};
```

So open-source protobufs are printed as bytes.

This was previously closed as "inactive": #1102

The issue still occurs at current master.

inazarenko:
fixes #2817 (allows open-source proto messages to be printed using DebugString).

Second attempt, since #2818 was rejected.


inazarenko:
Note that this required more metaprogramming magic that I'm really comfortable with, so I'm prepared for a few round-trips of feedback before this is ready to merge. Please let me know if this is the approach you had in mind @asoffer

inazarenko:
Gentle ping :)

googletest/include/gtest/internal/gtest-internal.h
```diff
@@ -880,11 +880,34 @@ class GTEST_API_ Random {
...
 
-// IsAProtocolMessage<T>::value is a compile-time bool constant that's
-// true if and only if T is type proto2::MessageLite or a subclass of it.
+// IsAProtocolMessage<T>::value is a compile-time bool constant that's true if
+// and only if T has methods DebugString() and ShortDebugString() that return
+// std::string.
 template <typename T>
-struct IsAProtocolMessage
-    : public std::is_convertible<const T*, const ::proto2::MessageLite*> {};
+class IsAProtocolMessage {
```
asoffer, line 887: Since we're no longer checking for being a protocol message but rather for DebugString and ShortDebugString, can we change the name to HasDebugStringAndShortDebugString?

inazarenko: Sure. Should I do it here, or will you pull the changes in through Google's internal source control?

asoffer:
@inazarenko this should be working now though as you can see there are merge conflicts because we made a few tweaks to the implementation. Leaving this for you to close.

Issue #2822, opened 2020-04-23
Visual Studio 2015 Windows XP (v140_xp) Platform Toolset


aidanoconnor:
Master and release-1.10.0 CMake builds fail when the Platform Toolset in Visual Studio 2015 is set to Windows XP (v140_xp). It is unable to find header **debugapi.h**
Tag release-1.8.1 does not have this issue. And the standard Platform Toolset (v140) does not have this issue.

```
C:\git\googletest-release-1.10.0\_build>cmake -G "Visual Studio 14 2015" -A "Win32" -T "v140_xp" ..\
-- Selecting Windows SDK version  to target Windows 10.0.17763.
-- The C compiler identification is MSVC 19.0.24215.1
-- The CXX compiler identification is MSVC 19.0.24215.1
-- Check for working C compiler: C:/Program Files (x86)/Microsoft Visual Studio 14.0/VC/bin/cl.exe
-- Check for working C compiler: C:/Program Files (x86)/Microsoft Visual Studio 14.0/VC/bin/cl.exe -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: C:/Program Files (x86)/Microsoft Visual Studio 14.0/VC/bin/cl.exe
-- Check for working CXX compiler: C:/Program Files (x86)/Microsoft Visual Studio 14.0/VC/bin/cl.exe -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Found PythonInterp: C:/Program Files (x86)/Python38-32/python.exe (found version "3.8.1")
-- Looking for pthread.h
-- Looking for pthread.h - not found
-- Found Threads: TRUE
-- Configuring done
-- Generating done
-- Build files have been written to: C:/git/googletest-release-1.10.0/_build

C:\git\googletest-release-1.10.0\_build>cmake --build . --config Debug
Microsoft (R) Build Engine version 14.0.25420.1
Copyright (C) Microsoft Corporation. All rights reserved.

  Checking Build System
  Building Custom Rule C:/git/googletest-release-1.10.0/googlemock/CMakeLists.txt
  gtest-all.cc
[...]

PR #2838, opened 2020-05-05
Export Test - Do Not Merge

vslashg:
Export Test - Do Not Merge

Addresses Github issue (https://github.com/google/googletest/issues/2822) by following the Microsoft documention (https://docs.microsoft.com/en-us/windows/win32/api/debugapi/nf-debugapi-isdebuggerpresent) to include capital-W Windows.h before including debugapi.h.

See "Header	debugapi.h (include Windows.h)"

### DIFF
--- a/googlemock/README.md
+++ b/googlemock/README.md
@@ -7,26 +7,26 @@ derive better designs of your system and write better tests.
 
 It is inspired by:
 
-*   [jMock](http://www.jmock.org/),
-*   [EasyMock](http://www.easymock.org/), and
-*   [Hamcrest](http://code.google.com/p/hamcrest/),
+*   [jMock](http://www.jmock.org/)
+*   [EasyMock](http://www.easymock.org/)
+*   [Hamcrest](http://code.google.com/p/hamcrest/)
 
-and designed with C++'s specifics in mind.
+It is designed with C++'s specifics in mind.
 
 gMock:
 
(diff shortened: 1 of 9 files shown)


mstorsjo:
This merge request broke cross compilation with mingw toolchains from case sensitive file systems, as `windows.h` is available only with a lower case `w` there.

I'm very suspicious about this fix - I doubt it fixes anything. I've asked the original issue poster in #2822 to test and clarify whether it made any difference.

The documentation linked, https://docs.microsoft.com/en-us/windows/win32/api/debugapi/nf-debugapi-isdebuggerpresent, which says "Header: debugapi.h (include Windows.h)" does not mean "you must include Windows.h with a capital W in order to include debugapi.h".

It says "the declaration of the function is in debugapi.h, but we recommend that you just include Windows.h instead and we guarantee that this function is visible just by doing that".

The issue in #2822 is that the debugapi.h header simply is missing. Including windows.h with a different capitalization is not going to change that. When using an older SDK targeting Windows XP, debugapi.h does not exist, and the definition of IsDebuggerPresent is in winbase.h.

Hence, the correct way to read the documentation about the function IsDebuggerPresent is "just include windows.h, even if the declaration currently is in debugapi.h - don't try to explicitly include debugapi.h as that is less portable".

Thus, the correct fix is to revert this change (remove the extra include of windows.h with a capital W) and just remove the include of debugapi.h.

PR #2839, opened 2020-05-06
On Cygwin, ensure that POSIX functions are available.

### LINKED ISSUE #813: gtest-port.h fileno not declared in scope
I'm getting errors in gtest-port.h. Have tried various flags none are working. 

Have searched and found similar errors, but no definitive solutions.

Any help is appreciated.

Thanks!

CYGWIN_NT-6.1 BaguaDragon 2.5.2(0.297/5/3) 2016-06-23 14:29 x86_64 Cygwin
g++ (GCC) 5.4.0

g++ **-static -std=c++0x -DGTEST_OS_CYGWIN=1** -pthread -MMD -c main.cpp -o build/main.o -I /home/ssi112/cpp/googletest-master/googletest -I /home/ssi112/cpp/googletest-master/googletest/include -I . -I ../src -I ../../class-library/include
In file included from /home/ssi112/cpp/googletest-master/googletest/include/gtest/internal/gtest-internal.h:40:0,
                 from /home/ssi112/cpp/googletest-master/googletest/include/gtest/gtest.h:58,
                 from main.cpp:9:
/home/ssi112/cpp/googletest-master/googletest/include/gtest/internal/gtest-port.h: In function ‘int testing::internal::posix::FileNo(FILE_)’:
/home/ssi112/cpp/googletest-master/googletest/include/gtest/internal/gtest-port.h:2340:51: error: ‘fileno’ was not declared in this scope
 inline int FileNo(FILE_ file) { return fileno(file); }
                                                   ^
/home/ssi112/cpp/googletest-master/googletest/include/gtest/internal/gtest-port.h: In function ‘char\* testing::internal::posix::StrDup(const char_)’:
/home/ssi112/cpp/googletest-master/googletest/include/gtest/internal/gtest-port.h:2346:57: error: ‘strdup’ was not declared in this scope
 inline char_ StrDup(const char\* src) { return strdup(src); }
                                                         ^
/home/ssi112/cpp/googletest-master/googletest/include/gtest/internal/gtest-port.h: In function ‘FILE\* testing::internal::posix::FDOpen(int, const char_)’:
/home/ssi112/cpp/googletest-master/googletest/include/gtest/internal/gtest-port.h:2374:71: error: ‘fdopen’ was not declared in this scope
 inline FILE_ FDOpen(int fd, const char\* mode) { return fdopen(fd, mode); }
                                                                       ^
In file included from main.cpp:12:0:
aggregate_test.hpp: In member function ‘virtual void AggregateTest_verifySplitAndFindAggregateFunctionality_Test::TestBody()’:
aggregate_test.hpp:11:32: error: conversion from ‘void’ to non-scalar type ‘std::string {aka std::basic_string<char>}’ requested
  std::string actual = aggregate(1, line);
                                ^
make: **\* [Makefile:28: build/main.o] Error 1

jfinkhaeuser:
Fixes #813 

To describe the problem outside of the issue, the C++ standards C++11 and later are based on ANSI C standards, which do not define some functions found on POSIX compliant systems.

The functions in this set that gtest seems to use are e.g. `fileno`, `strdup` and `mkstemp` that I found.

These functions are available on Cygwin, but hidden in the system headers when `-std=c++11`, etc. are passed. To un-hide them, either:

- Pass `-std=gnu++11`
- Pass `-D_POSIX_C_SOURCE=200809L` or a higher value.
- Pass `-D_GNU_SOURCE`, as GNU encompasses POSIX.

See `/usr/include/sys/features.h` for details.

The PR isn't ideal, as the defines have to be set before the *first* system header is included. As demonstrated by the `*_main.cc` test files, this includes standard C++ headers. To illustrate the issue, I've edited those files in the PR - but anyone including first system headers without setting this define will still run into the issue even with this PR applied.

Still, this is as far as I can get it today.


jfinkhaeuser:
@googlebot I signed it!

asoffer:
Thanks for this PR. We're a bit hesitant to accept this due to the brittle nature of ordered includes. Can you pass -D_GNU_SOURCE on the command line?

jfinkhaeuser:
I also don't like that it's brittle, and passing this flag is the workaround. It's just also not in the spirit of providing a package that's portable *in itself*.

Personally I'd like to see a documentation change that points out the need to pass the flag. But this PR as such I have no emotional attachment to.

vslashg:
Changes to documentation sound reasonable to me.  Do you have things you would edit beyond the readme.md?

We probably don't have to get into the "C++11 is based on ANSI C" bit; it's true but maybe off-topic.  I think it's sufficient to say that we require both C++11, and either Windows or POSIX.  A suggestion telling Cygwin users how to turn on POSIX functionality in their build files also feels welcome.

jfinkhaeuser:
You're right, that's the documentation change that would be needed. I can update this PR quickly.

tanzislam:
Encoding this compilation flag requirement in https://www.github.com/google/googletest/tree/master/googletest%2Fcmake%2FConfig.cmake.in may be helpful. References:
- https://people.freedesktop.org/~dbn/pkg-config-guide.html#writing
- https://cmake.org/cmake/help/latest/module/CMakePackageConfigHelpers.html

jfinkhaeuser:
No. This isn't about building googletest, it's about building downstream projects. This cannot be solved in `googletest/cmake/Config.cmake.in`, because that file is not part of the downstream projects.

PR #2866, opened 2020-05-26
[v1.10.x | backport] Fix `-Wgnu-zero-variadic-macro-arguments` in GMock

Bagira80:
Passing zero arguments to the variadic part of a macro is a GNU extension and triggers warnings when build projects using GMock with `-pedantic`.

 - Fix uses of `GMOCK_PP_INTERNAL_INTERNAL_16TH` to always receive at least 17 arguments. (This was triggered when `GMOCK_PP_NARG` or `GMOCK_PP_HAS_COMMA` were used with an argument containing no commas).
 - Fix `GMOCK_PP_HEAD` to append a dummy unused argument so that `GMOCK_PP_INTERNAL_HEAD_CAT` always has two arguments.

(backport of commit a09ea700d32bab83325aff9ff34d0582e50e3997)

### DIFF
--- a/googlemock/include/gmock/internal/gmock-pp.h
+++ b/googlemock/include/gmock/internal/gmock-pp.h
@@ -46,17 +46,17 @@ static_assert(
 #define GMOCK_PP_NARG(...)                                                    \
   GMOCK_PP_INTERNAL_NARG_CAT(                                                 \
       GMOCK_PP_INTERNAL_INTERNAL_16TH(__VA_ARGS__, 15, 14, 13, 12, 11, 10, 9, \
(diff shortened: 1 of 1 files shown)


derekmauro:
Unfortunately we can't just accept backports for arbitrary issues into our releases. We will fix genuine bugs, but this is just a compile warning. Release stability is important. If you want the latest and greatest fixes, just use the master branch.

Bagira80:
This is not an "arbitrary issue". It is a hard compilation error for users enabling the "-pedantic-errors" (GCC) compiler option.

This pull-request is no different than the former (latest) pull-request (#2723) accepted for this branch.

PR #2873, opened 2020-05-31
Set minimum C++ version to be C++11

seze4251:
Hey All,

I just pulled down GoogleTest on my MacOS and tried to build it locally.  I ran into some build errors with CMake because clang was defaulting to a C++ version less than C++11.  I edited the CMake files in GoogleTest, and GoogleMock to enforce a minimum version of C++11, which will allow GoogleTest to work out-of-the-box on MacOS.

Thanks,
Seth

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,9 +24,6 @@ Win32-Release/
 x64-Debug/
 x64-Release/
 
-# Eclipse
-.project
-
 # Ignore autoconf / automake files
 Makefile.in
 aclocal.m4
--- a/.gitignore
+++ b/.gitignore
@@ -24,6 +24,9 @@ Win32-Release/
 x64-Debug/
 x64-Release/
 
+# Eclipse
+.project
+
 # Ignore autoconf / automake files
 Makefile.in
 aclocal.m4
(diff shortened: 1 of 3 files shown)


seze4251:
@googlebot I signed it!

.gitignore
```diff
@@ -24,6 +24,9 @@ Win32-Release/
 x64-Debug/
 x64-Release/
 
+# Eclipse
```
derekmauro, line 27: Please revert the changes to this file. This really belong in a global file <https://docs.github.com/en/github/using-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer> otherwise this file will just be a collection of IDE specific directories.

googlemock/CMakeLists.txt
```diff
@@ -78,6 +78,11 @@ set(gmock_build_include_dirs
   "${gtest_SOURCE_DIR}")
 include_directories(${gmock_build_include_dirs})
 
+# Set minimum C++ version to C++11
+set(CMAKE_CXX_STANDARD 11)
```
derekmauro, line 82: If we accept this change, how would I build GoogleTest if I wanted C++20 support?

.gitignore
seze4251: reverted the .gitignore.  Putting the IDE stuff in a global git ignore is a really good idea.  My company just shoves them in our local repos, so I was used to that.  I will use your approach going forward.  Thanks!

googlemock/CMakeLists.txt
seze4251: The other pull request that just got merged is better.  I will close this.

PR #2891, opened 2020-06-10
Fix build issue for MinGW

### LINKED ISSUE #2885: Rtools 3.5 fails to build lines with std::wstring_convert in gtest-port.h
To get around the compiler error, 
```
diff --git a/googletest/include/gtest/internal/gtest-port.h b/googletest/include/gtest/internal/gtest-port.h
index 948a21f9..2df0f9ba 100644
--- a/googletest/include/gtest/internal/gtest-port.h
+++ b/googletest/include/gtest/internal/gtest-port.h
@@ -2046,7 +2046,7 @@ GTEST_DISABLE_MSC_DEPRECATED_PUSH_()
 inline int ChDir(const char* dir) { return chdir(dir); }
 #endif
 inline FILE* FOpen(const char* path, const char* mode) {
-#if GTEST_OS_WINDOWS
+#if GTEST_OS_WINDOWS && !GTEST_OS_WINDOWS_MINGW
   struct wchar_codecvt : public std::codecvt<wchar_t, char, std::mbstate_t> {};
   std::wstring_convert<wchar_codecvt> converter;
   std::wstring wide_path = converter.from_bytes(path);
```

zoddicus:
Fixes #2885

### DIFF
--- a/googletest/include/gtest/internal/gtest-port.h
+++ b/googletest/include/gtest/internal/gtest-port.h
@@ -2055,7 +2055,7 @@ GTEST_DISABLE_MSC_DEPRECATED_PUSH_()
 inline int ChDir(const char* dir) { return chdir(dir); }
 #endif
 inline FILE* FOpen(const char* path, const char* mode) {
-#if GTEST_OS_WINDOWS
+#if GTEST_OS_WINDOWS && !GTEST_OS_WINDOWS_MINGW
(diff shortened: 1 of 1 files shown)


dneto0:
Please take a look at this.  This is breaking my team's integrations.

CJ-Johnson:
Apologies for the delay! I've been swamped as of late.

This has now been imported. Please refrain from making further changes as they will be overwritten. (317661294)

PR #2902, opened 2020-06-19
Updated googletest issue tracker url.

mshingote:
This is a minor change.
In googletest CONTRIBUTING.md file, it is found that the existing issue tracker URL is the same as the project repo URL so In order to save user clicks update the exact URL of the issue tracker.


mshingote:
Can we merge this PR it to master?

PR #2903, opened 2020-06-20
Make EXPECT_THROW and EXPECT_NO_THROW macros more informative

### LINKED ISSUE #2878: Make EXPECT_THROW and EXPECT_NO_THROW macros more informative
This is a continuation of #2071.

We could add catch case for `std::exception` to `EXPECT_THROW` to improve case when thrown exception is of an incompatible type.

If RTTI is available, we can also report the exact exception type by means of [`GetTypeName`](https://github.com/google/googletest/blob/fff8dabbf6c5641e97a646d710862508746183a2/googletest/include/gtest/internal/gtest-type-util.h#L71). This would further improve `EXPECT_THROW` and `EXPECT_NO_THROW` macros.

AmatanHead:
`EXPECT_THROW` and `EXPECT_NO_THROW` will now print exception type and message when an unexpected `std::exception`-derived error is thrown.

Fixes #2878

### DIFF
--- a/googletest/include/gtest/internal/gtest-internal.h
+++ b/googletest/include/gtest/internal/gtest-internal.h
@@ -1293,6 +1293,15 @@ constexpr bool InstantiateTypedTestCase_P_IsDeprecated() { return true; }
 
 #if GTEST_HAS_EXCEPTIONS
 
+namespace testing {
+namespace internal {
+
+class NeverThrown: public std::exception {
+};
+
+}  // namespace internal
+}  // namespace testing
+
 #if GTEST_HAS_RTTI
 
 #define GTEST_EXCEPTION_TYPE_(e) \
@@ -1306,7 +1315,16 @@ constexpr bool InstantiateTypedTestCase_P_IsDeprecated() { return true; }
 #endif  // GTEST_HAS_RTTI
 
 #define GTEST_TEST_THROW_CATCH_STD_EXCEPTION_(statement, expected_exception) \
-  catch (std::exception const& e) { \
+  catch ( \
+      typename std::conditional< \
+          std::is_same< \
+              typename std::remove_cv< \
+                  typename std::remove_reference< \
+                      expected_exception>::type>::type, \
+              std::exception>::value, \
+          const ::testing::internal::NeverThrown&, \
+          const std::exception& \
+      >::type e) { \
     gtest_msg.value = \
         "Expected: " #statement " throws an exception of type " \
         #expected_exception ".\n  Actual: it throws "; \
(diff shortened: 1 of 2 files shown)


AmatanHead:
@mattcalabrese-google can we please have someone from the project team to review this PR?

mattcalabrese-google:
On it, sorry for the delay!

googletest/include/gtest/internal/gtest-internal.h
```diff
@@ -1291,16 +1291,49 @@ constexpr bool InstantiateTypedTestCase_P_IsDeprecated() { return true; }
...
+
+#else  // GTEST_HAS_RTTI
+
+#define GTEST_EXCEPTION_TYPE_(e) \
+  std::string{"an std::exception-derived error"}
+
+#endif  // GTEST_HAS_RTTI
+
+#define GTEST_TEST_THROW_CATCH_STD_EXCEPTION_(statement, expected_exception) \
+  catch (std::exception const& e) { \
```
mattcalabrese-google, line 1309: This will unfortunately fail if the user'ss expected_exception is, itself, std::exception, since we will have two catch clauses with the same type. One way we might avoid this is do a tiny bit of metaprogramming here and basically have it catch something like

```
catch(
    typename std::conditional<
        std::is_same<
            typename std::remove_cv<
                typename std::remove_reference<expected_exception>::type>::type,
            std::exception>,
            const SomeInternalTypeWithAWhatFunction&,
            const std::exception&
        >::type e)
```

mattcalabrese-google: There may be alternative solutions here, this is just the first solution I can think of.

AmatanHead: If I understand the standard correctly, it shouldn't. `catch` clauses are matched from top to bottom, and the first one that matches the exception is called. If we have two `catch`es with the same type, the later one will be ignored.

From cppreference:

> When an exception of type E is thrown by any statement in compound-statement, it is matched against the types of the formal parameters T of each catch-clause in handler-seq, in the order in which the catch clauses are listed. The exception is a match if any of the following is true: ...

mattcalabrese-google: True, my mistake -- it's not an error, but clang at least does warn on duplicate catches by default:

https://gcc.godbolt.org/z/DC5KSS

Regardless, I think we should do what we can to avoid such warnings here.

AmatanHead: I've fixed the issue and added a test for this case.

AmatanHead:
This PR was merged.

PR #2904, opened 2020-06-20
Add matchers for testing exception properties

### LINKED ISSUE #952: EXPECT_THROW should allow testing of exception contents
There should be a nice way for EXPECT_THROW to let me verify what's in my exception. Even testing the what() contents would already be helpful.

AmatanHead:
This PR adds matchers that accept a callable and verify that when invoked, it throws an exception with the given type and properties.

Fixes #952

### DIFF
--- a/googlemock/include/gmock/gmock-matchers.h
+++ b/googlemock/include/gmock/gmock-matchers.h
@@ -4812,7 +4812,8 @@ class ExceptionMatcherImpl {
 //
 //   EXPECT_THAT(
 //       []() { throw std::runtime_error("message"); },
-//       Throws
+//       Throws<std::runtime_error>(
+//           Property(&std::runtime_error::what, HasSubstr("message"))));
 
 template <typename Err>
 PolymorphicMatcher<internal::ExceptionMatcherImpl<Err>>
--- a/googlemock/include/gmock/gmock-matchers.h
+++ b/googlemock/include/gmock/gmock-matchers.h
@@ -4843,7 +4843,7 @@ ThrowsMessage(const MessageMatcher& messageMatcher) {
   // literals, e.g. ThrowsMessage("message").
   return MakePolymorphicMatcher(
       internal::ExceptionMatcherImpl<Err>{
-          Property("description", &std::exception::what,
+          Property("what", &std::exception::what,
                    MatcherCast<std::string>(messageMatcher))});
 }
 template <typename Err, typename Message = std::string>
@@ -4854,7 +4854,7 @@ ThrowsMessageHasSubstr(const internal::StringLike<Message>& message) {
       "expected an std::exception-derived class");
   return MakePolymorphicMatcher(
       internal::ExceptionMatcherImpl<Err>{
-          Property("description", &std::exception::what, HasSubstr(message))});
+          Property("what", &std::exception::what, HasSubstr(message))});
 }
 
 #endif  // GTEST_HAS_EXCEPTIONS
--- a/googlemock/include/gmock/gmock-matchers.h
+++ b/googlemock/include/gmock/gmock-matchers.h
@@ -4774,15 +4774,15 @@ class ExceptionMatcherImpl {
   ExceptionMatcherImpl(Matcher<const Err&> matcher)
       : matcher_(std::move(matcher)) {}
 
-  void DescribeTo(::std::ostream* os) const {
-    *os << "throws an exception of type " << GetTypeName<Err>();
+  void DescribeTo(std::ostream* os) const {
+    *os << "throws an exception which is a " << GetTypeName<Err>();
     if (matcher_.GetDescriber() != nullptr) {
       *os << " which ";
       matcher_.DescribeTo(os);
     }
   }
 
-  void DescribeNegationTo(::std::ostream* os) const {
+  void DescribeNegationTo(std::ostream* os) const {
     *os << "not (";
     DescribeTo(os);
     *os << ")";
@@ -4793,7 +4793,7 @@ class ExceptionMatcherImpl {
     try {
       (void)(std::forward<T>(x)());
     } catch (const Err& err) {
-      *listener << "throws an exception of type " << GetTypeName<Err>();
+      *listener << "throws an exception which is a " << GetTypeName<Err>();
       if (matcher_.GetDescriber() != nullptr) {
         *listener << " ";
         return matcher_.MatchAndExplain(err, listener);
@@ -4826,7 +4826,6 @@ class ExceptionMatcherImpl {
 // Throws()
 // Throws(exceptionMatcher)
 // ThrowsMessage(messageMatcher)
-// ThrowsMessageHasSubstr(message)
 //
 // This matcher accepts a callable and verifies that when invoked, it throws
 // an exception with the given type and properties.
@@ -4840,10 +4839,6 @@ class ExceptionMatcherImpl {
 //   EXPECT_THAT(
 //       []() { throw std::runtime_error("message"); },
 //       ThrowsMessage<std::runtime_error>(HasSubstr("message")));
-//
-//   EXPECT_THAT(
-//       []() { throw std::runtime_error("message"); },
-//       ThrowsMessageHasSubstr<std::runtime_error>("message"));
 //
 //   EXPECT_THAT(
 //       []() { throw std::runtime_error("message"); },
@@ -4881,16 +4876,6 @@ ThrowsMessage(const MessageMatcher& messageMatcher) {
       internal::ExceptionMatcherImpl<Err>{
           Property("what", &std::exception::what,
                    MatcherCast<std::string>(messageMatcher))});
-}
-template <typename Err, typename Message = std::string>
-PolymorphicMatcher<internal::ExceptionMatcherImpl<Err>>
-ThrowsMessageHasSubstr(const internal::StringLike<Message>& message) {
-  static_assert(
-      std::is_base_of<std::exception, Err>::value,
-      "expected an std::exception-derived class");
-  return MakePolymorphicMatcher(
-      internal::ExceptionMatcherImpl<Err>{
-          Property("what", &std::exception::what, HasSubstr(message))});
 }
 
 #endif  // GTEST_HAS_EXCEPTIONS
--- a/googlemock/include/gmock/gmock-matchers.h
+++ b/googlemock/include/gmock/gmock-matchers.h
@@ -4741,6 +4741,35 @@ class ExceptionMatcherImpl {
     }
   };
 
+  // If the matchee raises an exception of a wrong type, we'd like to
+  // catch it and print its message and type. To do that, we add an additional
+  // catch clause:
+  //
+  //     try { ... }
+  //     catch (const Err&) { /* an expected exception */ }
+  //     catch (const std::exception&) { /* exception of a wrong type */ }
+  //
+  // However, if the `Err` itself is `std::exception`, we'd end up with two
+  // identical `catch` clauses:
+  //
+  //     try { ... }
+  //     catch (const std::exception&) { /* an expected exception */ }
+  //     catch (const std::exception&) { /* exception of a wrong type */ }
+  //
+  // This can cause a warning or an error in some compilers. To resolve
+  // the issue, we use a fake error type whenever `Err` is `std::exception`:
+  //
+  //     try { ... }
+  //     catch (const std::exception&) { /* an expected exception */ }
+  //     catch (const NeverThrown&) { /* exception of a wrong type */ }
+  using DefaultExceptionType = typename std::conditional<
+      std::is_same<
+          typename std::remove_cv<
+              typename std::remove_reference<Err>::type>::type,
+          std::exception>::value,
+      const NeverThrown&,
+      const std::exception&>::type;
+
  public:
   ExceptionMatcherImpl(Matcher<const Err&> matcher)
       : matcher_(std::move(matcher)) {}
@@ -4771,14 +4800,7 @@ class ExceptionMatcherImpl {
       } else {
         return true;
       }
-    } catch (
-        typename std::conditional<
-            std::is_same<
-                typename std::remove_cv<
-                    typename std::remove_reference<Err>::type>::type,
-                std::exception>::value,
-            const NeverThrown&,
-            const std::exception&>::type const& err) {
+    } catch (DefaultExceptionType err) {
 #if GTEST_HAS_RTTI
       *listener << "throws an exception of type "
           << GetTypeName(typeid(err)) << " ";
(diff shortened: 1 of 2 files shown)


vlovich:
Rather than `ThrowsMessage`/`ThrowsMessageHasSubstr`, does it makes sense to have Throws take a generic matcher to evaluate the thrown exception? There could be a convenience matcher to retrieve `.what()` from an exception & then the composition facilities of gmock could be leveraged more powerfully (e.g. to introspect user-specific exceptions).

AmatanHead:
@vlovich this is exactly what's happening: `ThrowsMessage<E>(M)` is just a shortcut for `Throws<E>(Property("what", &std::exception::what, M)`. I've introduced these shortcuts because checking contents of the message seems to be the most frequent scenario. However, you're free to use `Throws` with an arbitrary matcher. See the last example in [this snippet](https://github.com/google/googletest/issues/952#issuecomment-638968347).

googlemock/include/gmock/gmock-matchers.h
```diff
@@ -4725,6 +4725,140 @@ PolymorphicMatcher<internal::variant_matcher::VariantMatcher<T> > VariantWith(
...
+//       []() { throw std::runtime_error("message"); },
+//       ThrowsMessage<std::runtime_error>(HasSubstr("message")));
+//
+//   EXPECT_THAT(
+//       []() { throw std::runtime_error("message"); },
+//       ThrowsMessageHasSubstr<std::runtime_error>("message"));
+//
+//   EXPECT_THAT(
+//       []() { throw std::runtime_error("message"); },
+//       Throws
```
vlovich, line 4815: This doesn't seem correct syntactically. Does this actually compile? Maybe include an instance of all of these functions in the tests?

AmatanHead:
> I do think that ThrowsMessageHasSubstr is a bit superfluous as the syntactic sugar benefit isn't that significant

Only if you have `using namespace testing` in your code =) Also, `ThrowsMessage("message")` will try and match message for equality, which is something you might not expect. Presence of `ThrowsMessageHasSubstr` makes this moment a bit clearer.

> On the other hand, for some reason I have approval permission?

I think GitHub allows anyone to leave a review, no matter if it's approval, need work or neutral. It doesn't allow merging PR though.

AmatanHead:
@CJ-Johnson can we please have someone from the project team to review this PR?

googlemock/include/gmock/gmock-matchers.h
```diff
@@ -4725,6 +4725,140 @@ PolymorphicMatcher<internal::variant_matcher::VariantMatcher<T> > VariantWith(
...
+// and MUST NOT BE USED IN USER CODE!!!
+namespace internal {
+
+template <typename Err>
+class ExceptionMatcherImpl {
+ public:
+  ExceptionMatcherImpl(Matcher<const Err&> matcher)
+      : matcher_(std::move(matcher)) {}
+
+ public:
```
CJ-Johnson, line 4741: Nit: duplicate public

googlemock/include/gmock/gmock-matchers.h
```diff
@@ -4725,6 +4725,140 @@ PolymorphicMatcher<internal::variant_matcher::VariantMatcher<T> > VariantWith(
...
+      (void)(std::forward<T>(x)());
+    } catch (const Err& err) {
+      *listener << "throws an exception of type " << GetTypeName<Err>();
+      if (matcher_.GetDescriber() != nullptr) {
+        *listener << " ";
+        return matcher_.MatchAndExplain(err, listener);
+      } else {
+        return true;
+      }
+    } catch (const std::exception& err) {
```
CJ-Johnson, line 4768: What happens when `Err` is of type `std::exception`? This will fail to compile, yes? (I think there should be a test for that as well, unless it is intentionally unsupported)

googlemock/include/gmock/gmock-matchers.h
```diff
@@ -4725,6 +4725,140 @@ PolymorphicMatcher<internal::variant_matcher::VariantMatcher<T> > VariantWith(
...
+#if GTEST_HAS_RTTI
+      *listener << "throws an exception of type "
+          << GetTypeName(typeid(err)) << " ";
+#else
+      *listener << "throws an std::exception-derived error ";
+#endif
+      *listener << "with description \"" << err.what() << "\"";
+      return false;
+    } catch (...) {
+      *listener << "throws an exception of some other type";
```
CJ-Johnson, line 4778: nit: I think "unknown" is a better word here than "other". What do you think?

googlemock/include/gmock/gmock-matchers.h
AmatanHead: `catch` clauses are matched from top to bottom, and the first one that matches the exception is called. If we have two catches with the same type, the later one will be ignored.

In this case (as opposed to [the same issue in macros](https://github.com/google/googletest/pull/2903#pullrequestreview-443969781)) it doesn't even trigger any warnings because we're in a template.

I've added a test for it.

AmatanHead: Hm. I was wrong about warning, MSVC does generate one. I'll use the same thing as with the macros then.

googlemock/include/gmock/gmock-matchers.h
```diff
@@ -4764,7 +4771,14 @@ class ExceptionMatcherImpl {
       } else {
         return true;
       }
-    } catch (const std::exception& err) {
+    } catch (
+        typename std::conditional<
```
CJ-Johnson, line 4775: Instead of defining this logic inline, could you define an internal template using declaration for this type? And then I would add comments to explain what it's doing.

Thanks!

AmatanHead: Done 👍

AmatanHead:
The build failed while trying to connect to a local Bazel server. I suppose this is a transient error that has nothing to do with the changes in this PR.

googlemock/include/gmock/gmock-matchers.h
```diff
@@ -4763,22 +4792,15 @@ class ExceptionMatcherImpl {
   bool MatchAndExplain(T&& x, MatchResultListener* listener) const {
     try {
       (void)(std::forward<T>(x)());
-    } catch (const Err& err) {
+    } catch (Err const& err) {
```
CJ-Johnson, line 4795: Nit: west cost

CJ-Johnson:
Are you able to re-run the CI?

AmatanHead:
@CJ-Johnson I'm not sure if there is a special command to restart CI so I've rebased the branch.

googlemock/test/gmock-matchers_test.cc
```diff
@@ -6950,6 +6950,188 @@ TEST_F(PredicateFormatterFromMatcherTest, DetectsFlakyShortCircuit) {
...
+  EXPECT_FALSE(
+      matcher.MatchAndExplain(
+          []() { (void)0; }, &listener));
+  EXPECT_THAT(
+      listener.str(),
+      testing::HasSubstr("does not throw any exception"));
+}
+
+INSTANTIATE_TEST_SUITE_P(AllMessagePredicates, ThrowsPredicateTest,
+    ::testing::Values(
```
CJ-Johnson, line 7123: Does this need to be fully qualified?

googlemock/include/gmock/gmock-matchers.h
```diff
@@ -4725,6 +4725,140 @@ PolymorphicMatcher<internal::variant_matcher::VariantMatcher<T> > VariantWith(
...
+namespace internal {
+
+template <typename Err>
+class ExceptionMatcherImpl {
+ public:
+  ExceptionMatcherImpl(Matcher<const Err&> matcher)
+      : matcher_(std::move(matcher)) {}
+
+ public:
+  void DescribeTo(::std::ostream* os) const {
```
CJ-Johnson, line 4777: Does this need to be fully qualified?

googlemock/include/gmock/gmock-matchers.h
```diff
@@ -4725,6 +4725,140 @@ PolymorphicMatcher<internal::variant_matcher::VariantMatcher<T> > VariantWith(
...
+    *os << "throws an exception of type " << GetTypeName<Err>();
+    if (matcher_.GetDescriber() != nullptr) {
+      *os << " which ";
+      matcher_.DescribeTo(os);
+    }
+  }
+
+  void DescribeNegationTo(::std::ostream* os) const {
```
CJ-Johnson, line 4785: Does this need to be fully qualified?

googlemock/include/gmock/gmock-matchers.h
```diff
@@ -4725,6 +4725,140 @@ PolymorphicMatcher<internal::variant_matcher::VariantMatcher<T> > VariantWith(
...
+
+template <typename Err>
+class ExceptionMatcherImpl {
+ public:
+  ExceptionMatcherImpl(Matcher<const Err&> matcher)
+      : matcher_(std::move(matcher)) {}
+
+ public:
+  void DescribeTo(::std::ostream* os) const {
+    *os << "throws an exception of type " << GetTypeName<Err>();
```
CJ-Johnson, line 4778: Since the type could be derived, I think this string should be `"throws an exception which is a "`

googlemock/include/gmock/gmock-matchers.h
```diff
@@ -4725,6 +4725,140 @@ PolymorphicMatcher<internal::variant_matcher::VariantMatcher<T> > VariantWith(
...
+ private:
+  const Matcher<const Err&> matcher_;
+};
+
+}  // namespace internal
+
+// Throws()
+// Throws(exceptionMatcher)
+// ThrowsMessage(messageMatcher)
+// ThrowsMessageHasSubstr(message)
```
CJ-Johnson, line 4829: Could `ThrowsMessageHasSubstr(x)` be spelled as `ThrowsMessage(HasSubstr(x))`? If so, do we need the former at all?

AmatanHead: Yes, but, as I've argued above, if you don't have `using namespace testing` in your code, it becomes `testing::ThrowsMessage(testing::HasSubstr(x))`, which is quite long. Also, `ThrowsMessage("message")` will try and match message for equality, which is something you might not expect. Presence of `ThrowsMessageHasSubstr` makes this moment a bit clearer.

CJ-Johnson: Composition is part of the benefit for matchers. We can't maintain the full matrix of combinations of matchers, so instead we break them into sub-components.

I'm not concerned by the length of the spelling.

CJ-Johnson: Just wanted to ping this thread. Are you able to make the change?

AmatanHead: @CJ-Johnson sorry for delay, it is fixed.

AmatanHead:
Rebased onto the current master.

CJ-Johnson:
Thank you for the PR, we have started internal review. Please don’t push any more changes into this PR as they might be overwritten. (325012153)

AmatanHead:
@CJ-Johnson, how's internal review going for this PR?

CJ-Johnson:
Ah yikes, 12 days already! I'm so sorry for the delay on that. It ended up not being "ready" yet so we've had to work on it a bit further (same API, modified implementation). I apologize for the delay since this came up in the middle of several ongoing tasks. I will do my best to get this done by the end of the week! Thanks for the reminder.

CJ-Johnson:
The change has been submitted internally. It will become available the next time we do an upstream push!

kuzkry:
Dear Googlers and @AmatanHead, do you think it's a good design to force users to bind function to its arguments? Because internally `ExceptionMatcherImpl` calls `(void)(std::forward<T>(x)());`, so `x` is expected to be callable with 0 arguments, therefore:
```c++
//  I can't do this (doesn't compile)
EXPECT_THAT(foo(1, 2, 3), ::testing::ThrowsMessage<std::runtime_error>("thrown from foo"));

// Instead I've gotta do that
EXPECT_THAT([]{ foo(1, 2, 3); }, ::testing::ThrowsMessage<std::runtime_error>("thrown from foo"));
```
What I mean here is that IMHO it doesn't compose nicely with `EXPECT_THROW` interface which doesn't require argument binding tricks and works like that:
```c++
EXPECT_THROW(foo(1, 2, 3), std::runtime_error);
```

AmatanHead:
@kuzkry `EXPECT_THAT` is not related to `EXPECT_THROW` in any way. In fact, `EXPECT_THAT` was originally a part of gmock, while `EXPECT_THROW` is native to gtest. It helps to think of it this way. `EXPECT_THAT(a, b)` means that we expect the value of `a` to satisfy condition `b`. Therefore `EXPECT_THAT(a, Throws<Error>())` means that the value of `a` throws an error when called. Then `EXPECT_THAT(foo(1, 2, 3), Throws<Error>())` would then mean that the value of `foo(1, 2, 3)` throws an error when called.

As for my reasoning for this implementation, it was the following:

- adding a new macro was discussed against in the [linked issue](https://github.com/google/googletest/issues/952#issuecomment-484135719);
- `EXPECT_THAT` is the only mechanism that'll allow us to implement this feature;
- we can't change the semantics of `EXPECT_THAT`;
- thus this is the only way we can implement this.

I'm not saying it is an ideal solution, but it is the best one I could come up with given the requirements.

adambadura:
Firstly, it seems the new `Throws*` matchers are not documented.

Secondly, do they work? I'm not able to make them work... I have the following code:
```c++
EXPECT_THAT(
    [&](){
        foo();
    },
    ::testing::ThrowsMessage<MyException>(
        ::testing::HasSubstr("message")
    )
);
```
however, while executing the test it fails:
```plain
Value of: [&](){ foo(); }
Expected: throws an exception which is a MyException which contains .what() that has substring "message"
  Actual: 8-byte object <F0-3C 27-01 00-00 00-00>, does not throw any exception
```

From other results, it seems to me the lambda in the `EXPECT_THAT` was not invoked. Instead, it was considered a value to be matched that didn't match.

I tried to call the lambda within the `EXPECT_THAT` but this resulted in a compilation error since the result of the lambda is `void` while `EXPECT_THAT` (unsurprisingly) expects some value.

doak:
@adambadura, did you include the gMock header (`gmock/gmock.h`)? I stumbled over the same, it was not obvious to me.

adambadura:
@doak, after more than half of a year I cannot be sure now. However, normally I do include `<gmock/gmock.h>` when using `EXPECT_THAT` or any matchers. So, I don't expect a missing include was the reason here.

doak:
Btw, it also works to include just `gmock/gmock-matchers.h`:
```
#include <gtest/gtest.h>
#include <gmock/gmock-matchers.h>

TEST(assertOnExpectedExceptionMessage, throwExceptionWithMessage_catchesExceptionWithCorrectMessage) {
    ASSERT_THAT(
            [&](){ throw std::runtime_error("bla"); },
            testing::ThrowsMessage<std::runtime_error>("bla"));
}

```

PR #2961, opened 2020-07-30
Removing tiny-dnn from "Who is using.."

FranciscoThiesen:
The tiny-dnn project is not using googletest anymore, for more details see (https://github.com/tiny-dnn/tiny-dnn/commit/d0d35ca2f4e75b3f647787e19ee7f31d0784b398).

Removing it from the "Who is Using Google Test?" list so as to keep it accurate.


FranciscoThiesen:
@googlebot I signed it!

PR #2994, opened 2020-08-25
Add timestamp to in old method mock macro guide

xerus2000:
I am working in a project with an older gMock version and was confused that the new way didn't work. Adding a hint when it was added can go a long way :)


xerus2000:
@googlebot I signed it!

Issue #2950, opened 2020-07-16
GOOGLETEST_VERSION is not defined in CMakeLists.txt


zhaoyang-star:
`GOOGLETEST_VERSION` is not defined in CMakeLists.txt. But it is used. So if I run command `mkdir mybuild && cd mybuild && cmake ..` the error happened:
```
CMake Warning at CMakeLists.txt:55 (project):
  VERSION keyword not followed by a value or was followed by a value that
  expanded to nothing.


CMake Error at CMakeLists.txt:130 (set_target_properties):
  set_target_properties called with incorrect number of arguments.


CMake Error at CMakeLists.txt:132 (set_target_properties):
  set_target_properties called with incorrect number of arguments.


-- Configuring incomplete, errors occurred!
```
So is there no `GOOGLETEST_VERSION` been defined in repo? I inserted one line `set(GOOGLETEST_VERSION 1.9.0)` and the error disappeared.

hyukmyeong:
Hello 

I'm not sure the exact context

But I think that the error looks like you omit project version not googletest version

If you want to check googletest version FindPkgConfig is a nice option
Because Current FindGTest in CMake is not enough for this

You can find how to use FindPkgConfig as below and if you once find it then you may know its version with variable GTest_VERSION
https://cmake.org/cmake/help/v3.0/module/FindPkgConfig.html

hyukmyeong:
I'm sorry for the above comment
GOOGLETEST_VERSION if defined in https://github.com/google/googletest/blob/master/CMakeLists.txt

and so I assume that you are build it in googletest/googletest not googletest

DinoStray:
always, I use XXX_VERSION like:
```
find_package(Libevent REQUIRED)
message(STATUS "Libevent version: ${Libevent_VERSION}")
find_package(ZeroMQ REQUIRED)
message(STATUS "ZeroMQ version: ${ZeroMQ_VERSION}")
```

But for gtest
```
find_package(GTest REQUIRED)
message(STATUS "GoogleTest version: ${GTest_VERSION}")
```
The GTest_VERSION is empty @hyukmyeong

hyukmyeong:
I'm not sure 
But how about checking if .pc file has the version information in the shell?

```
$ pkg-config gtest --modversion
1.9.0

$ cat local/lib/pkgconfig/gtest.pc
libdir=/usr/local/lib
includedir=/usr/local/include

Name: gtest
Description: GoogleTest (without main() function)
Version: 1.9.0
[...]

PR #3008, opened 2020-09-14
Improve the tutorial that may be confusing

hyukmyeong:
This patch updates the tutorial which can be confusing for users
This issue has been raised by the users from

https://github.com/google/googletest/issues/2950

I tried minimize the revision because I don't know exactly what the previous writer intended
and so almost all of the original contents have been preserved

### DIFF
--- a/googletest/README.md
+++ b/googletest/README.md
@@ -30,7 +30,8 @@ with
     cd build
     cmake ..             # Generate native build scripts for Google Test.
 
-If you want to build without Google Mock, you should replace the last command
+The above command also includes Google Mock by default. And so, if you want
+to build only Google Test, you should replace the last command
 with
 
     cmake .. -DBUILD_GMOCK=OFF
--- a/googletest/README.md
+++ b/googletest/README.md
@@ -22,21 +22,24 @@ another project.
 #### Standalone CMake Project
 
 When building Google Test as a standalone project, the typical workflow starts
-with:
+with
 
     git clone https://github.com/google/googletest.git -b release-1.10.0
-    cd googletest
+    cd googletest        # Main directory of the cloned repository.
     mkdir build          # Create a directory to hold the build output.
     cd build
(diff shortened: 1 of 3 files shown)


googletest/README.md
```diff
@@ -24,17 +24,20 @@ another project.
 When building Google Test as a standalone project, the typical workflow starts
 with:
 
-    mkdir mybuild       # Create a directory to hold the build output.
-    cd mybuild
-    cmake ${GTEST_DIR}  # Generate native build scripts.
+    git clone https://github.com/google/googletest.git -b release-1.10.0
```
romanwerpachowski, line 27: I would add here something like "Even if you want to build only one of the components (e.g. just `googletest`), you have to create your build directory in the main directory of the cloned repo. For people who are as thick as me :)

romanwerpachowski (approved):
Thanks!

googletest/README.md
hyukmyeong: I hope this additional revision would reflect what you commented
Please let me know if there are more parts that need to be corrected

romanwerpachowski (commented):
Thanks, that's a great improvement.

vslashg:
This looks good, but could you revert the changes adding a space between "GoogleTest" and "GoogleMock"?  I know we're inconsistent with this spelling in places, but we prefer to spell these as one word.

hyukmyeong:
@vslashg 

As I modified `README.md` in googletest directory, I couldn't avoid to modify other two `README.md`s for consistency.
(even though there are still some inconsistencies)
It is very cautious to revise the main title, but for now, I uploaded a patch for it.
Please let me know if there are some parts which can't be modified

Celthi:
This PR breaks a bunch of links such as the "gmock for dummies" in https://github.com/google/googletest/blob/master/googlemock/README.md

PR #3011, opened 2020-09-16
Move Unwrap helper to internal namespace

bgaifullin:
Allow to use this methods in custom matchers to avoid code duplication.


vslashg:
Thanks for the PR, but this feels more like a general metaprogramming facility, and not something that GoogleTest wants to provide.

ybgunjal2000:
Please unsubscribe me from your mailing list

PR #3012, opened 2020-09-16
Fix print protobuf message

bgaifullin:


vslashg:
We think this is reasonable, but because of differences in code between Google OSS projects and our internal forks of them, this change would break too many things on our end.

I will take a look at an alternate spelling for this, so that we can provide this feature in a compatible way.

bgaifullin:
I can check for both proto2::Message and google::protobuf::Message in trait `IsAProtocolMessage` to keep backward compatibility.

Zenix27 (approved):

vslashg:
The need for this CL is obviated by PR 2837.

PR #3022, opened 2020-09-23
Export Test - Do Not Merge

vslashg:
Export Test - Do Not Merge

Improve DoubleNearPredFormat output on bad epsilons

DoubleNearPredFormat will happily accept epsilon values (abs_error) that
are so small that they are meaningless. This turns EXPECT_NEAR into a
complicated and non-obvious version of EXPECT_EQ.

This change modifies DoubleNearPredFormat) so that when there is a
failure it calculates the smallest meaningful epsilon value, given the
input values, and then prints a message which explains what happened.

If a true equality test is wanted either pass a literal 0.0 as abs_error
or use EXPECT_EQ. If a check for being almost equal is wanted consider
using EXPECT_DOUBLE_EQ which, contrary to its name, verifies that the
two numbers are *almost* equal (within four ULPs).

With this change the flaky test mentioned in crbug.com/786046 gives this
output:

The difference between 4.2934311416234112e+18 and 4.2934311416234107e+18 is 512, where
4.2934311416234112e+18 evaluates to 4.2934311416234112e+18,
4.2934311416234107e+18 evaluates to 4.2934311416234107e+18.
The abs_error parameter 1.0 evaluates to 1 which is smaller than the minimum distance between doubles for numbers of this magnitude which is 512, thus making this EXPECT_NEAR check equivalent to EXPECT_EQUAL. Consider using EXPECT_DOUBLE_EQ instead.

Tested:
I confirmed that this change detects the bad epsilon value that caused
crbug.com/786046 in Chromium and added a test for the desired output.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,7 +80,7 @@ following notable projects:
 automated test-runner and Graphical User Interface with powerful features for
 Windows and Linux platforms.
 
-[Google Test UI](https://github.com/ospector/gtest-gbar) is test runner that
+[Google Test UI](https://github.com/ospector/gtest-gbar) is a test runner that
 runs your test binary, allows you to track its progress via a progress bar, and
 displays a list of test failures. Clicking on one shows failure text. Google
 Test UI is written in C#.
--- a/googlemock/docs/cook_book.md
+++ b/googlemock/docs/cook_book.md
@@ -1132,10 +1132,11 @@ Hamcrest project, which adds `assertThat()` to JUnit.
 
 ### Using Predicates as Matchers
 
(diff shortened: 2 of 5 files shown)


ybgunjal2000:
Please unsubscribe me from your mailing list

README.md
```diff
@@ -80,7 +80,7 @@ following notable projects:
 automated test-runner and Graphical User Interface with powerful features for
 Windows and Linux platforms.
 
-[Google Test UI](https://github.com/ospector/gtest-gbar) is test runner that
```
Maximuscari205, line 83: Let's test it and check it again

Maximuscari205 (approved):

PR #3024, opened 2020-09-24
Fix warning maybe-uninitialized

Thomas-Barbier-1A:


Maximuscari205:
Fixed

Thomas-Barbier-1A:
@Maximuscari205 which commit fix it? I still have the issue with Head of master with gcc (GCC) 11.0.0 20200803
I think this PR is necessary to fix the issue.

fthdrmzzz:
I agree with this change. Faced the same error.

PR #3027, opened 2020-09-27
Update cook_book.md

keshavgbpecdelhi:
"Please note that changing the default value for a type can make [you]* test....."
Here "you" word doesn't make sense.. rather "your" seems to make it understandable.


CJ-Johnson:
Thank you for the pull request. It has now been imported. Please refrain from making further changes, as they will be overwritten in the next upstream push. (334382611)

PR #3038, opened 2020-10-02
Refactor gtest-filepath

manavrion:
- Simplified FilePath constructors
- FilePath is MoveOnly class now
- FilePath::Set -> move assignment operator

### DIFF
--- a/googletest/include/gtest/internal/gtest-filepath.h
+++ b/googletest/include/gtest/internal/gtest-filepath.h
@@ -61,16 +61,12 @@ namespace internal {
 
 class GTEST_API_ FilePath {
  public:
-  FilePath() : pathname_("") { }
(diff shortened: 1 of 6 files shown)


googletest/include/gtest/internal/gtest-filepath.h
```diff
@@ -173,6 +165,9 @@ class GTEST_API_ FilePath {
   bool IsAbsolutePath() const;
 
  private:
+  FilePath(const FilePath& rhs) = default;
```
asoffer, line 168: A few questions here:
1. Why not allow users to copy FilePath? is it incorrect or we just believe it's inefficient?
2. Does the implementation need to make copies? If not, can we move these to the public section and make them `= delete`?

asoffer:
With no response for 3 months and failing builds, I am closing this PR. Please reopen if you wish to continue pursuing this.

PR #3044, opened 2020-10-07
Improve FilePath::Normalize method

manavrion:
The old algorithm allocates memory for normalization but we can do it in-place


CJ-Johnson:
Thank you for providing this PR! It has been imported for internal testing. Please do not make further changes as they may be overwritten. (339096163)

Issue #2809, opened 2020-04-12
--color_output with MinGW from MSYS2 adds escaping sequences


GbaLog:
Hello!  
I have installed GCC 9.2.0 from MSYS2 and I have compiled it with my tests.
When I run googletest with `--gtest_color=yes` flag I have output with escape sequences like `[0;32m`.  
Although on Windows, I think the `SetConsoleTextAttribute` function should be used.  
I did a little research and found that there is a check in the code that `GTEST_OS_WINDOWS_MINGW` wasn't detected.  
If it's not detected, works with Windows API functions, but if it's detected, it will try to use escape sequences.
If I remove that check, it works fine, and I can see the color output.  
My question is: why is this done? I don't think it's wrong, but I don't see the point.  
Please explain to me why and if possible explain how to fix it.  
  
`gcc -v` output if needed:  
```bash
Using built-in specs.
COLLECT_GCC=C:\msys64\mingw64\bin\gcc.exe
COLLECT_LTO_WRAPPER=C:/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/9.2.0/lto-wrapper.exe
Target: x86_64-w64-mingw32
[...]

Issue #24, opened 2015-07-28
needs support for parameterized tests


GoogleCodeExporter:
```
Google Test should support tests that are parameterized by:

- a set of values,
- a set of implementations of the same interface (in the inheritance
sense), and
- a set of types/values for instantiating some templates.

The above goals may require different constructs.



```

Original issue reported on code.google.com by `shiq...@gmail.com` on 6 Aug 2008 at 12:48

PR #3049, opened 2020-10-11
Fix color output with MinGW build under Console

tanzislam:
When built with MinGW, Google Test emits ANSI escape sequences for color output like it does on non-Windows platforms. This works with MSYS2's terminal ([`mintty.exe`](https://mintty.github.io/)) that intercepts stdout with a pipe and then renders color itself, but not with the native Windows Console (i.e. [`conhost.exe` in Windows 7 onwards, and `csrss.exe` earlier](https://devblogs.microsoft.com/commandline/windows-command-line-inside-the-windows-console/#so-whats-inside-the-windows-console)) that expects the Windows Console API to be used. This is arguably incorrect as the MinGW toolchain is intended for building native Windows programs, independent of any POSIX-like environment.

Change the color-output behavior of MinGW-built Google Test to be conditional:

 - If stdout is a TTY, use the Windows Console API. (This happens with Windows Console and Windows Terminal, as well as most 3rd-party consoles that screen-scrape the Windows Console.)
 - If stdout is not a TTY but the user specified `--gtest_color=yes`, use ANSI escape sequences. (This happens with `mintty.exe` and other 3rd-party consoles that directly read our stdout with a pipe, e.g. Travis CI's Windows logger.)

Behavior when built with any other compiler is unaffected. So Windows desktop toolchains (MSVC, C++Builder, etc.) will continue to use the Windows Console API unconditionally, while Cygwin, MSYS (not MinGW) and all other non-Windows toolchains will continue to use ANSI escape sequences unconditionally.

Fixes: #2809

### DIFF
--- a/googletest/src/gtest.cc
+++ b/googletest/src/gtest.cc
@@ -3167,8 +3167,7 @@ static void PrintTestPartResult(const TestPartResult& test_part_result) {
 
 // class PrettyUnitTestResultPrinter
 #if defined(GTEST_OS_WINDOWS) && !defined(GTEST_OS_WINDOWS_MOBILE) &&    \
-    !defined(GTEST_OS_WINDOWS_PHONE) && !defined(GTEST_OS_WINDOWS_RT) && \
-    !defined(GTEST_OS_WINDOWS_MINGW)
+    !defined(GTEST_OS_WINDOWS_PHONE) && !defined(GTEST_OS_WINDOWS_RT)
 
 // Returns the character attribute for the given color.
 static WORD GetColorAttribute(GTestColor color) {
@@ -3215,7 +3214,7 @@ static WORD GetNewColor(GTestColor color, WORD old_color_attrs) {
   return new_color;
 }
 
-#else
+#endif  // GTEST_OS_WINDOWS && !GTEST_OS_WINDOWS_MOBILE
 
 // Returns the ANSI color code for the given color. GTestColor::kDefault is
 // an invalid input.
@@ -3233,7 +3232,13 @@ static const char* GetAnsiColorCode(GTestColor color) {
   }
 }
 
-#endif  // GTEST_OS_WINDOWS && !GTEST_OS_WINDOWS_MOBILE
+static void EmitAnsiColorOutput(GTestColor color,
+                                const char* fmt,
+                                va_list args) {
+  printf("\033[0;3%sm", GetAnsiColorCode(color));
+  vprintf(fmt, args);
+  printf("\033[m");  // Resets the terminal to default.
+}
 
 // Returns true if and only if Google Test should use colors in the output.
 bool ShouldUseColor(bool stdout_is_tty) {
@@ -3241,7 +3246,7 @@ bool ShouldUseColor(bool stdout_is_tty) {
   const char* const gtest_color = c.c_str();
 
   if (String::CaseInsensitiveCStringEquals(gtest_color, "auto")) {
-#if defined(GTEST_OS_WINDOWS) && !defined(GTEST_OS_WINDOWS_MINGW)
+#if defined(GTEST_OS_WINDOWS)
     // On Windows the TERM variable is usually not set, but the
     // console there does support colors.
     return stdout_is_tty;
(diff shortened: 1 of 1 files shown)


tanzislam:
One of the AppVeyor jobs failed due to a Bazel issue, so retrying the PR build.

googletest/src/gtest.cc
```diff
@@ -3287,8 +3287,16 @@ static void ColoredPrintf(GTestColor color, const char* fmt, ...) {
     return;
   }
 
+#define GTEST_EMIT_ANSI_COLOR_OUTPUT              \
```
asoffer, line 3290: Can we make this a function instead of a macro?

googletest/src/gtest.cc
```diff
@@ -3287,8 +3287,16 @@ static void ColoredPrintf(GTestColor color, const char* fmt, ...) {
...
+  printf("\033[0;3%sm", GetAnsiColorCode(color)); \
+  vprintf(fmt, args);                             \
+  printf("\033[m");  // Resets the terminal to default.
+
 #if GTEST_OS_WINDOWS && !GTEST_OS_WINDOWS_MOBILE && \
-    !GTEST_OS_WINDOWS_PHONE && !GTEST_OS_WINDOWS_RT && !GTEST_OS_WINDOWS_MINGW
+    !GTEST_OS_WINDOWS_PHONE && !GTEST_OS_WINDOWS_RT
+# if GTEST_OS_WINDOWS_MINGW
```
asoffer, line 3297: Having a macro conditionally hide conditional control-flow is a bit tricky to read. Instead, would we #define a constant to either true or false and then use `if (using_mingw && stdout_is_tty)`
Or something similar?

tanzislam: Will do. It will be:

```cpp
if (!using_mingw || stdout_is_tty) {
```
as non-MinGW desktop Windows needs to do this unconditionally.

Switching away from the macro will also mean non-MinGW desktop Windows compiling `GTEST_EMIT_ANSI_COLOR_OUTPUT` and `GetAnsiColorCode()` but never using them.

googletest/src/gtest.cc
tanzislam: Yes. I was initially unsure of whether a `va_list` object (i.e. the `args` here) can be passed into a function. Turns out it can, as long as:

 - it is no longer used in the calling function afterwards
 - the `va_end()` stays in the calling function, matching the `va_start()`

tanzislam (commented):
Making requested changes ...

tanzislam:
The timeout failure in Test #24 (`googletest-param-test-test`) on `msvc-14-seh` in AppVeyor seems to have been introduced by 2cf1f99b9. Other PRs are also affected by it.

tanzislam:
Re-running CI with (hopefully) latest fixes to AppVeyor `msvc-14-seh`.

tanzislam:
Hi, is this good to merge? Do I need to make any further changes?

skywind3000:
I am facing same problem, don't know why it is still unmerged ??

tanzislam:
I've rebased this on the current `main` branch, but the Windows CI is already broken there with the same error message.

googletest/src/gtest.cc
```diff
@@ -3156,7 +3156,7 @@ static void PrintTestPartResult(const TestPartResult& test_part_result) {
 
 // class PrettyUnitTestResultPrinter
 #if GTEST_OS_WINDOWS && !GTEST_OS_WINDOWS_MOBILE && !GTEST_OS_WINDOWS_PHONE && \
```
Fangjun-Zhou, line 3158: It seems that google test is using stuff like
```cpp
#if defined(GTEST_OS_WINDOWS)
```

image

Is there a big difference between the old `#if GTEST_OS_WINDOWS` here?

Fangjun-Zhou (commented):
Question about `#if define` macro.

tanzislam:
Raised #4365 to address the latest CI failure.

googletest/src/gtest.cc
```diff
@@ -3281,11 +3286,12 @@ static void ColoredPrintf(GTestColor color, const char* fmt, ...) {
...
 
-  static const bool in_color_mode =
 #if GTEST_HAS_FILE_SYSTEM
-      ShouldUseColor(posix::IsATTY(posix::FileNo(stdout)) != 0);
+  static const bool stdout_is_tty = posix::IsATTY(posix::FileNo(stdout)) != 0;
+  static const bool in_color_mode = ShouldUseColor(stdout_is_tty);
 #else
-      false;
+  static const bool stdout_is_tty = false;
+  static const bool in_color_mode = false;
```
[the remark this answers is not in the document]
tanzislam: That is an existing problem with the `main` branch, independent of what I'm trying to fix here.

Issue #3063, opened 2020-10-19
Fuse-Source Script does not work after python 3


ReneOschmann:
Hi,

when I try to run the fuse_gmock_files.py after https://github.com/google/googletest/commit/72512aa89328f21fd18b636032c7a51d8d4ee068
I receive errors.

I'll make a pull request.

Kind regards
René

ReneOschmann:
Fixed with 252ee42

PR #3064, opened 2020-10-19
Fix fuse_gmock_files.py

ReneOschmann:
Issue: #3063

After upgrade to python3 (72512aa89328f21fd18b636032c7a51d8d4ee068)
the script fails.

Module fuse_gtest_files can only be imported, after the path is added to
sys.path.

The processed_files cannot be frozen sets as add is called in
ProcessFile.

### DIFF
--- a/googlemock/scripts/fuse_gmock_files.py
+++ b/googlemock/scripts/fuse_gmock_files.py
@@ -66,8 +66,6 @@ import os
 import re
 import sys
 
-import fuse_gtest_files
-
 __author__ = 'wan@google.com (Zhanyong Wan)'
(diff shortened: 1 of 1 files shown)


derekmauro:
Thanks for the PR. We already got a fix from an internal contributor for this (252ee42d3bb96e6d248063f622ece41ea5dce999).

ReneOschmann:
Hi Derek,

thanks for the notice. Could you please notify me through #3063 when it is available?

Kind regards
René

ReneOschmann:
NVM, I saw it is already in.

PR #3076, opened 2020-10-21
Added support for stack trace printing using libunwind on Linux

emmenlau:
This is a draft PR to get feedback on some recent work we have been doing. We use Google test a lot, its really great. But there is one problem that is a large bottleneck in our work: when implementing test-methods that get called for various inputs, its often hard or impossible to see for which inputs a test failed. We used various workarounds in the past, like extensive logging or custom stack tracers. But these workarounds just alleviate parts of the pain.

This PR tries to aim at a full solution, in the sense that Google test can print stack traces with file and line references for failed assertions.

In our tests (currently Linux only) this is a huge step forward and makes virtually impossible workflows very accessible and simple.

As a side-effect, the printed stack traces work in common IDEs like Visual Studio, Visual Studio Code and possibly many others.

Feedback would be highly appreciated!

### DIFF
--- a/googletest/CMakeLists.txt
+++ b/googletest/CMakeLists.txt
@@ -33,6 +33,9 @@ if (COMMAND pre_project_set_up_hermetic_build)
   pre_project_set_up_hermetic_build()
 endif()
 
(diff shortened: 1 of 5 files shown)


emmenlau:
Anyone?

Issue #3083, opened 2020-10-26
Describe how to trigger test failures when there are sanitizer errors


sfc-gh-ebrossard:
It took a bit of digging, but I found the magic hooks for UBSAN/ASAN/TSAN errors and was successfully able to instrument them to trigger individual test failures, as opposed just to crashing at the end of executing all tests. It would be nice to have this as part of the googletest documentation in case it's useful to other projects as well:

```
extern "C" {
void __ubsan_on_report() {
  FAIL() << "Encountered an undefined behavior sanitizer error";
}
void __asan_on_error() {
  FAIL() << "Encountered an address sanitizer error";
}
void __tsan_on_report() {
  FAIL() << "Encountered a thread sanitizer error";
}
}  // extern "C"
```

asoffer:
This looks really useful, and something you're now familiar with. None of us on the GoogleTest team have done this before. Would you mind sending a PR? Thanks!

sfc-gh-ebrossard:
Sure, created https://github.com/google/googletest/pull/3086.

axxel:
Found this in the official doc, very useful info. Thanks. Question: any chance/interest in having some kind of out-of-the-box support for this from inside `gtest/gtest.h`? I have whole bunch of individual cpp files with tests and would have to 'manually' provide those functions to all of them if I'm not mistaken.

PR #3086, opened 2020-10-27
Add instructions for sanitizer integration

sfc-gh-ebrossard:
See https://github.com/google/googletest/issues/3083 for context. This PR adds specific instructions for how to trigger test failures when sanitizers including UBSan, ASan, and TSan report errors.


sfc-gh-ebrossard:
I should mention that none of these hooks is particularly well-documented. I found the ASAN one via https://github.com/llvm-mirror/compiler-rt/blob/master/include/sanitizer/asan_interface.h#L263, but the others don't appear in the related headers, so I had to check e.g. `nm -C --defined-only libubsan.a | grep __ubsan_on`.

sfc-gh-ebrossard:
@asoffer Any feedback? It looks like I can't actually assign reviewers to the PR. Thanks!

asoffer:
Looks good to me. Pulling this in internally now.

Issue #3089, opened 2020-10-29
Only save original working directory if death tests are enabled


knutpett:
On a diskless system you cannot get the current directory. So if death tests are disabled anyway, there is no point trying to get current directory.

It will fail before main with: 
"Condition !original_working_dir_.IsEmpty() failed. Failed to get the current working directory."

asoffer:
Closing this issue. The referenced PR tracks this issue.

PR #3090, opened 2020-10-29
Only save original working directory if death tests are enabled

knutpett:
On a diskless system you cannot get the current directory. So if
death tests are disabled anyway, there is no point trying to
get current directory.

Without this fix, running tests on diskless systems will fail,
even when death tests are disabled.

Solves https://github.com/google/googletest/issues/3089

### DIFF
--- a/googletest/src/gtest-internal-inl.h
+++ b/googletest/src/gtest-internal-inl.h
@@ -682,6 +682,7 @@ class GTEST_API_ UnitTestImpl {
   void AddTestInfo(internal::SetUpTestSuiteFunc set_up_tc,
                    internal::TearDownTestSuiteFunc tear_down_tc,
                    TestInfo* test_info) {
+#if GTEST_HAS_DEATH_TEST
     // In order to support thread-safe death tests, we need to
     // remember the original working directory when the test program
     // was first invoked.  We cannot do this in RUN_ALL_TESTS(), as
@@ -694,6 +695,7 @@ class GTEST_API_ UnitTestImpl {
       GTEST_CHECK_(!original_working_dir_.IsEmpty())
           << "Failed to get the current working directory.";
     }
+#endif
 
     GetTestSuite(test_info->test_suite_name(), test_info->type_param(),
                  set_up_tc, tear_down_tc)


knutpett:
Seems like the appveyor build timed out on windows. How can I rerun?

kuzkry:
Hi @knutpett, the easiest way is to pretend you touched your commit by executing:
1) `git checkout working_dir_on_diskless`
2) `git commit --amend` and immediately close an editor
3) `git push --force ...` onto the same branch.

knutpett:
I keep getting "24 - googletest-param-test-test (Timeout)
ctest.exe : Errors while running CTest" from appveyor. Is introduced by me, or is this a known problem with this test timing out?

I've touched the PR twice, but I keeps on failing....

CJ-Johnson:
My apologies for the failing tests. It's a known issue and we've had trouble resolving it. For now, all that is required is the CLA and Travis.

Would you like us to import this PR? We can do additional internal testing, but importing it implies that the PR cannot be updated further.

knutpett:
@CJ-Johnson Yes, I'm happy with you importing this PR :)

CJ-Johnson:
Thank you for this PR! It has now been imported for internal testing. Please do not make additional changes as they will be overwritten. (341387352)

knutpett:
Thanks! What happens now? Does this mean that the change likely will be merged? How do I know when/if it is done? Should I close the PR?

CJ-Johnson:
Yes, if the internal testing passes, it will be merged and closed automatically. Please do not close it. :)

puetzk:
This caused a regression for me in finally updating our patches for winegcc (not submitted yet, trying to upstream the winegcc compiler improvements that enable it into winehq first).

UnitTest::GetInstance()->original_working_dir() is documented public API:
https://github.com/google/googletest/blob/main/docs/reference/testing.md#original_working_dir-unittestoriginal_working_dir and is used for things like UnitTestOptions::GetAbsolutePathToOutputFile. So stubbing it to just return an empty string breaks many other unit tests unnecessarily if you have a filesystem but not death tests

If there's a need to do something special for diskless systems, seems like they should have a distinct check.

gaspardpetit:
I ran in to the same regression, I will propose a fix in the upcoming days

gaspardpetit:
I undid this change in https://github.com/google/googletest/pull/3862
The original problem seems to be that on a system without a filesystem, `FilePath::GetCurrentDir()` would get called and cause `getcwd` to be called. 

https://github.com/google/googletest/pull/3862 moves this logic to `gtest-port.h` where a custom `GetCwd` function can be implemented. On a diskless system, it would be possible to simply return empty string or `./`.

Thus, I think  https://github.com/google/googletest/pull/3862 solves the original problem raised here without causing the side effects observed when `GTEST_HAS_DEATH_TEST` is set to `0`

PR #3094, opened 2020-10-30
Bump CMake minimum to 2.8.12

### LINKED ISSUE #3040: CMake 3.19.0 throws deprecation warning in googletest/googletest/CMakeLists.txt:56
```
CMake Deprecation Warning at googletest/googletest/CMakeLists.txt:56 (cmake_minimum_required):
  Compatibility with CMake < 2.8.12 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value or use a ...<max> suffix to tell
  CMake that the project does not need compatibility with older versions.
```

chuckatkins:
Fixes #3040

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,7 @@
 # Note: CMake support is community-based. The maintainers do not use CMake
 # internally.
 
-cmake_minimum_required(VERSION 2.8.8)
+cmake_minimum_required(VERSION 2.8.12)
 
 if (POLICY CMP0048)
   cmake_policy(SET CMP0048 NEW)
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -89,7 +89,7 @@ and their own tests from a git checkout, which has further requirements:
 
 *   [Python](https://www.python.org/) v2.3 or newer (for running some of the
     tests and re-generating certain source files from templates)
-*   [CMake](https://cmake.org/) v2.6.4 or newer
+*   [CMake](https://cmake.org/) v2.8.12 or newer
 
 ## Developing Google Test and Google Mock
 
--- a/googlemock/CMakeLists.txt
+++ b/googlemock/CMakeLists.txt
@@ -42,7 +42,7 @@ else()
   cmake_policy(SET CMP0048 NEW)
   project(gmock VERSION ${GOOGLETEST_VERSION} LANGUAGES CXX C)
 endif()
-cmake_minimum_required(VERSION 2.6.4)
+cmake_minimum_required(VERSION 2.8.12)
 
 if (COMMAND set_up_hermetic_build)
   set_up_hermetic_build()
--- a/googletest/CMakeLists.txt
+++ b/googletest/CMakeLists.txt
@@ -53,7 +53,7 @@ else()
   cmake_policy(SET CMP0048 NEW)
   project(gtest VERSION ${GOOGLETEST_VERSION} LANGUAGES CXX C)
 endif()
-cmake_minimum_required(VERSION 2.6.4)
+cmake_minimum_required(VERSION 2.8.12)
 
 if (POLICY CMP0063) # Visibility
   cmake_policy(SET CMP0063 NEW)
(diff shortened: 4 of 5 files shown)


chuckatkins:
From what I can tell the AppVeyor failure is unrelated.

atherkhan (approved):

acxz:
It has been a month since this PR has been submitted. @derekmauro can you take a quick look at it?

Robadob (approved):
Would appreciate this being merged, our CI has just automatically updated CMake version and hit failure due to this.

ybgunjal2000:
unsubscribe

CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 # Note: CMake support is community-based. The maintainers do not use CMake
 # internally.
 
-cmake_minimum_required(VERSION 2.8.8)
+cmake_minimum_required(VERSION 2.8.12)
```
hjmjohnson, line 4:
```cmake
# Adding ... to version rage to future proof code when using cmake > 3.18 that explicitly 
# deprecated 2.8.12 features
cmake_minimum_required(VERSION 2.8.12...)
```

MBNwaiser (changes requested):
Tell the try

CMakeLists.txt
hjmjohnson: https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html

Be-ing:
Ping. Lots of projects are stumbling on this bug on CI.

CMakeLists.txt
friendlyanon: Version range was added in 3.12 to this command and the goal is to support 2.8.12 as a minimum.

Also, empty max range?

remyjette: From the linked docs:

> The optional <max> version, if specified, must be at least the <min> version and affects policy settings as described below. If the running version of CMake is older than 3.12, the extra ... dots will be seen as version component separators, resulting in the ...<max> part being ignored and preserving the pre-3.12 behavior of basing policies on <min>.

MichaelRWolf:
Over 2 months ago, the 1 failing check contains 2 failed tests due to timeout.

Given the number of projects that are stumbling due to this, could someone please rerun the checks, and if the timeout clears, please merge this PR?  !!!

Thanks!

MichaelRWolf (commented):
This _minor_ change looks good to me.  It will certainly quiet many deprecation warnings that have been clogging many screens for over 2 months.  Anyone up for closing this by Christmas?

Holzhaus:
Whohoo!

mojca:
Would you be willing to cherry-pick this change into the release branch (v1.10.x)? Thanks in advance.

friendlyanon:
README says:
> Post 1.10.x googletest will follow [Abseil Live at Head philosophy](https://abseil.io/about/philosophy)

So I don't think there will be any more releases.  
My solution is to just fix the download at the commit this was merged.

Be-ing:
> Post 1.10.x googletest will follow Abseil Live at Head philosophy

Considering this critical bug fix took 3 months after reporting it, took an outside contributor to fix, and wasn't merged for another 2 months, all this despite many users speaking up, this doesn't inspire sufficient confidence to me for a "Live at Head" policy.

friendlyanon:
It's to be [expected](https://killedbygoogle.com/) things end up this way with Google.

I can recommend [doctest](https://github.com/onqtam/doctest) as an alternative.

Be-ing:
Thanks for the perspective. Just because a large company puts resources into a project does not mean it is well maintained or will be in the future...

mojca:
I just started testing doctest a few days ago and it seems awesome indeed. The author fixed a bug I reported in less than a day despite not having access to the compiler in question.

The only thing I'm missing a bit at the moment is a proper test adapter for VS (but it works well in VS Code).

lesomnus:
doctesg is good. but google/benchmark uses googletest.

LukeAI:
Is there any suggestion as to when we might see this in a release?

patrick96:
@LukeAI gtest doesn't have releases anymore. It's "live at head": https://github.com/google/googletest#live-at-head

PR #3114, opened 2020-11-11
Refactor finding python

marbre:
Use `find_package(Python ..)` if available, instead of the deprecated
`find_package(PythonInterp)` CMake module.

### DIFF
--- a/googletest/cmake/internal_utils.cmake
+++ b/googletest/cmake/internal_utils.cmake
@@ -244,7 +244,13 @@ function(cxx_executable name dir libs)
 endfunction()
 
 # Sets PYTHONINTERP_FOUND and PYTHON_EXECUTABLE.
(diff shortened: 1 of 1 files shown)


marbre:
@ScottTodd: FYI. This might relax the problems we experience within IREE.

marbre:
Another or rather an additional option that relaxes problems for projects embedding googletest would be to disable finding python if `gtest_build_tests` is to `OFF`. The `py_test` function is only called if `gtest_build_tests` is true.

abdulbade1991 (approved):

CJ-Johnson:
Thank you for this PR! It has now been imported for internal testing. Please do not modify it further as the changes will be overwritten. (342837512)

marbre:
Thanks for merging!

PR #3132, opened 2020-11-27
Add documentation for GTEST_SKIP()

arw2019:
xref https://github.com/google/googletest/pull/1544#issuecomment-431197802

#1544 added `GTEST_SKIP`. This PR adds an update to the documentation and a code sample.

### DIFF
--- a/googletest/docs/primer.md
+++ b/googletest/docs/primer.md
@@ -256,6 +256,7 @@ To create a test:
 3.  The test's result is determined by the assertions; if any assertion in the
     test fails (either fatally or non-fatally), or if the test crashes, the
     entire test fails. Otherwise, it succeeds.
+4.  To skip a test based on a runtime condition use the `GTEST_SKIP()` macro.
 
 ```c++
 TEST(TestSuiteName, TestName) {


sebkraemer:
I'd say leave out the second commit, there is a unit test already in `googletest/test/gtest_skip_test.cc`. This should also fix the build error. I created PR #3174 just before seeing this one here, it contains short sample code instead of a sample file.

I'd still appreciate the `GTEST_SKIP` mention in `primer.md` since my change is a section in `advanced.md` so I'm all for continuing this PR. If you don't want to do it but cla-agree, I could also cherry-pick commit 047b570 in my PR.

arw2019:
@googlebot I signed it!

arw2019:
> I'd say leave out the second commit, there is a unit test already in `googletest/test/gtest_skip_test.cc`. This should also fix the build error. I created PR #3174 just before seeing this one here, it contains short sample code instead of a sample file.
> 
> I'd still appreciate the `GTEST_SKIP` mention in `primer.md` since my change is a section in `advanced.md` so I'm all for continuing this PR. If you don't want to do it but cla-agree, I could also cherry-pick commit [047b570](https://github.com/google/googletest/commit/047b570ea87741835631777dd1fb35ef0e5e94c5) in my PR.

Sounds good to me! I've I removed the sample file.

derekmauro:
I think we prefer something like #3174. I personally don't think this belongs in the "simple tests" section at all.

Issue #2429, opened 2019-08-24
Add CMake aliases


lukasm91:
GTest exports the following targets:
- `GTest::gtest`
- `GTest::gtest_main`
- `GTest::gmock`
- `GTest::gmock_main`

This targets should also be available when adding gtest with `add_subdirectory` (or `FetchContent`), because this should behave the same way as adding GTest with `find_package`. So somewhere, we should add the aliases to these targets, i.e.
```
add_library(GTest::gtest ALIAS gtest)
add_library(GTest::gtest_main ALIAS gtest_main)
add_library(GTest::gmock ALIAS gmock)
add_library(GTest::gmock_main ALIAS gmock_main)
```
I am not sure, where I should add these aliases.

gennadiycivil:
Anything related to CMake is community supported. We do not use it internally and we are not in position to check. In general the maintainers would look for community consensus.

Mizux:
My 2 cents,
for *gtest* and *gtest_main*:
First this CMake install **NAMESPACE** is defined here:
https://github.com/google/googletest/blob/306f3754a71d6d1ac644681d3544d06744914228/googletest/CMakeLists.txt#L98
then used here:
https://github.com/google/googletest/blob/306f3754a71d6d1ac644681d3544d06744914228/googletest/CMakeLists.txt#L105

libraries are defined later using a custom function cxx_library:
https://github.com/google/googletest/blob/306f3754a71d6d1ac644681d3544d06744914228/googletest/CMakeLists.txt#L128-L129

note: gmock install rules seems to use an other macro/function:
https://github.com/google/googletest/blob/306f3754a71d6d1ac644681d3544d06744914228/googlemock/CMakeLists.txt#L121
defined  here (EDIT: just adding targets to the googletest target): https://github.com/google/googletest/blob/306f3754a71d6d1ac644681d3544d06744914228/googletest/cmake/internal_utils.cmake#L327
(ed notice the of variable *${targets_export_name}*)

so we should try to add the `add_library(... ALIAS ...)` to the function:
https://github.com/google/googletest/blob/306f3754a71d6d1ac644681d3544d06744914228/googletest/cmake/internal_utils.cmake#L147-L150
called by:
https://github.com/google/googletest/blob/306f3754a71d6d1ac644681d3544d06744914228/googletest/cmake/internal_utils.cmake#L197-L203

using the variable *${cmake_package_name}* which should be defined somewhere else IMHO...

piotrzarycki:
[...]

PR #3155, opened 2020-12-04
CMake: Add namespaced ALIAS library

### LINKED ISSUE #3125: Provide ALIAS targets to unify usage
If `googletest` is made part of the dependent's build tree (via `FetchContent`), then usage is different compared to using it from an install location (via `find_package`).

Ever since `FetchContent` has been introduced, the example has been using `googletest`. See [here](https://cmake.org/cmake/help/v3.11/module/FetchContent.html).

Usage differences in code:
```cmake
# make googletest part of hte build tree
include(FetchContent)
FetchContent_Declare(googletest
        GIT_REPOSITORY https://github.com/google/googletest.git
        GIT_TAG master)
FetchContent_MakeAvailable(googletest)

# no ALIAS target, must use non-namespaced target instead
target_link_libraries(my_target PRIVATE gtest_main)

# ----

# use googletest from an install location
find_package(GTest REQUIRED CONFIG)

# the install(EXPORT) command namespaces the targets
target_link_libraries(my_target PRIVATE GTest::gtest_main)
```

I see from another issue (https://github.com/google/googletest/issues/3040#issuecomment-717300633), that due to CMake 3.19 deprecating `<2.8.12` there are also plans to set that as the minimum version and conveniently 2.8.12 is also the version that introduced the `ALIAS` form for `add_library`.

ZedThree:
Fixes #3125

I've assumed that #3094 will be in shortly, and so there's no need to guard this like in other places, for example: https://github.com/google/googletest/blob/93748a946684defd1494d5585dbc912e451e83f8/googletest/cmake/internal_utils.cmake#L178

I've also had to make sure `cmake_package_name` is always set so that the namespace is consistent between exported targets and the alias libraries.

### DIFF
--- a/googletest/CMakeLists.txt
+++ b/googletest/CMakeLists.txt
@@ -92,10 +92,13 @@ include(cmake/internal_utils.cmake)
 
 config_compiler_and_linker()  # Defined in internal_utils.cmake.
 
+# Needed to set the namespace for both the export targets and the
+# alias libraries
+set(cmake_package_name GTest CACHE INTERNAL "")
+
 # Create the CMake package file descriptors.
 if (INSTALL_GTEST)
   include(CMakePackageConfigHelpers)
-  set(cmake_package_name GTest)
   set(targets_export_name ${cmake_package_name}Targets CACHE INTERNAL "")
   set(generated_dir "${CMAKE_CURRENT_BINARY_DIR}/generated" CACHE INTERNAL "")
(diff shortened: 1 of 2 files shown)


Bell-2319 (approved):

phcerdan:
Looks good!

Mizux:
should also close https://github.com/google/googletest/issues/2429#issuecomment-572767510

PR #3156, opened 2020-12-04
googletest-output-test: fix filtering out test counts

sthibaul:
3a460a26b7a9 ("TestCase->TestSuite refactoring") renamed "test cases" to
"test suites" but missed updating the RemoveTestCounts() that filters
them out to cope with varying operating system support.


sthibaul:
@googlebot I signed it!

PR #3173, opened 2020-12-15
Fix gmock_gen.py generation of template type names

### LINKED ISSUE #3172: gmock_gen.py generates template types with a wrong name
The otherwise very useful `gmock_gen.py` script does not produce _correct_ mocks when used with interfaces that contain template types that are used in the signature of one or more functions.

For example, when trying to generate a mock out of the following header:
```cpp
#pragma once

template <typename T>
class FooBar {
 public:
  virtual void bar(T t);
};
```

The script's output is:
```cpp
template <typename T0>
class MockFooBar : public FooBar<T0> {
 public:
  MOCK_METHOD(void, bar, (T t), (override));
};
```

The expected output would be something that compiles, e.g.:
```cpp
```cpp
template <typename T>
class MockFooBar : public FooBar<T> {
 public:
  MOCK_METHOD(void, bar, (T t), (override));
};
```
```

platisd:
When an interface includes template types, they get automatically
assigned a fixed name, i.e. T0, T1, T2 etc. This works fine as
long as the types are _not_ used in the signature of any of the
functions that need to be mocked.

Note the initial workaround would not have been if support for Python2
would be dropped. In Python 3.6 and later dictionaries are ordered
by default.

Fixes #3172

### DIFF
--- a/googlemock/scripts/generator/cpp/ast.py
+++ b/googlemock/scripts/generator/cpp/ast.py
@@ -38,6 +38,7 @@ except ImportError:
 
 import sys
 import traceback
+import collections
 
 from cpp import keywords
 from cpp import tokenize
@@ -1433,7 +1434,7 @@ class AstBuilder(object):
     pass  # Not needed yet.
 
   def _GetTemplatedTypes(self):
-    result = {}
+    result = collections.OrderedDict()
     tokens = list(self._GetMatchingChar('<', '>'))
     len_tokens = len(tokens) - 1    # Ignore trailing '>'.
     i = 0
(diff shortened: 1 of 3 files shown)


fowles (approved):
Thing change looks good.  I am going to do the stuff to merge it in tomorrow.

derekmauro:
It looks like this was already fixed by 95a9bdd9f91222e34d7ef50757e24351d73927b9.

platisd:
So did this happen to be introduced in parallel at some private branch?

fowles:
Yes, when I went to pull your change into our private branch, I discovered that a very similar change had already been made.

platisd:
Great minds think alike. :grin:
Good to see the fix merged. Thanks and take care!

PR #3183, opened 2020-12-24
Use proper feature test macro to test if library supports char8_t

georgthegreat:
Reference is here: https://en.cppreference.com/w/cpp/feature_test

This PR fixes the weird case of compiling with `clang++ -std=c++17 -fchar8_t`


georgthegreat:
@EricWF, could you, please, take a look?

Issue #2636, opened 2019-12-15
Feature Request: EXPECT_REGULAR, ASSERT_REGULAR macro's to test Regular type semantics


N-Dekker:
It is often good practice for a user-defined type to behave as a Regular type. As Titus Winters wrote in [Revisiting Regular Types](https://abseil.io/blog/20180531-regular-types) (2108): 
> After 20 years of experience, we're confident that the definition of a Regular type is a useful abstraction, and approximates the right "default semantics" for user-defined types.

However, it appears challenging for users to properly test that their types do indeed have the required semantics of a Regular type. I feel that it would be very useful for GoogleTest to provide some macro's to perform those tests.

At https://github.com/N-Dekker/gtest-regular/ I placed an example implementation of two macro's, `EXPECT_REGULAR` and `ASSERT_REGULAR`. Both macro's expect two arguments, allowing users to specify two different example values of the type they want to test. For example, when testing a user type:

    UserType example_value1(42);
    UserType example_value2{x, y, z};
    EXPECT_REGULAR(example_value1, example_value2);

Each macro call tests 12 different expectations, including:

- `a != b` equals `!(a == b)`
- `T() == T()`
- `T a(b)` implies `a == b`
- `a = b` implies `a == b`
- if `b != T()` then after `a = b; a = T();` still `b` remains unequal to `T()`
- if `a == b` then after self-assignment `a = a;` still `a` remains equal to `b`
- after `a = move(a)`, `a` should not be corrupted, so `a == a` remains true

The example implementation of these macro's is at https://github.com/N-Dekker/gtest-regular/blob/master/example_implementation/gtest-regular.h  It aims to produce a specific, informative message for each expectation failure. A test program is included at https://github.com/N-Dekker/gtest-regular (running at Azure Pipelines), displaying the output for various (intended) test failures.

Please consider adding such macro's to the GoogleTest framework!

sbenzaquen:
On Sun, Dec 15, 2019 at 7:10 AM Niels Dekker <notifications@github.com>
wrote:

> It is often good practice for a user-defined type to behave as a Regular
> type. As Titus Winters wrote in Revisiting Regular Types
> <https://abseil.io/blog/20180531-regular-types> (2108):
>
> After 20 years of experience, we're confident that the definition of a
> Regular type is a useful abstraction, and approximates the right "default
> semantics" for user-defined types.
>
> However, it appears challenging for users to properly test that their
> Regular types do indeed have the correct semantics. I feel that it would be
[...]

PR #3184, opened 2020-12-24
overload PrintTo for std::type_index, printing its name()

N-Dekker:
Typical use case:

    std::unique_ptr<AbstractProduct> product = FactoryMethod();
    ASSERT_EQ(std::type_index(typeid(*product)),
              std::type_index(typeid(ProductX)));

Possible output, including the name of the types:

> error: Expected equality of these values:
>  std::type_index(typeid(*product))
>    Which is: 8-byte object <D0-65 54-8C F6-7F 00-00> ("struct ProductY")
>  std::type_index(typeid(ProductX))
>    Which is: 8-byte object <40-64 54-8C F6-7F 00-00> ("struct ProductX")


kuzkry:
Hi! I have two findings here:
1) older `std::type_info` doesn't have its printer because it falls back to the byte printer. Do you plan on adding it too?
2) your code should probably depend on [`GTEST_HAS_RTTI`](https://github.com/google/googletest/blob/ca4b7c9ff4d8a5c37ac51795b03ffe934958aeff/googletest/include/gtest/internal/gtest-port.h#L79).

N-Dekker:
@kuzkry Thanks for your comment, Krystian.

> 1. older `std::type_info` doesn't have its printer because it falls back to the byte printer. Do you plan on adding it too?

Thanks for the suggestion. I think I'll amend this pull request tomorrow, to add an `std::type_info` overload as well! 

> 2. your code should probably depend on [`GTEST_HAS_RTTI`](https://github.com/google/googletest/blob/ca4b7c9ff4d8a5c37ac51795b03ffe934958aeff/googletest/include/gtest/internal/gtest-port.h#L79).

Are you sure it should? I mean, would it _harm_ to query the name of a type when `GTEST_HAS_RTTI` is false? The tests at https://travis-ci.org/github/google/googletest/jobs/751504899 (clang C++ NO_RTTI=ON ) seem to have passed. If `name()` only returns the compile-time type, it may still be helpful for some use cases. If `name()` just returns an empty string, it may not really harm either. What do you think?

N-Dekker:
@kuzkry Can you please check if I have addressed both of your findings correctly by my force-pushes?

googletest/include/gtest/gtest-printers.h
```diff
@@ -602,6 +604,22 @@ void PrintTo(const ::std::pair<T1, T2>& value, ::std::ostream* os) {
   *os << ')';
 }
 
+#if GTEST_HAS_RTTI
```
N-Dekker, line 607: @kuzkry I just added this `#if GTEST_HAS_RTTI` here, but now I realize it may not be necessary, as this code section does not use the `typeid` language feature. It only uses `std::type_info` and `std::type_index`, which appears acceptable, even with `-fno-rtti`. What do you think, shall I remove the `#if GTEST_HAS_RTTI` at this particular line of code from the pull request again?

kuzkry:
> @kuzkry Thanks for your comment, Krystian.

Heh, I see you took your time to find my actual name, nice :D

> > 1. older `std::type_info` doesn't have its printer because it falls back to the byte printer. Do you plan on adding it too?
> 
> Thanks for the suggestion. I think I'll amend this pull request tomorrow, to add an `std::type_info` overload as well!

Thanks too :) That looks good to me.

> > 2. your code should probably depend on [`GTEST_HAS_RTTI`](https://github.com/google/googletest/blob/ca4b7c9ff4d8a5c37ac51795b03ffe934958aeff/googletest/include/gtest/internal/gtest-port.h#L79).
> 
> Are you sure it should? I mean, would it _harm_ to query the name of a type when `GTEST_HAS_RTTI` is false? The tests at https://travis-ci.org/github/google/googletest/jobs/751504899 (clang C++ NO_RTTI=ON ) seem to have passed. If `name()` only returns the compile-time type, it may still be helpful for some use cases. If `name()` just returns an empty string, it may not really harm either. What do you think?

Good point, indeed. Therefore, it's not like it _should_, but it _could_ :) I think it does no harm really and you're right, we don't need it. If we keep it without `GTEST_HAS_RTTI`, it's one less `#ifdef` anyway. Maybe someone's gonna say that the slight advantage of using `GTEST_HAS_RTTI` is that code is more "greppable" or maybe that it's inconsistent with some logic / expectations of theirs on that switch (but it's merely a header switch anyway and some funny combinations like enabled RTTI in one's compiler and GTEST_HAS_RTTI=OFF are possible). I'm indifferent tbh but luckily this PR is gonna be looked into by Google maintainers :D

> @kuzkry Can you please check if I have addressed both of your findings correctly by my force-pushes?

Sure. I'm gonna reply in the thread of your inline comment.

Last but not least, I've just come up with this one. Did you know this?
> There is no guarantee that the same std::type_info instance will be referred to by all evaluations of the typeid expression on the same type, although std::type_info::hash_code of those type_info objects would be identical, as would be their std::type_index.
~ https://en.cppreference.com/w/cpp/language/typeid

So our equality matchers may fail on this:
```c++
const std::type_info& ti1 = typeid(A);
const std::type_info& ti2 = typeid(A);
 
EXPECT_EQ(&ti1, &ti2); // not guaranteed
```
But maybe it's too much for the testing framework and it shouldn't take care of it.

googletest/include/gtest/gtest-printers.h
kuzkry: As I wrote, I'm indifferent here :) But it's good you handled `GTEST_HAS_RTTI` in the unit test file :)

kuzkry (approved):
This is the point where I can only say I like it and wish you good luck with getting it approved by maintainers :)

N-Dekker:
For the record, it appears that the following compiles well, even with `-fno-rtti` (which I assume to correspond with `GTEST_HAS_RTTI = 0`):

```
#include <typeinfo>
#include <typeindex>

std::type_index f(std::type_index index, std::type_info info)
{
    if ((index == info) || (index.name() == info.name()))
    {
        return index;
    }
    throw index;
}

```
Tested on GCC 10.2 and Clang 11.0.0 at https://godbolt.org/z/jsvr1s

Whereas a simple `typeid(int)` fails at https://godbolt.org/z/vvrff8 (having `-fno-rtti`), saying:
> error: cannot use 'typeid' with '-fno-rtti'

googletest/include/gtest/gtest-printers.h
N-Dekker: > I'm indifferent here :) But it's good you handled `GTEST_HAS_RTTI` in the unit test file :)

@kuzkry Thanks, I just removed that one `#if GTEST_HAS_RTTI` from gtest-printers.h, as I like to avoid depending on preprocessor macros whenever possible  😃 But I will certainly leave the `#if GTEST_HAS_RTTI` in the unit test file (which does use `typeid`).

BTW The `#include <typeinfo>` and `<typeindex>` compile well without RTTI, no need to `#if GTEST_HAS_RTTI` them either 😃

N-Dekker:
> This is the point where I can only say I like it and wish you good luck with getting it approved by maintainers :)

@kuzkry Thank you very much for your approval, Krystian!

N-Dekker:
> > There is no guarantee that the same std::type_info instance will be referred to by all evaluations of the typeid expression on the same type, although std::type_info::hash_code of those type_info objects would be identical, as would be their std::type_index.
> > ~ https://en.cppreference.com/w/cpp/language/typeid
> 
> So our equality matchers may fail on this:
> 
> ```c++
> const std::type_info& ti1 = typeid(A);
> const std::type_info& ti2 = typeid(A);
>  
> EXPECT_EQ(&ti1, &ti2); // not guaranteed
> ```

@kuzkry Indeed! Fortunately, `EXPECT_EQ(ti1, ti2)` should be just fine. Right? (Guaranteed successfully passed.) Using `std::type_info::operator==(const type_info&)`.

kuzkry:
> > > There is no guarantee that the same std::type_info instance will be referred to by all evaluations of the typeid expression on the same type, although std::type_info::hash_code of those type_info objects would be identical, as would be their std::type_index.
> > > ~ https://en.cppreference.com/w/cpp/language/typeid
> > 
> > 
> > So our equality matchers may fail on this:
> > ```c++
> > const std::type_info& ti1 = typeid(A);
> > const std::type_info& ti2 = typeid(A);
> >  
> > EXPECT_EQ(&ti1, &ti2); // not guaranteed
> > ```
> 
> @kuzkry Indeed! Fortunately, `EXPECT_EQ(ti1, ti2)` should be just fine. Right? (Guaranteed successfully passed.) Using `std::type_info::operator==(const type_info&)`.

Yeah, no problem with it then. It becomes problematic when addresses are compared.

kuzkry:
Sorry to bother you, but I've found something about `GTEST_HAS_RTTI` here:
https://github.com/google/googletest/blob/d128fc8252d53baad6ea456fa08cbf9028d255f4/googletest/include/gtest/internal/gtest-port.h#L529

gtest-port.h includes `<typeinfo>` there conditionally (no other file in GoogleTest includes this btw.) so I think that's the place where you should move your `<typeindex>` inclusion and actually hide your printers under `#ifdef GTEST_HAS_RTTI`. Upon closer inspection, I think this fits better because everything related to RTTI relies on this switch and it's strictly followed rule.

kuzkry (commented):
Sorry to bother you, but I've found something about `GTEST_HAS_RTTI` here:
https://github.com/google/googletest/blob/d128fc8252d53baad6ea456fa08cbf9028d255f4/googletest/include/gtest/internal/gtest-port.h#L529

gtest-port.h includes `<typeinfo>` there conditionally (no other file in GoogleTest includes this btw.) so I think that's the place where you should move your `<typeindex>` inclusion and actually hide your printers under `#ifdef GTEST_HAS_RTTI`. Upon closer inspection, I think this fits better because everything related to RTTI relies on this switch and it's strictly followed rule.

kuzkry (changes requested):
I would like just to unapprove it but GitHub forces me to request changes.

N-Dekker:
> gtest-port.h includes `<typeinfo>` there conditionally (no other file in GoogleTest includes this btw.) so I think that's the place where you should move your `<typeindex>` inclusion and actually hide your printers under `#ifdef GTEST_HAS_RTTI`. Upon closer inspection, I think this fits better because everything related to RTTI relies on this switch and it's strictly followed rule.

@kuzkry Thanks for having another look at this pull request. Can you please elaborate a little bit more? Specifically, could those two `#include` directives in "gtest-printers.h" possibly cause any technical problems? 

https://github.com/google/googletest/blob/ae50871c5c1fbcd3e4fb35617cced5011166064a/googletest/include/gtest/gtest-printers.h#L109-L110

After having tried `-fno-rtti` on both Clang and GCC, as well as `/GR-` (disable run-time type info) on MSVC, I concluded that it appears OK to `#include` those headers, and use the Standard Library types `std::type_info` and `std::type_index`, _even_ when RTTI is disabled. (Only the use of the `typeid` language feature appeared troublesome, when RTTI is disabled.) Do you agree?

kuzkry:
Hey @N-Dekker, I'm sorry for not answering that earlier :( It slipped through my list.

I think you're right and what you did is fine 👍  I just meant we should probably try to conform to existing style and unify it with how RTTI is generally handled in GoogleTest.

I didn't want to go into details but gtest-port.h is quite special.
> // This file is fundamental to Google Test.  All other Google Test source
// files are expected to #include this.  Therefore, it cannot #include
// any other Google Test header.
~ from [line 37](https://github.com/google/googletest/blob/d128fc8252d53baad6ea456fa08cbf9028d255f4/googletest/include/gtest/internal/gtest-port.h#L37)

gtest-port.h exists so that GoogleTest can keep platform specific dilemmas away from the rest of GoogleTest code and make it more platform agnostic. And it makes sense that support for RTTI is determined there and the rest of the code simply conforms to that, hence the comment:
> // It's this header's responsibility to #include <typeinfo> when RTTI
// is enabled.
~ from [line 529](https://github.com/google/googletest/blob/d128fc8252d53baad6ea456fa08cbf9028d255f4/googletest/include/gtest/internal/gtest-port.h#L529)

Last but not least, you can always empirically verify that all RTTI-specific classes (from standard [`<typeinfo>`](https://en.cppreference.com/w/cpp/header/typeinfo) and [`<typeindex>`](https://en.cppreference.com/w/cpp/header/typeindex)) and `typeid()` , are indeed ALWAYS protected by `GTEST_HAS_RTTI` and I believe those are:
- [type_info](https://github.com/google/googletest/search?q=type_info)
- [bad_typeid](https://github.com/google/googletest/search?q=bad_typeid)
- [bad_cast](https://github.com/google/googletest/search?q=bad_cast)
- [type_index](https://github.com/google/googletest/search?q=type_index)
- [typeid](https://github.com/google/googletest/search?q=typeid)

N-Dekker:
@kuzkry Thanks you for your explanation, Krystian. Personally I would rather not add an extra `#ifdef`, it isn't technically necessary. However, I would certainly be willing to add such an `#ifdef` if it would help to get this pull request accepted. But then I would like to know if this pull request makes a chance anyway, one way or the other. 🤔

dinord:
I agree with Krystian's suggestion to guard the overloads with an `ifdef`:
- they are useless if RTTI isn't enabled anyway, guards are just an explicit way of pointing that out
- they'd be consistent with the rest of GoogleTest (see [GetTypeName](https://github.com/google/googletest/blob/1de637fbdd4ab0051229707f855eee76f5a3d5da/googletest/include/gtest/internal/gtest-type-util.h#L69) for almost identical example).

Otherwise, this PR looks good to me.

N-Dekker:
@dinord @kuzkry FYI The [force-push](https://github.com/google/googletest/compare/ae50871c5c1fbcd3e4fb35617cced5011166064a..59236b53cadbcb4b4cb18203bc97960465770ea6) that I just did only has a `git rebase master`, just to get up-to-date again. With my next force-push I hope to address the RTTI issue that you raised. Thanks so far!

N-Dekker:
@dinord @kuzkry Update: Addressed the RTTI issue with [my latest force-push](https://github.com/google/googletest/compare/59236b53cadbcb4b4cb18203bc97960465770ea6..ac3c2a8d0496893787015014a5abd397b766cce2), I hope it's OK to you!

kuzkry (approved):

N-Dekker:
@dinord Hi Dino! Thanks you for looking at this pull request of mine! I see you added a label, last week, "imported-into-CL", I'm interested to hear from you what that means! Anyway, I really hope this PR will make it onto the master branch 😃 What do you think?

dinord:
@N-Dekker your PR is going through our internal import process.

SaaadSaadAl:
S

N-Dekker:
@dinord Thank you very much for merging my pull request, Dino! It's the very first pull request of mine that made it onto the master branch of GoogleTest (while other attempts of mine, pull request #3191 and issue #2636, did not make it). I'm very proud to have made a contribution to GoogleTest!  😃

dinord:
Thank you for contributing to GoogleTest, Niels!

PR #3189, opened 2020-12-25
Fix gtest-help-test failure on GNU/kFreeBSD

ellert:


ellert:
@googlebot I signed it!

PR #3191, opened 2020-12-28
Print string for comparison operands only once when it is same for both

N-Dekker:
When both operands of a failed comparison get the very same string representation from `FormatForComparisonFailureMessage`, it is not necessary to print it twice. For example:

    auto x = std::numeric_limits<int64_t>::lowest();
    auto y = std::numeric_limits<int64_t>::min();
    ASSERT_NE(x, y);

Output before this commit:

> Expected: (x) != (y), actual: -9223372036854775808 vs -9223372036854775808

Output after this commit:

> Expected: (x) != (y), actual: -9223372036854775808 (same for both operands)

### DIFF
--- a/googletest/test/gtest_unittest.cc
+++ b/googletest/test/gtest_unittest.cc
@@ -62,6 +62,7 @@ TEST(CommandLineFlagsTest, CanBeAccessedInCodeOnceGTestHIsIncluded) {
 #include <time.h>
 
 #include <cstdint>
+#include <limits>
 #include <map>
 #include <ostream>
 #include <string>
@@ -71,6 +72,8 @@ TEST(CommandLineFlagsTest, CanBeAccessedInCodeOnceGTestHIsIncluded) {
 
 #include "gtest/gtest-spi.h"
 #include "src/gtest-internal-inl.h"
+
+const std::string suffix_for_same_operand_strings = " (same for both operands)";
 
 namespace testing {
 namespace internal {
@@ -3793,7 +3796,8 @@ TEST(AssertionTest, ASSERT_NE) {
   ASSERT_NE(6, 7);
   EXPECT_FATAL_FAILURE(ASSERT_NE('a', 'a'),
                        "Expected: ('a') != ('a'), "
-                       "actual: 'a' (97, 0x61) vs 'a' (97, 0x61)");
+                       "actual: 'a' (97, 0x61)" +
+                           suffix_for_same_operand_strings);
 }
 
 // Tests ASSERT_LE.
@@ -3807,8 +3811,8 @@ TEST(AssertionTest, ASSERT_LE) {
 // Tests ASSERT_LT.
 TEST(AssertionTest, ASSERT_LT) {
   ASSERT_LT(2, 3);
-  EXPECT_FATAL_FAILURE(ASSERT_LT(2, 2),
-                       "Expected: (2) < (2), actual: 2 vs 2");
+  EXPECT_FATAL_FAILURE(ASSERT_LT(2, 2), "Expected: (2) < (2), actual: 2" +
+                                            suffix_for_same_operand_strings);
 }
 
 // Tests ASSERT_GE.
@@ -3822,8 +3826,8 @@ TEST(AssertionTest, ASSERT_GE) {
 // Tests ASSERT_GT.
 TEST(AssertionTest, ASSERT_GT) {
   ASSERT_GT(2, 1);
-  EXPECT_FATAL_FAILURE(ASSERT_GT(2, 2),
-                       "Expected: (2) > (2), actual: 2 vs 2");
+  EXPECT_FATAL_FAILURE(ASSERT_GT(2, 2), "Expected: (2) > (2), actual: 2" +
+                                            suffix_for_same_operand_strings);
 }
 
 #if GTEST_HAS_EXCEPTIONS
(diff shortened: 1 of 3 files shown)


N-Dekker:
@kuzkry Hi Krystian! As you reviewed my previous pull request (#3184) so carefully, would you also like to look into this one?

kuzkry (approved):
Hey, sure. Since you're doing 2 things here I would suggest to create 2 separate PRs. Have a look at these pull requests for my GoogleTest fork:
https://github.com/kuzkry/googletest/pull/2
https://github.com/kuzkry/googletest/pull/3

What do you think? I think this separation makes it crystal clear.

But in general, I like this change.

N-Dekker:
> Since you're doing 2 things here I would suggest to create 2 separate PRs.

Nice suggestion, but what would you think of 2 separate commits within a single PR? Otherwise, the second PR (avoiding to print the same string twice) would depend on the first one (in order to get the right string for any operand type).

kuzkry:
> (...) what would you think of 2 separate commits within a single PR?

In theory, if my understanding of [the 2nd rule of contribution rules](https://github.com/google/googletest/blob/ca4b7c9ff4d8a5c37ac51795b03ffe934958aeff/CONTRIBUTING.md#contributing-a-patch) is correct, then yes, you should. I think those are really two separate things that happen to touch the same code :) Additionally, it enables maintainers to take the 1st commit and reject other(s) they dislike.

The bigger picture I see here is this:
- GitHub inability to mark PRs as dependent on others is a well-known [issue](https://github.com/isaacs/github/issues/959). People have been requesting a solution for a long time. This doesn't work nicely with the "one change, one PR" rule of GoogleTest but it is what it is. I understand your concern and what's worse, I've had times when maintainers were prematurely closing my PRs as they thought those were duplicates.
- 2 PRs x 1 commit each here is fine. Sometimes, you may find yourself on the opposite side, i.e. you'll try to deliver multiple commits under one PR as it makes no sense to separate them (it would break some logic cause it to lose meaning / be incomplete).

N-Dekker:
>  2 PRs x 1 commit each here is fine.

@kuzkry  Honesty I'm not so sure. I would rather not present the two commits as "independent" separate PRs for the master branch (if that's what you mean). Because the first commit _"reuse CmpHelperOpFailure()"_ is intended to be merely a coding style improvement, right? While the second commit is the one that aims to change the behavior, for any type of operand (by printing _"same for both operands"_). However, when the two commits are committed in reverse order, both commits will cause behavior changes!!!

More specifically: the _"reuse CmpHelperOpFailure()"_ commit affects the `CmpHelper##op_name` functions from "gtest.cc", that are called when comparing "big integers" like `int64_t`. If the _"same for both operands"_ commit is merged onto the master before the _"reuse CmpHelperOpFailure()"_ commit, it will have the _"same for both operands"_ behavior for `int`, but not for `int64_t`. I don't think that would be desirable.

Note that I just added a (missing) unit test for `int64_t`, to clarify my point. This unit test will fail without the _"reuse CmpHelperOpFailure()"_ change. https://github.com/google/googletest/compare/0ef094adc866d565cda807bfb48941fc407fd14d..cfd65970d07c7b8d30f82fcb0c9afe0e626269ae

You see the problem? Instead of creating two independent PRs, it would still be OK to me to make this 1 PR of 2 commits... but would that possibly still get your approval as well?  😃

googletest/test/gtest_unittest.cc
```diff
@@ -4551,9 +4555,23 @@ TEST(ExpectTest, EXPECT_EQ_0) {
 TEST(ExpectTest, EXPECT_NE) {
   EXPECT_NE(6, 7);
 
+  constexpr auto lowest_int64 = ::std::numeric_limits<::std::int64_t>::lowest();
+  constexpr auto min_int64 = (::std::numeric_limits<::std::int64_t>::min)();
```
kuzkry, line 4559: Extra pair of parenthesis not needed here. Compare with the previous line.

kuzkry:
> However, when the two commits are committed in reverse order, both commits will cause behavior changes!!!

Haha, actually yes. Obviously order would matter here but that's the case when there's dependency. Anyway, I respect you don't want complications.

> You see the problem? Instead of creating two independent PRs, it would still be OK to me to make this 1 PR of 2 commits... but would that possibly still get your approval as well? 😃

Sure. Just fix that small format issue please and it'll be perfect.

googletest/test/gtest_unittest.cc
N-Dekker: @kuzkry Are you sure the extra pair of parenthesis not needed here? When I remove them, VS2019 says:

> gtest_unittest.cc(4559,69): error C2220: the following warning is treated as an error
> gtest_unittest.cc(4559,69): warning C4003: not enough arguments for function-like macro invocation 'min'
> gtest_unittest.cc(4559,69): error C2589: '(': illegal token on right side of '::'
> gtest_unittest.cc(4559,69): error C2062: type 'unknown-type' unexpected
> gtest_unittest.cc(4559,69): error C2059: syntax error: ')'
> gtest_unittest.cc(4563,1): error C3536: 'min_int64': cannot be used before it is initialized

As it tries to call the `min(a,b)`macro from "minwindef.h".

kuzkry: LOL, I didn't expect that 😄 I only verified it on my Linux machine. You're right

kuzkry (approved):

googletest/test/gtest_unittest.cc
N-Dekker: @kuzkry No problem, thanks! Instead of the extra parentheses, I could have added `#define NOMINMAX`, or `#undef min`, but some people may not like those hashtags either 😸  

Thanks also for your approval  😃

kuzkry:
Hey! FYI, I find some of this code spurious and I'm gonna attempt to remove it. Depending what gets first, your PR may conflict with #3204. Happy new year btw. :)

N-Dekker:
> FYI, I find some of this code spurious and I'm gonna attempt to remove it. Depending what gets first, your PR may conflict with #3204.

@kuzkry Thanks for informing me about our potential conflict, Krystian! Now I just hope that mine gets merged first 😸 Happy new year anyway!  :)

N-Dekker:
> Depending what gets first, your PR may conflict with #3204.

I guess such a (possible) conflict could be fixed by simply doing a `git rebase -i HEAD~2`, and then removing the first commit from this PR, [Reuse existing CmpHelperOpFailure() in GTEST_IMPL_CMP_HELPER_ (gtest.cc)](https://github.com/google/googletest/pull/3191/commits/724a650b5baf9fd504340d992531ad2f192eeed4), right?

kuzkry:
That's right :) It's good you're doing notes cause some things take time to be delivered here so you could forget about this merge conflict 10 times 😄

derekmauro:
Thanks for the pull request. If we were to do this over, I think we'd print a message similar to the one you are proposing. Unfortunately though, we have to deal with [Hyrum's Law](https://www.hyrumslaw.com) in the sense that there may be code depending on the message that has been baked in for a while. The cost of the churn here outweighs the cost of having a better message.

Sorry about this.

N-Dekker:
@derekmauro Thanks for having reviewed my pull request. Of course, I'm very disappointed about the rejection of my proposed improvement.

> there may be code depending on the message that has been baked in for a while

Do you possibly have an example of a useful piece of code in the real world, that would be broken by improving the error message of `EXPECT_NE` and `ASSERT_NE`?

N-Dekker:
> Unfortunately though, we have to deal with Hyrum's Law

FYI, I just informed Hyrum about his law being dealt with: https://twitter.com/Niels_Dekker/status/1350450521132453888

PR #3195, opened 2020-12-29
Using auto instead of container::const_iterator

### LINKED ISSUE #3194: ElementsAre doesn't compile if container::const_iterator is not declared
I cannot use `ElementsAre` with `gsl::span`
```cpp
TEST(demo, elements_are_on_gsl_span)
{
   std::array<int, 4> data = { 1, 2, 3, 4 };
   gsl::span<int, 4> span(data);
   ASSERT_THAT(span, ::testing::ElementsAre(1, 2, 3, 4));
}
```
because the current implementation of `ElementsAreMatcherImpl` explicitely references `typename StlContainer::const_iterator` 
```cpp
  bool MatchAndExplain(Container container,
                       MatchResultListener* listener) const override {
    // To work with stream-like "containers", we must only walk
    // through the elements in one pass.

    const bool listener_interested = listener->IsInterested();

    // explanations[i] is the explanation of the element at index i.
    ::std::vector<std::string> explanations(count());
    StlContainerReference stl_container = View::ConstReference(container);
    typename StlContainer::const_iterator it = stl_container.begin();
    ...
```
which does not exist for `gsl::span`.

Could we use auto instead?
```cpp
auto it = stl_container.begin();
```

Inujel:
Fixes #3194

### DIFF
--- a/googlemock/include/gmock/gmock-matchers.h
+++ b/googlemock/include/gmock/gmock-matchers.h
@@ -2370,7 +2370,6 @@ class ContainerEqMatcher {
     typedef internal::StlContainerView<
         typename std::remove_const<LhsContainer>::type>
         LhsView;
-    typedef typename LhsView::type LhsStlContainer;
     StlContainerReference lhs_stl_container = LhsView::ConstReference(lhs);
     if (lhs_stl_container == expected_)
       return true;
@@ -2379,8 +2378,7 @@ class ContainerEqMatcher {
     if (os != nullptr) {
       // Something is different. Check for extra values first.
       bool printed_header = false;
-      for (typename LhsStlContainer::const_iterator it =
-               lhs_stl_container.begin();
+      for (auto it = lhs_stl_container.begin();
            it != lhs_stl_container.end(); ++it) {
         if (internal::ArrayAwareFind(expected_.begin(), expected_.end(), *it) ==
             expected_.end()) {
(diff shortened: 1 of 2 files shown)


Inujel:
@googlebot I fixed it.

googletest/src/gtest-internal-inl.h
```diff
@@ -290,7 +290,7 @@ inline int CountIf(const Container& c, Predicate predicate) {
   // Implemented as an explicit loop since std::count_if() in libCstd on
   // Solaris has a non-standard signature.
```
N-Dekker, line 291: Just wondering, would this workaround (a hand-written "count-if", below here) still be necessary, or is `std::count_if` nowadays fine on any supported platform? The workaround was introduced by zhanyong.wan, April 12, 2011 "Fixes Sun C++ compiler errors (by Pasi Valminen)", https://github.com/google/googletest/commit/b8c0e16eeb7496f71480c6a060144b0e050edcf5

googlemock/include/gmock/gmock-matchers.h
```diff
@@ -4597,7 +4594,7 @@ UnorderedPointwise(const Tuple2Matcher& tuple2_matcher,
 
   // Create a matcher for each element in rhs_container.
   ::std::vector<internal::BoundSecondMatcher<Tuple2Matcher, Second> > matchers;
-  for (typename RhsStlContainer::const_iterator it = rhs_stl_container.begin();
+  for (auto it = rhs_stl_container.begin();
```
N-Dekker, line 4597: Would a range-based `for` loop be OK here as well? (Just asking. I'm not a maintainer.)

Inujel: I think so!

andrewkcorcoran:
Any progress getting this approved/merged? As stated by the linked issues this is preventing gtest from being used with gsl::span.

andrewkcorcoran (approved):

ladislas:
would need this as well! :)

Ket3r:
Any news on when this will be merged?

samkearney:
Ran across this issue today, +1 for merging this!

PR #3199, opened 2020-12-30
Split big test

ellert:
The gmock-matchers_test has grown so that it no longer compiles on a machine with 2 GB of RAM.
This PR splits the test in two, so that the compilation of each part uses less memory.
The changes are split over two commits so that git can keep track of the changes and show the correct history when using git blame. If the whole change is done in one commit the history is lost for one of the parts.

This change is proposed in order to resolve the failing builds of googletest on the sh4 architecture in the Debian Linux distribution.


ellert:
Rebased to resolve conflicts.

ellert:
Closing in favour of #3653.

PR #3200, opened 2020-12-30
Port to GNU/Hurd

ellert:
This PR defines GTEST_OS_GNU_HURD for compiling on Debian GNU/Hurd, and adds it to relevant #ifdefs in the code.

Note that since this PR makes additional changes to the file changed in #3189 it was branched of that branch in order to avoid a merga conflict later. Once that PR is in master this one can be rebased if so desired.


googletest/include/gtest/internal/gtest-port.h
```diff
@@ -1331,6 +1335,9 @@ class ThreadWithParam : public ThreadWithParamBase {
   void Join() {
     if (!finished_) {
       GTEST_CHECK_POSIX_SUCCESS_(pthread_join(thread_, nullptr));
+      // Yield to the joining thread to ensure its destructors are run
+      // before marking the thread as finished.
+      sched_yield();
```
fowles, line 1340: Relying on sched_yield to get interleaving is almost never correct.  The joined thread should have its dtors run before join returns regardless.  I have serious concerns about this

fowles (commented):
At a high level, we do not wish to make Hurd an officially supported platform because we don't have continuous integration for it.  That said, most of this patch is pretty reasonable and we would accept it on a best effort support basis.  I do have one real concern called out in the code that we need to resolve however.

googletest/include/gtest/internal/gtest-port.h
ellert: I agree that the call to sched_yield was the one arguable change in this PR.
The reason for adding it was that without it the TEST(ThreadLocalTest, DestroysManagedObjectAtThreadExit) test fails.
It might be a better idea to simply disable this test rather than to try to force it to work.

Rebased to master and removed the call to sched_yield and put the one failing test inside a #if !defined(GTEST_OS_GNU_HURD) instead.

pinotree:
@fowles while I can understand the POV about supported platforms, it seems like GoogleTest already supports many other OSes not listed as supported in the README (e.g. AIX, the various BSDs, etc). If those OSes are already community-supported on "best effort" from you, I don't see how adding Hurd would be worse.

Also, the issue I hit is that, since the recommended way of using GoogleTest is _embedding_ it in the sources of your own project (sigh), then every time there is a project that uses an embedded GoogleTest we (Hurd porters) need to debug and fix the GoogleTest failures, and apply manually a patch similar to this one. This is honestly a tiring and time consuming job, and having GoogleTest fixed upstream would at least help by telling projects to update their embedded copy (in case they don't update it regularly).

@ellert considering that #3189 was merged, can you please rebase this once again on master?

fowles:
Sorry for the long period of silence here.  I had my alerts misconfigured and failed to notice the update.  I am broadly ok with this but don't know how important this behavior is to functioning correctly on all systems.  Most likely it will make tests more flakey on GNU/HURD, which in the end is more your problem than mine.

fowles:
Assigning to Andy for final review and merge

googletest/include/gtest/internal/gtest-port.h
asoffer: Could you say more about why the test doesn't work on this platform? At the very least we should have a comment explaining why the test cannot pass on this platform, and maybe some notes on workarounds if it affects users who need the thread local destructors to run. But ideally fixing the underlying issue is preferable.

ellert: When I run the test on exodar.debian.net it fails every time:
```
$ googletest/googletest-port-test
[ ... ]
[==========] 49 tests from 16 test suites ran. (5090 ms total)
[  PASSED  ] 48 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] ThreadLocalTest.DestroysManagedObjectAtThreadExit
```
However, if I run only the thread-local test it succeeds:
```
$ googletest/googletest-port-test --gtest_filter=ThreadLocalTest.*
[ ... ]
[==========] 9 tests from 1 test suite ran. (20 ms total)
[  PASSED  ] 9 tests.
```
So there seems to be some interference with other tests.

ellert: If I move the death tests to a separate file there is no more interference, and all tests succeed.
```
      Start 26: googletest-port-death-test
26/62 Test #26: googletest-port-death-test .................   Passed    7.34 sec
      Start 27: googletest-port-test
27/62 Test #27: googletest-port-test .......................   Passed    3.13 sec
```
I have removed the #ifndef disabling the DestroysManagedObjectAtThreadExit test and added a commit that does the above split. Is this acceptable?

PS. Can you address the memory usage during compilation? To build the package on the Debian hurd test machine, which had 3 GB of memory, I have to apply the proposed change from #3199 in order not to run out of memory.

pinotree: @ellert I can reproduce the failure on exodar, however on my VM all the tests work fine (master + the current patch #1 in this series). I'd say there might be something else going on, specific to Hurd, and to discuss it in a Hurd list (bug-hurd upstream, for example).

Let's go ahead with the plain GNU/Hurd enablement, and in case fixes can be applied later on. WDYT?

ellert: @pinotree Thank you for redoing the test on exodar and confirming my findings. If the issue with the interference is not a general problem that happens for all hurd installations, I agree that maybe we can ignore it for now and only apply the first commit of the PR. It would be interesting to know why it happens on exodar though, and also if it happens on the Debian build servers. If it does happen on the build servers I guess the second commit of the PR can be added as a patch to the Debian source package until another solution is found.

I have removed the second commit from the PR.

ellert:
Hi. Any news?

ellert:
This seems to have gotten stuck.

fowles:
I will poke Andy internally

PR #3207, opened 2021-01-06
Prevent unintended parameter generation

hyukmyeong:
If `INSTANTIATE_TEST_SUITE_P` or `INSTANTIATE_TYPED_TEST_SUITE_P` are used  
with the same `prefix + test_suite_name` in more than one translation unit,  
they can be run multiple times even though repeat options are not used  
  
As guide documents encourage users to call the macros with unique prefixes  
It looks like they should be run only once if same `prefix`es are used  
  
To do that, I removed `static` from the variables which are defined after the corresponding tests are registered  
to inform users of the multiple definition at link time


Ernel1997:
#3578

PR #3215, opened 2021-01-11
Protect against -Wundef

adambadura:
`-Wundef` detects cases where an undefined symbol is evaluated in the context of `#if` and defaults to `0`. While this default behavior could sometimes be a nice shortcut it can also hide issues like lack of proper configuration or a misspelled macro.

Furthermore, clients adding the library by CMake `add_subdirectory` might be hit by the problem of `-Wundef` triggered in the library itself. Protecting against it also simplifies client scenarios by not requiring the client to disable the flag in the library builds.

For "detection macros" this is done by either adding proper `#else` cases (in simplest scenarios) or by appending a block:
```C
#ifndef <VAR>
# define <VAR> 0
#endif
```
which avoids complexity while ensuring the macro is always defined.

For "configuration macros" this is done by adding a similar block prior to the first use to ensure it is always defined.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,19 +22,19 @@ matrix:
     - os: linux
       dist: bionic
       compiler: gcc
-      env: BUILD_TYPE=Debug CXX_FLAGS="-std=c++11 -Wdeprecated"
+      env: BUILD_TYPE=Debug CXX_FLAGS="-std=c++11 -Wdeprecated -Wundef"
     - os: linux
       dist: bionic
       compiler: clang
-      env: BUILD_TYPE=Release CXX_FLAGS="-std=c++11 -Wdeprecated" NO_EXCEPTION=ON NO_RTTI=ON COMPILER_IS_GNUCXX=ON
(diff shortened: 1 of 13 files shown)


derekmauro:
Sorry for the delayed response on this one. Unfortunately we can't take this as is because it is a breaking change. For example, any code doing
```
#ifdef GTEST_HAS_DEATH_TEST
```
will be broken by this.

The style used in GoogleTest is inconsistent at times. Some macros are always defined to 0 or 1, others are either defined to 1 or undefined. Resolving this is going to be more complicated.

adambadura:
@derekmauro, so... should I abandon this pull request? Or do you see any way to work around it? Can we perhaps determine which symbols might be used this way? Or go the other way and stop relying on value where it wasn't clearly specified?

derekmauro:
I haven't completely thought it through, but my first suggestion would be to look into changing `#if` to `#ifdef` or `#if defined()` for the macros that are either defined to 1 or undefined.

Let us know if you are interested in doing this.

adambadura:
@derekmauro, I will try to rewrite it. It may take some time thou, so if you are in a rush with this fix, do it yourself then.

thughes:
This should be fixed as of commit d92a270d2d9d6bb7ed3bf7dc3f698028550a65ef.

higher-performance:
Closing given this appears to no longer be an issue.

PR #3231, opened 2021-01-16
Fix link to Google Test Primer

dreamer:
Markdown file was moved to a different directory. When googling for
"gtest primer", the result is 404 page.


ericschmidtatwork:
Thanks for the PR, we've brought in the change internally. Please don’t push any more changes here as they might be overwritten.
(353312701)

PR #3262, opened 2021-02-05
[gmock_gen.py] Do not include void into mocked method arguments

### LINKED ISSUE #3261: [GoogleMock] Mocked method with explicit "void" argument does not compile
When explicitly declaring `void` as the argument of a function that takes no arguments, things get messy.

See example below:
```cpp
class Foo {
 public:
  virtual bool bar(void) = 0;
};
```
A mock generated like this, does not compile:
```cpp
class MockFoo : public Foo {
 public:
  MOCK_METHOD(bool, bar, (void), (override));
  //                       ^------- That's a problem!
};
````
The error message:
> error: static_assert failed due to requirement '::testing::tuple_size<std::tuple<> >::value == 1' "This method does not take 1 arguments. Parenthesize all types with unprotected commas."

I tried going down the rabbit hole but I was soon deeper than I'm comfortable with. It might boil down to the following snippet from `gmock-generated-function-mockers.h` not working as intended with a single `(void)` argument:

```cpp
// The type of argument N of the given function type.
// INTERNAL IMPLEMENTATION - DON'T USE IN USER CODE!!!
#define GMOCK_ARG_(tn, N, ...) \
    tn ::testing::internal::Function<__VA_ARGS__>::template Arg<N-1>::type
```

Additionally, in the unit tests of `gmock_gen.py` found in `gmock_class_tests.py`, there's the following test case which implies that such mocks should be supported:

```python
  def testExplicitVoid(self):
    source = """
class Foo {
 public:
  virtual int Bar(void);
};
"""
    self.assertEqualIgnoreLeadingWhitespace(
        'MOCK_METHOD(int, Bar, (void), (override));',
        self.GenerateMethodSource(source))
```

platisd:
If a function that takes no arguments explicitly states (void)
then do not include it in the mocked method argument list
since it triggers static assertions that expect no arguments
to be present.

Fixes #3261

### DIFF
--- a/googlemock/scripts/generator/cpp/gmock_class.py
+++ b/googlemock/scripts/generator/cpp/gmock_class.py
@@ -132,7 +132,8 @@ def _GenerateMethods(output_lines, source, class_node):
       args = []
       for p in node.parameters:
         arg = _GenerateArg(source[p.start:p.end])
-        args.append(_EscapeForMacro(arg))
+        if arg != 'void':
+          args.append(_EscapeForMacro(arg))
 
       # Create the mock method definition.
       output_lines.extend([
--- a/googlemock/scripts/generator/cpp/gmock_class_test.py
+++ b/googlemock/scripts/generator/cpp/gmock_class_test.py
@@ -156,7 +156,7 @@ class Foo {
 };
 """
     self.assertEqualIgnoreLeadingWhitespace(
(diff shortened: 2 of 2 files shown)


platisd:
This does not _really_ fix the underlying issue, but at least mitigates the problem when using the `gmock_gen.py` script.

PR #3313, opened 2021-03-12
Move failure reporter to variable

IhnoL:
We are using GoogleMock for mocking our Unreal Engine dependencies.
As the default FailureReporter can't be overridden in the current implementation, we adapted it a bit to make it work.

I'm not sure if this could be merged as is. Anyway it would be nice to have these or similar changes in the codebase, to enable developers to override the FailureRepoter with their custom implementation. In our case this custom implementation just logs errors to Unreal Engine instead of std::out.

### DIFF
--- a/googlemock/src/gmock-internal-utils.cc
+++ b/googlemock/src/gmock-internal-utils.cc
@@ -107,6 +107,14 @@ class GoogleTestFailureReporter : public FailureReporterInterface {
   }
 };
 
+static FailureReporterInterface* failure_reporter = nullptr;
+
+// Sets the failure reporter.
+// Enables overriding of the default failure reporter with a custom one that implements the interface
+GTEST_API_ void SetFailureReporter(FailureReporterInterface* const in_failure_reporter) {  
+  failure_reporter = in_failure_reporter;  
+}
+
 // Returns the global failure reporter.  Will create a
 // GoogleTestFailureReporter and return it the first time called.
 GTEST_API_ FailureReporterInterface* GetFailureReporter() {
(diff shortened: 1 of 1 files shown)


IhnoL:
Hello? Is anyone out there?

IhnoL:
Up

sbenzaquen:
Sorry for the delay.
`FailureReporterInterface` is an internal API, so even if we were to add this function you can't implement a `FailureReporterInterface` without using `testing::internal`.
Note that the failure reporter is not the one posting to `std::cout`. All it does is post an event to the registered `TestEventListener`.
The event listeners is the public API. This is where you can add your own listener to post the events to any backend you prefer.

IhnoL:
Thanks for the hint!

This seems to do the job:

class WTFListener : public testing::EmptyTestEventListener
{
	void OnTestPartResult(const testing::TestPartResult& result) override
	{
		if (result.type() == testing::TestPartResult::kFatalFailure
			|| result.type() == testing::TestPartResult::kNonFatalFailure)
		{
			const auto Message = FString(result.message());
			UE_LOG(LogTemp, Error, TEXT("%s"), *Message);
		}
	}
};

void FTestModule::StartupModule()
{
	UE_LOG(LogTemp, Log, TEXT("TestModule has started!"));
	::testing::TestEventListeners& Listeners = ::testing::UnitTest::GetInstance()->listeners();
	Listeners.Append(new WTFListener());
}

>We use only gmock and don't run our tests via gtest. Maybe I overlooked it for that reason.<

PR #3319, opened 2021-03-17
Update nicestrictnaggy gmock cook_book links

a-sully:
An incorrect link is output to the user when these conditions are met.


ericschmidtatwork:
Thanks for this pull request, we've brought in the change internally and it's going through review. Please don’t push any more changes here as they might be overwritten.
(363647553)

PR #3325, opened 2021-03-20
Add missing ; to code example

rscohn2:


rscohn2:
@googlebot I signed it!

derekmauro:
The semi-colon isn't missing. That is an empty function definition. No semi-colon is needed.

PR #3335, opened 2021-03-27
Add CMake alias targets

NewProggie:
Currently, projects may use `target_link_libraries(... GTest::gtest)` etc. when consuming googletest via `find_package()` and `target_link_libraries(... gtest)` when consuming it via `add_subdirectory()`.
This PR enables consuming projects to always refer to be using `GTest::gtest` etc. regardless of how this project was integrated.


derekmauro:
#3155 is nearly identical and was just imported. It should be re-exported to GitHub this week.

PR #3345, opened 2021-04-05
Allow a port to set GTEST_INTERNAL values manually

donny-dont:
Add `!defined` checks around macros determining whether a port supports a `GTEST_INTERNAL` utility. That way more exotic platforms can manually set this value if desired.

This issue cropped up after WebKit updated its checked in copy of googletest. The platform has an expectation that if you are using `<any>` that RTTI is enabled but WebKit explicitly disables RTTI.

### DIFF
--- a/googletest/include/gtest/internal/gtest-port.h
+++ b/googletest/include/gtest/internal/gtest-port.h
@@ -2277,11 +2277,16 @@ using Any = ::absl::any;
 }  // namespace internal
 }  // namespace testing
 #else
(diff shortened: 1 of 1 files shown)


derekmauro (changes requested):
We want to avoid having manual build modes as they create a combinatorial expulsion of modes to test in.

Can the check be based existing macros somehow? Perhaps some combination of `GTEST_HAS_RTTI` and `__macro_defined_by_your_compiler__`?

derekmauro:
Closing due to inactivity.

PR #3361, opened 2021-04-14
Workaround for IAR chrono library

arBmind:
The IAR compiler is popular on embedded systems because it produces very compact binaries. The C++ standard library is different. Therefore this patch is required.

* Use time_point from clock class instead of the template.

dlib uses the same time_point structure for all clocks. But the template specialisation would generate a new type for each clock. Compilation fails because of the type mismatches.

### DIFF
--- a/googletest/src/gtest.cc
+++ b/googletest/src/gtest.cc
@@ -1044,7 +1044,7 @@ class Timer {
   }
 
  private:
-  std::chrono::time_point<std::chrono::steady_clock> start_;
+  std::chrono::steady_clock::time_point start_;
 };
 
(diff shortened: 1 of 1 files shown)


arBmind:
This is a minor tweak, not a substantial contribution. It's too much effort to go through all the hassles and get authorization to sign the CLA for that.
Take it as inspiration and if you like it, apply the changes or dismiss it.

The IAR compiler is popular on embedded systems because it produces very compact binaries. The C++ standard library is different. Therefore this patch is required.

derekmauro:
By policy we have to close PRs with no CLA, but I will look into using it as inspiration for a trivial change.

PR #3376, opened 2021-04-19
Update CMake scripts

roehling:
This PR simplifies the CMake scripts by removing legacy code. The
changes in particular are:

* Require CMake 3.8 as minimum version to avoid messy workarounds for
  ancient CMake releases.
* Do not use the GTEST_API_ macro with GMock. It's a different library,
  and that reuse is fundamentally broken, with a few flimsy workarounds
  in CMake to cover up the problem.
* Make the project more vendoring-friendly; in particular, default to
  not building the internal unit tests if Googletest is included as
  subproject, but build them by default for stand-alone compilation.
* Remove a lot of very opinionated compiler flag settings. Virtually
  all of these settings are irrelevant for normal users, and may even
  interfere with inherited build flags if Googletest is included as
  subproject.


roehling:
I've updated the scripts for the new 1.11.0 release and further separated the GMock targets to aid downstream packagers in having separate devel packages for Googletest and Googlemock.

googlemock/include/gmock/internal/gmock-port.h
```diff
@@ -60,6 +60,27 @@
...
+#ifndef GMOCK_API_
+
+#ifdef _MSC_VER
+# if GMOCK_LINKED_AS_SHARED_LIBRARY
+#  define GMOCK_API_ __declspec(dllimport)
+# elif GMOCK_CREATE_SHARED_LIBRARY
+#  define GMOCK_API_ __declspec(dllexport)
+# endif
+#elif __GNUC__ >= 4 || defined(__clang__)
+# define GMOCK_API_ __attribute__((visibility ("default")))
```
joshiayush, line 75: I believe the reason why `GTEST_API_` macro was there because of the fact that mocking is a subset of `googletest` and should be served under as its public API. The whole point of `__attribute__((visibility ("default"))` is to expose those symbols for dynamic linkage, done either way makes no difference but separates one from another. In my opinion `GTEST_API_` should not be removed, again not an expert, just an opinion.

PR #3392, opened 2021-04-30
fixed includes in gtest_all

marpit:
In gtest_all.cc file is wrong path to include files. This file and included files are in same directory, but gtest-all.cc include this files from src directory.


Leandro545 (commented):
Confirma

dinord:
Does the original specification (`src/foo.h`) not work for your project?

marpit:
> Does the current include style (`src/foo.h`) not work for your project?

`
fatal error: src/gtest.cc: No such file or directory
...
`

because of these files are in same directory as gtest_all file, so this path link to src directory, but this directory does not exist in this folder where is located gtest_all file. 

If I apply this fix, compilation is ok and all files are found in correct path. 

`
// The following lines pull in the real gtest *.cc files.
#include "gtest.cc"
#include "gtest-death-test.cc"
#include "gtest-filepath.cc"
#include "gtest-matchers.cc"
#include "gtest-port.cc"
#include "gtest-printers.cc"
#include "gtest-test-part.cc"
#include "gtest-typed-test.cc"
`

If you open src folder in repository: https://github.com/google/googletest/tree/master/googletest/src you can see that all files in this folder, so include path is wrong.

PR #3393, opened 2021-05-04
isalnum -> IsAlNum for correct handling of signed chars

JC3:
I spotted this in https://github.com/assimp/assimp/pull/3880/commits/7dd7a053a91322fad88cdf958c6d0b3b7b91cb90 and figured I'd fix it here, too. 

If this is not the right thing to do, please lmk so I can undo it in assimp, too. Seems right, though. It's the only spot in gtest where a ctype call was made directly.


JC3:
@googlebot I signed it!

Issue #3390, opened 2021-04-29
GTEST_CAN_STREAM_RESULTS_ issue with PGI opecnacc compiler


PhilipDeegan:
Hi there, 

I'm having trouble with this line using socket.h while attempting to compile with the PGI compiler from nvidia

https://github.com/google/googletest/blob/252ce9c52d304659eff6be558209c811b7191963/googletest/include/gtest/internal/gtest-port.h#L631

```
"/usr/include/x86_64-linux-gnu/bits/socket.h", line 282: error: incomplete type is not allowed
      __extension__ unsigned char __cmsg_data __flexarr; /* Ancillary data.  */
```

would a PR be considered surrounding this line with a "#ifndef GTEST_CAN_STREAM_RESULTS_" ?

Thanks

dinord:
Could you please clarify your suggestion to use `#ifndef` to fix this?

`GTEST_CAN_STREAM_RESULTS_` is defined for an allowlist of platforms, and the platform you're compiling against appears to be in that list.

We would consider a pull request exempting the PGI openacc compiler, if you can prove it cannot link against the implementation of `bits/socket.h` (e.g. with a link to code or official documentation).

PhilipDeegan:
> Could you please clarify the suggestion to use #ifndef to fix this?

```
#ifndef GTEST_CAN_STREAM_RESULTS_
#defne GTEST_CAN_STREAM_RESULTS_ 1
#endif
```

This would allow us to disable this feature as we have no need for it.

> if you can prove it cannot link against the implementation of bits/socket.h (e.g. with a link to code or official documentation).

I don't think I can, but perhaps nvidia can: https://forums.developer.nvidia.com/t/nvc-google-test-fails-to-compile/177067

dinord:
That sounds reasonable, even without additional evidence (since it wouldn't impact other users). Feel free to send the PR.

PhilipDeegan:
cmake was adding in a "-A" compile flag to gtest when it probably shouldn't have been.

Without this flag everything compiles ok.

PR #3394, opened 2021-05-05
gtest-port.h - ifndef for disabling stream results

PhilipDeegan:
commit description:

```
only define GTEST_CAN_STREAM_RESULTS_ if not defined
Non GCC compilers can have issues with this (eg pgcc)
such errors may include:

/usr/include/x86_64-linux-gnu/bits/socket.h", line 282: error: incomplete type is not allowed
      __extension__ unsigned char __cmsg_data __flexarr; /* Ancillary data.  */
```

issue: https://github.com/google/googletest/issues/3390


akeles85 (approved):

PhilipDeegan:
It appears that this "incomplete type is not allowed" error only appears with pgc++ when the flag "-A" is in use.

```
-A                  Accept ANSI standard C++
```

this PR may be unnecessary.

PR #3414, opened 2021-05-22
Remove Cocoa use on iOS.

pwnall:
(currently creating PR to check if it passes tests)

https://github.com/google/googletest/commit/252ce9c52d304659eff6be558209c811b7191963 introduced a dependency on Cocoa (Foundation.h, Objective C++) in googletest for iOS.

This PR aims to remove the dependency by replacing the functionality with a lower-level API that does not require Objective C++.

### DIFF
--- a/googletest/src/gtest-port.cc
+++ b/googletest/src/gtest-port.cc
@@ -80,10 +80,6 @@
 # include <zircon/syscalls.h>
 #endif  // GTEST_OS_FUCHSIA
 
-#if GTEST_OS_IOS
-#import <Foundation/Foundation.h>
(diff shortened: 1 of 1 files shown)


derekmauro:
FYI it isn't necessary to send both an internal CL and a pull request. Internal CLs are far easier for us to handle, and will get exported back to github.

pwnall:
@derekmauro Thank you for the quick response!

I put this up to get the CI feedback. If the internal repo covers all the checks, I'll stop putting up PRs in the future.

derekmauro:
The internal repo now has the full CI.

PR #3429, opened 2021-06-03
Make a matcher ElementsAreArray applicable for std ranges

1KoT1:
Hi!
I propose to make a matcher ElementsAreArray applicable for std ranges.
It is small changes in the code but it allow to make assertions of containers easy.
Compilers gcc 5.0, clang 5.0, MSVC 2015 support 'auto' so you don't need to enhance requirements for compilers. And users with modern compillers can use modern features.

### DIFF
--- a/googlemock/include/gmock/gmock-matchers.h
+++ b/googlemock/include/gmock/gmock-matchers.h
@@ -3527,8 +3527,8 @@ inline internal::ElementsAreArrayMatcher<T> ElementsAreArray(
 }
 
 template <typename Container>
-inline internal::ElementsAreArrayMatcher<typename Container::value_type>
-ElementsAreArray(const Container& container) {
+inline auto ElementsAreArray(const Container& container)
+    -> decltype(ElementsAreArray(container.begin(), container.end())) {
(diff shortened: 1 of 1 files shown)


1KoT1:
@googlebot I signed it!

1KoT1:
@googlebot I signed it!

CJ-Johnson:
template <typename Container>
     inline auto
     ElementsAreArray(const Container& container) {
       return ElementsAreArray(container.begin(), container.end());
     }

As written, this will not work in C++11. However, with a small change, it will be ok.

     template <typename Container>
     inline auto ElementsAreArray(const Container& container)
         -> decltype(ElementsAreArray(container.begin(), container.end())) {
       return ElementsAreArray(container.begin(), container.end());
     }

Please change the diff to use this instead :)

MARY203475 (approved):

1KoT1:
@CJ-Johnson, thanks.

CJ-Johnson:
Thank you for PR! It has now been imported for internal testing. Please refrain from making additional changes. The PR will close once the change is upstreamed. (379509119)

PR #3453, opened 2021-06-22
remove DEBUG_POSTFIX

### LINKED ISSUE #2174: In debug build: gtest.pc.in provides -lgtest flag but libgtestd.a is built.
When building gtest with CMAKE_BUILD_TYPE Debug, googletest/cmake/gtest.pc.in contains hard-coded` -lgtest` flag, but libgtest**d**.a file is built.

I think that this flag should be templated.

r00t-:
while this is not synced with the pkg-config file, it only breaks things.

fix #2174

### DIFF
--- a/googletest/cmake/internal_utils.cmake
+++ b/googletest/cmake/internal_utils.cmake
@@ -154,10 +154,6 @@ function(cxx_library_with_type name type cxx_flags)
   set_target_properties(${name}
     PROPERTIES
     COMPILE_FLAGS "${cxx_flags}")
-  # Generate debug library name with a postfix.
-  set_target_properties(${name}
-    PROPERTIES
-    DEBUG_POSTFIX "d")
   # Set the output directory for build artifacts
   set_target_properties(${name}
     PROPERTIES


asoffer:
Ping. Please sign the CLA so we can accept this PR.

r00t-:
i will have to talk to my employer before i can sign anything, will try.
in the meantime i suggest you simply steal my "code".

Neumann-A:
This breaks FindGTest.cmake from CMake for multi config generators......

r00t-:
cmake hardocdes the "gtestd" name:
https://gitlab.kitware.com/cmake/cmake/-/blob/ffa4c4b527f78b6e1f6c267f05ae3469be659d0f/Modules/FindGTest.cmake#L412

i don't think it should be relying on that. (but i'm no expert in that domain.)

Neumann-A:
> i don't think it should be relying on that. 

It has to. there is no other way to differentiate between libs otherwise. Especially if they are installed into the same prefix which is often assumed.
The best way to solve this issue is to fix the pkg-config file, since you are probably supplying it with different PKG_CONFIG_PATHs already. The trick is to move the gneration of the pkg-config in the install or build phase instead of the configure phase because then generator expressions are available, e.g. `TARGET_FILE_BASE_NAME` or `TARGET_LINKER_FILE_BASE_NAME`

asoffer:
Closing this PR becuase we do not have a signed CLA.

googletest/cmake/internal_utils.cmake
```diff
@@ -154,10 +154,6 @@ function(cxx_library_with_type name type cxx_flags)
   set_target_properties(${name}
     PROPERTIES
     COMPILE_FLAGS "${cxx_flags}")
-  # Generate debug library name with a postfix.
-  set_target_properties(${name}
-    PROPERTIES
```
Bat3an, line 159: 1.. (:String)../;../..;./

googletest/cmake/internal_utils.cmake
```diff
@@ -154,10 +154,6 @@ function(cxx_library_with_type name type cxx_flags)
   set_target_properties(${name}
     PROPERTIES
     COMPILE_FLAGS "${cxx_flags}")
-  # Generate debug library name with a postfix.
-  set_target_properties(${name}
```
Bat3an, line 158: /\//(N)

googletest/cmake/internal_utils.cmake
```diff
@@ -154,10 +154,6 @@ function(cxx_library_with_type name type cxx_flags)
   set_target_properties(${name}
     PROPERTIES
```
Bat3an, line 155: .:.."".""..(0)../(1..0../(2)../

r00t-:
a similar fixed was now merged as https://github.com/google/googletest/commit/ca3ffd90f2726f6c0c6e790f05c7c88491a16466 :tada:

PR #3454, opened 2021-06-22
Update library.json

Vertexwahn:
Fix indentation

PR #3467, opened 2021-06-28
Remove deprecated file

Vertexwahn:
The PR removes `library.json `. This file is to be considered as deprecated (see [here](https://github.com/google/googletest/pull/3454#issuecomment-869982771)).


derekmauro:
Thanks, but we have already made this change to our internal repo, and it will be reflected on Github shortly.

PR #3470, opened 2021-06-30
Allow empty Type<> type list

### LINKED ISSUE #3469: Allow emtpy ::testing::Types<> list in Typed Tests

No, but it was working in earlier versions of googletest, e.g. in version 1.10.0.

The features is convenient if the `Types<...>` type list is constructed conditionally, e.g.
```cpp
my_types = typename std::conditional<have_types, some_types, Types<>>::type;
```

**Describe the proposal**

Allow the following code to compile
```cpp
TYPED_TEST_SUITE(TypedTest3, Types<>);
TYPED_TEST(TypedTest3, A) {}
```
It will do nothing, i.e. not add a test.

No.

havogt:
Fixes #3469.


Rosemoreira2910 (approved):
Hf

sbenzaquen:
We have had many complaints in the past of cases where the test pass but didn't actually run anything. Eg with a filter that does nothing, or with parameterized tests that were never instantiated.
I don't know if adding yet one more way of having trivial PASS is a good idea.
Can you give more details as to why you want to instantiate with no types?

havogt:
> We have had many complaints in the past of cases where the test pass but didn't actually run anything. Eg with a filter that does nothing, or with parameterized tests that were never instantiated.
> I don't know if adding yet one more way of having trivial PASS is a good idea.

In general I totally agree. For this particular case I would argue the risk for average users is rather low and advanced users should be responsible for the risk they want to take. The average user will typically specify types explicitly like `TYPED_TEST_SUITE(foo, Types<int,char>);`. Writing explicitly `Types<>` is relatively obvious and of course doesn't make any sense. For advanced users, it enables the use of template meta-programming techniques to construct the type list and to avoid macro magic or avoid having to move the logic to the build system.

> Can you give more details as to why you want to instantiate with no types?

My concrete example is roughly like the following. We have a set of programs (tests) that can be executed in different ways (backends). Different programs require different features and not all backends support all features. So the type list is constructed with something like `std::conditional_t<some_test::can_run_on_backend<backend>, Types<SomeType>, Types<>>`.

As mentioned in #3469, empty `Types<>` list where support in earlier versions e.g. 1.10.0 and to me it felt like an oversight in a refactoring as it introduces an arbitrary limitation to generic programming.

Issue #3476, opened 2021-07-06
gcc 11 C++20 compilation warning redundant redeclaration of ‘constexpr’ static data member


Thomas-Barbier-1A:
**Describe the bug**

When compiling Gtest with gcc 11 and C++20 (-std=gnu++20) we have the following compilation warning:
```
include/gmock/internal/gmock-internal-utils.h:504:18: error: redundant redeclaration of ‘constexpr’ static data member ‘testing::internal::Function<R(Args ...)>::ArgumentCount’ [-Werror=deprecated]
  504 | constexpr size_t Function<R(Args...)>::ArgumentCount;
      |                  ^~~~~~~~~~~~~~~~~~~~
/data/mwrep/res/osp/Gmock/21-0-0-0/include/gmock/internal/gmock-internal-utils.h:493:27: note: previous declaration of ‘testing::internal::Function<R(Args ...)>::ArgumentCount’
  493 |   static constexpr size_t ArgumentCount = sizeof...(Args);
```

**Steps to reproduce the bug**

```
git clone -b master https://github.com/google/googletest.git
cd googletest
mkdir _build
cd _build
cmake ${BuildTarget} \
    -Dgtest_build_samples=ON \
    -Dgtest_build_tests=ON \
    -Dgmock_build_tests=ON \
    -DCMAKE_INSTALL_PREFIX=%{buildroot} \
    -DCMAKE_PREFIX_PATH:PATH="%{deps_prefix}" \
    -DBUILD_SHARED_LIBS=ON \
    -DCMAKE_SKIP_RPATH=OFF \
    -DBUILD_GMOCK=ON \
    -DINSTALL_GTEST=ON \
    -G"Unix Makefiles" ..
make 
```

**Does the bug persist in the most recent commit?**

Yes

**What operating system and version are you using?**

Ubuntu 20.04.2 LTS

**What compiler and version are you using?**
g++ (GCC) 11.1.1 20210517

**What build system are you using?**

cmake version 3.16.3

**Additional context**

I think the problem is a pure C++ problem (although I am not sure if it is affect C++20 only, or also older dialect

PR #3477, opened 2021-07-06
Fix redundant redeclaration of ‘constexpr’ static data member

Thomas-Barbier-1A:
Fix issue 3476
‘testing::internal::Function<R(Args ...)>::ArgumentCount'
Issue happening at least with gcc 11.1.1a and -std=gnu++20
include/gmock/internal/gmock-internal-utils.h:504:18: error: redundant redeclaration of ‘constexpr’ static data member ‘testing::internal::Function<R(Args ...)>::ArgumentCount’ [-Werror=deprecated]
  504 | constexpr size_t Function<R(Args...)>::ArgumentCount;
      |                  ^~~~~~~~~~~~~~~~~~~~
/data/mwrep/res/osp/Gmock/21-0-0-0/include/gmock/internal/gmock-internal-utils.h:493:27: note: previous declaration of ‘testing::internal::Function<R(Args ...)>::ArgumentCount’
  493 |   static constexpr size_t ArgumentCount = sizeof...(Args);
https://github.com/google/googletest/issues/3476

### DIFF
--- a/googlemock/include/gmock/internal/gmock-internal-utils.h
+++ b/googlemock/include/gmock/internal/gmock-internal-utils.h
@@ -446,8 +446,10 @@ struct Function<R(Args...)> {
   using MakeResultIgnoredValue = IgnoredValue(Args...);
 };
 
+#if __cplusplus <= 201703L
 template <typename R, typename... Args>
 constexpr size_t Function<R(Args...)>::ArgumentCount;
+#endif
 
 #ifdef _MSC_VER
 # pragma warning(pop)


Thomas-Barbier-1A:
If you prefer I can protect the removal with an ifdef 
```c++
#ifdef __cplusplus <= 
```

derekmauro:
This doesn't build with earlier C++ standards. I think the `#if` might be necessary, but I'm not seeing this warning in GCC 11.1.

Thomas-Barbier-1A:
I have added an ```#ifdef``` so the hcange will not happen for C++ <= 17
I get the warning with gcc (GCC) 11.1.1 20210517 and CXXFLAGS  "-std=gnu++20 -fno-working-directory -ggdb3".
However I get it while using gtest in a project not while compiling gtest itself

asoffer:
Thank you so much for looking in to this. Unfortunately, we are not able to accept this PR. This definition is required in earlier versions of the language that we still support, and the definition is correct even in C++20, and likely to stay in the language for decades. Removing it, even conditionally, is cumbersome and not worth the maintenance burden on our end. If there is a way to suppress -Werror=deprecated diagnostics emanating from GoogleTest, we would suggest that.

Anucha7736zxc:
> Fix issue 3476
> ‘testing::internal::Function<R(Args ...)>::ArgumentCount'
> Issue happening at least with gcc 11.1.1a and -std=gnu++20
> include/gmock/internal/gmock-internal-utils.h:504:18: error: redundant redeclaration of ‘constexpr’ static data member ‘testing::internal::Function<R(Args ...)>::ArgumentCount’ [-Werror=deprecated]
> 504 | constexpr size_t Function<R(Args...)>::ArgumentCount;
> | ^~~~~~~~~~~~~~~~~~~~
> /data/mwrep/res/osp/Gmock/21-0-0-0/include/gmock/internal/gmock-internal-utils.h:493:27: note: previous declaration of ‘testing::internal::Function<R(Args ...)>::ArgumentCount’
> 493 | static constexpr size_t ArgumentCount = sizeof...(Args);
> #3476

Issue #3468, opened 2021-06-29
Compilation error, ‘SleepMilliseconds’ was not declared in this scope


Mr-Leshiy:
**Describe the bug**

I have got a compilation error
```
[ 24%] Linking CXX executable gtest_throw_on_failure_ex_test
/home/leshiy/Project/googletest/googletest/test/googletest-port-test.cc: In member function ‘virtual void testing::internal::GetThreadCountTest_ReturnsCorrectValue_Test::TestBody()’:
/home/leshiy/Project/googletest/googletest/test/googletest-port-test.cc:318:5: error: ‘SleepMilliseconds’ was not declared in this scope
  318 |     SleepMilliseconds(100);
      |     ^~~~~~~~~~~~~~~~~
[ 24%] Built target gtest_throw_on_failure_ex_test
[ 24%] Building CXX object googletest/CMakeFiles/gtest_skip_test.dir/test/gtest_skip_test.cc.o
[ 24%] Linking CXX executable gtest-death-test_ex_nocatch_test
```

**Steps to reproduce the bug**

Build project with the following lines
```sh
mkdir build
cd build
cmake .. -Dgtest_build_tests=ON -Dgtest_build_samples=ON -Dgtest_disable_pthreads=ON -Dgmock_build_tests=ON
make

```

**What build system are you using?**
```
[leshiy@leshiy-g33579 googletest]$ cmake -version
cmake version 3.20.3

CMake suite maintained and supported by Kitware (kitware.com/cmake).
```

Schallbert:
Hi @Mr-Leshiy, I searched for this function in GoogleTest's source code and found this in `googletest/include/gtest/internal/gtest-port.h`:

```c
// Defines synchronization primitives.
#if GTEST_IS_THREADSAFE
# if GTEST_HAS_PTHREAD
// Sleeps for (roughly) n milliseconds.  This function is only for testing
// Google Test's own constructs.  Don't use it in user tests, either
// directly or indirectly.
inline void SleepMilliseconds(int n) {
  const timespec time = {
    0,                  // 0 seconds.
    n * 1000L * 1000L,  // And n ms.
  };
  nanosleep(&time, nullptr);
}
# endif  // GTEST_HAS_PTHREAD
```

It looks like you're trying to use a GTEST-internal function within your test suite. If you just want to make gtest wait, maybe this thread from [stackoverflow](https://stackoverflow.com/questions/46467980/add-delay-to-gtest-test-case) can help.

[...]

PR #3482, opened 2021-07-07
Fix compilation error that occurs when gtest_disable_pthreads is set

hyukmyeong:
As reported here https://github.com/google/googletest/issues/3468  
  
`SleepMilliseconds()` is defined when `#if GTEST_HAS_PTHREAD` is `true`.  
But there is a case where it is called even though `#if GTEST_HAS_PTHREAD` is `false` which causes a compilation error.  
  
This PR fixes that error by adding `#if GTEST_HAS_PTHREAD` where `SleepMilliseconds()` is called without its definition


rioj7:
Why have you removed the text

```
GTEST_HAS_PTHREAD - Define it to 1/0 to indicate that <pthread.h> is/isn't available.
```

The value of the macro is important

hyukmyeong:
@rioj7 
Thanks for letting me know, that was a big mistake  
I updated the patch, could you please review it again?

derekmauro:
`SleepMilliseconds()` has been removed, so this is no longer necessary.

We would like to eventually remove `GTEST_HAS_PTHREAD`.

Issue #3460, opened 2021-06-25
proposal for simple free function mocking


aweisi:
simplify mocking for C functions or c++ free functions with the following approach 

1- add macros to generate Mock types (MOCK_FUNCTION)
2- add runtime support for verification

Pull Request is in preparation

asoffer:
Unfortunately, I do not know of any safe way to mock free-functions. At it's core, mocking requires some sort of indirection to choose between the mock and real implementation. This can be done in with virtual functions, templates (where the indirection happens at compile-time), or even an if-statement. But fundamentally in each case, the function itself needs to be aware it is being mocked (by making it a virtual method, a template, adding an if-statement, etc). This makes free functions not mockable. By their very nature, they do not have an injection point enabling them to dispatch to a different implementation.

A common approach to make something like this sort of work is to provide two implementations and let the build system choose. This once again has an indirection (happening at build time, like with templates). The One Definition Rule in C++ makes this dangerous, and build systems don't generally support this sort of thing very well.

Because of this, the only solutions we've seen to this are either not allowed (undefined behavior) or dangerous parts of the build system. We suggest passing function pointers through and having them point to either the real or the mock implementation. This can be done with templates if runtime performance is a concern.

aweisi:
first thanks for your review and feedback.  The Main use case of approch is when external Code not ready or available for Host platform, therefore Not linked in the Test Case at all. The primary target Environment is embedded Systems aka. bare Metal and run tests on Host System.  only Firmware Code ist used in the Tests No Middleware or HAL.
Nowdays build System Like cmake can easliy used to generate such Tests in scale.

radoslav006:
That's the same use case for me. The mocked part of the code could be not included in the test build or defined as a weak symbol by build system with a macro.
For those who also land here from web and uses gtest for C (or C++ with free functions):
 - If you have to use gtest, add an additional lib that solves this problem  - fake function framework (https://github.com/meekrosoft/fff)
- If you don't have to use gtest, replace it with some decent framework. For C I can recommend Ceedling (https://github.com/ThrowTheSwitch/Ceedling).

PR #3496, opened 2021-07-23
simplify free function mocking

aweisi:
implementation of the proposal #3460


aweisi:
@googlebot I fixed it

PR #3497, opened 2021-07-23
simplify free function mocking

### LINKED ISSUE #3460: proposal for simple free function mocking

aweisi:
resolve #3460 and provide implementation of  proposal for simple free function mocking

### DIFF
--- a/docs/gmock_cook_book.md
+++ b/docs/gmock_cook_book.md
@@ -4299,3 +4299,75 @@ expectations.
 Although `std::function` supports unlimited number of arguments, `MockFunction`
 implementation is limited to ten. If you ever hit that limit... well, your
 callback has bigger problems than being mockable. :-)
+
+# Mocking free functions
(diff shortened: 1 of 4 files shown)


asoffer:
Unfortunately we will not be able to accept this approach. If I'm understanding it correctly, mocking happens by only providing a declaration but no definition for the function to mock. The mock provides the definition, which means the *actual* definition must not be linked into the binary (to avoid One Definition Rule violations). This puts significant constraints on how a build system can build the test targets. Specifically, it means that you cannot know what is in a build target on it's own because someone downstream my want to mock it. This does not work well with Bazel, Google's most common use-case. I suspect it can be made to work with CMake, but my impression is that it would be difficult at best.

PR #3505, opened 2021-07-30
Use ::testing::A::B or testing::A::B depending on the situation

hyukmyeong:
Currently there is a mix of using and not using full qualification for the `testing` namespace.
This PR makes the appropriate method is used depending on the situation.
(References : https://abseil.io/tips/119, https://abseil.io/tips/130)

1. for source code inside the `testing` namespace
`testing::A::B` and `using testing::A::B` (for using-declarations)

2. for source code outside the `testing` namespace, comments, and documents
`::testing::A::B` and `using ::testing::A::B` (for using-declarations)


googlemock/include/gmock/gmock-matchers.h
```diff
@@ -2650,7 +2650,7 @@ class QuantifierMatcherImpl : public MatcherInterface<Container> {
   template <typename InnerMatcher>
   explicit QuantifierMatcherImpl(InnerMatcher inner_matcher)
       : inner_matcher_(
-           testing::SafeMatcherCast<const Element&>(inner_matcher)) {}
+           ::testing::SafeMatcherCast<const Element&>(inner_matcher)) {}
```
dinord, line 2653: To quote https://abseil.io/tips/119:
> 1) Never declare namespace aliases or convenience using-declarations at namespace scope in header files, only in `.cc` files.
> 2) Declare namespace aliases and using-declarations inside the innermost namespace, whether named or anonymous. (Do not add an anonymous namespace just for this purpose.)
> 3) When declaring namespace aliases and using-declarations, use fully qualified names (with leading `::`) unless you are referring to a name inside the current namespace.
> 4) For other uses of names, avoid fully qualifying when reasonable

This would fall under (4), so we would prefer not to fully qualify `SafeMatcherCast` here.

hyukmyeong: Thank you for sharing the tips.
I have modified source code as follows referring the tips.
Please let me know if there are any mistakes.

1. source code (inside the `testing` namespace)
`using testing::xxx`
`testing::xxx`

2. source code (outside the `testing` namespace), comments, and documents
`using ::testing::xxx`
`::testing::xxx`

docs/advanced.md
```diff
@@ -78,11 +78,11 @@ write a predicate function that returns `AssertionResult` instead of `bool`. For
 example, if you define `IsEven()` as:
 
 ```c++
-testing::AssertionResult IsEven(int n) {
+::testing::AssertionResult IsEven(int n) {
```
dinord, line 81: From https://abseil.io/tips/119:
> 1) Never declare namespace aliases or convenience using-declarations at namespace scope in header files, only in `.cc` files.
> 2) Declare namespace aliases and using-declarations inside the innermost namespace, whether named or anonymous. (Do not add an anonymous namespace just for this purpose.)
> 3) When declaring namespace aliases and using-declarations, use fully qualified names (with leading `::`) unless you are referring to a name inside the current namespace.
> 4) For other uses of names, avoid fully qualifying when reasonable

This would be (4), since it's presumably outside `namespace testing`, but neither a namespace alias nor using declaration.
This should apply for the vast majority of the PR that doesn't start with `typedef` or `using`.

hyukmyeong: Thank you for the review!

As you said, for (4), it should be referred to https://abseil.io/tips/130.
And I found that 130 has the followings

> ...
> 
> The current style guide suggests the last option, but allows for the old style (namespaces match package names) if necessary. This is largely because the Google didn’t want to cause too much anxiety or trigger anyone re-namespacing things.
> ...
> 
> Obviously this isn't too bad with std - nobody should ever be adding a nested namespace std - but what about more common namespaces? How about things like testing?

I interpreted it as `::` can be added if a namespace is not guaranteed to be completely unique. And so, I thought `::` should be added if `testing` is used outside its namespace since some projects might already have `testing` in their nested namespaces.

PR #3531, opened 2021-08-19
Simplify example code with c++11

theidexisted:


theidexisted:
@googlebot I signed it!

sbenzaquen:
(392042170)

PR #3620, opened 2021-10-20
Start a simple github actions CI

hzeller:
Running on major platforms.

I'd like to contribute a starting point for a CI that can run on pushes and pull requests, which is in particular useful to make sure that things continue working on all platforms.

Note, the Windows run actually fails right now and it looks like it might actually be worthwhile looking into that. One test looks like a CR to CRLF mismatch. And there is some death test message that doesn't fit. I'll leave that to the googletest team to work out.

### DIFF
--- a/.github/workflows/gtest-ci.yml
+++ b/.github/workflows/gtest-ci.yml
@@ -1,0 +1,40 @@
+name: ci
+
+on:
+  push:
+  pull_request:
+
+jobs:
+  Linux:
+    runs-on: ubuntu-latest
+    steps:
+
+    - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+
+    - name: Tests
+      run: bazel test --test_output=errors //...
+
+  MacOs:
+    runs-on: macos-latest
+    steps:
+
+    - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+
+    - name: Tests
+      run: bazel test --test_output=errors //...
+
+
+  Windows:
+    runs-on: windows-latest
+    steps:
+
+    - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+
+    - name: Tests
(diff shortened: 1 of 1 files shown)


hzeller:
It might be needed to enable actions for this repository before they can run.

dinord:
We've decided to accept this PR, and start using GitHub actions as a best-effort community contribution CI that can be triggered by open source users, and not just Googlers.
For now, we'll still keep using Kokoro CI as the source of truth for breakages.

.github/workflows/gtest-ci.yml
```diff
@@ -0,0 +1,40 @@
+name: ci
+
+on:
+  push:
+  pull_request:
```
danielkrupinski, line 5: `on: [push, pull_request]` is also an option

.github/workflows/gtest-ci.yml
```diff
@@ -0,0 +1,40 @@
...
+  Windows:
+    runs-on: windows-latest
+    steps:
+
+    - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+
+    - name: Tests
+      run: bazel test --test_output=errors //...
```
danielkrupinski, line 40: Duplications can be avoided by using a [strategy matrix](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix):
```
jobs:
  build:
    name: ${{ matrix.name }}
    runs-on: ${{ matrix.os }}
    strategy:
      matrix:
        os: [ubuntu-latest, macos-latest, windows-latest]
        include:
          - os: ubuntu-latest
            name: Linux
          - os: macos-latest
            name: MacOs
          - os: windows-latest
            name: Windows

    steps:
    - uses: actions/checkout@v2
      with:
        fetch-depth: 0
    - name: Tests
      run: bazel test --test_output=errors //...

Issue #2932, opened 2020-07-07
Generated XML test result files are not outputting UTF-8 encoded characters correctly using --gtest-output:xml


MJHylkema:
We have a .cpp google test file, with UTF-8 encoded characters in some of our test names, for example:

`TEST(stringReplace, Can_find_matching_characters_differing_case_Àà)`

When this is executed using the `--gtest-output:xml` flag, it is generating an .xml file which is declared as having UTF8 encoding (`<?xml version=“1.0” encoding=“utf-8”?>`), however those UTF8 characters are each being output as single byte ASCII characters. 

This is causing issues when attempting to consume these files through an Azure Pipeline, using the [PublishTestResults](https://github.com/microsoft/azure-pipelines-tasks/tree/master/Tasks/PublishTestResultsV2) task.

vslashg:
This is something we can't prioritize right now, but we would welcome patches to address it.

Issue #1931, opened 2018-10-24
--gtest_output=xml does not show UTF8 symbols


lietusme:
Running GTEST test on Windows x64:
EXPECT_STREQ("abc", "ąbč");

Using console prints out expectations, while not formatted, at least expected in a way Unicode symbols are usually mangled:

      Expected: "abc"
To be equal to: "─àb─ì"
      Which is: "\xC4\x85" "b\xC4\x8D"

However, this is what is outputed to XML:
      Expected: "abc"
To be equal to: "b"
      Which is: "\xC4\x85" "b\xC4\x8D"

This caused a bit of confusion few times. As one may wonder why GTEST picked up "b" as string to compare and whats up with Unicode symbols being hidden.

Pinpointed cause to be XmlUnitTestResultPrinter::RemoveInvalidXmlCharacters() as it strips away any unfit characters.. And XML could technically support UTF8 characters

gennadiycivil:
Thank you very much for this report. The best way to approach this would be to create a proper PR and submit it for consideration.

PR #3626, opened 2021-10-25
Do not include char with negative value in invalid XML characters

LWisteria:
When `char` is `signed`, depending on compiler implementation, some non-ASCII characters are treated as invalid (e.g. alpha α = 0xCE = 206 = -50)

cf.
* When `char` is `signed` https://wandbox.org/permlink/Lq0kDiowEPZYqXv2
* When `char` is `unsigned` https://wandbox.org/permlink/hu2G0X9ld2YW2Ajd
* https://en.cppreference.com/w/c/language/arithmetic_types#Character_types

Related issue
* #1931
* #2932


derekmauro:
Fixed with https://github.com/google/googletest/commit/6c8a386513b70cb43df997b6406ae5c39f5c9b88 instead.

PR #3635, opened 2021-10-30
Fix memory leak false alarm from windows CRT

RunpingLi:
The thread_to_thread_locals allocated from GetThreadLocalsMapLocked() function will be immortal, however windows CRT will report it as memory leak.
Propose this change to fix the false alarm from windows CRT.


RunpingLi:
@googlebot I signed it!

Issue #3637, opened 2021-10-30
Display Disabled Test Names In Verbose Output


limitedAtonement:
https://github.com/google/googletest/issues/452 is related, but the current proposal does not include the unrelated "Display test fixture" functionality.

**Does the feature exist in the most recent commit?** no.

Ideally, explain why a combination of existing features cannot be used instead.

We recommend using the latest commit from GitHub in your projects.

:heavy_check_mark: living at HEAD as requested

**Why do we need this feature?**

This feature has shown itself to be desired by the community over the years: 
https://bugs.eclipse.org/bugs/show_bug.cgi?id=497892

https://groups.google.com/g/googletestframework/c/jYGXVcrqsX8

It will help testing quality by, instead of hiding disabled tests behind a message like "YOU HAVE 1527 DISABLED TESTS", a larger list showing which tests exactly are disabled. According to the docs (https://github.com/google/googletest/blob/master/docs/advanced.md):

> This feature should only be used for temporary pain-relief. You still have to fix the disabled tests at a later date. As a reminder, googletest will print a banner warning you if a test program contains any disabled tests.

This "reminder" is more effective when it's harder to ignore.

To my knowledge, there is no existing feature or combination of features to show disabled test names.

**Describe the proposal**

Include a detailed description of the feature, with usage examples.

if `./test` has output like this:

```
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from activateBuildings
[ RUN      ] activateBuildings.pinkMarket
[       OK ] activateBuildings.pinkMarket (0 ms)
[----------] 1 test from activateBuildings (0 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test suite ran. (0 ms total)
[  PASSED  ] 1 test.

  YOU HAVE 2 DISABLED TESTS
```

    ./test --gtest_print_disabled_test_names;

would display output like this:

```
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from activateBuildings
[ RUN      ] activateBuildings.pinkMarket
[       OK ] activateBuildings.pinkMarket (0 ms)
[----------] 1 test from activateBuildings (0 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test suite ran. (0 ms total)
[  PASSED  ] 1 test.

  YOU HAVE 2 DISABLED TESTS
[ DISABLED ] activateBuildings.DISABLED_fixedPedlar
[...]

PR #3638, opened 2021-10-30
Re #3637 Show Disabled Tests in testing output

limitedAtonement:
A new flag is herein created, --gtest_print_disabled_test_names,
whose default value is false (so that by default, the behaviour
will not change).

The functionality is implemented by a new member function on the
TestEventListener interface (which is responsible for printing
testing output) which is called when a disabled test is encountered.

The PrettyUnitTestResultPrinter has the canonical implementation of this
added function. The BriefUnitTestResultPrinter and the
EmptyTestEventListener get a nullary implementation. The
JsonUnitTestResultPrinter and XmlUnitTestResultPrinter
inherit that trivial implementation from the EmptyTestEventListener.

### DIFF
--- a/googletest/include/gtest/gtest.h
+++ b/googletest/include/gtest/gtest.h
@@ -1123,6 +1123,9 @@ class TestEventListener {
   // Fired before the test starts.
   virtual void OnTestStart(const TestInfo& test_info) = 0;
 
+  // Fired when skipping a test
+  virtual void OnTestSkipped(const TestInfo& test_info) = 0;
+
   // Fired after a failed assertion or a SUCCEED() invocation.
   // If you want to throw an exception from this function to skip to the next
   // TEST, it must be AssertionException defined above, or inherited from it.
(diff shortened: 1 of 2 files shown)


limitedAtonement:
@googlebot I signed it!

CJ-Johnson:
Thank you for the PR! My apologies for commenting on the associated issue without first seeing this PR.

We would like to adopt this feature without the flag. Could you remove the flag code, such that the `DISABLED_*` tests unconditionally display in this manner?

limitedAtonement:
All done. The current pull request adds disabled test names to the pretty printer output (see the commit message for more information).

limitedAtonement:
To implement the new function, `OnTestSkipped`, I was copying `OnTestStart`, so with these changes, we will have:

image

(which won't be readily visible in the diff unless you get more context.)

I'm happy to make the switch, but `PrintF` and `Flush` are not implemented with the other functions in https://github.com/google/googletest/blob/master/googletest/include/gtest/internal/gtest-port.h#L2091 so I get

```
src/test/googletest/googletest/src/gtest.cc:3503:10: error: ‘PrintF’ is not a member of ‘testing::internal::posix’
```
Perhaps the standard functions are good enough without specializing them for the various other configurations? Let me know what you would like.

CJ-Johnson:
I see! In that case, let's not worry about it. I have imported this change for internal testing, so please do not make any additional modifications as they will be overwritten. (407083802)

vsoltan:
Hey, I was searching for this and saw that these changes are not in the main branch. Was it later decided to skip this feature? 

Thanks!

PR #3641, opened 2021-11-02
EXPECT_EXIT_IF_SUPPORTED/ASSERT_EXIT_IF_SUPPORTED

McCzarny:
Currently I'm missing `*_DEATH_IF_SUPPORTED` equivalents for `*_EXIT` macros. It would be very helpful for codebases used in multiple platforms. In my case the test is failing to build, because iOS has no definition of `EXPECT_EXIT` an its predicate.
It's only a minimal change to get information if it's worth to continue working on this change?

What's missing:
- documentation update
- moving type declaration of the predicate types
- extend `GTEST_UNSUPPORTED_DEATH_TEST` to verify predicate correctness in a similar way it's done for statement and regex.

### DIFF
--- a/googletest/include/gtest/gtest-death-test.h
+++ b/googletest/include/gtest/gtest-death-test.h
@@ -331,10 +331,18 @@ class GTEST_API_ KilledBySignal {
     EXPECT_DEATH(statement, regex)
 # define ASSERT_DEATH_IF_SUPPORTED(statement, regex) \
     ASSERT_DEATH(statement, regex)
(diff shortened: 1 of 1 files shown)


McCzarny:
@googlebot I signed it!

derekmauro:
The short answer here is we think `*_DEATH_IF_SUPPORTED` was a mistake and we'd rather not add new macros that may silently make tests pass. We think it would be better to make the tests get explicitly skipped and print a warning.

Issue #3627, opened 2021-10-25
Consider making "switch" statements cover all cases


anthonygraca:
**Describe the bug**

I was playing around with some compiler flags recommended from the [Airbus Security Lab](https://airbus-seclab.github.io/c-compiler-security/) and it looks like using the ```-Wswitch-enum``` flag triggers some errors/warning in the googletest code base.

I took a look at the [C++ Core Guidelines](https://github.com/isocpp/CppCoreGuidelines/blob/036324/CppCoreGuidelines.md#enum2-use-enumerations-to-represent-sets-of-related-named-constants) and they have a rule under Enum.2 that says developers should enforce the rule by "[flagging instances of] switch-statements where the cases cover most but not all enumerators of an enumeration". Handling all cases is also recommended by the MISRA guidelines as well, which would allow googletest to be more likely used in code bases involving automobile and aerospace applications.

Solution: This is an easy fix. I would add sufficient tests and I would fill in the new cases with a ```[[fallthrough]]``` to the logic handled in the default case. If this issue is approved as valid, I can prepare a PR within a day. A fall-through is desired in this case as described in the [google c++ style guide.](https://google.github.io/styleguide/cppguide.html#Loops_and_Switch_Statements)

**Steps to reproduce the bug**

A developer can reproduce this bug by adding the following line near the top of his/her CMakeLists.txt 
```set(CMAKE_CXX_FLAGS ${CMAKE_CXX_FLAGS} "-Wswitch-enum"}```

Generate makefiles:
```
mkdir mybuild
cd mybuild
cmake -Dgtest_build_tests=ON -Dgmock_build_tests=ON ${GTEST_REPO_DIR}
cmake -DPYTHON_EXECUTABLE=path/to/python ...
```
Calling make should cause warnings too show.
```
make
```

**Alternatively with Bazel**
```bazel test --cxxopt='-Wswitch-enum' --test_output=errors //...```

**Does the bug persist in the most recent commit?**

This bug is found with commit 16f637fbf4ffc3f7a01fa4eceb7906634565242f from October 11 (which is the latest on master as of writing).

**What operating system and version are you using?**

I am using an M1 MacBook Pro running macOS Big Sur Version 11.6

**What compiler and version are you using?**

I am using Clang++ Version 13.0.0

**What build system are you using?**

I am using CMake 3.21.3

**Additional context**

N/A

PR #3643, opened 2021-11-02
Applied C++ Core Guidelines Enum.2 Rule. Problem: switch statements s…

anthonygraca:
…hould cover all cases. Solution: Added missing enumerations in switch statements and added [[fallthrough]] attribute to denote that the default case is the desired case for these missing enumerations.

### DIFF
--- a/googletest/src/gtest.cc
+++ b/googletest/src/gtest.cc
@@ -3254,6 +3254,8 @@ static const char* GetAnsiColorCode(GTestColor color) {
       return "2";
     case GTestColor::kYellow:
       return "3";
+    case GTestColor::kDefault:
+      [[fallthrough]];
     default:
       return nullptr;
   }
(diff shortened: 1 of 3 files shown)


anthonygraca:
#3627

googletest/src/gtest.cc
```diff
@@ -3259,6 +3259,8 @@ static const char* GetAnsiColorCode(GTestColor color) {
       return "2";
     case GTestColor::kYellow:
       return "3";
+    case GTestColor::kDefault:
```
Bat3an, line 3262: String

googletest/test/gtest_unittest.cc
```diff
@@ -146,6 +146,27 @@ TEST_F(StreamingListenerTest, OnTestPartResult) {
...
+  // Meta characters in the failure message should be properly escaped.
+  EXPECT_EQ(
+      "event=TestPartResult&file=foo.cc&line=42&message=failed%3D%0A%26%25\n",
+      *output());
+}
+
+TEST_F(StreamingListenerTest, OnTestPartResultWithSkip) {
+  *output() = "";
+  streamer_.OnTestPartResult(TestPartResult(
+      TestPartResult::kSkip, "foo.cc", 42, "failed=\n&%"));
```
Bat3an, line 163: No

Bat3an (changes requested):

googletest/src/gtest.cc
anthonygraca: Not sure what you are referring to here.

googletest/test/gtest_unittest.cc
anthonygraca: What about this code snippet that makes you say "no"?

PR #3644, opened 2021-11-03
Export Test - Do Not Merge

CJ-Johnson:
Export Test - Do Not Merge

Replace the multiple implementations of Notification with a single
portable implementation.

The also removes the awkward loop with sleep in Notification and will
allow the removal of SleepMilliseconds.


Burberrysmart (commented):
Duplicate of #

PR #3648, opened 2021-11-04
added clang sanitizers as github actions

tbonza:
Had a downstream project and the [Clang MemorySanitizer](https://clang.llvm.org/docs/MemorySanitizer.html) got tripped up. I noticed that the github actions are not currently configured so I'm hoping to start a discussion with this pull request. For example, it make sense to include Bazel unit tests before merging this branch.

### DIFF
--- a/.github/.libcxx-setup.sh
+++ b/.github/.libcxx-setup.sh
@@ -1,0 +1,24 @@
+#!/usr/bin/env bash
+
+# Checkout LLVM sources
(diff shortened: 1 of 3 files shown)


dinord:
We ended up accepting best-effort GitHub Actions CI in https://github.com/google/googletest/pull/3620. Would you mind rebasing and expanding it with clang sanitizers?

tbonza:
I've added the change but I'm unable to rebase due to the amount of time it would take to fix merge conflicts from unrelated upstream commits.

PR #3655, opened 2021-11-09
Export Test - Do Not Merge

dinord:
Export Test - Do Not Merge

Update example for SetUpTestSuite/TearDownTestSuite to use modern C++ standards.

Currently it is using an outdated C++ construct (defining static member variables separately from the declaration).


suhassangangire:
Merge it

Issue #854, opened 2016-08-23
Static build needs -fPIC CXXFLAG with CMake


xor-gate:
With the introduction of the new release 1.8.0 compilation runned fine but the linking failed.

System info:
- Googletest 1.8.0
- Debian 8 AMD64
- GCC 4.9.2 (debian)
- CMake 3.5.1

Error:

```
[ 44%] Linking CXX executable time
/usr/bin/ld: gtest-prefix/src/gtest-build/libgtest.a(gtest-all.cc.o): relocation R_X86_64_32S against `_ZTVN7testing8internal17TestEventRepeaterE' can not be used when making a shared object
; recompile with -fPIC
gtest-prefix/src/gtest-build/libgtest.a: error adding symbols: Bad value
collect2: error: ld returned 1 exit status
tests/CMakeFiles/time.dir/build.make:97: recipe for target 'tests/time' failed
make[3]: *** [tests/time] Error 1
CMakeFiles/Makefile2:1672: recipe for target 'tests/CMakeFiles/time.dir/all' failed
make[2]: *** [tests/CMakeFiles/time.dir/all] Error 2
```

We use CMake [`ExternalProject_Add`](https://cmake.org/cmake/help/v3.2/module/ExternalProject.html#command:externalproject_add)

``` cmake
# gtest.cmake

ExternalProject_Add(
        gtest
        DOWNLOAD_COMMAND ""
        SOURCE_DIR ${PROJECT_SOURCE_DIR}/common/gtest/googletest
        CMAKE_ARGS -DCMAKE_BUILD_TYPE=Release
        INSTALL_COMMAND ""
)
ExternalProject_Get_Property(gtest binary_dir)
ExternalProject_Get_Property(gtest source_dir)
set(gtest_main_library ${binary_dir}/libgtest_main.a)
set(gtest_library ${binary_dir}/libgtest.a)
set(gtest_include_dir "${source_dir}/include")
include_directories(SYSTEM "${gtest_include_dir}")

# application
include("gtest.txt")
add_executable(${test} ${test}.cpp)
add_dependencies(${test} di gtest)
target_link_libraries(${test} ${gtest_library})
target_link_libraries(${test} ${gtest_main_library})
```

I patch the googletest CMakeLists.txt localy with (from [here](https://github.com/BRAINSia/BRAINSTools/blob/master/SuperBuild.cmake#L34-L46)):

```
#-----------------------------------------------------------------------------
# Add needed flag for gnu on linux like enviroments to build static common libs
[...]

PR #3661, opened 2021-11-13
CMake add gtest_enable_pic option

zchrissirhcz:
Hi, googletest developers

This Pull Request add an CMake option `gtest_enable_pic`, which default to `OFF` thus keep same behaviour as before.

The purpose of this PR is to fix issue #854. The original author of #854 provide an global-affecting patch, this PR provide a better implementation, only affect specific targets, thus won't affect other targets.

A practical scenario is: on Ubuntu system, compile googletest with Clang, but compile & linking own project to googletest with GCC, which gives same error as #854 reported. If same scenario, people can then specify `-Dgtest_enable_pic=ON` when building with CMake. Hope this helps.

### DIFF
--- a/googletest/CMakeLists.txt
+++ b/googletest/CMakeLists.txt
@@ -79,29 +79,6 @@ else()
 
 endif()
 
-include(CheckCXXCompilerFlag)
-function(add_flag_or_print_warning flag name)
-  check_cxx_compiler_flag("-Werror ${flag}" "CXX_SUPPORTS_${name}")
-  if (CXX_SUPPORTS_${name})
-    message(STATUS "Building with ${flag}")
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${flag}" PARENT_SCOPE)
-  else()
-    message(WARNING "${flag} is not supported.")
-  endif()
-endfunction()
-
-if (gtest_enable_pic)
-  if (XCODE)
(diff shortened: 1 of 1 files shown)


guruantree:
opsi gtest_enable_pic #3661

derekmauro:
One issue we've found with build options is that they create a combinatoric explosion of build modes to test in, and they end up never getting tested. Therefore we have a pretty high bar for adding new build modes.

Would using `CXXFLAGS="-fPIC"` or something similar solve your problem instead?

zchrissirhcz:
Using `CXXFLAGS="-fPIC"` will globally affect all targets. Consider if googletest is put into some project instead of using its library, the CXXFLAGS will affect other targets, will would not be what people expect it would be.

Lounarok:
@zchrissirhcz I tried to use "cmake .. -Dgtest_enable_pic=ON" but it seems not working. Am I using the wrong syntax?

zchrissirhcz:
> @zchrissirhcz I tried to use "cmake .. -Dgtest_enable_pic=ON" but it seems not working. Am I using the wrong syntax?

The usage is correct. You may do some other checks:
1. Use googletest source code of this branch: https://github.com/zchrissirhcz/googletest/tree/cmake_add_pic_option
2. Clean previous cmake build output, do a clean build.

Lounarok:
@zchrissirhcz thanks for the reply! 
I didn't notice it's not merged into main. I was using main branch to test it and that might be the issue.

PR #3664, opened 2021-11-15
Fix a typo

kuzkry:
@cramertj, I like [your change](https://github.com/google/googletest/commit/1a3e2a265f35e604378156cd675b11ef0a184627) but let's improve it a bit. :)

~Everyone:
I'm not really sure if removal of that comma is correct (English is not my native language), but that's how I think it works. It would be nice if someone could review that.~


cramertj:
This is known as an "Oxford comma." Opinions differ as to when and where it should be used.

kuzkry:
Thanks @cramertj, it's good to know that. I restored the comma as it was.

Issue #3663, opened 2021-11-15
'master' tag missing from github breaking cmake builds


Bam4d:
**Describe the bug**

It looks like the "master" tag has been dropped, so all builds that use cmake to clone the repositories are breaking.

I think its an easy fix but this bug will probably help other people that run into this issue.

```
CMake Error at googletest-download/googletest-prefix/tmp/googletest-gitclone.cmake:40 (message):
  Failed to checkout tag: 'master'
```

**Steps to reproduce the bug**

using the standard cmake builder for googletest:

```
cmake_minimum_required(VERSION 3.10.0)

project(googletest-download NONE)

include(ExternalProject)
ExternalProject_Add(googletest
  GIT_REPOSITORY    https://github.com/google/googletest.git
  GIT_TAG           master
  SOURCE_DIR        "${GOOGLE_TEST_DIR}/googletest-src"
  BINARY_DIR        "${GOOGLE_TEST_DIR}/googletest-build"
  CONFIGURE_COMMAND ""
  BUILD_COMMAND     ""
  INSTALL_COMMAND   ""
  TEST_COMMAND      ""
)

```

As soon as i try to build my project the above cmake error occurs.

**What operating system and version are you using?**

Ubuntu Linux 20.04

**What build system are you using?**

cmake version 3.21.2

**Additional context**

if I change GIT_TAG to 'main' the problem fixes itself for me.

crtrott:
Wow wonder how many repos that broke ...

theShmoo:
Thanks for the fast issue description and proposed fix!

Wanted to report it myself after my CI build broke.

Great community here 👍

muthu-sioux:
Maybe makes sense to update the readme. It still refers to `master` :-)

assafpr:
@Bam4d , Any reason you don't apply the suggested fix? (change GIT_TAG)? or do i miss something as a brand new user?

Bam4d:
@assafpr It's totally fine to just change the tag (I did this in my own projects). I made the bug report because I imagined many people would run into this issue and wanted to show that the fix was quick and easy :)

assafpr:
@Bam4d , I'll try to push it to the origin repository if it is OK with you (it is your fix after all) and if @j-blue-arz is not going to try and push the fix he did in https://github.com/j-blue-arz/labyrinth/commit/8806f073c394a7c0506d08551820b8d49e0c14a4 .

j-blue-arz:
[...]

PR #3669, opened 2021-11-16
fixed branch from master to main in readme.md

assafpr:


assafpr:
this is actually related to issue https://github.com/google/googletest/issues/3663 .  If a user follows the current instructions, it's clone will fail, while following as in the PR does not suppose to fail.

PR #3676, opened 2021-11-19
Silence almost all clang-tidy warnings

OleksandrKvl:
Hi, I got tired `// NOLINT`ing  GoogleTest macros in my work environment (and convincing my colleagues to do so) and decided to fix it. Here's a brief description how it's done.
I used as strict clang-tidy config as possible to fix as many warnings as possible:
```
Checks:          '*,
                  -readability-identifier-naming,
                  -llvmlibc*,
                  -readability-function-cognitive-complexity'
```
Of course I don't use all the macros in my projects so I configured `googletest` with enabled `gtest_build_tests, gmock_build_tests`. I use `vscode` with `clangd-13` so I can see all the warnings on the fly. Then I just went through all the files in `googletest/test` and `googlemock/test` and fixed everything I noticed. However, there are some macros which I can't fix or they somehow don't produce warnings in one places but do so in other ones. Unfixed macros I'm aware of:
`EXPECT_FATAL_FAILURE` and `EXPECT_NONFATAL_FAILURE`. They produce these warnings: `Kernel performance could be improved by unrolling this loop with a '#pragma unroll' directive. clang-tidy(altera-unroll-loops).` and `Calling a function that uses a default argument is disallowed. clang-tidy(fuchsia-default-arguments-calls)`. Both checks are pretty insane and I don't expect them to be widely used. Everything else looks fine. Let me know if you'll notice something else not fixed. I've not squashed commits to keep it more clear for reviewers, of course I can later do it if you wish.


OleksandrKvl:
Signed CLA.

Anucha7736zxc:
Re: [google/googletest] Silence almost all clang-tidy warnings (PR #3676)

ในวันที่ ส. 20 พ.ย. 2021 04:40 น. Oleksandr Koval ***@***.***>
เขียนว่า:

TheStormN (approved):

derekmauro:
I'm against accepting this pull-request as-is. It seems to just use the `// NOLINT` hammer indiscriminately.

What I'd rather see is obviously the issues be fixed if possible. But if that isn't possible, individual pull-requests tackling each issue, explaining what the issue is, and explaining why the issue can't be fixed and needs a `// NOLINT` instead. If there are actual issues, they are just getting hidden by this pull-request.

I also think another option is having clang-tidy exclude warnings coming from "system headers", which usually means files you have no control over. See https://stackoverflow.com/a/47611238 for example.

OleksandrKvl:
> I also think another option is having clang-tidy exclude warnings coming from "system headers", which usually means files you have no control over. See https://stackoverflow.com/a/47611238 for example.

I know about system  headers, it doesn't work here because GoogleTest relies on macros which are expanded in client files, not in system headers and clang-tidy checks are not so smart to detect their real location.

> What I'd rather see is obviously the issues be fixed if possible.

That's the point, there are no real issues, most fixes are related to `goto`, `public/protected` members, this is just the design of the library. Another popular warning is the usage of the default function arguments which comes from the rarely used clang-tidy check (but I'd like to fix as much as possible). They are all from the realm of good coding practices and are not the real issues of the framework. There are ~20 fixed macros, do you propose to create 20+ PRs, each with just a couple of comments and clang-tidy warning which doesn't represent the real issue?

harrism:
There is one clang-tidy check that gets flagged that this PR doesn't address: `Variable name 't' is too short, expected at least 3 charactersclang-tidy(readability-identifier-length)`. This occurs on line 474 of gtest-param-test.h.  Changing the variable name `t` to `args` in lines 474-475 (line numbers from present PR, not original line numbers) fixes it:

```
auto t = std::make_tuple(__VA_ARGS__);                                  \
      static_assert(std::tuple_size<decltype(t)>::value <= 2,                 \
                    "Too Many Args!");
```

Besides that, this PR fixes all of the clang-tidy warnings we have in our project that are caused by gtest, so we would love to see this merged soon. 

@derekmauro would you be happier if explicit clang-tidy checks were used in the NOLINTS? FOr example, `/* NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables) */`

I agree with @OleksandrKvl that some of the causes of the clang-tidy warnings are difficult to fix without major changes to GTest. For example the use of `goto` and non-const global static variables like in `INSTANTIATE_TEST_SUITE_P`. But at least specifying the clang-tidy check that is being disabled is better than blanket application of NOLINT.

harrism:
BTW, I have my own branch where I fixed a lot of clang-tidy warnings. I tried to actually fix the warnings where possible, rather than just NOLINTing. @OleksandrKvl feel free to crib from this diff and bring in the actual fixes (I see missed the t->args change mentioned above though).  

For example, I added a default destructor override to `class GTEST_TEST_CLASS_NAME_` in `TEST_P`: 
```
GTEST_TEST_CLASS_NAME_(test_suite_name, test_name)() override = default;             \
```

https://github.com/google/googletest/compare/main...harrism:fix-clang-tidy-nolint?expand=1

Our list of checks is more constrained, so I didn't need nearly as many changes.

OleksandrKvl:
@harrism thanks for the feedback.

> There is one clang-tidy check we use in our codebase that this PR doesn't address: Variable name 't' is too short, expected at least 3 charactersclang-tidy(readability-identifier-length).

This one is from not yet released `clang-tidy-14`, probably I have to build it by-hand and recheck everything.

> I added a default destructor override to class GTEST_TEST_CLASS_NAME_ in TEST_P:

Somehow, I always thought that googletest targets pre-C++11, using `=default` is definitely better than having empty bodies.

OleksandrKvl:
@derekmauro is there any sense in updating this PR or you are going to fix everything internally? Because I can see that it's already has conflict because one warning has been fixed by the Abseil Team.

AStefan93:
Bump? I would like to see this PR merged as well. It's frustrating to see so many warnings triggered by google test macros.

derekmauro:
As is, this change is not acceptable. I also want to warn that we are not going to be working around every arbitrary warning that some tool issues.

jason-assc:
> As is, this change is not acceptable. I also want to warn that we are not going to be working around every arbitrary warning that some tool issues.

Calling clang-tidy an arbitrary tool is just insanely myopic.

AStefan93:
Is there a way to track and fix this on clang side?
Maybe work on adding a feature or smth to clang to exclude specific macros or have something specific to google-test?

TheStormN:
I can't see why is such a big deal for Google devs to accept PR's like this one. As mentioned clang-tidy is not just an arbitrary tool. How other testing frameworks are friendly with such kind of tools? Like Catch or Doctest, etc, but for Google clang-tidy means nothing?

peakschris:
I'd appreciate if you could reopen this. It is impacting us too.

Issue #3474, opened 2021-07-02
Add verbose on flag for passing assertions


lb0nseu:
**Does the feature exist in the most recent commit?**

This feature exist for gmock with flag --gmock_verbose=LEVEL where level of verbosing can be defined. But there nothing like this for gtest.

**Why do we need this feature?**

In ASPICE standard there is need to provide test report with extended log with full information of values used during tests even if tests passes. 

**Describe the proposal**

--gtest_verbose=LEVEL 

info: gTest will print all informational messages, warnings, and errors (most verbose). At this setting, gTest will also log any calls to the ASSERT_*/EXPECT_* macros. 
warning: gGock will print both warnings and errors (less verbose); This is the default.
error: gTest will print errors only (least verbose).

Output of this verbose should be also added to xml report.

**Is the feature specific to an operating system, compiler, or build system version?**

No

derekmauro:
I think this is a reasonable feature request, but we are unable to prioritize it. We would consider accepting a PR for this feature.

m42e:
I started working on it. You might want to have a look at: https://github.com/m42e/googletest/tree/output-succeeded
Currently it is missing for throw and boolean checks, but works for equals. 

Let me know what you think. And feel free to add more things.

WilliamTambellini:
+1
note: boost test has a log level for long: 
https://www.boost.org/doc/libs/1_60_0/libs/test/doc/html/boost_test/utf_reference/rt_param_reference/log_level.html

NimoTh:
--gmock_verbose exists now. I think we can close this issue.

PR #3678, opened 2021-11-21
Output succeeded

m42e:
This change allows to include succeeded assertions into the output (xml, json, terminal).

It introduces a flag called `--gtest_output_succeed`.

It also adds type to the failures in the json output. Please let me know if this should be done in another way (like a separate array for succeeded ones).

Would be happy if you can give me some feedback.

I think this basically covers what is requested in https://github.com/google/googletest/issues/3474

### DIFF
--- a/googlemock/test/gmock-matchers_test.cc
+++ b/googlemock/test/gmock-matchers_test.cc
@@ -7238,7 +7238,7 @@ TEST_F(PredicateFormatterFromMatcherTest, ShortCircuitOnSuccess) {
   AssertionResult result = RunPredicateFormatter(kInitialSuccess);
   EXPECT_TRUE(result);  // Implicit cast to bool.
   std::string expect;
-  EXPECT_EQ(expect, result.message());
+  EXPECT_NE(expect, result.message());
 }
 
 TEST_F(PredicateFormatterFromMatcherTest, NoShortCircuitOnFailure) {
--- a/googletest/include/gtest/gtest.h
+++ b/googletest/include/gtest/gtest.h
@@ -156,6 +156,9 @@ GTEST_DECLARE_string_(stream_result_to);
 GTEST_DECLARE_string_(flagfile);
 #endif  // GTEST_USE_OWN_FLAGFILE_FLAG_
 
+// This flags control whether Google Test prints succeeded assertions.
+GTEST_DECLARE_bool_(output_succeeded);
+
 namespace testing {
 
 // Silence C4100 (unreferenced formal parameter) and 4805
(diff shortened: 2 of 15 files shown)


googletest/include/gtest/gtest_pred_impl.h
```diff
@@ -242,9 +260,15 @@ AssertionResult AssertPred4Helper(const char* pred_text,
...
-  return AssertionFailure()
+  AssertionResult result = [&]{
+    if (pred(v1, v2, v3, v4)) {
+      return AssertionSuccess();
+    }else {
+      return AssertionFailure();
+    }
+  }();
+
+  return std::move(result)
```
asoffer, line 271: I don't believe there is an r-value reference qualified overload for `operator<<`.

googletest/test/gtest_unittest.cc
```diff
@@ -4284,25 +4284,29 @@ TEST(SuccessfulAssertionTest, SUCCEED) {
 // Tests that Google Test doesn't track successful EXPECT_*.
 TEST(SuccessfulAssertionTest, EXPECT) {
   EXPECT_TRUE(true);
-  EXPECT_EQ(0, GetUnitTestImpl()->current_test_result()->total_part_count());
+  EXPECT_EQ(1, GetUnitTestImpl()->current_test_result()->total_part_count());
```
asoffer, line 4287: We need to test both flag settings.

googlemock/test/gmock-matchers_test.cc
```diff
@@ -7238,7 +7238,7 @@ TEST_F(PredicateFormatterFromMatcherTest, ShortCircuitOnSuccess) {
   AssertionResult result = RunPredicateFormatter(kInitialSuccess);
   EXPECT_TRUE(result);  // Implicit cast to bool.
   std::string expect;
-  EXPECT_EQ(expect, result.message());
+  EXPECT_NE(expect, result.message());
```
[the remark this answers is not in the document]
m42e: As the success message contains a string it no longer equals an empty string.

googletest/include/gtest/gtest_pred_impl.h
m42e: Ok, would you suggest to apply the operators first and `return std::move(...)` afterwards?

googletest/test/gtest_unittest.cc
m42e: The flag will not have an influence on the stored information, only on the output itself, see https://github.com/google/googletest/pull/3678/files/9c2d5c84300a271a4cc503ee606120ce57b52453?authenticity_token=GA9v%2FZzi5u38QObqbF5CCdesUinLL%2B3vls9ZNOti9nV7Osp479Bdved2P357pvWozUiRI8F%2B8zd30q2PjY2dog%3D%3D&file-filters%5B%5D=.cc#diff-bdb2e6c83624322651521250ab8b6249c086965371a73c75ed19f40e82d05b53R3787

googletest/include/gtest/gtest_pred_impl.h
danielkrupinski: `std::move()` shouldn't be used here at all.

googletest/include/gtest/gtest.h
```diff
@@ -1618,11 +1631,9 @@ AssertionResult CmpHelperOpFailure(const char* expr1, const char* expr2,
...
 AssertionResult CmpHelper##op_name(const char* expr1, const char* expr2, \
                                    const T1& val1, const T2& val2) {\
-  if (val1 op val2) {\
-    return AssertionSuccess();\
-  } else {\
-    return CmpHelperOpFailure(expr1, expr2, val1, val2, #op);\
-  }\
+  AssertionResult result = (val1 op val2) ? AssertionSuccess() : AssertionFailure();\
+  result << CmpHelperOpMessage(expr1, expr2, val1, val2, #op);\
+  return std::move(result);\
```
danielkrupinski: Do not use `std::move()` on returned object - https://pvs-studio.com/en/docs/warnings/v828/

Issue #3674, opened 2021-11-17
Enable user-defined templated PrintTo to not be ambiguous


MakersF:
I'm proposing to make the following compile

```c++
namespace user {

struct A {};

template<typename T, typename = std::enable_if_t<...>>
void PrintTo(const T& t, std::ostream* os) { ... }
}

testing::PrintToString(user::A{});
```

which currently fails due to an ambiguous overload with the `PrintTo `defined in gtest.

**Does the feature exist in the most recent commit?**

No

**Why do we need this feature?**

Enable to define `PrintTo` methods for templates, using SFINAE to constrain them on the right type.

Typical case:

Library authors can define a concept (intended as an set of requirements a type must satisfy) that users of the library need to implement.
Given this concept, they might be able to print the type in an effective way.
It would be good to enable the library to define a `PrintTo` that all the types implementing the concept can use.

The possible workarounds are

1. Create a wrapper type which defines the `operator<<`. This cannot be used for `final` types (my current situation). It also requires the user to wrap every object into such wrapper.
2. Define a `PrintTo` for each type. The library author not be able to, if the type is created by the user. For the user, this might be very tiresome if there are many types
3. Use C++20, using a requires clause on the function. Since require clauses are considered more specific than unconstrained templates, this fixes the ambiguous problem. Unfortunately C++20 won't be available in production to many people for quite a while.

See https://stackoverflow.com/questions/25146997/teach-google-test-how-to-print-eigen-matrix for a user struggling with the ambiguity of the template.

Google itself has this problem: protobuf printer is integrated inside google test, I suspect because it's not possible to add it as an external component, since it does concept detection on any type.

**Describe the proposal**

Resolve the template ambiguity.

It can be done by removing the `PrintTo` (from https://github.com/google/googletest/blob/1b18723e874b256c1e39378c6774a90701d70f7a/googletest/include/gtest/gtest-printers.h#L438 ), and change the `UniversalPrinter` to detect whether a valid `PrintTo` function exists, and call that if so, otherwise call the fallback.
Special care needs to be taken because of implicit conversions that would make the `PrintTo` function selected when currently the fallback is used.

I have managed to get this to compile and pass all the existing tests, and I'm currently testing it against a very large code base to ensure it's a backward compatible change.

I'm looking to upstream the change if there is interest.

**Is the feature specific to an operating system, compiler, or build system version?**

No

asoffer:
[...]

PR #3682, opened 2021-11-25
Allow users to provide a templated overload of PrintTo

MakersF:
It's currently not possible to provide an overload of `PrintTo` for all the types supporting a specific concept

```
template<typename T, typename = std::enable_if_t<MatchesConcept<T>>>
void PrintTo(const T& t, std::ostream* os) {
  // print using the concept
}
```

because this is as constrained as the fallback implementation defined in gtest, so the compiler considers them ambiguous.
With this change, gtest doesn't define a fallback implementation of `PrintTo()`, but instead checks if one is callable, and if it doesn't uses the fallback behaviour.

---

This change addresses #3674 and it's largely backward compatible.
These are the situations in which this is not backward compatible

1. When printing type `T`, there is a `PrintTo` for type `Q`, `T` is convertible to `Q`.
Before: fallback behaviours, After: `PrintTo(Q)`

There is a way to be fully backward compatible, but the implementation is more complicated (it involves simulating the old behaviour to detect whether the fallback or PrintTo would have been called), so I'm proposing this change first

### DIFF
--- a/googletest/include/gtest/gtest-printers.h
+++ b/googletest/include/gtest/gtest-printers.h
@@ -420,41 +420,61 @@ std::string FormatForComparisonFailureMessage(
 // We define UniversalPrinter as a class template (as opposed to a
 // function template), as we need to partially specialize it for
 // reference types, which cannot be done with function templates.
+// When PrintTo() is not specialized or overloaded for type T, it
+// prints the given value using the << operator if it has one;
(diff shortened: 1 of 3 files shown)


MakersF:
Closing as this is a simplification and results in currently working code not to compile. I opened another PR which has the fully backward compatible solution

PR #3695, opened 2021-12-05
Enable `PrintTo` template overloads to be defined by the user

MakersF:
Enable PrintTo template overloads to be defined by the user

A user cannot define a template for PrintTo which just takes the type directly.
This is because gtest already defines an implementation for the function, which prints the type with the fallback strategy.

This prevents users from providing a PrintTo function like

```c++
template<typename T, typename = std::enable_if_t<...>>
void PrintTo(const T&, std::ostream*) { ... }
```

as it's ambiguous with the PrintTo template defined by gtest.

This change resolves this.

The current behaviour depends on the overload resolution priority for calling the right PrintTo function.
This change removes the default implementation of PrintTo, and instead detects whether there is a PrintTo function that can be called with the type. If it exists, it calls that, otherwise it uses the fallback behaviour.

Unfortunately it's not that simple: due to implicit conversions, once the PrintTo (unconstrained) template function is removed, many types will implicitly convert and use one of the other overloads of PrintTo defined by gtest, while before it would have used the unconstrained template, as it was a better match.

To address this, this change checks whether the type would ahve called the unconstrained template, if it existed.
In such situations, the fallback behaviour will be used, to remain backward compatible.

To check if the unconstrained template would have been called, such a function is added in a private namespace, and decltype is used to detect which overload would have been called.

---

Addresses #3674 and it's a more comprehensive solution than #3682

### DIFF
--- a/googletest/include/gtest/gtest-printers.h
+++ b/googletest/include/gtest/gtest-printers.h
@@ -420,13 +420,10 @@ std::string FormatForComparisonFailureMessage(
 // We define UniversalPrinter as a class template (as opposed to a
 // function template), as we need to partially specialize it for
 // reference types, which cannot be done with function templates.
-template <typename T>
-class UniversalPrinter;
-
-// Prints the given value using the << operator if it has one;
-// otherwise prints the bytes in it.  This is what
-// UniversalPrinter<T>::Print() does when PrintTo() is not specialized
-// or overloaded for type T.
+//
+// When PrintTo() is not specialized or overloaded for a type T,
+// UniversalPrinter<T>::Print prints the given value using the
+// << operator if it has one; otherwise prints the bytes in it.
 //
 // A user can override this behavior for a class type Foo by defining
 // an overload of PrintTo() in the namespace where Foo is defined.  We
@@ -435,9 +432,7 @@ class UniversalPrinter;
 // or there is already a << operator but it doesn't do what the user
 // wants).
 template <typename T>
-void PrintTo(const T& value, ::std::ostream* os) {
-  internal::PrintWithFallback(value, os);
-}
+class UniversalPrinter;
 
 // The following list of PrintTo() overloads tells
 // UniversalPrinter<T>::Print() how to print standard types (built-in
@@ -450,7 +445,7 @@ inline void PrintTo(char c, ::std::ostream* os) {
   // When printing a plain char, we always treat it as unsigned.  This
   // way, the output won't be affected by whether the compiler thinks
   // char is signed or not.
-  PrintTo(static_cast<unsigned char>(c), os);
+  UniversalPrint(static_cast<unsigned char>(c), os);
 }
 
 // Overloads for other simple built-in types.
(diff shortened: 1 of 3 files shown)


MakersF:
I tested this change against a codebase with 100k+ tests, and everything compiled (1) and there were no regressions in the results.

(1) one test didn't compile because an implicit copy constructor from a different type is not sfinae friendly (it reports that it's convertible but then a static assert fails the compilation)

googletest/include/gtest/gtest-printers.h
```diff
@@ -677,30 +672,107 @@ void PrintTo(const ::std::pair<T1, T2>& value, ::std::ostream* os) {
...
+    ::std::declval<const T&>(), ::std::declval<::std::ostream*>()));
+
+}  // namespace print_to_with_unconstrained_print
+
+// This is going to be:
+//  1. True: when the unconstrained PrintTo would be the preferred one, if it
+//  existed
+//  2. False: when there is a PrintTo that is preferred over the unconstrained
+//  one, or if the call is ambiguous
+template <class T>
```
asoffer, line 710: There's a lot going on here that I think can be simplified.

Rather than using the sentinel and the detector idiom, would it work to

```
template <typename T>
auto CallPrintTo(T&& t, std::ostream* os, int) -> decltype(PrintTo(std::forward<T>(t), os)) {
  return PrintTo(std::forward<T>(t), os);
}
template <typename T>
void CallPrintTo(const T& t, std::ostream* os, char) {
  PrintToFallback(t, os);
}
```
And then call `CallPrintTo(t, os, 0);`?

I might be missing something here entirely, but I believe that would remove a lot of the machinery.

MakersF: I think what you are proposing is similar to #3682. Is that correct?

The problem is implicit conversions. e.g. `int` implicitly converts to `char`, so even if `PrintTo` is only defined for `char`, this will detect it as valid for `int`, and will call that instead of the fallback (which instead is the current behaviour).

If the problem was only for standard types, we could fix it in gtest side by providing all the possible overloads, but the problem exists for any class which is implicitly convertible to another one (and especially derived to base conversions).

This is the behaviour currently implemented in gtest: https://godbolt.org/z/MYzz6foEo
This is with the code you proposed: https://godbolt.org/z/3849svGx6

If we want to maintain backward compatibility, we'd need to check if `PrintTo` is invokable without implicit conversions, and the current (complicated, I agree), machinery is the only way I found.

Do you have some idea of how we could do that check in a simpler way?

MakersF: Any opinion on the above?

The alternative is to offer an extra customization point, which gets called with precedence and allows the user to define a SFINAE template?
I didn't initially propose that because I didn't want to complicate the interface for the user, but if that's preferable we can go down that route

hwilliams265:
@MakersF, @asoffer: what's the status of this PR? I am also running into https://github.com/google/googletest/issues/3674, and this would be a very desirable enhancement to get in.

Issue #3659, opened 2021-11-12
Wrong c++20 feature test macro is used


AntonPilyak:
**Describe the bug**
There is the following code in gtest-printers.h header file:

#ifdef __cpp_char8_t
GTEST_IMPL_FORMAT_C_STRING_AS_STRING_(char8_t, ::std::u8string);    
GTEST_IMPL_FORMAT_C_STRING_AS_STRING_(const char8_t, ::std::u8string);
#endif

However, __cpp_char8_t macro does not guarantee that std::u8string type is available, __cpp_lib_char8_t should be used for this check. As a result, users might get the following error in some cases:

tools/google-test-1.11/googletest/include/gtest/gtest-printers.h:389:55: 
error: no member named 'u8string' in namespace 'std'

**Steps to reproduce the bug**

My setup for reproducing the issue is: RHEL8.4, clang 13, libstd++ 6.0.25 (system default).
I think should be reproducible with GCC GCC 8.1.0, though, not sure about this.

Just try to use gtest in this setup - it'll lead to a build failure

**Does the bug persist in the most recent commit?**

yes

**What operating system and version are you using?**

RHEL8.4

**What compiler and version are you using?**

clang 13 + libstdc++ 6.0.25

**What build system are you using?**

cmake

**Workaround**:
As a temporary workaround, you can just edit the header, replacing cpp_char8_t for __cpp_lib_char8_t in a few places:
#ifdef cpp_char8_t // <---- should be changed to __cpp_lib_char8_t
//...std::u8string usage...
#endif

PR #3703, opened 2021-12-13
Use `__cpp_lib_char8_t` macro instead of `__cpp_char8_t` which does not guarantee to represent UTF-8 code units

joshiayush:
Fixes https://github.com/google/googletest/issues/3659

Since `__cpp_char8_t` does not guarantee to represent UTF-8 code units we should use `__cpp_lib_char8_t` macro which gives library support for `char8_t` code units otherwise use of `__cpp_char8_t` will lead to build failure.

For reference [see](https://en.cppreference.com/w/cpp/feature_test).

### DIFF
--- a/googletest/src/gtest.cc
+++ b/googletest/src/gtest.cc
@@ -5835,9 +5835,7 @@ bool UnitTestImpl::RunAllTests() {
     return true;
   }
 
-  random_seed_ = GTEST_FLAG_GET(shuffle)
-                     ? GetRandomSeedFromFlag(GTEST_FLAG_GET(random_seed))
-                     : 0;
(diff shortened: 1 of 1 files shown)


MikeChan-HK (approved):

MikeChan-HK (approved):

joshiayush:
@asoffer I'm not sure why this PR is not merged yet could you please give me some feedback at least.

joshiayush:
@MikeChan-HK I'm not sure why it failed on windows could you please tell me if there's something wrong in my PR although you've already approved it but if that failing action is an issue I'll try to resolve it as well.

PR #3706, opened 2021-12-16
Annotate headers included directly by gtest.h with IWYU private

kirillbobyrev:
We're implementing the inclusion cleanup tooling in clangd and would like googletest to support some basic IWYU pragmas to facilitate automatic cleanup suggestions.


kirillbobyrev:
Marking more headers would be nice but I don't know if we should mark everything as "private, include `gtest/gtest.h`".

PR #3717, opened 2021-12-30
FIX #3659 -- Use `__cpp_lib_char8_t` macro to support `UTF-8` code units

### LINKED ISSUE #3659: Wrong c++20 feature test macro is used

joshiayush:
Since `__cpp_char8_t` does not guarantee to represent UTF-8 code units we should use `__cpp_lib_char8_t` macro instead to support UTF-8 code units otherwise use of `__cpp_char8_t` will lead to build failure.

Fixes https://github.com/google/googletest/issues/3659

For reference [see](https://en.cppreference.com/w/cpp/feature_test).

### DIFF
--- a/googletest/include/gtest/gtest-printers.h
+++ b/googletest/include/gtest/gtest-printers.h
@@ -383,7 +383,7 @@ GTEST_IMPL_FORMAT_C_STRING_AS_POINTER_(const char32_t);
 
 GTEST_IMPL_FORMAT_C_STRING_AS_STRING_(char, ::std::string);
 GTEST_IMPL_FORMAT_C_STRING_AS_STRING_(const char, ::std::string);
-#ifdef __cpp_char8_t
+#ifdef __cpp_lib_char8_t
 GTEST_IMPL_FORMAT_C_STRING_AS_STRING_(char8_t, ::std::u8string);
 GTEST_IMPL_FORMAT_C_STRING_AS_STRING_(const char8_t, ::std::u8string);
 #endif
@@ -471,7 +471,7 @@ GTEST_API_ void PrintTo(char32_t c, ::std::ostream* os);
 inline void PrintTo(char16_t c, ::std::ostream* os) {
   PrintTo(ImplicitCast_<char32_t>(c), os);
 }
-#ifdef __cpp_char8_t
+#ifdef __cpp_lib_char8_t
 inline void PrintTo(char8_t c, ::std::ostream* os) {
   PrintTo(ImplicitCast_<char32_t>(c), os);
 }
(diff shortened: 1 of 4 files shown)


joshiayush:
@derekmauro Sorry my previous commit https://github.com/google/googletest/pull/3703 was directly on the main branch of my fork and I did not realise that 😅, yes I know I sound stupid.

joshiayush:
@MikeChan-HK Like my previous PR https://github.com/google/googletest/pull/3703 this one again didn't pass Windows, please guide me I'm not familiar with gmock yet. The reason why I asked you is because you approved PR https://github.com/google/googletest/pull/3703 regardless of the failing action.

joshiayush:
@asoffer Hey is this PR even visible from your end or I'm stuck in some kind of parallel earth? I'm not sure but in case you see this don't forget to drop a review from your side down here.

CJ-Johnson:
This is failing on windows, and the reason seems to be that it's inserting `\r` before each `\n`, causing the output to be different from the golden file. Is this change fit for Windows?

joshiayush:
> This is failing on windows, and the reason seems to be that it's inserting `\r` before each `\n`, causing the output to be different from the golden file. Is this change fit for Windows?

Yes, that's how Windows define a line feed. What should be done in this case?

Anucha7736zxc (commented):
<div class="container">
				Copyright Blog of BIGO LIVE@2022. All rights reserved.			</div>

joshiayush:
@dinord

joshiayush:
@dinord

connoranderson:
What happened to this PR and why didn't it get reviewed?

PR #3732, opened 2022-01-13
FIX #2174 -- remove `DEBUG_POSTFIX`

### LINKED ISSUE #2174: In debug build: gtest.pc.in provides -lgtest flag but libgtestd.a is built.

joshiayush:
While this is not synced with the pkg-config file, it only breaks things

Fix #2174

### DIFF
--- a/googletest/cmake/internal_utils.cmake
+++ b/googletest/cmake/internal_utils.cmake
@@ -154,10 +154,6 @@ function(cxx_library_with_type name type cxx_flags)
   set_target_properties(${name}
     PROPERTIES
     COMPILE_FLAGS "${cxx_flags}")
-  # Generate debug library name with a postfix.
-  set_target_properties(${name}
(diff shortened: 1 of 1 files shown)


Adams214565 (approved):

stilllman:
This commit broke the CMake workflow, as described in #3843

B3ret:
> This commit broke the CMake workflow, as described in #3843

Also ran into this.
Until this is fixed, a workaround is to cmake-configure with the CMAKE_DEBUG_PREFIX flag, that will be used for all targets:

`cmake .... -DCMAKE_DEBUG_POSTFIX="d" ...`

PR #3740, opened 2022-01-20
Silence warnings with Clang/Clang-Cl on Windows

Bagira80:
On Windows *Clang* and *Clang-Cl* (which both build against the MSVC ABI) complain about using Microsoft-specific extensions to C++. If warnings are treated as errors this can be problematic:

```
In file included from ...gtest-source/googletest/src/gtest-all.cc:41:
...gtest-source/googletest/src/gtest.cc:2591:3: error: extension used [-Werror,-Wlanguage-extension-token]
  __try {
  ^
1 error generated.
```

This PR silences these warnings and therefore prevents such problems.

### DIFF
--- a/googletest/src/gtest.cc
+++ b/googletest/src/gtest.cc
@@ -2576,6 +2576,10 @@ template <class T, typename Result>
 Result HandleSehExceptionsInMethodIfSupported(
     T* object, Result (T::*method)(), const char* location) {
 #if GTEST_HAS_SEH
+#ifdef __clang__
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wlanguage-extension-token"
+#endif
   __try {
     return (object->*method)();
   } __except (internal::UnitTestOptions::GTestShouldProcessSEH(  // NOLINT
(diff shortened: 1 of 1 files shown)


asoffer:
Different users have different warnings enabled and maintaining compatibility with all such options puts burden that our small team simply cannot sustain. We sometimes try to be helpful for warnings encountered in header files, given that the language does not provide a great mechanism for limiting warnings there, but supporting warning cleanliness in sources files for all possible warnings is unfortunately not something we can reasonably promise.

Bat3an (approved):

Bagira80:
> Different users have different warnings enabled and maintaining compatibility with all such options puts burden that our small team simply cannot sustain. ... but supporting warning cleanliness in sources files for all possible warnings is unfortunately not something we can reasonably promise.

Dear @asoffer, I fully understand that your small team does not have the time to come up with fixes for all possible warnings with all possible compiler combinations.

But you might have mistaken this ticket for an issue/feature request to achieve that, while in fact it is a PR that provides you with a ready fix for one such problem with a specific compiler (Cland-cl).

And as I see, @Bat3an already approves merging it (although he did so after you closed this PR).

Please consider reopening and merging this PR.

Issue #3614, opened 2021-10-14
gtest_filter slow


anholt:
**Describe the bug**

Running va-api's 17,415 gtest tests is slow (~25 minutes on Chrome OS's "grunt" boards), so I'm parallelizing it by sharding to multiple threads running multiple tests per gtest process invocation (https://gitlab.freedesktop.org/anholt/deqp-runner/-/merge_requests/24).  By specifying the tests to run on each shard as a `gtest_filter` separated by ':', we're now spending more CPU in `FilterMatchesTest()` than in the actual tests.  On my desktop, perf says I've got 35% of my CPU spent under main() of my unit tests, and 30% under `FilterMatchesTest()`

If at startup googletest could parse the filter list and toss the non-glob-containing filters into a set, leaving just a vec of glob-containing filters to go through `PatternMatchesString()`, that should be make the filter cheap for the case of "lots of names of specific tests to run"

**Steps to reproduce the bug**
```
git clone from that deqp-runner branch
cargo install --path .

# replace test_va_api here with a handy gtest binary with 17k tests in it
perf record -g gtest-runner run --gtest ~/test_va_api --output <output_dir>
perf report -g
```

**Does the bug persist in the most recent commit?**

Tested on fd873f6d7d22e231a083a7fc570f95961c36e85c, hoping that might help, but no luck (I'm not using globs, so the fixes to globbing didn't do much).

**What operating system and version are you using?**

debian unstable

**What compiler and version are you using?**

`gcc version 11.2.0 (Debian 11.2.0-9) `

**What build system are you using?**

meson "debugoptimized" buildtype (-O2 but with assertions)

anholt:
#3615 did not fix this bug, this bug was about the search for non-pattern-match test names in an array instead of using a better datastructure for those.

dinord:
Thanks for pointing it out. This was closet by a robot based on "Fixes #3614" in the description of #3615.

IYP-Programer-Yeah:
@anholt Would appreciate it if you would also take a look at the new merge request and confirm that it is what your desired solution would be. It would be much better if you could also test it out and provide some feedback.

anholt:
Thanks for the fix!  This is an excellent improvement for Mesa CI (which is, incidentally, using google resources :) )

dinord:
We merged https://github.com/google/googletest/pull/3746 (thanks @IYP-Programer-Yeah). Is `gtest_filter` still too slow?

anholt:
[...]

PR #3746, opened 2022-01-29
Do constant time matching for exact match filters.

IYP-Programer-Yeah:
We can use O(1) look up for patterns that are exact names of the tests. This issue arose from #3614 but I am not sure if it is something that the maintainers would want in the library.

### DIFF
--- a/googletest/src/gtest.cc
+++ b/googletest/src/gtest.cc
@@ -50,6 +50,7 @@
 #include <map>
 #include <ostream>  // NOLINT
 #include <sstream>
+#include <unordered_set>
 #include <vector>
 
 #include "gtest/gtest-assertion-result.h"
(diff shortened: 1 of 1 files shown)


anholt:
Reduces gtest-runner runtime with libva-utils on radeonsi by ~40%.  Now we're spending 95% of CPU time in our actual tests.

googletest/src/gtest.cc
```diff
@@ -734,13 +740,19 @@ class UnitTestFilter {
   // Constructs a filter from a string of patterns separated by `:`.
   explicit UnitTestFilter(const std::string& filter) {
     // By design "" filter matches "" string.
-    SplitString(filter, ':', &patterns_);
+    SplitString(filter, ':', &glob_patterns_);
```
dinord, line 743: Instead of initializing `glob_patterns_` with all patterns and then erasing, could we use a local named `all_patterns` instead and then std::move respective ranges into `glob_patterns_` and `exact_match_patterns_` (the latter maybe using `std::inserter`)?

PR #3793, opened 2022-04-07
Disable new GCC 12 warning.

marxin:
The change disables a new warning in GCC 12:
`warning: infinite recursion detected [-Werror=infinite-recursion]`

similarly to what MSVC does.


asoffer:
Thanks for this PR. Rather than explicitly disable the warning, we've decided to explicitly call compiler builtins to indicate unreachability. The change should land at some point in the next day or so.

marxin:
Works for me, let me then close this pull request.

marxin:
> The change should land at some point in the next day or so.

Can you please provide a link for the change?

dinord:
https://github.com/google/googletest/commit/733f875989a6fba05f4c366ea7b2ba4054994a7e

PR #3807, opened 2022-04-26
Use TEST_TMPDIR (if set) on MacOS

muro:


derekmauro:
Just a note on this one: this is probably a change we want to make, and in fact the is an identical change pending internally (cl/369671161) but it seems to have stalled because we have tests that are relying on some weird quirk about the filesystem on MacOS. It should be possible to make this change, but we need to fix an issue first.

muro:
Thanks - I'll take a look at the internal CL once back, made this change while on vacation an without internal access.

muro:
Sorted the internal change, should now be ready to go in. Maybe that's automatic since the internal change is now submitted.

PR #3833, opened 2022-05-12
Add a manual anchor

balshetzer:
gMock outputs text like:
NOTE: You can safely ignore the above warning unless this call should not happen.  Do not suppress it by blindly adding an EXPECT_CALL() if you don't mean to enforce the call.  See https://github.com/google/googletest/blob/master/docs/gmock_cook_book.md#knowing-when-to-expect for details.

This pull request adds the manual anchor to make that link work.

### DIFF
--- a/docs/gmock_cook_book.md
+++ b/docs/gmock_cook_book.md
@@ -1485,7 +1485,7 @@ mock object and gMock.
 
 ## Setting Expectations
 
-### Knowing When to Expect {#UseOnCall}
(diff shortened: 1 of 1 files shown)


blogric:
Create Like this **[About us Generator Tool Script](https://aswriter.com/about-us-generator-tool-script/)**

ProKashif (approved):

defihook (commented):
WhenTheCodingCodingGIF

PR #3860, opened 2022-05-29
Add wrappers for printf, vprintf and fflush(stdout)

gaspardpetit:
The new wrappers allow a platform port to use something different than stdout, for example using OutputDebugString;

This is a follow up on https://github.com/google/googletest/pull/1201 which was rejected because it relied on macros; in this PR, we instead define `gtest_printf`, `gtest_vprintf` and `gtest_flush_stdout` functions.

### DIFF
--- a/googletest/src/gtest-port.cc
+++ b/googletest/src/gtest-port.cc
@@ -1293,8 +1293,8 @@ bool ParseInt32(const Message& src_text, const char* str, int32_t* value) {
     msg << "WARNING: " << src_text
         << " is expected to be a 32-bit integer, but actually"
         << " has value \"" << str << "\".\n";
-    printf("%s", msg.GetString().c_str());
-    fflush(stdout);
+    gtest_printf("%s", msg.GetString().c_str());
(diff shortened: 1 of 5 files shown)


googletest/src/gtest-port.cc
```diff
@@ -1310,8 +1310,8 @@ bool ParseInt32(const Message& src_text, const char* str, int32_t* value) {
     msg << "WARNING: " << src_text
         << " is expected to be a 32-bit integer, but actually"
         << " has value " << str << ", which overflows.\n";
-    printf("%s", msg.GetString().c_str());
-    fflush(stdout);
+    gtest_printf("%s", msg.GetString().c_str());
+    gtest_flush_stdout();
```
Bat3an, line 1314: ('String')..//: . . ; . <> . {Object}. . . .{}\\. .

Bat3an (approved):

gaspardpetit:
Absolutely no interest for this? :(

defihook (commented):
WhenTheCodingCodingGIF

PR #3861, opened 2022-05-29
Guard functions in gtest-port.h

gaspardpetit:
Guard functions in gtest-port.h so they can be redefined in custom/gtest-port.h

Maybe I am misunderstanding how ports should be made, but it seems like some settings can be set in internal/custom/gtest_port.h` to override what is set in `internal/gtest_port.h` although what can be done there is fairly limited. I observe that different ports keep adding cases in `internal/gtest_port.h` and the complexity of that file keeps increasing.

It seems like adding the guards proposed in this PR will simplify the ports. In my case, I have ports for Playstation 4 and 5 which I cannot submit back because of licensing and these guards allow me to implement almost everything I need exclusively under `internal/custom/gtest_port.h`.

I think this should be pushed further - things like Windows mobile, should be moved to their own `internal/custom/gtest_port-windows_mobile.h` to limit the growth in complexity of the `internal/gtest_port.h` and make it easier to add/remove platforms.

(Arguably, windows mobile could be removed ;-) )

### DIFF
--- a/googletest/include/gtest/internal/gtest-port.h
+++ b/googletest/include/gtest/internal/gtest-port.h
@@ -1956,68 +1956,96 @@ namespace posix {
 
 #if GTEST_OS_WINDOWS
 
+#define GTEST_STATSTRUCT_ 1
 typedef struct _stat StatStruct;
 
 #ifdef __BORLANDC__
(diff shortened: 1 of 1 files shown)


defihook (commented):
WhenTheCodingCodingGIF

PR #3862, opened 2022-05-30
Move getcwd logic to gtest-port.h

gaspardpetit:
Move getcwd logic to gtest-port.h so it can be redefined for each port

The logic in `[gtest-filepath.cc` around obtaining the current directory is heavy on platform checks, and in my case I have to alter this .cc file for some of my ports. Given that port logic is intended to be done in `gtest-port.h` (and the `custom/gtest-port.h`), I am moving the implementation of `GetCwd` function there.

I also moved the separators and `GTEST_PATH_MAX_` over since these are also port specific.

### DIFF
--- a/googletest/include/gtest/internal/gtest-port.h
+++ b/googletest/include/gtest/internal/gtest-port.h
@@ -1879,11 +1879,27 @@ GTEST_API_ size_t GetThreadCount();
 
 #if GTEST_OS_WINDOWS
 #define GTEST_PATH_SEP_ "\\"
-#define GTEST_HAS_ALT_PATH_SEP_ 1
+#define GTEST_PATH_SEPCHAR_ '\\'
+// On Windows, '\\' is the standard path separator, but many tools and the
+// Windows API also accept '/' as an alternate path separator. Unless otherwise
+// noted, a file path can contain either kind of path separators, or a mixture
+// of them.
+#define GTEST_PATH_SEP_ALT_ "/"
+#define GTEST_PATH_SEPCHAR_ALT_ '/'
+#if GTEST_OS_WINDOWS_MOBILE
+// Windows CE doesn't have a current directory. You should not use
+// the current directory in tests on Windows CE, but this at least
+// provides a reasonable fallback.
+#define GTEST_PATH_CWD_ "\\"
+#else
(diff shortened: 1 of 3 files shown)


gaspardpetit:
Appended the PR so that `GetCwd` returns a `std::string`, that way we don't have to deal  with platform specific string copies

gaspardpetit:
I had amended this PR to also revert https://github.com/eidosmontreal/googletest/commit/e5686bb18b52dc29312a7b79c3d6ca2c826c31ed since it resolves the same issue from https://github.com/google/googletest/issues/3089 without introducing a crash when `GTEST_HAS_DEATH_TEST` is `0`

derekmauro (commented):
We are actively moving away from customization points (`custom/gtest-port.h`). Can I ask what you are trying to do (in both this pull request and some of the others you've submitted)? We might be able to point you in a different direction.

gaspardpetit:
Hi @derekmauro 

I have to maintain a couple of platforms that are now officially supported here. Unfortunately I cannot send PR for these because they are covered by license agreements (ex. Sony, Nintendo). With the sum of the following changes
- https://github.com/google/googletest/pull/3861
- https://github.com/google/googletest/pull/3862
- https://github.com/google/googletest/pull/3865
- https://github.com/google/googletest/pull/3866

I am able to define these platforms completely in a `custom/gtest-port.h` file. On my side I started maintaining platform specific `gtest-port.h` - ex. `gtest-port-platformA.h` and my `gtest-port.h` simply finds which platform we are running on and then proceeds to include the right `gtest-port-platformX.h` file.

This setup it pretty clean for me, I can pull from googletest without having to worry about merge conflicts.

It also means that I can share my `custom/gtest-port-platformA.h` file on the private forums with other developers under the platform license agreement.

Prior to these PR, I have the changes in a couple of files and it means everytime googletest changes these functions I have to merge.

May I ask how you plan on replacing the `custom/gtest-port.h` solution? Using `#ifdef` all over the codebase seems like a terrible approach. Over time, some platforms will disappear, so isolating the ports to a single file seemed like a great solution to me.

defihook (commented):
WhenTheCodingCodingGIF

PR #3863, opened 2022-05-30
Fix "unused variable" warning

gaspardpetit:
Fix "unused variable" warning when GTEST_HAS_STREAM_REDIRECTION is set to false in gtest-port.cc

The three guarded constants are used only when `GTEST_HAS_STREAM_REDIRECTION` is enabled and currently trigger the following warnings when `GTEST_HAS_STREAM_REDIRECTION` is disabled:

```
1\googletest\googletest\src\gtest-port.cc(99,11): warning : unused variable 'kStdOutFileno' [-Wunused-const-variable]
1>const int kStdOutFileno = STDOUT_FILENO;
1>          ^
1>\googletest\googletest\src\gtest-port.cc(100,11): warning : unused variable 'kStdErrFileno' [-Wunused-const-variable]
1>const int kStdErrFileno = STDERR_FILENO;
```

### DIFF
--- a/googletest/src/gtest-port.cc
+++ b/googletest/src/gtest-port.cc
@@ -90,6 +90,7 @@
 namespace testing {
 namespace internal {
 
+#if GTEST_HAS_STREAM_REDIRECTION
 #if defined(_MSC_VER) || defined(__BORLANDC__)
 // MSVC and C++Builder do not provide a definition of STDERR_FILENO.
 const int kStdOutFileno = 1;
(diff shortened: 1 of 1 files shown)


googletest/src/gtest-port.cc
```diff
@@ -90,6 +90,7 @@
 namespace testing {
 namespace internal {
 
+#if GTEST_HAS_STREAM_REDIRECTION
```
derekmauro, line 93: Instead of placing the guards here, can these variables simply be moved closer to where they are used, near CaptureStdout()/CaptureStderr(), where they would already be inside the GTEST_HAS_STREAM_REDIRECTION block?

dreamingskey:
I am author to creation and wait to be recognized as creator to this point and christopher paul vincet born march 30 1984  you will not leave me out

googletest/src/gtest-port.cc
gaspardpetit: Absolutely!

PR #3864, opened 2022-05-30
Log test results to logcat on Android

gaspardpetit:
Similar to what is being done on Windows where results are printed using `OutputDebugStringA`, this PR sends the results to the `logcat` logging facility for Android (see https://developer.android.com/studio/command-line/logcat and https://developer.android.com/ndk/reference/group/logging)


ProKashif (approved):

mathewprabakar:
This seems to be very useful when the application using gtest is packaged as an APK.
Any reason why this hasn't been merged for 2 years?

PR #3865, opened 2022-05-30
Move the LocalTime implementation to gtest-port.h so it can be customized for each port;

gaspardpetit:
The logic to convert `time_t` to `tm` is unfortunately quite platform dependant; This PR moves the selection of the right `localtime` function (`localtime`, `localtime_r` or `localtime_s`) to `gtest-port.h` so it is easier to customize on new ports.


defihook (commented):
WhenTheCodingCodingGIF

PR #3879, opened 2022-06-08
Use `-lm` on linux as well.

### LINKED ISSUE #3878: Linker error compiling with clang on linux
When compiling tests with bazel with clang on linux, I get a linker error `undefined reference to nextafter`.

Yes.

Linux.  The specifics don't matter because we're using a custom sysroot, described below.

`clag` version 13.0.0 with `--sysroot` pointing to a sysroot built using `crosstool-ng` based on centos7.

bazel version 5.1.1.

**Additional context**

This is basically the same bug as #3601, but the fix for that one made the incorrect assumption that this was only a problem on BSD.  The correct solution is to change https://github.com/google/googletest/blob/0e402173c97aea7a00749e825b194bfede4f2e45/BUILD.bazel#L123-L125
to be simply
```starlark
    "//conditions:default": ["-lm", "-pthread"], 
```
It won't do any harm when using compilers like gcc which add `-lm` implicitly, but it will prevent breakage for compilers which do not do this.

adam-azarchs:
Not all compilers on linux add it implicitly.
It won't do any harm on those which do.

Closes #3878

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -120,15 +120,10 @@ cc_library(
     linkopts = select({
         ":qnx": ["-lregex"],
         ":windows": [],
-        ":freebsd": [
+        "//conditions:default": [
             "-lm",
             "-pthread",
         ],
-        ":openbsd": [
-            "-lm",
-            "-pthread",
-        ],
-        "//conditions:default": ["-pthread"],
     }),
     deps = select({
         ":has_absl": [


dreamingskey:
I don't want them atached to me

ProKashif (approved):

dangars:
Hi,
There has been progress with this PR :) ?
I confirm that clang requires `-lm`

adam-azarchs:
This is just waiting for someone with permission to actually review and (hopefully) merge it (the test failures shown in the checks output were present on master as well at the time they ran).  Our team has been using this branch for a long time now.

dangars:
Hi,

I think that Bazel 6 allows to do a selection depending on the compiler (gcc or clang), see `@bazel_tools//tools/cpp:compiler`

Maybe `-lm` can be added only for clang, but I think that this would require Bazel 6.

adam-azarchs:
Yes I think that would require bazel 6.  And even after it's been released, it'll be a while before bazel 6 can be assumed.

defihook (commented):
WhenTheCodingCodingGIF

PR #3899, opened 2022-06-16
Provide implicit cast of MockFunction to std::function

jerylvaz:
Add a type cast operator overload to allow implicit conversion of `MockFunction` to `std::function`. This reduces test code verbosity such as from
`Foo(mock_function.AsStdFunction());`
to
`Foo(mock_function);`


asoffer:
Thanks for the PR. I think we want to keep this explicit. I could imagine allowing a static_cast, but I do not believe we want this to be an implicit conversion. Typically we want implicit conversions only in situations where you do not care which value you are working with (the before or after value). What you are able to do with a MockFunction versus a std::function or any other invocable object differs sufficiently that we want this to be explicit.

PR #3901, opened 2022-06-17
Add MACROs for EXPECT and ASSERT

eabreu26:
This is to avoid clashing with user own code for the Expects and ASSERTs of the STRs


nicola-lunghi:
I think this is not valid

defihook (commented):
WhenTheCodingCodingGIF

Issue #3909, opened 2022-06-23
warning: no previous declaration for 'testing::internal::CallReaction testing::internal::intToCallReaction(int)


yesudeep:
**Describe the bug**

I get this warning when trying to use googletest. It's rather benign but I'd like to 
address this warning to improve the signal to noise ratio of the compiler in the console.

```
INFO: From Compiling googlemock/src/gmock-spec-builders.cc:
external/com_google_googletest/googlemock/src/gmock-spec-builders.cc:438:14: warning: no previous declaration for 'testing::internal::CallReaction testing::internal::intToCallReaction(int)' [-Wmissing-declarations]
  438 | CallReaction intToCallReaction(int mock_behavior) {
      |              ^~~~~~~~~~~~~~~~~
```

**Steps to reproduce the bug**

```shell
$ git clone https://github.com/yesudeep/foobar
$ bazel test ...
```

**Does the bug persist in the most recent commit?**

Yes.

**What operating system and version are you using?**

ClearLinux

```shell
$ uname -a
Linux meow 5.18.5-1158.native #1 SMP Thu Jun 16 14:59:00 PDT 2022 x86_64 GNU/Linux
```

**What compiler and version are you using?**

```shell
$ gcc --version
gcc (Clear Linux OS for Intel Architecture) 12.1.1 20220619 releases/gcc-12.1.0-157-ge8df0d960b
Copyright (C) 2022 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

$ clang --version
clang version 14.0.4
Target: x86_64-generic-linux
Thread model: posix
InstalledDir: /usr/bin

```

**What build system are you using?**

```shell
$ bazel --version
bazel 5.2.0
```

**Additional context**

None.

higher-performance:
Should be resolved in 9406a60c7839052e4944ea4dbc8344762a89f9bd.

PR #3910, opened 2022-06-23
Address a missing declaration warning to improve signal

yesudeep:
https://github.com/google/googletest/issues/3909

### DIFF
--- a/googletest/README.md
+++ b/googletest/README.md
@@ -25,7 +25,7 @@ When building GoogleTest as a standalone project, the typical workflow starts
 with
 
 ```
-git clone https://github.com/google/googletest.git -b release-1.11.0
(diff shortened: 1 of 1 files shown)


googlemock/src/gmock-spec-builders.cc
```diff
@@ -68,6 +68,8 @@ namespace internal {
 // mockers, and all expectations.
 GTEST_API_ GTEST_DEFINE_STATIC_MUTEX_(g_gmock_mutex);
 
+CallReaction intToCallReaction(int mock_behavior);
```
derekmauro, line 71: Since this function is only called from this file, it does not need a declaration. It only needs to be declared static of be placed in the anonymous namespace to suppress the warning.

defihook (commented):
WhenTheCodingCodingGIF

PR #3916, opened 2022-06-26
Allow -Werror -Wconversion

### LINKED ISSUE #3915: Unable to compile with -Werror -Wconversion
We are hitting compile errors when building with `-Werror=conversion` since release 1.12.0
```
googletest/googlemock/src/gmock-internal-utils.cc:204:61: error: conversion from ‘int’ to ‘char’ may change value [-Werror=conversion]
googletest/googletest/src/gtest-printers.cc:318:16: error: conversion from ‘int’ to ‘char’ may change value [-Werror=conversion]

```
I will try to propose a fix soon.

asmodai27:
Explicitly convert from int to char where error occurred

Closes #3915


martukas:
Would you not want to amend the build/CI scripts to have `-Werrow -Werror=conversion` to prevent similar errors creeping into the code in the future?

PR #3852, opened 2022-05-23
gtest.cc Allow the host to control the termination of a unit test's lifetime

venik:
The PR switches static pointer to a smart pointer. API change is minimal. It allows us to destroy UnitTest object
earlier (before DLL unload phase). Additionally I fixed the comment for `GetInstance()`. All function static members are thread-safe since c++11. Conditional for Borland C++
was removed because current googletest can’t be compiled with Borland C++.

PR #3928, opened 2022-07-01
Cleanup of gtest.cc and a couple of unittests

venik:
Changes:
1. Unused parameter was removed from `::testing::internal::GetCurrentOsStackTraceExceptTop`. This code was reworked more than a decade ago aaebfcdc4005afb22b68df61b58edd1ccc002913.
2. In a couple of tests return type was replaced with `const auto&`. It's a little bit more convenient because uncouples the implementation of the tests from the return type of `::GetInstance()`.

```
make test
Running tests...
...
100% tests passed, 0 tests failed out of 63

Total Test time (real) =   7.67 sec
```

### DIFF
--- a/googlemock/src/gmock-internal-utils.cc
+++ b/googlemock/src/gmock-internal-utils.cc
@@ -180,8 +180,7 @@ GTEST_API_ void Log(LogSeverity severity, const std::string& message,
       std::cout << "\n";
     }
     std::cout << "Stack trace:\n"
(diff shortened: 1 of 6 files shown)


venik (commented):
This is a partial split from https://github.com/google/googletest/pull/3852. It cleans up code a little bit without the change to the `::GetInstance()` implementation.

Mistrz82 (approved):

PR #3930, opened 2022-07-04
gmock-actions: suppress warning about unreachable code

gjasny:
MSVC warns about unreachable code in `BuiltInDefaultValueGetter::Get`.
The `C4702` warning has been disabled globally in the GoogleTest project itself
but might pop up in downstream code.

### DIFF
--- a/googlemock/include/gmock/gmock-actions.h
+++ b/googlemock/include/gmock/gmock-actions.h
@@ -148,7 +148,7 @@
 
 #ifdef _MSC_VER
 #pragma warning(push)
-#pragma warning(disable : 4100)
+#pragma warning(disable : 4100 4702)
 #endif
 
 namespace testing {


dinord:
Could you please identify the specific codepaths that trigger this warning? We would prefer to fix those instead of disabling it.

The warning is disabled now, but we would prefer to re-enable it if possible.

asoffer:
Thanks for the explanation and sorry for the delayed response. Given that this is already disabled globally for the project, we would prefer not to add additional handling related to this warning. Given that different compilers have different behaviors regarding specification of return types in these circumstances, turning off the warning globally as we have done is the best solution. I realize this isn't ideal if someone reenables the warning of their own project, but I don't think we can manage to generally support all of these sorts of warnings in our headers.

PR #3952, opened 2022-07-21
Allow using re2 without absl dependency

georgthegreat:
googletest 1.12 does not allow using re2 without abseil dependency out of the box.

In our case re2 is the only option to enable regexp support and absl dependency is not favourable.
This PR uses testing::internal::StringView typedef instead of the unconditional dependency on absl::string_view.

### DIFF
--- a/googletest/include/gtest/internal/gtest-port.h
+++ b/googletest/include/gtest/internal/gtest-port.h
@@ -393,8 +393,11 @@ typedef struct _RTL_CRITICAL_SECTION GTEST_CRITICAL_SECTION;
 #if GTEST_HAS_ABSL
 // When using Abseil, RE2 is required.
 #include "absl/strings/string_view.h"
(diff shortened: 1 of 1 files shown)


googletest/include/gtest/internal/gtest-port.h
```diff
@@ -393,8 +393,11 @@ typedef struct _RTL_CRITICAL_SECTION GTEST_CRITICAL_SECTION;
 #if GTEST_HAS_ABSL
 // When using Abseil, RE2 is required.
```
georgthegreat, line 394: I am quite unsure about the logic behind this statement.
It looks like it should be reversed, i. e.

> When using RE2, Abseil is required.

derekmauro:
Using Abseil is going to be a hard dependency soon, so we aren't going to relax this requirement.

PR #3961, opened 2022-07-28
modify some contents in sample2.cc

qiyoumin:
Hi,
1. Copy constructor is used for constructing a new object. So `c_string_ = nullptr`, I don't know the effect of `delete[] c_string_`.
2. I don't know why has `c_string == c_string_` comment, and don't have self-assignment situation?


qiyoumin:
If it needs to self-assign, first determine whether the self-assignment performance will be higher?

ProKashif (approved):

qiyoumin:
I'm sorry, but I want to know why not merge this branch that approved changes. Is there any other arrangement or is there a problem with the operation on my side? Thank for your response.

PR #3980, opened 2022-08-15
Add  Flag for Successful Compilation on GCC < 6.1.0

### LINKED ISSUE #3979: Does Not Compile With GCC < 6.1.0
**Describe the bug**

GCC < 6.1.0 uses C++98 as default C++ mode. Googletest requires C++11. Trying to compile with GCC < 6.1.0 causes a lot of errors/warnings. Adding the `--std=c++11` or `--std=c++14` flag fixes the issue. I suggest the latter, since that is the default behavior of more recent GCC versions.

Attempt to build googletest from the most recent commit using an ancient GCC:

    mkdir build
    cd build
    cmake ..
    make VERBOSE=1

Yes.

    Linux ubuntu 4.4.0-116-generic #140-Ubuntu SMP Mon Feb 12 21:23:47 UTC 2018 aarch64 aarch64 aarch64 GNU/Linux

    gcc (Ubuntu/Linaro 5.5.0-12ubuntu1~16.04) 5.5.0 20171010

I know it's old, but it's what I'm working with.

    cmake version 3.5.1

andrej:
Fixes #3979.

### DIFF
--- a/googletest/cmake/internal_utils.cmake
+++ b/googletest/cmake/internal_utils.cmake
@@ -93,6 +93,11 @@ macro(config_compiler_and_linker)
     set(cxx_no_rtti_flags "-fno-rtti")
   elseif (CMAKE_COMPILER_IS_GNUCXX)
     set(cxx_base_flags "-Wall -Wshadow")
+    # Up until version 6.1.0, GCC's default C++ mode is C++98. After GCC 6.1.0,
+    # C++14 is the default mode. We require at least C++11.
+    if(CMAKE_CXX_COMPILER_VERSION VERSION_LESS 6.1.0)
+      set(cxx_base_flags "--std=c++14")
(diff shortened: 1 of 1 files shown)


derekmauro:
We can't take this for a few reasons.

1) Because of some of the `string_view` hackery, `-std=X` affects the ABI and thus it needs we can't set it for the user.
2) We already have code that sets the minimum `-std=X` (without overriding the user choice) for CMake >= 3.8 so that GoogleTest compiles by default.
3) From https://github.com/google/oss-policies-info/blob/main/foundational-cxx-support-matrix.md, GCC < 7 and CMake < 3.10 are unsupported.

andrej:
Gotcha. I should have found that compatibility matrix myself, sorry about that. Thanks for looking it over anyways.

Issue #2442, opened 2019-09-03
Suppress Clang-Tidy warnings


hanspacket:
When we run clang-tidy on our source code, it finds hundreds of issues related to usages of gtest. We currently have to disable a few clang-tidy checks because of this. In the longer run, we want to enable these checks again. I believe the reported issues are false positives, so it would be beneficial for us if gtest would suppress these warnings.

I found out (among others in https://github.com/google/googletest/issues/853) that gtest already contains 'NOLINT' comments to suppress clang-tidy warnings. I would like to point out a few additional places to add them (based on the code in version 1.8.1):

- clang-tidy warning 'cert-err58-cpp' 
  - gtest_internal.h
    - affects usages of TEST and TEST_F
    - add nolint comment to the assignment on line 1319
  - gtest-param-test.h
    - affects usages of TEST_P and INSTANTIATE_TEST_CASE_P
    - add nolint comment to the assignment on lines 1396/1397
    - add nolint comment to the assignment on line 1421
- clang-tidy warning 'cppcoreguidelines-avoid-goto'
  - gtest-internal.h
    - affects usages of EXPECT_NO_THROW and similar
    - add nolint comment to lines 1231, 1237, 1250, 1268, 1294
  - gtest-death-test-internal.h
    - affects usages of EXPECT_DEATH and similar
    - add nolint comment to lines 196, 204

If I can help out in any other way, let me know.

Zitrax:
Two more I hit:

* `cppcoreguidelines-owning-memory`
  * gtest.h - (initializing non-owner argument of type 'testing::internal::TestFactoryBase *' with a newly created 'gsl::owner<>' )
* `cppcoreguidelines-special-member-functions`
  * gtest-param-test.h - (class 'SleepTest_Sleep_Test' defines a copy constructor and a copy assignment operator but does not define a destructor, a move constructor or a move assignment operator)

andrew-wils:
One more:
- `cppcoreguidelines-avoid-non-const-global-variables`
-- gtest.h: - Variable 'test_info_' provides global access to a non-const object; consider making the pointed-to data 'const'

Trass3r:
>     * clang-tidy warning 'cert-err58-cpp'

Would `noexcept` help with that?

jaques-sam:
@hanspacket 
> If I can help out in any other way, let me know.

You already know where to place the NOLINTs, just go ahead! This has been open for too long 😉

TheStormN:
[...]

PR #3983, opened 2022-08-19
Suppress a clang-tidy false positive

funous:
In some tests we got the following error:

  /usr/local/include/gmock/gmock-spec-builders.h:988:3: note: Potential leak of memory pointed to by field '_M_pi'

This seems to be a false positive, so we simply suppress it.

See #2442.

### DIFF
--- a/googlemock/include/gmock/gmock-spec-builders.h
+++ b/googlemock/include/gmock/gmock-spec-builders.h
@@ -985,7 +985,7 @@ class TypedExpectation<R(Args...)> : public ExpectationBase {
     return WillOnce(Action<F>(ActionAdaptor{
         std::make_shared<OnceAction<F>>(std::move(once_action)),
     }));
-  }
+  }  // NOLINT
 
   // Fallback overload: accept Action<F> objects and those actions that define
   // `operator Action<F>` but not `operator OnceAction<F>`.


googlemock/include/gmock/gmock-spec-builders.h
```diff
@@ -985,7 +985,7 @@ class TypedExpectation<R(Args...)> : public ExpectationBase {
     return WillOnce(Action<F>(ActionAdaptor{
         std::make_shared<OnceAction<F>>(std::move(once_action)),
     }));
-  }
+  }  // NOLINT
```
funous, line 988: It's also possible to avoid the `NOLINT` with the following workaround:
```
     // Use a helper variable to avoid a clang-tidy false positive - see
     // https://github.com/llvm/llvm-project/issues/55219#issuecomment-1117665920
     auto action_ptr = std::make_shared<OnceAction<F>>(std::move(once_action));
     return WillOnce(Action<F>(ActionAdaptor{
       std::move(action_ptr),
     }));
```

Given that this is a know clang-tidy issue, I'm not sure whether this PR should even get merged to googletest. If it should, is there any preference between the explicit suppression and the workaround?

ProKashif: ?

ProKashif (approved):

googlemock/include/gmock/gmock-spec-builders.h
funous: To add context, I'm not sure whether `// NOLINT` comments are the preferred approach to mitigate this in googletest, given https://github.com/google/googletest/pull/3676#issuecomment-1030228060 on a closed PR that would have added a bunch of NOLINT directives. The code I suggest is an alternative way to silence clang-tidy by avoiding the false-positive - it doesn't need the NOLINT when used. It comes from the suggested workaround in the LLVM issue tracker for this false positive.

My question is what is the preference from these options:
1. Use the NOLINT comment (merge this PR as-is)
2. Use the code mentioned in my previous comment, which avoids the false positive without the need for an explicit NOLINT. This avoids the possibly too-wide linter suppression, but adds some complexity.
3. None of the above, perhaps wait for clang-tidy to address the false-positive

Which option is preferred?

asoffer: NOLINT will surpress all warnings which is overly broad. I do not think we should be contorting our code to avoid false positives in a tool. I think the best course of action here is to file a bug with clang-tidy and wait for it to be fixed.

funous: Looking elsewhere in this repo, there's already a bunch of places where `// NOLINT` comments are used. Interestingly, a big part of these comes from the initial commit from 14 years ago. There is a bunch of PRs related to clang-tidy specific suppressions, most of them closed (typically due to inactivity). There is this one recent commit that made it in though: https://github.com/google/googletest/commit/4c5650f68866e3c2e60361d5c4c95c6f335fb64b. The difference I see on the merged one is that it's related to a true positive, rather than a false one.

Do I take it that in general, clang-tidy specific suppressions are accepted, as long as they are related to actual errors, but suppressions for false positives are rejected? I got mixed messages from the PR, as it was actually approved right away, but with question mark comments.

asoffer: We don't have official policy on this, and I'm sure our disposition has changed over time. I realize that's not a super helpful answer, but let me try to explain our current thinking:

GoogleTest is a project that gets included into many others and is therefore compiled and analyzed with nearly every configuration of warnings and clang-tidies. It's simply not possible for us to comply with every such request to avoid warning generation. If a warning uncovers an actual bug, or we can simultaneously improve the readability of our code and avoid the warning, we are willing to do so.

In this case in particular, the warning is incorrect and so adding suppression actually harms readability: Readers now need to understand why the code is correct and that there is some existing tool that is run by others in a configuration we do not directly test which incorrectly diagnoses the code as buggy. In other words, it would exist to circumvent a non-issue we do not test.

asoffer:
I'm going to close this since we do not intend to accept the additional NOLINT directive. If the code can be refactored to improve readability and happens to avoid this particular false positive, we would welcome such a pull request.

PR #3984, opened 2022-08-19
bazel: move -std=c++14 to .bazelrc

keith:
Bazel defaults to -std=c++11 on macOS which gtest doesn't support.


keith:
interested to see from CI if this is preferred over the env var used to test multiple versions

stonigrl (approved):

stonigrl (approved):

PR #3989, opened 2022-08-29
Add support for streaming results on Windows (fixes #3935).

### LINKED ISSUE #3935: Support --gtest_stream_result_to on Windows

No

There is no other built-in streaming support, and currently this feature excludes Windows. This feature provides the best way to communicate results back from a compiled test exe in a programmatic way while the test is running.

**Describe the proposal**

Update
[StreamingListener](https://github.com/google/googletest/blob/86add13493e5c881d7e4ba77fb91c1f57752b3a4/googletest/include/gtest/internal/gtest-port.h#L620)
to include this define on Windows (and make other changes needed for compilation to succeed on Windows; Windows supports sockets but some includes/APIs may be different).

Yes; Windows (but not any specific version).

davidmatson:


ProKashif (approved):

stonigrl (approved):

stonigrl (approved):

davidmatson:
@asoffer - would you mind approving the workflow so it can run? Thanks!

davidmatson:
@asoffer  - I fixed the workflow errors locally; if you wouldn't mind re-approving to run again, that would be great - thanks!

davidmatson:
@asoffer - re-configured to include gmock tests and fixed thoso too; if you wouldn't mind re-approving to run again, that would be great - thanks!

davidmatson:
@asoffer - I'm not sure what's happening with the failing tests. The only one that failed before locally seemed completely unrelated to code that's changing in this PR. I rebased on the latest in upstream main and added a commit that fixes that PR locally, but I'm not sure why that would be needed unless that test is already failing upstream.

There's also one failing in the workflow that isn't failing locally:
```
//googletest/test:gtest_unittest                                         FAILED in 1 out of 2 in 0.3s
...
 C:/users/runneradmin/_bazel_runneradmin/olwzt263/execroot/com_google_googletest/bazel-out/x64_windows-fastbuild/testlogs/googletest/test/gtest_unittest/shard_2_of_2/test.log
```
Any idea what's happening here?

davidmatson:
@asoffer - any thoughts on this PR - thanks!

googletest/src/gtest-internal-inl.h
```diff
@@ -1074,24 +1085,44 @@ class StreamingListener : public EmptyTestEventListener {
...
       MakeConnection();
     }
 
     ~SocketWriter() override {
+#if GTEST_OS_WINDOWS
+      if (sockfd_ != INVALID_SOCKET) CloseConnection();
+#else
       if (sockfd_ != -1) CloseConnection();
+#endif
     }
```
krishnacx, line 1102: Can we have single invocation of CloseConnection after #endif

davidmatson:
@asoffer - could you review this PR? Thanks!

googletest/src/gtest-internal-inl.h
davidmatson: I'll leave it up to the maintainers but I tend to think that separating a braceless body from the condition by preprocessor directives is more worth avoiding.

davidmatson:
It looks like this PR is stuck; abandoning.

PR #3993, opened 2022-08-31
work around GCC 6~11 ADL bug

### LINKED ISSUE #3992: Work around GCC name lookup bug with GCC 6~11 (StreamPrinter::PrintValue)

No.

GCC versions up to (including) 11 have a bug where GCC will sometimes find an operator in the global namespace, when doing ADL for a type defined in a namespace: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=51577

This currently prevents code like this from working with GTest:
```c++
namespace foo {
    struct bar { int x; };

    bool operator == (bar a, bar b) {
        return a.x == b.x;
    }
}

std::ostream& operator <<(std::ostream& os, foo::bar) {
    return os;
}

void fun() {
    EXPECT_EQ(foo::bar{1}, foo::bar{2});
}
```
https://godbolt.org/z/Ken6ha16z

In this example, the `RawBytesPrinter` should be selected because the `operator <<` is not applicable for ADL. However GCC <= 11 will select the `StreamPrinter` but then fail when compiling the actual call of `StreamPrinter::PrintValue`.

We have a large code base and some of our tests failed to compile because of this. Updating to GCC 12 is unfortunately not possible for us at this time. We currently have a local patch with a workaround for this issue and would like to upstream it if possible.

**Describe the proposal**

Change `StreamPrinter::PrintValue` to do the SFINAE check via the return type instead of using a dummy template parameter. For some reason, GCC seems to do the name lookup property there.

GCC versions 6 through 11

pgroke-dt:
see https://gcc.gnu.org/bugzilla/show_bug.cgi?id=51577
ADL seems to work properly when we do the SFINAE check via the return type, but not when using a dummy template parameter

fix #3992


googletest/include/gtest/gtest-printers.h
```diff
@@ -205,12 +205,14 @@ struct StreamPrinter {
...
-            // ADL (possibly involving implicit conversions).
-            typename = decltype(std::declval<std::ostream&>()
-                                << std::declval<const T&>())>
-  static void PrintValue(const T& value, ::std::ostream* os) {
+                !std::is_member_pointer<T>::value>::type>
+  // Only accept types for which we can find a streaming operator via
+  // ADL (possibly involving implicit conversions).
+  // (Use SFINAE via return type, because it seems GCC < 12 doesn't handle name
+  // lookup properly when we do it in the template parameter list.)
+  static decltype(std::declval<std::ostream&>() << std::declval<const T&>(),
```
asoffer, line 213: Nitpick:
Lets write this with trailing retrun type so we have access to the variables and can avoid std::declval:
static auto PrintValue(const T& value, ::std::ostream* os)  -> decltype((void)(*os << value)) { ... }

asoffer:
Generally looks good. Please sign the CLA, and we'll get this merged.

pgroke-dt:
The CLA should be signed, but somehow it still doesn't work. The issue is being discussed here: https://github.com/google/closure-compiler/pull/3988
In the meantime I will update the PR to use trailing return type as you requested.

pgroke-dt:
@asoffer Sorry for taking so long, but the CLA check is finally green now.

pgroke-dt:
@asoffer Hi. I made the changes you requested and the CLA check is green. Is there anything blocking us from moving forward with this?

pgroke-dt:
@asoffer @derekmauro I'd really like to proceed with this. Please let me know if there's something I need to do before this can be merged.

Krzmbrzl:
@asoffer @derekmauro @dinord could someone of you please look into this? This bug is what is keeping me from moving to following HEAD of googletest in my projects as they fail to compile :eyes:

bencsikandrei:
is there any news on this one?

Krzmbrzl:
Unfortunately not and I have the feeling that my pings to @asoffer @derekmauro and @dinord didn't get through :shrug:

PR #4005, opened 2022-09-12
Don't set GTEST_HAS_CXXABI_H_ to 1 if __NuttX__ is defined

xiaoxiang781216:
since NuttX(https://github.com/apache/incubator-nuttx) doesn't support cxxabi

### DIFF
--- a/googletest/include/gtest/internal/gtest-port.h
+++ b/googletest/include/gtest/internal/gtest-port.h
@@ -780,7 +780,7 @@ typedef struct _RTL_CRITICAL_SECTION GTEST_CRITICAL_SECTION;
 
 // _LIBCPP_VERSION is defined by the libc++ library from the LLVM project.
 #if !defined(GTEST_HAS_CXXABI_H_)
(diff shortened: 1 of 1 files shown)


googletest/include/gtest/internal/gtest-port.h
```diff
@@ -780,7 +780,7 @@ typedef struct _RTL_CRITICAL_SECTION GTEST_CRITICAL_SECTION;
 
 // _LIBCPP_VERSION is defined by the libc++ library from the LLVM project.
 #if !defined(GTEST_HAS_CXXABI_H_)
-#if defined(__GLIBCXX__) || (defined(_LIBCPP_VERSION) && !defined(_MSC_VER))
+#if !defined(__NuttX__) && (defined(__GLIBCXX__) || defined(_LIBCPP_VERSION) && !defined(_MSC_VER))
```
derekmauro, line 783: Please restore the extra parenthesis. We don't want to have to think about order of operations.

xiaoxiang781216: Done.

PR #4007, opened 2022-09-12
Add extern "C" to main entry point

xiaoxiang781216:
since the bare metal or RTOS environment need demangled name

### DIFF
--- a/googlemock/src/gmock_main.cc
+++ b/googlemock/src/gmock_main.cc
@@ -60,7 +60,7 @@ void loop() { RUN_ALL_TESTS(); }
 
 GTEST_API_ int _tmain(int argc, TCHAR** argv) {
 #else
-GTEST_API_ int main(int argc, char** argv) {
+extern "C" GTEST_API_ int main(int argc, char** argv) {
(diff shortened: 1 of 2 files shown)


derekmauro (commented):
It is my understanding that `main()` is special and never mangled. Can you elaborate on what is going on here?

xiaoxiang781216:
> It is my understanding that `main()` is special and never mangled. Can you elaborate on what is going on here?

Since all source codes are built into one image in bare metal or RTOS environment, we can't have main entry point more than once. To fix this limitation, one approach is:

1. Add -Dmain=xxx_main when compile source files which contain main entry point
2. Create a table contain all possible entry point like this:
```
{"gtest", gtest_main, ...},
{"gmock", gmock_main, ...},
```
A special shell will lookup this table to find and launch the entry point.
`extern "C"` mark can fix the issue in this case.

nicola-lunghi:
You should simply not link with great main and add your own main function see the examples

xiaoxiang781216:
Yes, this is one of approach.

defihook (commented):
WhenTheCodingCodingGIF

PR #4016, opened 2022-09-21
Port GoogleTest to QuRT (Hexagon RTOS)

Maratyszcza:
This change makes possible to build and use Google Test framework on Hexagon DSPs

### DIFF
--- a/googletest/include/gtest/internal/gtest-port.h
+++ b/googletest/include/gtest/internal/gtest-port.h
@@ -385,7 +385,8 @@ typedef struct _RTL_CRITICAL_SECTION GTEST_CRITICAL_SECTION;
 // On Android, <regex.h> is only available starting with Gingerbread.
 #define GTEST_HAS_POSIX_RE (__ANDROID_API__ >= 9)
 #else
-#define GTEST_HAS_POSIX_RE (!GTEST_OS_WINDOWS && !GTEST_OS_XTENSA)
+#define GTEST_HAS_POSIX_RE \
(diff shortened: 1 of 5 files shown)


googletest/src/gtest_main.cc
```diff
@@ -32,18 +32,26 @@
 #include "gtest/gtest.h"
 
 #if GTEST_OS_ESP8266 || GTEST_OS_ESP32
-#if GTEST_OS_ESP8266
```
derekmauro, line 35: This change seems unrelated to the issue this CL is addressing.

googletest/include/gtest/internal/gtest-port.h
```diff
@@ -385,7 +385,8 @@ typedef struct _RTL_CRITICAL_SECTION GTEST_CRITICAL_SECTION;
 // On Android, <regex.h> is only available starting with Gingerbread.
 #define GTEST_HAS_POSIX_RE (__ANDROID_API__ >= 9)
 #else
-#define GTEST_HAS_POSIX_RE (!GTEST_OS_WINDOWS && !GTEST_OS_XTENSA)
+#define GTEST_HAS_POSIX_RE \
+  !(GTEST_OS_WINDOWS || GTEST_OS_XTENSA || GTEST_OS_QURT || GTEST_OS_QURT)
```
derekmauro, line 389: It looks like GTEST_OS_QURT was added twice here.

derekmauro (changes requested):

googletest/src/gtest_main.cc
Maratyszcza: Reverted

PR #4021, opened 2022-09-27
Fix Clang compiler settings when used with option GTEST_FORCE_SHARED_CRT=ON

mangh:
…CRT=ON

Clang (on Windows) selects MSVC RuntimeLibrary with the following defines:
* _MT for MultiThreaded (static runtime lib),
* _MT and _DLL for MultiThreadedDLL (dynamic runtime lib) See: https://learn.microsoft.com/en-us/visualstudio/msbuild/clangcompile-task?view=vs-2022.

The fix allows Clang to behave on Windows similar to MSVC compiler:
* by default it compiles for MultiThreaded (defines _MT only),
* but it compiles for MultiThreadedDLL (defines both _MT and _DLL) when BUILD_SHARED_LIBS or GTEST_FORCE_SHARED_CRT option is used.

Without the fix Clang always compiles for MultiThreadedDLL (with both _MT and _DLL defined).

### DIFF
--- a/googletest/cmake/internal_utils.cmake
+++ b/googletest/cmake/internal_utils.cmake
@@ -21,8 +21,8 @@ endif (POLICY CMP0054)
 # This must be a macro(), as inside a function string() can only
 # update variables in the function scope.
 macro(fix_default_compiler_settings_)
-  if (MSVC)
-    # For MSVC, CMake sets certain flags to defaults we want to override.
+  if (CMAKE_CXX_COMPILER_ID MATCHES "MSVC|Clang")
+    # For MSVC and Clang, CMake sets certain flags to defaults we want to override.
     # This replacement code is taken from sample in the CMake Wiki at
     # https://gitlab.kitware.com/cmake/community/wikis/FAQ#dynamic-replace.
     foreach (flag_var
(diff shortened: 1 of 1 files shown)


mangh:
I have tried twice to create Google account and enter CLA  - going each time through a multipage application form - and I have failed. I am certainly not going to do it for the 3rd time as it is simply waste of time. Thus it is up to you what to do with my pull request. I really don't care...

mangh:
Having the GoogleTest fixed as above
I run the command below for `MultiThreaded`:
```cmd
cmake -G Ninja -DCMAKE_BUILD_TYPE=Release ..
```
and - for `MultiThreadedDLL` - the command:
```cmd
cmake -G Ninja -DCMAKE_BUILD_TYPE=Release -Dgtest_force_shared_crt=On ..
```
Without the fix, the 1st command behaves the same as the 2nd one: it creates `MultiThreadedDLL` - clang is started with the `-D_DLL` option that should be removed when `BUILD_SHARED_LIBS` and `gtest_force_shared_crt` are NOT requested. See attached log files for more details:
[gtest-clang-md-release.log](https://github.com/google/googletest/files/9703530/gtest-clang-md-release.log)
[gtest-clang-mt-release.log](https://github.com/google/googletest/files/9703531/gtest-clang-mt-release.log)

The other story is when GoogleTest is build from within CMake e.g. via `FetchContent`. Then - as it seems to me now - it is enough to set properly `CMAKE_MSVC_RUNTIME_LIBRARY` and `gtest_force_shared_crt` does NOT influence GoogleTest at all.

mangh:
As you can see I have not used Visual Studio generator to build Googletest. I have just run CMake from the simple command line (NOT the `x64 Native Tools Command Prompt for VS 2022` or alike) having Clang LLVM in the `PATH` and then - after GoogleTest has been built - I entered include and library paths into VS project settings.

PR #4023, opened 2022-09-28
Fix memory leak when NiceMock is used.

igor-epam:


DanielJerolm:
Hi,

we encountered the same problem for StrictMocks and NiceMocks.

Our company-policy requires us to have no so-called draconian memory leaks.
The solution of igor-epam is simple and works for us. Why is it rejected?

PR #4024, opened 2022-09-29
Fixed missing definition of GTEST_NO_TAIL_CALL_ with old clang

seanm:
There was a missing else branch for:

#if __has_attribute(disable_tail_calls)

and so old versions of clang ended up with GTEST_NO_TAIL_CALL_ totally undefined, and various syntax errors ensued.

### DIFF
--- a/googletest/include/gtest/internal/gtest-port.h
+++ b/googletest/include/gtest/internal/gtest-port.h
@@ -770,6 +770,8 @@ typedef struct _RTL_CRITICAL_SECTION GTEST_CRITICAL_SECTION;
 // Ask the compiler not to perform tail call optimization inside
 // the marked function.
 #define GTEST_NO_TAIL_CALL_ __attribute__((disable_tail_calls))
+#else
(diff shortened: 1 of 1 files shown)


seanm:
Sorry, it's too much work to get my employer's legal people involved for such a trivial patch...  Can this CLA thing be overridden for such a trivial patch?

dzenanz:
For trivial (a few lines) contributions, most companies don't require special internal legal approval for a developer to sign a CLA. If you don't want to do it @seanm, I could resubmit the patch.

seanm:
@dzenanz by 'companies' not sure if you are referring to Google or my employer.

If the latter, it looks like way too much work.  Even *seeing* the [form](https://cla.developers.google.com/clas/new?domain=DOMAIN_GOOGLE&kind=KIND_CORPORATE) apparently requires a google account, which requires reading all kinds of other terms...

If the former, let's see if anyone responds, otherwise, yes, please, resubmit it. Thanks.

PR #4025, opened 2022-09-29
Fixed missing definition of GTEST_NO_TAIL_CALL_ with old clang

dzenanz:
There was a missing else branch for:

#if __has_attribute(disable_tail_calls)

and so old versions of clang ended up with GTEST_NO_TAIL_CALL_ totally undefined, and various syntax errors ensued.

Closes #4024.

### DIFF
--- a/googletest/include/gtest/internal/gtest-port.h
+++ b/googletest/include/gtest/internal/gtest-port.h
@@ -770,6 +770,8 @@ typedef struct _RTL_CRITICAL_SECTION GTEST_CRITICAL_SECTION;
 // Ask the compiler not to perform tail call optimization inside
 // the marked function.
 #define GTEST_NO_TAIL_CALL_ __attribute__((disable_tail_calls))
+#else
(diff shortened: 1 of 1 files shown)


ProKashif (approved):
LGTM ✅

derekmauro:
https://github.com/google/oss-policies-info/blob/main/foundational-cxx-support-matrix.md

Since this PR is for an unsupported version of clang (it was added in clang 3.8), it will not be merged into the `1.12.x` branch. If you want to re-target it for the `main` branch I will merge it.

seanm:
> Since this PR is for an unsupported version of clang (it was added in clang 3.8), it will not be merged into the `1.12.x` branch. If you want to re-target it for the `main` branch I will merge it.

Yeah, it's an old clang, but:

- googletest 1.11 built with it, so this is a regression
- the change is very small and very safe
- this is the *only* thing stopping it from building with that old clang
- the `main` branch doesn't have this code, so is not affected

PR #4028, opened 2022-10-02
Enforce printability of all types (optionally)

### LINKED ISSUE #4027: Enforce printability of all types (optionally)

For failed assertions, GoogleTest tries to print the arguments if it can, falling back to printing raw bytes if not. This is convenient for writing tests, but if you have a policy that requires all types to be printable, it's very hard to ensure this. You basically have to manually make each assertion fail, inspecting the test output to see whether raw bytes are printed.

Furthermore, even if you manually ensured this once, it can easily break thereafter, for example when code is moved around in refactorings. Nobody will notice until the next time a test fails, and only prints raw bytes.

And finally, this can lead to ODR issues that are extremely hard to figure out. For example, if you have your printers in separate header files (because you don't want them to be visible to production code), then forgetting to include this header in one test file can break printing for other test files too, even if those do include the printer header.

For all these reasons it may be desirable to configure gtest to not have a raw bytes fallback printer, so that compilation fails if a type used in an assertion is not printable. This must be optional of course.
 
**Describe the proposal**

One way to implement the feature is to provide an optional define, say GTEST_FALLBACK_PRINTER; it can be defined to a function that will be called when no other printer could be found. Defining this to a function that doesn't exist achieves the desired effect of causing a compilation error.

The Catch2 test framework has a [very similar feature](https://github.com/catchorg/Catch2/blob/devel/src/catch2/catch_tostring.hpp#L131) that can be used in exactly the same way.

No.

stefanhaller:
Introduce a new customization point GTEST_FALLBACK_PRINTER which can be defined to a function that is used as a fallback for types that are not printable, instead of the default raw bytes printer. One especially useful way to use this is to define it to a symbol that doesn't exist (e.g. "#define GTEST_FALLBACK_PRINTER MISSING_PRINTER") to enforce printability of all types that are used in gtest checks at compile time. See #4027 for why this can be useful.

The first commit of the branch, "Add ScopedEnumPrinter", is somewhat independent and could be broken out into a PR of its own. Let me know if you'd like me to do that. I'm including it here because it becomes more important when you enforce printabilty at compile time, because in that case it fixes compilation errors for scoped enums, rather than just printing them as raw bytes, which is not quite as severe.

Fixes #4027.

### DIFF
--- a/googletest/include/gtest/gtest-printers.h
+++ b/googletest/include/gtest/gtest-printers.h
@@ -250,6 +250,16 @@ struct ConvertibleToIntegerPrinter {
   }
 };
 
+struct ScopedEnumPrinter {
+  // Print scoped enums. These are not implicitly convertible to int, so they
+  // are not covered by ConvertibleToIntegerPrinter above.
+  template <typename T,
+            typename = typename std::enable_if<std::is_enum<T>::value>::type>
+  static void PrintValue(T value, ::std::ostream* os) {
+    *os << static_cast<internal::BiggestInt>(value);
+  }
+};
+
 struct ConvertibleToStringViewPrinter {
 #if GTEST_INTERNAL_HAS_STRING_VIEW
   static void PrintValue(internal::StringView value, ::std::ostream* os) {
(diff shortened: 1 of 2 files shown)


asoffer:
Closing. See comments in #4027

PR #4043, opened 2022-10-28
Check code formatting with GitHub Action

shenxianpeng:
I see that there is a .clang-format configuration file under the code repository, and the[ contribution documentation](https://github.com/google/googletest/blob/main/CONTRIBUTING.md#style) reminds developers that they need to use clang-format to check to format. 

So it would help googletest to have a CI to check the code formatting of the pull request every time.

### DIFF
--- a/.github/workflows/cpp-linter.yml
+++ b/.github/workflows/cpp-linter.yml
@@ -1,0 +1,20 @@
+name: cpp-linter
+
+on: pull_request
+
+jobs:
+  cpp-linter:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


dinord:
Thank your for the proposal. Unfortunately, we do not have the capacity to maintain this presubmit, especially since we wouldn't use it ourselves.

We've accepted GitHub Actions presubmits that run tests due to very high popular demand, but without a promise to support them. We still need to review changes and triage issues around them, which takes some of our time. But we are not ready to accept other types of presubmits at this time.

Issue #2930, opened 2020-07-07
Question: Colored output on Windows powershell with CTest


shrgabor:
I'm using CMake and CTest to build and run my tests. 

On Linux I can produce color googletest output with CTest, like this:
`GTEST_COLOR=1 ctest -V`

On Windows I tried something similar in powershell:
`$Env:GTEST_COLOR=1; ctest -V`

**However the output is still not colored.**

I'm using Googletest 1.10.0 and CMake 3.17.3

derekmauro:
I don't think any of us use powershell. If someone can figure this out, we might accept a patch.

tanzislam:
@shrgabor Which compiler did you use to build Google Test? If it's using MinGW then #3049 should help.

Virendra2406087:
I was able to reproduce this issue on the latest versions.

### Environment
- OS: Windows 11
- Visual Studio Build Tools 2026
- CMake 4.4.0
- GoogleTest: latest `main` branch

### Reproduction

Running the test executable directly:

```powershell
gtest_unittest.exe --gtest_color=yes
```

produces colored output.

Running through CTest:

```powershell
ctest -C Debug -V
```

produces plain (non-colored) output, even after setting:

```powershell
$Env:GTEST_COLOR="yes"
```

### Investigation

I added some temporary debugging inside `ColoredPrintf()`.

When the test is run via `ctest`, I observed:

```text
GetConsoleMode = 0
mode = 0
```

This suggests that `stdout` is not attached to a Windows console when launched by CTest, so `SetConsoleTextAttribute()` cannot apply console colors.

This suggests the behavior may be related to how CTest launches or captures the test process on Windows. I'm sharing these observations in case they help identify the root cause.

Tiancheng-Xu:
I would like to take this on. I will first add a Windows/CTest regression that captures the current behavior, then make the narrowest fallback: when color is explicitly forced (`GTEST_COLOR=yes` or `--gtest_color=yes`) but Win32 console coloring is unavailable because stdout is captured, emit ANSI color sequences instead of silently disabling color. This intentionally differs from the closed #4050 approach: no new public `ansi` option, and the existing auto/no-color behavior remains unchanged. I will validate it on Windows CI and keep the patch scoped to this issue.

PR #4050, opened 2022-11-03
Add "ansi" to supported gtest_color values

sergio-nsk:
Fixes uncolored output on Windows PowerShell with CTest (#2930)

Fixes uncolored output on GitHub and other Windows terminals.

CTest writes googlest standard output and error streams to standard output and a file. Windows `_isatty()` returns 0 in this case and `ShouldUseColor()` returns `false`. Windows PowerShell supports ANSI escape sequences.

Reviewed-by: Nathan Moinvaziri <nathan@nathanm.com>
Reviewed-by: steve-tucker <steve-tucker@users.noreply.github.com>

### DIFF
--- a/googletest/src/gtest-internal-inl.h
+++ b/googletest/src/gtest-internal-inl.h
@@ -75,6 +75,8 @@ GTEST_DECLARE_bool_(death_test_use_fork);
 namespace testing {
 namespace internal {
 
+enum class GTestColorMode { kNo, kYes, kAnsi };
+
 // The value of GetTestTypeId() as seen from within the Google Test
 // library.  This is solely for testing GetTestTypeId().
 GTEST_API_ extern const TypeId kTestTypeIdInGoogleTest;
@@ -89,8 +91,9 @@ GTEST_API_ extern bool g_help_flag;
 // Returns the current time in milliseconds.
 GTEST_API_ TimeInMillis GetTimeInMillis();
 
-// Returns true if and only if Google Test should use colors in the output.
-GTEST_API_ bool ShouldUseColor(bool stdout_is_tty);
+// Returns the color mode kNo if and only if Google Test should not use colors
+// in the output.
+GTEST_API_ GTestColorMode ShouldUseColor(bool stdout_is_tty);
 
 // Formats the given time in milliseconds as seconds. If the input is an exact N
 // seconds, the output has a trailing decimal point (e.g., "N." instead of "N").
--- a/googletest/src/gtest.cc
+++ b/googletest/src/gtest.cc
@@ -3215,7 +3215,7 @@ static WORD GetNewColor(GTestColor color, WORD old_color_attrs) {
   return new_color;
 }
 
-#else
+#endif  // GTEST_OS_WINDOWS && !GTEST_OS_WINDOWS_MOBILE
 
 // Returns the ANSI color code for the given color. GTestColor::kDefault is
 // an invalid input.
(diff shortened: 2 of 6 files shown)


sergio-nsk:
GitHub log before with `GTEST_COLOR: 1`

image

GitHub log after with `GTEST_COLOR: ansi`

image

mentornuh (approved):

dinord:
Instead of adding an API for forcing ANSI, could we detect whether the output terminal supports it or not?
For example, by updating the ifndef around it, checking some environment variable, etc.

sergio-nsk:
> Instead of adding an API for forcing ANSI, could we detect whether the output terminal supports it or not? For example, by updating the ifndef around it, checking some environment variable, etc.

I doubt it's possible on Windows.

If it does not run on Windows, it uses `$TERM` to detect if the output terminal supports ANSI colors. It works even if the output stream is redirected to a file or a pipe like `| tee` does it.

If it runs on Windows, it is able to detect if the output terminal supports colors. But it does not work if the output stream is redirected to a file or a pipe, what happens if `ctest`, GitHub runners or `| tee`  are used.

googletest/src/gtest-internal-inl.h
```diff
@@ -74,6 +74,8 @@ GTEST_DECLARE_bool_(death_test_use_fork);
 namespace testing {
 namespace internal {
 
+enum class GTestColorMode { kNo, kYes, kAnsi };
```
dinord, line 77: I'm worried about the difference between `yes` and `ansi`. Both of these are used to turn on colors, but in different ways. If we were to design GoogleTest from scratch, we'd probably want to provide a flag with possible values such as `ansi`, `no`, `auto`, and `not_ansi` (probably something more informative for the last one).

Instead of expanding the existing flag, could we add another one for using ANSI color escape sequences? I'm okay with it being a force flag (on/off) or a feature flag (on/off/auto).

sergio-nsk: The enum is made with the logic 0 - coloring is disabled, !0 - coloring is enabled, 1 or 2 - it's a detail how it is enabled.

sergio-nsk: I could make it as another flag `--gtest_color_ansi` and keep `--gtest_color` unchanged. but did not want to add yet another environment variable and to make a larger changeset.

dinord: I'd recommend adding another flag, e.g. something like `--gtest_use_ansi_color` and keep the existing one unchanged.
Most users shouldn't care about the new flag, but by providing it, we leave an escape hatch for those that fail to auto-detect ANSI colors.

tanzislam:
I believe #3049 fixes this problem without introducing any options.

sergio-nsk:
> I believe #3049 fixes this problem without introducing any options.

Unfortunately it does not fix the coloring issues.
* `gtest_host.exe --gtest_help --gtest_color=yes > out.txt`
* `gtest_host.exe --gtest_help --gtest_color=yes | tee out.txt`
*  Git hub test run logs

Neither above is able to color the output. Running them in Windows PowerShell additionally produces UTF-16 uncolored outputs. cmd.exe. #3049 is showing the working redirection `>test.txt`, but I can't confirm it.

tanzislam:
@sergio-nsk I've now rebased #3049 on the latest `main` branch; could you please try again? When built with MinGW's GCC, if `--gtest_color=yes` is supplied (or `GTEST_COLOR=yes` is exported) it should emit colors using ANSI escape sequences when stdout isn't a TTY (e.g. when redirected or piped).

Which compiler are you using on Windows? My PR has the logic only for MinGW toolchain, although I contemplated generalizing it to all Windows compilers.

I've also confirmed the color output (using ANSI escape sequences) working in a PowerShell window with both `> out.txt` and `| tee out.txt` redirection. However, the text file does end up getting UTF-16LE encoding under PowerShell, even if colors weren't used -- this seems to be a peculiarity of the PowerShell environment, and [as per this page](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_character_encoding?view=powershell-7.2#changing-the-default-encoding) I was able to fix this by running `$PSDefaultParameterValues['Out-File:Encoding'] = 'utf8'`.

googletest/src/gtest.cc
```diff
@@ -3222,7 +3222,7 @@ static WORD GetNewColor(GTestColor color, WORD old_color_attrs) {
   return new_color;
```
captain-hoity-toity, line 3222: Was the "random" key on your keyboard stuck when naming this function?

mentornuh: That's wasn't necessary. I choose not to used random

Issue #3529, opened 2021-08-17
Support custom user-defined logic for tests filtering 


victordk13:
**Does the feature exist in the most recent commit?**

Currently googletest allows to filter tests with `--gtest_filter=<positive mask:...>-<negative_mask:...>` based on suite/test names.

**Why do we need this feature?**

If there are a lot of tests and different testing scenarios, it may become cumbersome to filter tests based on name masks alone.
Suppose, you categorized your tests based on speed and added postfix to test names like `TEST(suite, test_SLOW)`.
Then you decided to mark flaky tests: `TEST(suite, test_SLOW_FLAKY)`.
You may now run slow flaky tests with e.g. `--gtest_filter=*_SLOW*_FLAKY*:*_FLAKY*_SLOW*`, but as I said, it's not easily expandable and will become cumbersome fast (two categories are manageable via postfixes, but not many more).

**Describe the proposal**

I propose supporting a custom user-defined logic for tests filtering. It can be implemented as easily as adding a callback in [UnitTestImpl::FilterTests()](https://github.com/google/googletest/blob/0134d73a4902574269ff2e42827f7573d3df08ae/googletest/src/gtest.cc#L6097)
and providing means to pass this callback to gtest.

Users can implement assigning, collecting, processing of additional test info (categories, tags, etc.) however they like (e.g. with custom macros that link this info to the tests). The only obstacle standing in their way is that now they cannot filter tests according to their own logic (concretely, modify private fields `TestInfo::matches_filter_` and/or `TestInfo::should_run_` of each test - and the best moment to adjust these values is in `UnitTestImpl::FilterTests()` itself).

sbenzaquen:
GoogleTest supports skipping tests with `GTEST_SKIP();`.
You can do so from the test, or from the fixture (eg in `SetUp`).
Example: https://gcc.godbolt.org/z/sPhTxnn5G

It is not exactly the same as filters because the test will show in the output as skipped, though.
But this is something you can do today, and you can use any runtime logic you want to skip the tests.

Filtering can't really be that dynamic because it is used in several parts of the framework and it is expected to be deterministic.

leonardopsantos:
I think such feature would be a great addition because `--gtest_filter=...` and markers (in th e Pytest sense) aren't exactly the same. I understand that by adding custom filters like @sbenzaquen suggested probably covers most cases, I agree with @victordk13 that manually managing many categories quickly gets cumbersome.

asoffer:
[...]

PR #4069, opened 2022-11-22
Add custom filter support

zloylos:
https://github.com/google/googletest/issues/3529

### DIFF
--- a/googletest/test/gtest_custom_filter.cc
+++ b/googletest/test/gtest_custom_filter.cc
@@ -1,0 +1,46 @@
+// Copyright 2022 Google Inc.
+// All Rights Reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are
+// met:
+//
+//     * Redistributions of source code must retain the above copyright
+// notice, this list of conditions and the following disclaimer.
+//     * Redistributions in binary form must reproduce the above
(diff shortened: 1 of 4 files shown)


zloylos:
I'd like to discuss interface, probably better will be passing TestSuit and TestInfo instead of just names.

googletest/test/gtest_custom_filter.cc
```diff
@@ -0,0 +1,46 @@
...
+TEST(GTestCustomFilterTest, DummyOne) {}
+TEST(GTestCustomFilterTest, DummyTwo) { FAIL(); }
+TEST(GTestCustomFilterTestTwo, DummyThree) { FAIL(); }
+
+int main(int argc, char** argv) {
+  testing::InitGoogleTest(&argc, argv);
+
+  testing::SetCustomFilter(
+      [](const std::string& test_suit_name, const std::string& test_name) {
+        return test_suit_name != "GTestCustomFilterTestTwo" &&
```
krishnacx, line 41: `return (test_suit_name != "GTestCustomFilterTestTwo") &&
               (test_name != "DummyTwo");` looks better

asoffer:
Thanks for the PR. I don't think we will be able to accept it. Adding programatic filtering can be done either via `GTEST_SKIP`, or by manually invoking the appropriate tests via the `testing::UnitTest` API.

PR #4070, opened 2022-11-24
Getting googletest code can be compiled with -Wundef option

lygstate:
Signed-off-by: Yonggang Luo <luoyonggang@gmail.com>


tigthor (approved):
I have reviewed this, this is a good option

lygstate:
The windows CI also fixed by this MR

dinord:
1. This contains changes beyond the minimal code required to enable -Wundef (e.g. edits to build files, adding underscores to struct fields, etc.). Could we reduce it to just change the spurious macros?
2. Since we use the pattern `#if defined(__FOO) && __FOO` for external macros, could we use it for googletest macros as well? Instead of attempting to define them to 0.

lygstate:
> 1. This contains changes beyond the minimal code required to enable -Wundef (e.g. edits to build files, adding underscores to struct fields, etc.). Could we reduce it to just change the spurious macros?
> 2. Since we use the pattern `#if defined(__FOO) && __FOO` for external macros, could we use it for googletest macros as well? Instead of attempting to define them to 0.

define them to 0 is better than `#if defined(__FOO) && __FOO` because we can use `-Wundef` to detect if the invalid macro are used

lygstate:
@thughes Please take a look at this MR

googletest/include/gtest/internal/gtest-port-arch.h
```diff
@@ -115,4 +115,100 @@
 #define GTEST_OS_QURT 1
 #endif  // __CYGWIN__
 
+// Defines macros that will be used but not defined to 0
```
krishnacx, line 118: Defines -> Define

googletest/include/gtest/internal/gtest-port.h
```diff
@@ -648,6 +656,81 @@ typedef struct _RTL_CRITICAL_SECTION GTEST_CRITICAL_SECTION;
...
+#ifndef GTEST_DONT_DEFINE_EXPECT_FALSE
+#define GTEST_DONT_DEFINE_EXPECT_FALSE 0
+#endif
+#ifndef GTEST_DONT_DEFINE_TEST_F
+#define GTEST_DONT_DEFINE_TEST_F 0
+#endif
+#ifndef GTEST_DONT_DEFINE_TEST
+#define GTEST_DONT_DEFINE_TEST 0
+#endif
+
```
krishnacx, line 726: Empty line can be ommitted.

lygstate:
cc @dinord

thughes:
This should now be fixed as of commit d92a270d2d9d6bb7ed3bf7dc3f698028550a65ef.

lygstate:
@thughes There is an issue, the bazel didn't guard -Wundef option
      # Enable Warn if an undefined identifier is evaluated in an #if directive. Such identifiers are replaced with zero.
        "//:msvc_compiler": [],
        "//conditions:default": ["-Wundef"],

thughes:
@lygstate Should be fixed in commit 678c1c73de2cf004d58f5dbbef299083997d924a and a798c2f10200b6293a5cc236b5f41b26c1ae7378.

Issue #4071, opened 2022-11-24
CMake Error (dev): uninitialized variable 'cxx_strict_flags'.


shrek1402:
**Describe the bug**

I get an error message when building a project

**Steps to reproduce the bug**

It's important that we are able to reproduce the problem that you are
experiencing. Please provide all code and relevant steps to reproduce the
problem, including your `BUILD`/`CMakeLists.txt` file and build commands. Links
to a GitHub branch or [godbolt.org](https://godbolt.org/) that demonstrate the
problem are also helpful.

**Does the bug persist in the most recent commit?**

Yes

I use preset:
```
"configurePresets": [
    {
      "name": "cmake-pedantic",
      "hidden": true,
      "warnings": {
        "dev": true,
        "deprecated": true,
        "uninitialized": true,
        "unusedCli": true,
        "systemVars": false
      },
      "errors": {
        "dev": true,
        "deprecated": true
      }
    },
```

```
$ cmake --preset=dev
Preset CMake variables:

  BUILD_MCSS_DOCS="ON"
  CMAKE_CXX_EXTENSIONS="OFF"
  CMAKE_CXX_FLAGS="/utf-8 /w14165 /w44242 /w44254 /w44263 /w34265 /w34287 /w44296 /w44365 /w44388 /w44464 /w14545 /w14546 /w14547 /w14549 /w14555 /w34619 /w34640 /w24826 /w14905 /w14906 /w14928 /w4503
8 /W4 /permissive- /volatile:iso /Zc:preprocessor /Zc:__cplusplus /Zc:externConstexpr /Zc:throwingNew /EHsc"
  CMAKE_CXX_STANDARD="17"
  CMAKE_CXX_STANDARD_REQUIRED="ON"
  super_lib_DEVELOPER_MODE="ON"

-- Selecting Windows SDK version 10.0.19041.0 to target Windows 10.0.19044.
CMake Error (dev) at build/dev/_deps/googletest-src/googletest/cmake/internal_utils.cmake:144 (set):
  uninitialized variable 'cxx_strict_flags'
Call Stack (most recent call first):
  build/dev/_deps/googletest-src/googletest/CMakeLists.txt:93 (config_compiler_and_linker)
This error is for project developers. Use -Wno-error=dev to suppress it.

CMake Error (dev) at build/dev/_deps/googletest-src/googletest/cmake/internal_utils.cmake:144 (set):
  uninitialized variable 'cxx_strict_flags'
Call Stack (most recent call first):
[...]

PR #4075, opened 2022-11-28
Fix uninitialized variable 'cxx_strict_flags'

shrek1402:


shrek1402:
Fix issue https://github.com/google/googletest/issues/4071

asoffer:
Thanks for the PR. Keeping GoogleTest warning-free is an impossible task, and one we do not think is valuable. Where there is actual undefined behavior we will absolutely clean things up. Indeed we run builds of googletest frequently with sanitizers to catch such issues. However these warnings trigger on perfectly valid code that we do not see the value in chasing down.

PR #4091, opened 2022-12-16
fix: One tricky way to silence all clang-tidy warnings (#2442)

MiaoDX:
Related issue: https://github.com/google/googletest/issues/2442

Hi,

I know googletest team do not care about clang-tidy that much. 

But we are really tired of the clang-tidy's complains for googletest.

Hopefully, maybe, some can benefit from this trick.

Good luck.

### DIFF
--- a/googletest/include/gtest/gtest.h
+++ b/googletest/include/gtest/gtest.h
@@ -2164,7 +2164,7 @@ constexpr bool StaticAssertTypeEq() noexcept {
 // Define this macro to 1 to omit the definition of TEST(), which
 // is a generic name and clashes with some other libraries.
 #if !GTEST_DONT_DEFINE_TEST
-#define TEST(test_suite_name, test_name) GTEST_TEST(test_suite_name, test_name)
+#define TEST(test_suite_name, test_name) GTEST_TEST(test_suite_name, test_name)  // NOLINT
 #endif
 
 // Defines a test that uses a test fixture.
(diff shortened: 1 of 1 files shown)


googletest/include/gtest/gtest.h
```diff
@@ -2164,7 +2164,7 @@ constexpr bool StaticAssertTypeEq() noexcept {
 // Define this macro to 1 to omit the definition of TEST(), which
 // is a generic name and clashes with some other libraries.
 #if !GTEST_DONT_DEFINE_TEST
-#define TEST(test_suite_name, test_name) GTEST_TEST(test_suite_name, test_name)
+#define TEST(test_suite_name, test_name) GTEST_TEST(test_suite_name, test_name)  // NOLINT
```
MiaoDX, line 2167: The TEST itself is one define, will be expand to one long line. 

Append NOLINT to it to suppress all warnings...

asoffer:
Thanks for the PR. We feel your pain regarding clang-tidy warnings showing up inside GoogleTest, however, we believe pretty strongly that these are issues that should be fixed in clang-tidy, rather than GoogleTest. Insofar as GoogleTest can adhere to best practices, we desire to do so. But if clang-tidy is triggering in a situation that is not generally useful, we believe the best course of action is to fix clang-tidy, rather than GoogleTest. This has the added benefit of improving clang-tidy for all use-cases (e.g., other macro-based libraries). We do a fair amount of work on clang-tidy ourselves, and a common technique is to simply avoid triggering on any code snippet inside a macro, or on code snippets in macros generated in a different file. Internally this approach has worked quite well for us.

MiaoDX:
@asoffer Hi, thanks. 

Tricks here is `simply avoid triggering on any code snippet inside a macro, or on code snippets in macros generated in a different file` for us, as user of both gtest and clang-tidy.
And of course, as I looked at other issues, I was pretty sure this won't be acceptable for you. Just put here to ease the pain of some other poor guy.

And, is there any more detail for `Internally this approach has worked quite well for us`? How are you `avoid triggering on any code snippet inside a macro`?

samopolacek:
@asoffer can you share more info about your internal approach?

LazyWolfLin:
@MiaoDX Thank you for your sharing. I spent an afternoon on this error #3983.
@asoffer Is it possible to include this tricky in the F&Q?

PR #4092, opened 2022-12-19
Update primer.md

zwillb:
The 3 sample fixture variables weren't equally treated in the  Setup()  method, which was confusing.

I presume it was not intended.


dinord:
This was actually intended. 
See `TEST_F(QueueTest, DequeueWorks) ` several lines below:
```
  int* n = q0_.Dequeue();
  EXPECT_EQ(n, nullptr);
```

We do get pull requests identical to this one from time to time, so we'll add a comment to point out that `q0` is intentionally empty.

Issue #4098, opened 2022-12-21
[Bug]: can't compile gtest-matchers.cc on Linux with gcc via bazel 5 or bazel 6


dneto0:
### Describe the issue

On Linux x86-64 where /usr/bin/gcc is GCC 12.2.0
Building with bazel fails because the C++ compiler is compiling for C++03

INFO: Found 7 targets...
ERROR: /project/googletest/BUILD.bazel:80:11: Compiling googletest/src/gtest-matchers.cc failed: (Exit 1): gcc failed: error executing command (from target //:gtest) /usr/bin/gcc -U_FORTIFY_SOURCE -fstack-protector -Wall -Wunused-but-set-parameter -Wno-free-nonheap-object -fno-omit-frame-pointer '-std=c++0x' -MD -MF ... (remaining 34 arguments skipped)

Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
In file included from googletest/include/gtest/gtest-printers.h:114,
                 from googletest/include/gtest/gtest-matchers.h:48,
                 from googletest/src/gtest-matchers.cc:35:
googletest/include/gtest/internal/gtest-internal.h:635:58: error: wrong number of template arguments (0, should be 1)
  635 |   typedef ::std::map<std::string, CodeLocation, std::less<>> RegisteredTestsMap;
      |                                                          ^
In file included from /usr/include/c++/12/bits/unique_ptr.h:37,
                 from /usr/include/c++/12/memory:76,
                 from googletest/include/gtest/gtest-matchers.h:43:
/usr/include/c++/12/bits/stl_function.h:403:12: note: provided for 'template<class _Tp> struct std::less'
  403 |     struct less : public binary_function<_Tp, _Tp, bool>
      |            ^~~~
googletest/include/gtest/internal/gtest-internal.h:635:59: error: template argument 3 is invalid
  635 |   typedef ::std::map<std::string, CodeLocation, std::less<>> RegisteredTestsMap;
      |                                                           ^~
googletest/include/gtest/internal/gtest-internal.h: In member function 'bool testing::internal::TypedTestSuitePState::AddTestName(const char*, int, const char*, const char*)':
[...]

PR #4100, opened 2022-12-21
BUILD.bazel: Add CXX_OPTS to specify -std=c++14 when needed

dneto0:
Googletest requires a C++14 compiler.
But Bazel can default to givin GCC option -std=c++0x which is unable to compile googletest.

Specify -std=c++14 for any compiler that isn't MSVC and isn't on QNX. I don't know what C++ compiler Bazel invokes for QNX.

Fixes: #4098

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -69,11 +69,20 @@ config_setting(
     values = {"define": "absl=1"},
 )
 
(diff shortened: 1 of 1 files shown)


dneto0:
I'm proposing this with an internal CL.

PR #4106, opened 2022-12-30
TI CGT support (No-completed)

pradeep-hn:
- Added the support for Texas Instruments CGT compilation to run on TI targets

Signed-Off-By: Pradeep HN <p-hn@ti.com>


defihook (commented):
WhenTheCodingCodingGIF

Issue #3583, opened 2021-09-27
GoogleMock segfaults if built with option "-D gtest_disable_pthreads=ON"


ingo-h:
**Describe the bug**

I built GoogleTest without **pthread** on Linux and successful compile tests with it using mocks. When running these tests all of them crash with a final message "Segmentation fault". They should pass.

**Steps to reproduce the bug**

I cloned GoogleTest and stay in its root directory. Here I use this simple test program:

    ~$ cat test_simple.cpp
    #include "gmock/gmock.h"
    #include "gtest/gtest.h"

    // simple mocked class
    // -------------------
    class Foo {
        virtual int GetInt() const = 0;
    };

    class MockFoo : public Foo {
      public:
        MOCK_METHOD(int, GetInt, (), (const, override));
    };

    // simple testsuite
    //-----------------
    TEST(simpleTestSuite, simpleMockTest) {
        MockFoo mockedFoo;

        EXPECT_CALL(mockedFoo, GetInt()).Times(1);
        EXPECT_EQ(mockedFoo.GetInt(), 0);
    }

    // main entry
    // ----------
    int main(int argc, char** argv) {
        ::testing::InitGoogleTest(&argc, argv);
        return RUN_ALL_TESTS();
    }

Now verify that it runs successful with default configure/build of GoogleTest:

    ~$ cmake -S . -B build
    ~$ cmake --build build

and compile the simple test:

    ~$ g++ -std=c++11 -otest_simple -I./googletest/include/ -I./googlemock/include/ ./test_simple.cpp ./build/lib/libgtest.a ./build/lib/libgmock.a -lpthread

run it:

    ~$ ./test_simple
    [==========] Running 1 test from 1 test suite.
    [----------] Global test environment set-up.
    [----------] 1 test from simpleTestSuite
    [ RUN      ] simpleTestSuite.simpleMockTest
    [       OK ] simpleTestSuite.simpleMockTest (0 ms)
    [----------] 1 test from simpleTestSuite (0 ms total)

    [----------] Global test environment tear-down
    [==========] 1 test from 1 test suite ran. (0 ms total)
    [  PASSED  ] 1 test.

So far so good. Now build GoogleTest with option `-D gtest_disable_pthreads=ON` and compile the simple test:

    ~$ rm -rf build
    ~$ cmake -S . -B build -D gtest_disable_pthreads=ON
    ~$ cmake --build build

On compiling the simple test I do not need option `-lpthread` of course.

[...]

PR #3693, opened 2021-12-03
Properly export GTEST_HAS_PTHREAD compile definition with CMake

dalg24:
Fix #3692

PR #4107, opened 2022-12-30
Add GTEST_HAS_PTHREAD to target interface

gitplcc:
Add -DGTEST_HAS_PTHREAD compiler flags to GTest::* targets as pkg-config do

### DIFF
--- a/googletest/cmake/internal_utils.cmake
+++ b/googletest/cmake/internal_utils.cmake
@@ -187,6 +187,8 @@ function(cxx_library_with_type name type cxx_flags)
         $<INSTALL_INTERFACE:GTEST_LINKED_AS_SHARED_LIBRARY=1>)
     endif()
   endif()
+  target_compile_definitions(${name} INTERFACE
+    $<INSTALL_INTERFACE:GTEST_HAS_PTHREAD=$<BOOL:${GTEST_HAS_PTHREAD}>>)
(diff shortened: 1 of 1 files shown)


defihook (commented):
WhenTheCodingCodingGIF

tpadioleau:
Up, can this merged ? the problem is still present

googletest/cmake/internal_utils.cmake
```diff
@@ -187,6 +187,8 @@ function(cxx_library_with_type name type cxx_flags)
         $<INSTALL_INTERFACE:GTEST_LINKED_AS_SHARED_LIBRARY=1>)
     endif()
   endif()
+  target_compile_definitions(${name} INTERFACE
+    $<INSTALL_INTERFACE:GTEST_HAS_PTHREAD=$<BOOL:${GTEST_HAS_PTHREAD}>>)
```
tpadioleau, line 191: An alternative implementation would be to reuse `GTEST_HAS_PTHREAD_MACRO`

```cpp
target_compile_definitions(${name} INTERFACE $<INSTALL_INTERFACE:${GTEST_HAS_PTHREAD_MACRO}>)
```

tpadioleau:
Please @derekmauro can you have a look ? It supersedes #3693 and likely fixes #3583

PR #4118, opened 2023-01-06
Add bzlmod module file

hanneskaeufler:
Motivation:
https://github.com/bazelbuild/bazel-central-registry/issues/124

Open questions:
* Is the googletest project willing to maintain the bzlmod file?

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -1,0 +1,9 @@
+module(
+    name = "googletest",
+    compatibility_level = 1,
+    version = "0.0.0",
+)
+
+bazel_dep(name = "abseil-cpp", repo_name = "com_google_absl", version = "20210324.2")
+bazel_dep(name = "platforms", version = "0.0.5")
+bazel_dep(name = "rules_cc", version = "0.0.1")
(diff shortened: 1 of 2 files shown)


MODULE.bazel
```diff
@@ -0,0 +1,10 @@
+module(
+    name = "googletest",
+    compatibility_level = 1,
+    version = "0.0.0",
```
hanneskaeufler, line 4: In order not to have to maintain a version number multiple times (or at all), it seems to be a somewhat often used practice to have `0.0.0` here and just increment the version number in the bazel central registry.

hanneskaeufler:
@derekmauro Anything I can do to move this forward? thanks!

Vertexwahn:
@derekmauro Can we bring this forward?

MODULE.bazel
Vertexwahn: The version attribute is optional - maybe leave it out, instead of using 0.0.0

Vertexwahn:
@derekmauro Please review!

derekmauro (commented):
I'm not really sure how this is supposed to work.

We prefer to live at head. This means, for example, that sometimes we might use a feature in the latest head version of Abseil that isn't part of any release. How is bzlmod supposed to deal with this?

Vertexwahn:
@derekmauro Good point! There is  `archive_override` to override a specific version with a new one: https://github.com/google/googletest/pull/4368

hanneskaeufler:
Closing in favor of https://github.com/google/googletest/pull/4368, thanks for picking this up Vertexwahn!

PR #4123, opened 2023-01-16
Fix `-Wdouble-promotion` compilation warning

RaisinTen:
This fixes this compilation warning which occurred at 17 lines in `googletest/include/gtest/gtest-printers.h`:

```sh
/.../googletest/include/gtest/gtest-printers.h:511:9: error:
      implicit conversion increases floating-point precision: 'float' to 'double' [-Werror,-Wdouble-promotion]
    if (val >= 100000.0) {  // 100,000 to 999,999
        ^~~ ~~
```

Signed-off-by: Darshan Sen <raisinten@gmail.com>

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -12,6 +12,7 @@ Chandler Carruth <chandlerc@google.com>
 Chris Prince <cprince@google.com>
 Chris Taylor <taylorc@google.com>
 Dan Egnor <egnor@google.com>
+Darshan Sen <raisinten@gmail.com>
 Dave MacLachlan <dmaclach@gmail.com>
 David Anderson <danderson@google.com>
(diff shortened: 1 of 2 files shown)


RaisinTen:
Hey @thughes, since you submitted https://github.com/google/googletest/commit/5cd81a7848203d3df6959c148eb91f1a4ff32c1d recently, I thought it would be a nice idea to ask you to take a look.

RaisinTen:
@asoffer could you PTAL?

asoffer:
Apologies for the extremely delayed response. Looking at the changes here, we do not believe there is significant value in avoiding this particular warning. The extra casts, in our opinion, are noise, calling attention to operations that are safe and unsurprising.

Issue #4124, opened 2023-01-20
[Bug]: CMake fails to build gmock python tests


bogdan-lab:
### Describe the issue

CMake build does not produce googlemock python tests even when the python interpreter is found.

### Steps to reproduce the problem

Try to build gmock with tests using CMake and check if it will create `gmock_leak_test` and `gmock_output_test`.

### What version of GoogleTest are you using?

I am using current HEAD (`ec25eea8`)

### What operating system and version are you using?

Distributor ID:	Ubuntu
Description:	Ubuntu 22.04.1 LTS
Release:	22.04
Codename:	jammy

### What compiler and version are you using?

g++ (Ubuntu 11.3.0-1ubuntu1~22.04) 11.3.0

### What build system are you using?

cmake version 3.19.0-rc2

PR #4125, opened 2023-01-20
Fix CMake build of gmock python tests

bogdan-lab:
Old code set python variables in the scope which is not accessible for gmock build. Therefore py_test function in gmock was ignored. Now we set variables in the correct scope using corresponding macro.

This PR closes issue https://github.com/google/googletest/issues/4124

### DIFF
--- a/googlemock/CMakeLists.txt
+++ b/googlemock/CMakeLists.txt
@@ -135,6 +135,8 @@ if (gmock_build_tests)
   # 'make test' or ctest.
   enable_testing()
 
+  set_python_interpreter(PYTHON_EXECUTABLE)
+
   if (MINGW OR CYGWIN)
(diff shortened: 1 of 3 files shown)


derekmauro:
I'm aware that the ci/Windows issue is an old one. We consider the GitHub checks to be unofficial and the CI with the Google logo to be the official CI.

The issue in ci/Windows appears harmless. It appears to have something to do with slightly different output on various systems. If that were the only issue I would just accept this PR.

This PR is basically bringing the same issue to the official CI. This is the one we use for internal changes, and if it breaks it will block submission, so I can't merge this yet. I think someone is looking into fixing this.

Thank you for the PR. I hope to get it merged soon.

bogdan-lab:
Close because it looks like https://github.com/google/googletest/commit/fd36851c8d383a2b711905eb488692da1152a3d8 solved the problem

PR #4146, opened 2023-02-09
Fixes the test gmock_output_test.py with MSVC in a better way

lygstate:
cc @thughes


googletest/include/gtest/internal/gtest-type-util.h
```diff
@@ -67,6 +67,22 @@ inline std::string CanonicalizeForStdLibVersioning(std::string s) {
       s.erase(strlen("std"), end - strlen("std"));
     }
   }
+
+  /* Strip redundant space in typename to match MSVC */
```
krishnacx, line 71: space -> spaces

googletest/include/gtest/internal/gtest-type-util.h
```diff
@@ -67,6 +67,22 @@ inline std::string CanonicalizeForStdLibVersioning(std::string s) {
...
+  const std::string &toSearch = ", ";
+  const std::string &replaceStr = ",";
+  size_t pos = 0;
+  for (;;) {
+    // Get the next occurrence from the current position
+    pos = s.find(toSearch, pos);
+    if (pos == std::string::npos) {
+      break;
+    }
+    // Replace this occurrence of Sub String
```
krishnacx, line 82: sub string

googletest/include/gtest/internal/gtest-type-util.h
```diff
@@ -85,6 +101,19 @@ inline std::string GetTypeName(const std::type_info& type) {
   const std::string name_str(status == 0 ? readable_name : name);
   free(readable_name);
   return CanonicalizeForStdLibVersioning(name_str);
+#elif defined(_MSC_VER)
+  // Strip struct and class due to differences between
+  // MSVC and other compilers. std::pair<int,bool> is printed as
+  // "struct std::pair<int,bool>" when using MSVC vs "std::pair<int, bool>" with
+  // other compilers.
+  std::string s = name;
+  if (s.rfind("struct ", 0) == 0) {
```
krishnacx, line 110: Why not just find ?

googletest/include/gtest/internal/gtest-type-util.h
```diff
@@ -85,6 +101,19 @@ inline std::string GetTypeName(const std::type_info& type) {
...
+    s = s.substr(strlen("struct "));
+  }
+  if (s.rfind("class ", 0) == 0) {
```
krishnacx, line 113: else if ?

googletest/include/gtest/internal/gtest-type-util.h
lygstate: Only strip the leading "struct " and "class ", so uses rfind == 0 to ensure that

asoffer:
Can you describe what the issue is and why this fix is better (and what it's better than?)

lygstate:
> Can you describe what the issue is and why this fix is better (and what it's better than?)

updated the reason

lygstate:
@thughes @asoffer ping

asoffer:
Sorry, I'm still not sure I understand what the problem is here. Could you answer:
* What does gmock_output_test.py do on MSVC?
* What do you expect it to do?
* Why is this solution to that problem better than the existing solution?

lygstate:
> Sorry, I'm still not sure I understand what the problem is here. Could you answer:
> 
> * What does gmock_output_test.py do on MSVC?

For MSVC, gmock_output_test.py  output struct std::pair<int,bool>, for GCC, it's output
std::pair<int, bool>, it's not the same, my intention is getting these to be same by removing
`struct ` for MSVC's outptu, and strip redundant space for GCC.

> * What do you expect it to do?
ditto.


> * Why is this solution to that problem better than the existing solution?
The new solution can use `std::pair<int,bool>` directly, the exiting solution is avoid to use it, 
this solution would be more robust that output consistence typename.
As a by-product, 
```c
#ifdef _MSC_VER
#define ERROR_DESC "class std::runtime_error"
#else
#define ERROR_DESC "std::runtime_error"
#endif
````
can be simplified to
```c
#define ERROR_DESC "std::runtime_error"
```

googletest/include/gtest/internal/gtest-type-util.h
```diff
@@ -67,6 +67,22 @@ inline std::string CanonicalizeForStdLibVersioning(std::string s) {
       s.erase(strlen("std"), end - strlen("std"));
     }
   }
+
+  /* Strip redundant spaces in typename to match MSVC */
+  /* For example, std::pair<int, bool> -> std::pair<int,bool> */
+  const std::string &toSearch = ", ";
```
asoffer, line 73: Do not rely on lifetime extension if it's not necessary.
In this case, `const char to_search[] = ", ";` would be preferred.

Also, please use snake_case instead of camelCase for local variables in accordance with Google's style guide.

googletest/include/gtest/internal/gtest-type-util.h
```diff
@@ -67,6 +67,22 @@ inline std::string CanonicalizeForStdLibVersioning(std::string s) {
...
+  const std::string &replaceStr = ",";
+  size_t pos = 0;
```
asoffer, line 75: Nit. Let's use `while (true) { ... }`.

googletest/include/gtest/internal/gtest-type-util.h
```diff
@@ -85,6 +101,20 @@ inline std::string GetTypeName(const std::type_info& type) {
...
+#elif defined(_MSC_VER)
+  // Strip struct and class due to differences between
+  // MSVC and other compilers. std::pair<int,bool> is printed as
+  // "struct std::pair<int,bool>" when using MSVC vs "std::pair<int, bool>" with
+  // other compilers.
+  std::string s = name;
+  // Only strip the leading "struct " and "class ", so uses rfind == 0 to 
+  // ensure that
+  if (s.rfind("struct ", 0) == 0) {
+    s = s.substr(strlen("struct "));
```
asoffer, line 113: Super pedantic nit: Let's use `std::strlen`

asoffer (commented):
Just a few nits, but otherwise looks great. Thank you!

googletest/include/gtest/internal/gtest-type-util.h
lygstate: the existing code in gtest-type-util.h are using `strlen`, not `std::strlen`, so I'll keep use strlen,
if std::strlen are needed, I think it's should in another MR and change all usage of `strlen` to `std::strlen`

lygstate:
@derekmauro @asoffer ping?

asoffer:
Apologies for the delay. Pulling this in.

PR #4163, opened 2023-02-21
Fix error in_death_test_child_process: undeclared identifier

sergio-nsk:
The error occurs if `!GTEST_HAS_DEATH_TEST` on Windows.

I could not make the change better. If I make
```c++
bool in_death_test_child_process = false;
#if GTEST_HAS_DEATH_TEST
  in_death_test_child_process =
      GTEST_FLAG_GET(internal_run_death_test).length() > 0;
```
then the error about an assigned but unused variable occurs on Linux.


sergio-nsk:
IDK how to change committer email in PR.

PR #4164, opened 2023-02-21
Fix error in_death_test_child_process: undeclared identifier

sergio-nsk:
The error occurs if `!GTEST_HAS_DEATH_TEST` on Windows.

I could not make the change better. If I make
```c++
bool in_death_test_child_process = false;
#if GTEST_HAS_DEATH_TEST
  in_death_test_child_process =
      GTEST_FLAG_GET(internal_run_death_test).length() > 0;
```
then the error about an assigned but unused variable occurs on Linux.

### DIFF
--- a/googletest/src/gtest.cc
+++ b/googletest/src/gtest.cc
@@ -5416,6 +5416,9 @@ int UnitTest::Run() {
   impl()->set_catch_exceptions(GTEST_FLAG_GET(catch_exceptions));
 
 #if GTEST_OS_WINDOWS
+#if !GTEST_HAS_DEATH_TEST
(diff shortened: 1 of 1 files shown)


horenmar:
I think the fix in #4165 (implemented for https://github.com/microsoft/vcpkg/pull/29772) is better.

sergio-nsk:
@horenmar The fix with `#if GTEST_OS_WINDOWS && GTEST_HAS_DEATH_TEST` completely excludes the code block that tunes output to stderr not only for death tests. IMHO it's not correct.

kolumnin:
The family links at the mobile device access is unregistered.

PR #4169, opened 2023-02-25
gmock: Detect absl libary exist by #if not #ifdef

xiaoxiang781216:
align with the detection done by googletest


lygstate:
It's already done in https://github.com/google/googletest/pull/4070

thughes:
This should no longer be needed as of commit d92a270d2d9d6bb7ed3bf7dc3f698028550a65ef.

PR #4180, opened 2023-03-02
Environment variables are accessible in UWP/WinRT apps

sergio-nsk:
Environment variables, for example `GTEST_FILTER`, `GTEST_OUTPUT` can be set before getting WinRT app run. GoogleTest can read environment variables and use them. It's easier than setting and passing command line parameters to WinRT app.


sergio-nsk:
@higher-performance Yes, of course I tested it. My first approach was using command line parameters, but this is so an extra hard task for UWP/WinRT applications, then I tried using environment variables, and discovered that disabled code block, enabled it, and it worked perfectly. Honestly, I took a look at [`GetEnvironmentVariable()`](https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-getenvironmentvariable) and I don't see what can prevent `getenv()` from working:
<html>
<body>


Minimum supported client | Windows XP [desktop apps \| UWP apps]
-- | --
Minimum supported server | Windows Server 2003 [desktop apps \| UWP apps]


</body>
</html>
image

PR #4224, opened 2023-04-21
Ankit-gautam23/issue#4221

ankit-gautam23:
Rename struct Result -> NonMovableStruct to avoid shadowing #4221


dinord:
Let's do this in https://github.com/google/googletest/pull/4221 to preserve the discussion and give credit to the original author.

PR #4231, opened 2023-05-07
Update README.md

yagneshprajapati:
feature section in table format.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ This repository is a merger of the formerly separate GoogleTest and GoogleMock
 projects. These were so closely related that it makes sense to maintain and
 release them together.
 
-### Getting Started
+### | Getting Started |
 
 See the [GoogleTest User's Guide](https://google.github.io/googletest/) for
 documentation. We recommend starting with the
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -41,7 +41,7 @@ This repository is a merger of the formerly separate GoogleTest and GoogleMock
 projects. These were so closely related that it makes sense to maintain and
 release them together.
 
-### Getting Started
+### | Getting Started |
```
derekmauro, line 44: Why are there | chars around Getting Started?

README.md
```diff
@@ -50,18 +50,19 @@ documentation. We recommend starting with the
...
-*   User-defined assertions.
-*   Death tests.
-*   Fatal and non-fatal failures.
-*   Value-parameterized tests.
-*   Type-parameterized tests.
-*   Various options for running the tests.
-*   XML test report generation.
+| Feature | Description |
+| --- | --- |
+| xUnit test framework | Googletest is based on the xUnit testing framework, a popular architecture for unit testing |
```
derekmauro, line 55: Can you restore the xUnit link?

derekmauro (changes requested):
Can you update the description of this change to describe what is changing? "Update README.md" is too general a description. Please describe how it is changing.

README.md
yagneshprajapati: it indicate that "Getting Started" is a heading in a larger body of text, also drawing attention to it.

README.md
yagneshprajapati: yes done!!

yagneshprajapati:
now looks good

README.md
derekmauro: Please take a look at the preview of the file: https://github.com/google/googletest/blob/a7fa9b37082993107a98181e631cafcdbe484f22/README.md#-getting-started-

I don't think the markdown is rendering as you intend.

yagneshprajapati:
> See my comment on the rendering.

yes agree about the | char is may be not well. so my recent commit is ok?

PR #4251, opened 2023-05-23
Add Instant leak check option

Matthdonau:
This PR adds functionality to instantly invoke a mock object leak check (as currently already done at program end) in order to provide a more fine grained leak check over a test suite. If the leak check detects still allocated mock objects it will fail the current test and report the leaks. However it wont affect the end of program leak check in any way.


Matthdonau:
@dinord Is there something I can provide to move this forward?

defihook (commented):
great

PR #4252, opened 2023-05-23
For sample1_unittest.cc etc, update comments to use new "test-suite, test-case" nomenclature.

chjfth:
According to 2019 updates of [GoogleTest Primer](http://google.github.io/googletest/primer.html), we have two naming change:

* a **test-case** now becomes **test-suite** .
* a **TEST_F()** now becomes a **test-case** .

I totally agree with this change, because the new naming is clearer and more natural. 

Although 3 years has passed, most code comments in `sample1_unittest.cc` ... `sample8_unittest.cc` has not been updated to reflect the new naming, so I offer to do the updates.

BTW, how do you call a `EXPECT_EQ` or a `EXPECT_TRUE` inside a test-case? This naming is not very clear yet. I offer to call it a test-assertion.

### DIFF
--- a/googletest/samples/sample1_unittest.cc
+++ b/googletest/samples/sample1_unittest.cc
@@ -48,24 +48,25 @@ namespace {
 
 // Step 2. Use the TEST macro to define your tests.
 //
-// TEST has two parameters: the test case name and the test name.
+// TEST has two parameters: the test-suite name and the test-case name.
 // After using the macro, you should define your test logic between a
 // pair of braces.  You can use a bunch of macros to indicate the
-// success or failure of a test.  EXPECT_TRUE and EXPECT_EQ are
-// examples of such macros.  For a complete list, see gtest.h.
+// success or failure of a test(it means a test-assertion here).  
+// EXPECT_TRUE and EXPECT_EQ are examples of such macros.  
+// For a complete list, see gtest.h.
 //
 // <TechnicalDetails>
 //
-// In Google Test, tests are grouped into test cases.  This is how we
+// In Google Test, tests are grouped into test-suites.  This is how we
 // keep test code organized.  You should put logically related tests
-// into the same test case.
+// into the same test-suite.
 //
-// The test case name and the test name should both be valid C++
+// The test-suite name and the test-case name should both be valid C++
 // identifiers.  And you should not use underscore (_) in the names.
 //
-// Google Test guarantees that each test you define is run exactly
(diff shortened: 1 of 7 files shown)


dinord:
Instead of `test-suite` and `test-case`, can we drop the dashes and write `test suite` and `test case`?

chjfth:
> Instead of `test-suite` and `test-case`, can we drop the dashes and write `test suite` and `test case`?

I really think test-suite is more friendly to the reader. Seeing “test-suite”, the reader is 100% sure it is ONE word representing an atomic meaning, even if the reader is a thorough novice to GoogleTest.

On the other hand, typing "test-suite" is a bit more tiring than "test suite". That may be (one of) the reason existing doc writes them without a hyphen.

dinord:
I don't think `test suite` is any less atomic than `test-suite`. As a reader, I'd parse those words together regardless of whether they had a hyphen or not.
I'd break the tie towards `test suite` as it is plain English. with `test-suite`, future editors and writers of auxiliary documentation might forget the hyphen, which would hurt searchability and possibly even confuse some readers.

dinord:
> Let's consider this: In a long paragraph of text, if there are hyphen-ed words like
> 
> * test-program
> * test-suite
> * test-case
> * test-assertion
> 
> Don't they more vividly convey the idea that these words are highly related and have very specific meaning in a testing framework?
Maybe, but my inner voice would still read them out loud as "test program", "test suite", etc.

> If someone could/would design GoogleTest from the beginning, do you (as a doc reader first-time getting to know GoogleTest) prefer to see the hyphen-ed words or the non-hyphen-ed ones？
I don't think we can answer this without polling actual users (and this is hard to do right), which makes this somewhat subjective for now.

> As for searchability, if we formally use "test-case" from the very beginning, then, searching for "test-case" will definitely produce results from that dedicated meaning, without the interference of casual word "test case". Furthermore, searching for "test-" would conveniently give results of all of [test-program, test-suite, test-case and test-assertion], which is bonus.
On the other hand, the presence of the hyphen is a new rule, imposed by GoogleTest on the reader. Such rules tend to increase the complexity of documentation. Careful nomenclature can sometimes be worth it (e.g. "lambda function" is one such example), but it comes at a cost.

> Technical documentation is hard for readers, I'd just like to make it a bit easier.
Me too.

dinord:
Closing due to inactivity, please re-open if you'd like to continue the naming discussion.

PR #4256, opened 2023-05-25
Build and test googletest with IntelLLVM

niranjan-nilakantan:
Make IntelLLVM use the same flags as Clang or MSVC (as appropriate).

Suppress warnings-as-error for the Windows build.

Use -ffp-model=precise or /fp:precise to override IntelLLVM's default choice of **fast**.


dinord:
Does IntelLLVM not supprot `-Wchar-subscripts`?

Issue #4254, opened 2023-05-24
[Bug]: bazel passes '-std=c++0x' which is too old


pogo59:
### Describe the issue

Trying to build googletest with bazel, I see it passing `'-std=c++0x'` on the gcc command line, which of course fails the version check for C++14 or later in gtest-port.h.  I am a bazel noob but I don't see anything in the BUILD files related to this.

### Steps to reproduce the problem

```
$ bazel build //:gtest
INFO: Analyzed target //:gtest (1 packages loaded, 59 targets configured).
INFO: Found 1 target...
ERROR: /home/probinson/ssd/projects/google/googletest/BUILD.bazel:80:11: Compiling googletest/src/gtest-typed-test.cc failed: (Exit 1): gcc failed: error executing command (from target //:gtest) /usr/lib/icecc/bin/gcc -U_FORTIFY_SOURCE -fstack-protector -Wall -Wunused-but-set-parameter -Wno-free-nonheap-object -fno-omit-frame-pointer '-std=c++0x' -MD -MF ... (remaining 34 arguments skipped)

Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
In file included from googletest/include/gtest/internal/gtest-internal.h:42:0,
                 from googletest/include/gtest/gtest-typed-test.h:174,
                 from googletest/src/gtest-typed-test.cc:30:
googletest/include/gtest/internal/gtest-port.h:280:2: error: #error C++ versions less than C++14 are not supported.
 #error C++ versions less than C++14 are not supported.
  ^~~~~
Target //:gtest failed to build
Use --verbose_failures to see the command lines of failed build steps.
INFO: Elapsed time: 0.481s, Critical Path: 0.30s
INFO: 15 processes: 15 internal.
FAILED: Build did NOT complete successfully
```

### What version of GoogleTest are you using?

$ git rev-parse HEAD
e49c6b946a44f9a58586bfc76f75687d8f77e725

### What operating system and version are you using?

Ubuntu 18.04

### What compiler and version are you using?

```
$ gcc -v
Using built-in specs.
COLLECT_GCC=/usr/bin/gcc
COLLECT_LTO_WRAPPER=/usr/lib/gcc/x86_64-linux-gnu/7/lto-wrapper
OFFLOAD_TARGET_NAMES=nvptx-none
OFFLOAD_TARGET_DEFAULT=1
Target: x86_64-linux-gnu
[...]

PR #4258, opened 2023-05-25
Enable c++14 by default in bazel build issue: #4254

pgawro:
This change is related to issue   #4254


pgawro:
It was a bit of a weird solution for me as well. At leas OP had example and my comment in #4254 helped him

PR #4263, opened 2023-06-01
Avoid stack overflow in CreateDirectoriesRecursively

### LINKED ISSUE #4262: [Bug]: Stack overflow in CreateDirectoriesRecursively
### Describe the issue

Asking google test to output to a non existing path, on a system with no notion of "./" being a path causes FilePath::CreateDirectoriesRecursively to end up in an infinite loop with stack overflow.

Somewhat hard.
- Run on an embedded system with file support, but no concept on a current directory named "./".
- Ask google-test to output xml to a non existing path.

609281088cfefc76f9d0ce82e1ff6c30cc3591e5

Its also patch in a small way to work around: https://github.com/google/googletest/issues/1372

rt-kernel

arm-eabi-gcc (rt-collab toolbox 2022.1-0-gf81ad3e) 10.4.0

cmake version 3.25.0

### Additional context

When creating directories recursively the stop condition is an empty pathname:
https://github.com/google/googletest/blob/04cf2989168a3f9218d463bea6f15f8ade2032fd/googletest/src/gtest-filepath.cc#L338

However, RemoveFileName() function will never end up with an empty pathname, but a ./ separator.
https://github.com/google/googletest/blob/04cf2989168a3f9218d463bea6f15f8ade2032fd/googletest/src/gtest-filepath.cc#L215

elupus:
Avoid stack overflow in CreateDirectoriesRecursively by checking for end condition of RemoveFileName()

Fixes #4262

### DIFF
--- a/googletest/src/gtest-filepath.cc
+++ b/googletest/src/gtest-filepath.cc
@@ -335,7 +335,7 @@ bool FilePath::CreateDirectoriesRecursively() const {
     return false;
   }
 
-  if (pathname_.length() == 0 || this->DirectoryExists()) {
+  if (pathname_.length() == 0 || pathname_ == kCurrentDirectoryString || this->DirectoryExists()) {
     return true;
   }
 


defihook (commented):
great

PR #4264, opened 2023-06-01
Allow CreateDirectoriesRecursively() to fail on report

elupus:
Only warn on failure to create path to report file.

When running in a semihosted ARM environment. There is no stat support nor mkdir, so existens checks for directories will always fail, however file open may still work.


pgawro (approved):

defihook (commented):
GreatGreatJobGIF

PR #4265, opened 2023-06-01
Add back support for platformio

ciband:
The version of googletest available to the Platformio framework is stale due to the removal of the library.json file from the googletest repo.  To continue to support this community, it would be nice if we could add this file back.

### DIFF
--- a/library.json
+++ b/library.json
@@ -1,0 +1,51 @@
+{
+  "name": "googletest",
+  "keywords": "unit-testing, testing, tdd, testing-framework, gtest, gmock",
+  "description": "Google Testing and Mocking Framework",
+  "license": "BSD-3-Clause",
+  "homepage": "https://google.github.io/googletest/",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/google/googletest.git"
+  },
+  "version": "1.13.0",
+  "frameworks": "*",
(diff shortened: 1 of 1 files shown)


asoffer:
Do you have thoughts on how this can be kept up to date/tested? We do not have any experience with platformio and are not well-poised to maintain this. Is it critical that this file be in this git repository, or could it be maintained elsewhere?

ciband:
When the feature was first added, I also added CI tests that would at least compile the code to make sure it builds.  Running the googletest unit tests would be impractical as we would need to run them on the target embedded hardware.  I could certainly create a follow up PR to add those CI tests back.

If the library.json is stored in the repo, then the process for maintaining the library would be to update its version and any relevant changes to the include/exclude files, and then re-publish it with the PlatformIO framework.

Details on that here: https://docs.platformio.org/en/latest/librarymanager/creating.html#publishing

A future state could potentially automate this step as a part of the googletest CI/CD pipeline when you guys do releases.

If the file is not in this repo, then it would just have to be maintained locally and manually republished every time a googletest released.

ciband:
@asoffer, is there anything else you need for this PR, or are there any other questions I can answer?

bishalbaaniya123 (approved):

PR #4270, opened 2023-06-02
Suppress avoid-goto Clang-tidy warning in DeathTest and exception test macros

jerylvaz:


dinord:
We typically don't accept `NOLINT`s in our headers. Although this is a macro, the `goto` is just an internal implementation detail.

I'd file a bug against Clang-Tidy to provide a way to disable it these types of warnings.

PR #4288, opened 2023-06-19
cmake: Raise min to 3.6

juan-lunarg:
From the CMake 3.27 release notes:
Compatibility with versions of CMake older than 3.5 is now deprecated and will be removed from a future version. Calls to cmake_minimum_required() or cmake_policy() that set the policy version to an older value now issue a deprecation diagnostic.

This PR also removes manually setting policy CMP0048. This is redundant since the CMake min is already 3.X

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,11 +1,7 @@
 # Note: CMake support is community-based. The maintainers do not use CMake
 # internally.
 
-cmake_minimum_required(VERSION 3.5)
-
-if (POLICY CMP0048)
-  cmake_policy(SET CMP0048 NEW)
-endif (POLICY CMP0048)
+cmake_minimum_required(VERSION 3.6)
 
 if (POLICY CMP0069)
   cmake_policy(SET CMP0069 NEW)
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -1,11 +1,7 @@
 # Note: CMake support is community-based. The maintainers do not use CMake
 # internally.
 
-cmake_minimum_required(VERSION 3.5)
-
-if (POLICY CMP0048)
-  cmake_policy(SET CMP0048 NEW)
-endif (POLICY CMP0048)
```
juan-lunarg, line 8: Manually setting this policy is only necessary if the minimum is 2.X

It's safe to remove this code since the min is 3.6

https://cmake.org/cmake/help/latest/policy/CMP0048.html#policy:CMP0048

juan-lunarg:
@dinord why was the min raised to 3.13?

I'm fine with raising the min to 3.13 just curious:
- https://github.com/google/googletest/commit/4fed5f285030085b096c930ff03e42c7814739b2 <- Me
- https://github.com/google/googletest/commit/29836977ef1c664588daeb035591c18a912e7c50 <- Copybara-Service

dinord:
We used your change to bump the min to 3.13 because that's the oldest officially supported version of CMake, as per https://opensource.google/documentation/policies/cplusplus-support.

juan-lunarg:
Cool beans 👍🏾

PR #4290, opened 2023-06-20
cmake: Clean up policy code

juan-lunarg:
Now that the min is 3.13 these policies don't need to be set manually anymore.

CMP0054 - 3.1
CMP0063 - 3.3
CMP0069 - 3.9
CMP0077 - 3.13


dinord:
Thanks for the cleanup!

PR #4291, opened 2023-06-20
ci: Fix Windows CI

juan-lunarg:
For context this is the action:
https://github.com/ilammy/msvc-dev-cmd

We use it for various Vulkan repos to setup CMake + Ninja on Windows.


juan-lunarg:
Ah I misunderstood the issue. Sorry for the noise.

PR #4293, opened 2023-06-21
cmake: Remove remaining checks for CMAKE_VERSION

juan-lunarg:
Remove conditional code that doesn't need to exist anymore.

### DIFF
--- a/googlemock/CMakeLists.txt
+++ b/googlemock/CMakeLists.txt
@@ -100,18 +100,14 @@ else()
   target_link_libraries(gmock_main PUBLIC gmock)
   set_target_properties(gmock_main PROPERTIES VERSION ${GOOGLETEST_VERSION})
 endif()
-# If the CMake version supports it, attach header directory information
-# to the targets for when we are part of a parent build (ie being pulled
-# in via add_subdirectory() rather than being a standalone build).
-if (DEFINED CMAKE_VERSION AND NOT "${CMAKE_VERSION}" VERSION_LESS "2.8.11")
-  string(REPLACE ";" "$<SEMICOLON>" dirs "${gmock_build_include_dirs}")
-  target_include_directories(gmock SYSTEM INTERFACE
-    "$<BUILD_INTERFACE:${dirs}>"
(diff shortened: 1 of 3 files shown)


googlemock/CMakeLists.txt
```diff
@@ -135,11 +131,7 @@ if (gmock_build_tests)
   enable_testing()
 
   if (MINGW OR CYGWIN)
-    if (CMAKE_VERSION VERSION_LESS "2.8.12")
-      add_compile_options("-Wa,-mbig-obj")
-    else()
-      add_definitions("-Wa,-mbig-obj")
```
juan-lunarg, line 141: `add_definitions` is used to add defines to the code like `NOMINMAX` or `WIN32_LEAN_AND_MEAN`  using it for compiler options isn't correct.

Also `add_definitions` isn't recommended for usage anymore:
https://cmake.org/cmake/help/latest/command/add_definitions.html#command:add_definitions

googletest/cmake/internal_utils.cmake
```diff
@@ -291,34 +273,22 @@ endfunction()
...
-          --build_dir=${CMAKE_CURRENT_BINARY_DIR}/\${CTEST_CONFIGURATION_TYPE} ${ARGN})
-    endif()
-    # Make the Python import path consistent between Bazel and CMake.
-    set_tests_properties(${name} PROPERTIES ENVIRONMENT PYTHONPATH=${CMAKE_SOURCE_DIR})
-  endif(PYTHONINTERP_FOUND)
+  if (NOT Python3_Interpreter_FOUND)
+    return()
+  endif()
+
+  get_cmake_property(is_multi "GENERATOR_IS_MULTI_CONFIG")
```
juan-lunarg, line 280: Note `GENERATOR_IS_MULTI_CONFIG` was added in 3.9 and should be used here instead of `CMAKE_CONFIGURATION_TYPES`.

https://cmake.org/cmake/help/latest/prop_gbl/GENERATOR_IS_MULTI_CONFIG.html

juan-lunarg:
There are now no more instances of `CMAKE_VERSION` in the codebase 💯

dinord:
Thanks for the cleanup! Just for posterity, explicitly running python3 in cmake is based on the fact that googletest's python tests are actually written in python3 (as of a few years ago).

PR #4297, opened 2023-06-21
ci: Add CMake to Open Source GitHub Actions

juan-lunarg:
Linux-CMake-Min:
Test against 3.13 to prevent breaking backcompat.

Linux-CMake:
Tests a large matrix of combinations.

Notably testing against newest version of CMake available,
and testing single/multi configuration generator.

### DIFF
--- a/.github/workflows/gtest-ci.yml
+++ b/.github/workflows/gtest-ci.yml
@@ -6,12 +6,13 @@ on:
 
 env:
   BAZEL_CXXOPTS: -std=c++14
+  CTEST_OUTPUT_ON_FAILURE: 1
 
 permissions:
   contents: read
 
 jobs:
-  Linux:
+  Linux-Bazel:
     runs-on: ubuntu-latest
     steps:
 
@@ -22,7 +23,7 @@ jobs:
     - name: Tests
       run: bazel test --cxxopt=-std=c++14 --features=external_include_paths --test_output=errors ...
 
-  macOS:
+  macOS-Bazel:
     runs-on: macos-latest
     steps:
 
@@ -34,7 +35,7 @@ jobs:
       run:  bazel test --cxxopt=-std=c++14 --features=external_include_paths --test_output=errors ...
 
 
-  Windows:
+  Windows-Bazel:
     runs-on: windows-latest
     steps:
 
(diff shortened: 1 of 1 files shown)


.github/workflows/gtest-ci.yml
```diff
@@ -7,6 +7,9 @@ on:
 env:
   BAZEL_CXXOPTS: -std=c++14
 
+permissions:
+  contents: read
+
```
juan-lunarg, line 12: For some context:

https://github.com/KhronosGroup/Vulkan-Headers/pull/400

.github/workflows/cmake.yml
```diff
@@ -0,0 +1,29 @@
...
+  cmake_min:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        config: ["Debug", "Release"]
+    steps:
+      - uses: actions/checkout@v3
+      - uses: lukka/get-cmake@latest
+        with:
+          cmakeVersion: 3.13.0
```
juan-lunarg, line 25: This is the main thing I'm trying to add. We added this in the Vulkan Validation Layers since we found it very easy for developers using newer CMake version to add code that didn't respect the minimum.

dinord:
As maintainers, we do not use GitHub actions CI. Instead, we have a proprietary workflow that runs the scripts in https://github.com/google/googletest/tree/main/ci. Our existing GHA are purely maintained by the open-source community.

Despite that, sometimes users believe GoogleTest is broken because GHA is broken, even though our internal workflows are green. Can we make the CMake  CI as similar as possible to the internally used script?
https://github.com/google/googletest/blob/ec4fed93217bc2830959bb8e86798c1d86956949/ci/linux-presubmit.sh#L45

.github/workflows/cmake.yml
dinord, line 1: 1. Since we have a cmake workflow now, can we rename existing ones to clarify that they are bazel?
2. Can we stuff this job into [gtest-ci.yml](https://github.com/google/googletest/blob/ec4fed93217bc2830959bb8e86798c1d86956949/.github/workflows/gtest-ci.yml) to make it easier to keep track of the platforms that are being tested when editing the config?

.github/workflows/cmake.yml
```diff
@@ -0,0 +1,29 @@
...
+    push:
+    pull_request:
+        branches:
+            - main
+
+permissions:
+    contents: read
+
+jobs:
+  cmake_min:
```
dinord, line 16: Can we make this name consistent with other job names, or make them consistent with it?

.github/workflows/cmake.yml
```diff
@@ -0,0 +1,29 @@
...
+    runs-on: ubuntu-latest
```
dinord, line 17: Can we include the platform in the job's name? Just "linux" is probably sufficient.

juan-lunarg:
Closing this PR since I wasn't aware of the internal CI beforehand and am not sure how much contributing this would add to the project.

Issue #4306, opened 2023-06-30
[Bug]: GCC Warnings with -Wuseless-cast


jonathanmcdermid:
### Describe the issue

When building for GCC with -Wuseless-cast, there are some warnings in \googletest\include\gtest\internal\gtest-port.h

`return static_cast<int>(read(fd, buf, count));`

### Steps to reproduce the problem

1. Enable -Wuseless-cast flag on a GCC build
2. Do the build

### What version of GoogleTest are you using?

v1.13

### What operating system and version are you using?

Windows 11

### What compiler and version are you using?

GCC

### What build system are you using?

minGW64

pgawro:
Hi @jonathanmcdermid 
I could reproduce that problem also on Linux can you check if proposed changes by me works also under windows?

jonathanmcdermid:
Hi @pgawro I am still getting warnings for useless casts using your changes. I had to remove these two casts.
https://github.com/pgawro/googletest/blob/abe672c483bd1bd26b28977f1a00e2a758db5677/googletest/include/gtest/internal/gtest-port.h#L2160C6-L2160C6
https://github.com/pgawro/googletest/blob/abe672c483bd1bd26b28977f1a00e2a758db5677/googletest/include/gtest/internal/gtest-port.h#L2163

pgawro:
Hi @jonathanmcdermid Thanks for pointing that out. I looks like I skipped that unintentionally. I will be obliged if you could check if that works for you. I have updated commit.

jonathanmcdermid:
The new version works as expected, thanks.

jonathanmcdermid:
I've tried using your version on linux though and am still having issues. According to the existing comment, this const cast is redundant on C++17 and onward. https://github.com/pgawro/googletest/blob/63d67622ea58dd5b9f5529ae9d7a8737419b0fe2/googletest/src/gtest-port.cc#L1100C61-L1100C61

jonathanmcdermid:
The latest version works!

pgawro:
Now the question is if that is suitable to merge. @derekmauro Can you take a look to PR #4306

derekmauro:
https://github.com/google/googletest/pull/4307#issuecomment-1640275519

PR #4307, opened 2023-06-30
Fix for #4306 -Wuseless-cast

pgawro:


derekmauro:
This doesn't build in my tests. Some of these casts are necessary on some architectures since not all types have the same sizes everywhere. I don't think we should support this warning.

pgawro:
Ok, that as I thought

Ryanf55:
Would you accept pull requests that conditionally remove the casts on some platforms?

PR #4313, opened 2023-07-11
Fixed variables that could be declared 'const'

kimvaleen:
While running the clang-tidy checker named misc-const-correctness, I discovered that these variables should be declared 'const'. Please consider adding this to your next patch


sab222 (commented):
good luck 👍

pgawro (approved):

Issue #4321, opened 2023-07-25
[Bug]: Compile gtest-all.cc comes with error


lygstate:
### Describe the issue

```
In file included from ../src/gtest/src/gtest-all.cc:49:
../src/gtest/src/gtest.cc: In function 'void testing::internal::ColoredPrintf({anonymous}::GTestColor, const char*, ...)':
../src/gtest/src/gtest.cc:3308:19: error: '%s' directive argument is null [-Werror=format-overflow=]
 3308 |   printf("\033[0;3%sm", GetAnsiColorCode(color));
      |                   ^~
```

### Steps to reproduce the problem

https://gitlab.freedesktop.org/mesa/mesa/-/issues/9404

### What version of GoogleTest are you using?

[gtest tip](https://github.com/google/googletest/commit/40412d85124f7c6f3d88454583c4633e5e10fc8c)

### What operating system and version are you using?

GCC on Debian 12.

### What compiler and version are you using?

GCC on Debian 12.

### What build system are you using?

GCC on Debian 12.

pgawro:
I added small tweak. Can you check if that is helpful for you?

okias:
@pgawro thank you, fixed the issue for us!

pgawro:
Since PR is merged I think we can close this ticket

PR #4322, opened 2023-07-25
googletest: ansi color

pgawro:
Adds support for kDefault option and also prevents from returning nullptr by choosing default color.

Issue: #4321

### DIFF
--- a/googletest/src/gtest.cc
+++ b/googletest/src/gtest.cc
@@ -3228,7 +3228,8 @@ static const char* GetAnsiColorCode(GTestColor color) {
     case GTestColor::kYellow:
       return "3";
     default:
-      return nullptr;
+      assert(false);
+      return "9" ;
   }
 }
 


okias:
@pgawro thank you, fixed the issue for us!

googletest/src/gtest.cc
```diff
@@ -3217,8 +3217,7 @@ static WORD GetNewColor(GTestColor color, WORD old_color_attrs) {
 
 #else
 
-// Returns the ANSI color code for the given color. GTestColor::kDefault is
-// an invalid input.
+// Returns the ANSI color code for the given color.
```
derekmauro, line 3220: I think the code fix for the warning is fine, but I don't want to expand the support for kDefault. That is to say, we should keep the current contract for this function. Can we add an `assert(false)` on line 3231 before returning "9"?

derekmauro (changes requested):

googletest/src/gtest.cc
okias: ping @pgawro

pgawro: @okias Right I made code changes but forget about comment. Sorry for that. Now it's up to @derekmauro since requested changes are made

PR #4327, opened 2023-08-02
Suppress clang-tidy warnings about goto statements

### LINKED ISSUE #2442: Suppress Clang-Tidy warnings

georgthegreat:
Partially fixes #2442

### DIFF
--- a/googletest/include/gtest/internal/gtest-death-test-internal.h
+++ b/googletest/include/gtest/internal/gtest-death-test-internal.h
@@ -227,6 +227,7 @@ inline Matcher<const ::std::string&> MakeDeathTestMatcher(
             #statement,                                                        \
             ::testing::internal::MakeDeathTestMatcher(regex_or_matcher),       \
             __FILE__, __LINE__, &gtest_dt)) {                                  \
+      /* NOLINTNEXTLINE(cppcoreguidelines-avoid-goto, hicpp-avoid-goto) */     \
       goto GTEST_CONCAT_TOKEN_(gtest_label_, __LINE__);                        \
     }                                                                          \
     if (gtest_dt != nullptr) {                                                 \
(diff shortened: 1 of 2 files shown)


dinord:
We typically don't accept `NOLINT`s in our headers. Although this is a macro, the `goto` is just an internal implementation detail.

I'd file a bug against Clang-Tidy to provide a way to disable it these types of warnings.

georgthegreat:
> I'd file a bug against Clang-Tidy to provide a way to disable it these types of warnings.

But using NOLINTNEXTLINE is the way for clang-tidy to disable these types of warning.
You just have to add itto your code.

Once you will get rid of these goto statements, the NOLINT stanzas can be removed too.

georgthegreat:
And, btw, 

> We typically don't accept NOLINTs in our headers

This just is not true.
There are NOLINT marks all around the code.

@dinord, can we reopen and merge this?

PR #4330, opened 2023-08-03
Remove public includes of `<iomanip>`

IncludeGuardian:
This commit removes `<iomanip>` from public GoogleTest header files.

As `<iomanip>` is not a common included file, its content is unlikely to be included in translation units other than through GoogleTest includes.

By reducing the number of include directives public headers in GoogleTest, this may reduce the time taken to compile tests as it would reduce the amount of work that the preprocessor and compiler front-end need to do.

### DIFF
--- a/googletest/include/gtest/gtest.h
+++ b/googletest/include/gtest/gtest.h
@@ -51,7 +51,6 @@
 
 #include <cstddef>
 #include <cstdint>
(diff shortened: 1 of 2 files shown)


joshqti:
Can we revert this change?  This broke our automated build system.  That header is used.

PR #4334, opened 2023-08-07
Fix RETest/1.ImplicitConstructorWorks on non-ABSL platforms

ngie-eign:
The last regular expression specified in the test is not technically POSIX compatible. Use `[[:alnum:]_]` instead of `\w+`; the latter is a Perl-compatible regular expression.

### DIFF
--- a/googletest/test/googletest-port-test.cc
+++ b/googletest/test/googletest-port-test.cc
@@ -418,8 +418,8 @@ TYPED_TEST(RETest, ImplicitConstructorWorks) {
   const RE simple(TypeParam("hello"));
   EXPECT_STREQ("hello", simple.pattern());
 
-  const RE normal(TypeParam(".*(\\w+)"));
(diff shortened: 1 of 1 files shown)


DanAlbert:
Which platform were you trying to fix? aiui this was a breaking change to how gtest interprets regex on most platforms. The two I know which don't support the perl-regex classes with `REG_EXTENDED` are NetBSD and Android (because Android's implementation comes from NetBSD). Both can be fixed without breaking compatibility by using `REG_EXTENDED | REG_GNU`. I've sent an (internal, annoyingly) CL to fix that, but if this needs to work on some other platform where that fix doesn't work then that's probably not the right fix.

svmhdvn:
I don't think this merge fixes the issue correctly on FreeBSD. See https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=241130, where it shows that the '\w' expression is not valid on FreeBSD either. The FreeBSD port still shows errors in the test suite. Can you take a look?

Issue #4336, opened 2023-08-10
[FR]: std::stacktrace support


strager:
### Does the feature exist in the most recent commit?

no

### Why do we need this feature?

I want to get pretty stack traces in my tests without depending on Abseil.

In addition, Microsoft STL's implementation of std::stacktrace provides more information than Abseil, such as file and line numbers.

I think this feature would be generally useful, thus I think this feature should be built into googletest instead of users leveraging the (currently broken #1020) `OsStackTraceGetterInterface` extension point.

### Describe the proposal.

Approach A:
Change the existing `OsStackTraceGetter` class to use std::stacktrace if available, falling back to Absel or a no-op implementation if std::stacktrace unavailable.

Problem with approach A:
This would involve adding `#if`s in header files which might lead to ODR problems.

Approach B:
1. Split `OsStackTraceGetter` into two classes: `AbslStackTraceGetter` (which uses Abseil; conditionally defined) and `NoopStackTraceGetter` (always defined).
2. Add `StdStackTraceGetter` (which uses std::stacktrace; conditionally defined).
3. Make `UnitTestImpl::os_stack_trace_getter` choos `StdStackTraceGetter` if available, falling back to `AbslStackTraceGetter` or `NoopStackTraceGetter` if std::stacktrace is unavailable.

(This seems to be how `OsStackTraceGetterInterface` was meant to be used.)

### Is the feature specific to an operating system, compiler, or build system version?

Currently, std::stacktrace is available on recent MSVC versions for Windows. It might be available with effort on GCC/libstdc++. libc++ does not have an implementation.

Boost has a somewhat-portable implementation which looks similar to std::stacktrace. I don't propose using Boost's version.

strager:
I am willing to implement this feature myself. I prefer approach B. As an alternative, we could make `OsStackTraceGetterInterface` a public API and leave users to define their own std::stacktrace-using implementations (i.e. implement #1020).

strager:
I implemented approach B in #4337.

derekmauro:
If I recall correctly there are two uses of stack tracing in GoogleTest. One of them is to print a stack trace when an expectation fails. The other is to print a stack trace when your test crashes (via the environment variable `GTEST_INSTALL_FAILURE_SIGNAL_HANDLER=1`).

`std::stacktrace` makes no guarantees about signal safety, while Abseil's is async signal safe, at least on Linux. The lack of signal safety is a problem for the crash handler, even if it happens to work some of the time.

[...]

PR #4337, opened 2023-08-10
Print stacktraces using std::stacktrace if available

strager:
Reviewers: See issue #4336 for some design considerations. This PR chooses approach B. The first commit changes the design, and the second commit adds the std::stacktrace feature. If this should be split into two separate pull requests (one commit per PR), let me know.

If std::stacktrace is available (currently only GCC (opt-in feature) and MSVC), use it for printing stack traces.

If both Abseil and std::stacktrace are available, prefer std::stacktrace.

Advantages of std::stacktrace over Abseil:

* Includes file and line information
* Includes DLL information
* Better stack trace quality on Windows

Disadvantages of std::stacktrace compared to Abseil:

* Runs more slowly on Windows
* Includes information which might not be relevant such as byte offsets
* Printed lines are longer thus harder to skim

Before (Windows using Abseil):

    [ RUN      ] FooTest.Xyz
    [snip]\googletest-filter-unittest_.cc(49): error: Failed
    Expected failure.
    Stack trace:
      00007FF69EF90E8D: testing::internal::HandleSehExceptionsInMethodIfSupported<testing::Test,void>
      00007FF69EF90AC3: testing::internal::HandleExceptionsInMethodIfSupported<testing::Test,void>
      00007FF69EF5C01B: testing::Test::Run
      00007FF69EF5CCC5: testing::TestInfo::Run
      00007FF69EF5D6EC: testing::TestSuite::Run
    ... Google Test internal frames ...

    [  FAILED  ] FooTest.Xyz (21 ms)

After (Windows using MSVC STL's std::stacktrace):

    [ RUN      ] FooTest.Xyz
    [snip]\googletest-filter-unittest_.cc(49): error: Failed
    Expected failure.
    Stack trace:
      00007FF6770283A9: googletest_filter_unittest_!`anonymous namespace'::FooTest_Xyz_Test::TestBody+0x89 [snip]\googletest-filter-unittest_.cc:49
      00007FF677080E8D: googletest_filter_unittest_!testing::internal::HandleSehExceptionsInMethodIfSupported<testing::Test,void>+0x1D [snip]\gtest.cc:2609
      00007FF677080AC3: googletest_filter_unittest_!testing::internal::HandleExceptionsInMethodIfSupported<testing::Test,void>+0x73 [snip]\gtest.cc:2652
      00007FF67704C01B: googletest_filter_unittest_!testing::Test::Run+0xAB [snip]\gtest.cc:2698
    ... Google Test internal frames ...

    [  FAILED  ] FooTest.Xyz (127 ms)

### DIFF
--- a/googletest/src/gtest.cc
+++ b/googletest/src/gtest.cc
@@ -4968,9 +4968,9 @@ void StreamingListener::SocketWriter::MakeConnection() {
 const char* const OsStackTraceGetterInterface::kElidedFramesMarker =
     "... " GTEST_NAME_ " internal frames ...";
 
-std::string OsStackTraceGetter::CurrentStackTrace(int max_depth, int skip_count)
-    GTEST_LOCK_EXCLUDED_(mutex_) {
 #ifdef GTEST_HAS_ABSL
+std::string AbslStackTraceGetter::CurrentStackTrace(int max_depth, int skip_count)
+    GTEST_LOCK_EXCLUDED_(mutex_) {
   std::string result;
 
   if (max_depth <= 0) {
@@ -5010,16 +5010,11 @@ std::string OsStackTraceGetter::CurrentStackTrace(int max_depth, int skip_count)
   }
 
   return result;
-
-#else   // !GTEST_HAS_ABSL
-  static_cast<void>(max_depth);
-  static_cast<void>(skip_count);
-  return "";
-#endif  // GTEST_HAS_ABSL
 }
+#endif  // GTEST_HAS_ABSL
 
-void OsStackTraceGetter::UponLeavingGTest() GTEST_LOCK_EXCLUDED_(mutex_) {
 #ifdef GTEST_HAS_ABSL
+void AbslStackTraceGetter::UponLeavingGTest() GTEST_LOCK_EXCLUDED_(mutex_) {
   void* caller_frame = nullptr;
   if (absl::GetStackTrace(&caller_frame, 1, 3) <= 0) {
     caller_frame = nullptr;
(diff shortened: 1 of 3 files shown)


snnn:
Recently, we found a problem that: after gtest started taking dependency on abseil, the "::testing::InitGoogleTest" function will call SymInitialize Windows API indirectly. However, the API can only be called once per process. For example, 
```c++
#include <iostream>
#include <Windows.h>
#include <DbgHelp.h>

int main()
{
    HANDLE process = GetCurrentProcess();
    SymSetOptions(SYMOPT_DEFERRED_LOADS | SYMOPT_UNDNAME);
    if (!SymInitialize(process, nullptr, true)) {
        return -1;
    }
    if (!SymInitialize(process, nullptr, true)) {
        return -1;
    }
    return 0;
}

```

The second SymInitialize call would definitely fail.  So, if the library we are testing also has its own backtrace implementation, the extra "SymInitialize" function call would cause trouble.  I wonder how it is handled in VC++ runtime's "<stacktrace>" library.

snnn:
I think std::stacktrace has two benefits compared to the existing one:
1. You don't need to worry about initialization/cleanup. The C++ runtime should handle it.
2. It dynamically loads dbghelp.dll. So the dependency is optional.

zrno:
@derekmauro any plans to merge this?

yuslepukhin:
> Recently, we found a problem that: after gtest started taking dependency on abseil, the "::testing::InitGoogleTest" function will call SymInitialize Windows API indirectly. However, the API can only be called once per process. For example,
> 
> ```c++
> #include <iostream>
> #include <Windows.h>
> #include <DbgHelp.h>
> 
> int main()
> {
>     HANDLE process = GetCurrentProcess();
>     SymSetOptions(SYMOPT_DEFERRED_LOADS | SYMOPT_UNDNAME);
>     if (!SymInitialize(process, nullptr, true)) {
>         return -1;
>     }
>     if (!SymInitialize(process, nullptr, true)) {
>         return -1;
>     }
>     return 0;
> }
> ```
> 
> The second SymInitialize call would definitely fail. So, if the library we are testing also has its own backtrace implementation, the extra "SymInitialize" function call would cause trouble. I wonder how it is handled in VC++ runtime's "<stacktrace>" library.

Is this getting some attention? All that is need to be done is not to use `GetCurrentProcess()`

googletest/src/gtest.cc
```diff
@@ -5010,24 +5014,87 @@ std::string OsStackTraceGetter::CurrentStackTrace(int max_depth, int skip_count)
...
+    caller_frame = stack[0].native_handle();
+  }
+
+  MutexLock lock(&mutex_);
+  caller_frame_ = caller_frame;
+}
 #endif  // GTEST_HAS_ABSL
+
+std::string NoopStackTraceGetter::CurrentStackTrace(int max_depth, int skip_count) {
+  static_cast<void>(max_depth);
```
AbdulTade, line 5091: Are these neccessary ?

Issue #4167, opened 2023-02-23
[Bug]: CaptureStream (stdout, stderr) not working on recent Android version for app environment


richard1122:
### Describe the issue

While running GTest from JNI, we may need to `CaptureStdout` to a file and redirect to logcat.

Right now it uses `/data/local/tmp/`, which is writeable when it's a standalone executable by running through `adb shell`, but not when it's in dalvik/art.

Related: https://github.com/google/googletest/blob/39a26e12d67ed6c21feeb606372bfee39a8e6d53/googletest/src/gtest-port.cc#L1049-L1061

### Steps to reproduce the problem

Adb environment:
```bash
adb shell
emu64a:/ $ touch /data/local/tmp/temp-file-adb
emu64a:/ $ echo "content" > /data/local/tmp/temp-file-adb
emu64a:/ $ cat /data/local/tmp/temp-file-adb
content
```

APP environment (Use run-as to emulate):
```bash
run-as com.richard.testing
touch /data/local/tmp/temp-file-app
touch: '/data/local/tmp/temp-file-app': Permission denied
```

### What version of GoogleTest are you using?

03597a01ee50ed33e9dfd640b249b4be3799d395

### What operating system and version are you using?

Android 13, arm64

### What compiler and version are you using?

Android NDK 25.1.8937393
Clang:
```
Android (8490178, based on r450784d) clang version 14.0.6 (https://android.googlesource.com/toolchain/llvm-project 4c603efb0cca074e9238af8b4106c30add4418f6)
Target: x86_64-apple-darwin22.3.0
Thread model: posix
InstalledDir: /Users/richard/Library/Android/sdk/ndk/25.1.8937393/toolchains/llvm/prebuilt/darwin-x86_64/bin
```

### What build system are you using?

3.25.2

### Additional context

I found we have existing code to get temp folder path from ENV, and fallback to `/data/local/tmp`: https://github.com/google/googletest/blob/2057566e4e16c88f1fea4d6c96b2e2bfb87507a6/googletest/src/gtest.cc#L6801-L6802

If you think this is a good to keep the same logic in `CaptureStream`? Which will make it works in ADB and APP environment. If so, I can open a PR later.

PR #4342, opened 2023-08-18
Prefer $TMPDIR to /data/local/tmp on Android

tanzislam:
Newer devices can have the latter location read-only. (I observed this with Termux on a non-rooted Pixel 6.)

Fixes: #4167


dinord:
I generally agree with the change, but maybe we could reuse an existing utility function by declaring it in `gtest-port.h`: https://github.com/google/googletest/blob/460ae98267b83db4ca2730d368d53f8aee3b636e/googletest/src/gtest.cc#L6835

PR #4345, opened 2023-08-21
Fixed #4341

### LINKED ISSUE #4341: Unencrypted HTTP Links to external websites

no

Multiple markdown files (*.md) across the project contain public labeled links that hold unencrypted HTTP links, e.g. [Link Label](http://)

Hence, when a user clicks such a link, an unencrypted HTTP request is made.

A passive eavesdropper with the ability to monitor the network traffic of the target could potentially intercept and peruse these unencrypted requests and track the target's browsing activity on these websites.

Among the links one can find: code.google.com, groups.google.com,  github.com and wikipedia.org.

### Describe the proposal.

Change public links so that they use HTTPS
For example edit line 103 on googletest-main/README.md 
so that (**http://www.chromium.org/**) will change into (**https://www.chromium.org/**)

no

Roshan4518Git:
Solved issue number #4341

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,7 +100,7 @@ tools.
 In addition to many internal projects at Google, GoogleTest is also used by the
 following notable projects:
 
-*   The [Chromium projects](http://www.chromium.org/) (behind the Chrome browser
+*   The [Chromium projects](https://www.chromium.org/) (behind the Chrome browser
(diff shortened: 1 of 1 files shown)


dinord:
Fixed by #4343.

PR #4346, opened 2023-08-21
Update README.md

mahaamimran:
An easy guide to macOS users to install, build, compile and execute test cases.


dinord:
This is very similar to our cmake quickstart doc: http://google.github.io/googletest/quickstart-cmake.html

I realize that this PR suggests adding macOS terminal specifics, but we have no interest in keeping those up to date, and would rather have the quickstart pages (for cmake and bazel) be our canonical tutorials.

PR #4353, opened 2023-08-25
fix: correct threadsafe preprocessor check

ciband:
Corrected logic in checking the define for GTEST_IS_THREADSAFE.  The block was being entered when GTEST_IS_THREADSAFE was defined at all vs. being defined and set to a value of 1.


derekmauro:
The code is correct as written. Please see https://github.com/google/googletest/blob/8a6feabf04bec8fb125e0df0ad1195c42350725f/googletest/include/gtest/internal/gtest-port.h#L154-L174

ciband:
@derekmauro , I see now.  Sorry for my mistake.  Thank you.

PR #4358, opened 2023-08-30
Enable auto colors on MinGW

orgads:
MinGW is not MSYS. The binary runs natively, just like MSVC.


orgads:
Rebased #3049, looks much more complete than my solution.

PR #4368, opened 2023-09-05
Add MODULE.bazel to support bzlmod

Vertexwahn:
Life with [Bzlmod](https://bazel.build/external/overview#bzlmod) is easier if every dependency provides a `MODULE.bazel` file. Therefore, the main purpose of this PR is to introduce a `MODULE.bazel` file in googletest. The `MODULE.bazel` can coexist with the `WORKSPACE` file. Users have to enable bzlmod via `--enable_bzlmod` flag (more details to the Bzlmod migration plan can be found [here](https://blog.bazel.build/2023/07/24/whats-new-with-bzlmod.html#high-level-roadmap-for-the-future)).

This PR is similar to https://github.com/google/googletest/pull/4118#pullrequestreview-1611267844

Differences:
- Uses newer versions of modules
- Uses `archive_ovrride` to support live at head philosophy
- Update checkout action (v3 -> v4) in GitHub workflow
- Reformates old WORKSPACE file using `buildifier`

Further remarks:
- [module_dot_bazel.patch](https://github.com/google/googletest/pull/4368/files#diff-e208d606896d1d6088d3b4a50b0657bc53aa829efc32d2995bc5d9d2c345da7e) is needed since abseil-cpp does not provide a `MODULE.bazel` file unit now - this is a chicken egg situation between the public GitHub repos of googletest and abseil

I recommend closing https://github.com/google/googletest/pull/4118#pullrequestreview-1611267844 and merging this PR

### DIFF
--- a/.github/workflows/gtest-ci.yml
+++ b/.github/workflows/gtest-ci.yml
@@ -4,40 +4,45 @@ on:
   push:
   pull_request:
 
-env:
-  BAZEL_CXXOPTS: -std=c++14
-
 jobs:
   Linux:
     runs-on: ubuntu-latest
     steps:
 
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v4
       with:
         fetch-depth: 0
 
     - name: Tests
       run: bazel test --cxxopt=-std=c++14 --features=external_include_paths --test_output=errors ...
 
+    - name: Tests with bzlmod
+      run: bazel test --cxxopt=-std=c++14 --features=external_include_paths --test_output=errors --enable_bzlmod ...
+      
   macOS:
     runs-on: macos-latest
     steps:
 
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v4
       with:
         fetch-depth: 0
 
     - name: Tests
-      run:  bazel test --cxxopt=-std=c++14 --features=external_include_paths --test_output=errors ...
+      run: bazel test --cxxopt=-std=c++14 --features=external_include_paths --test_output=errors ...
 
+    - name: Tests with bzlmod
+      run: bazel test --cxxopt=-std=c++14 --features=external_include_paths --test_output=errors --enable_bzlmod ...
 
   Windows:
     runs-on: windows-latest
     steps:
 
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v4
       with:
         fetch-depth: 0
 
     - name: Tests
       run: bazel test --cxxopt=/std:c++14 --features=external_include_paths --test_output=errors ...
+
+    - name: Tests with bzlmod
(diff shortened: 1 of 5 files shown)


Vertexwahn:
@derekmauro Please review!

Vertexwahn:
@derekmauro Should I open a PR on abseil-cpp?

derekmauro:
> As long as both googletest and abseil-cpp are published to the BCR, the end users can always specify the latest version for both of them, and Bzlmod will resolve the dependencies. I don't see any chicken-egg problem?

Sure, that works. But what if you are only using GoogleTest directly? You might not even know that you depend on Abseil.

derekmauro:
> @derekmauro Should I open a PR on abseil-cpp?

Let's figure out how to resolve this first.

meteorcloudy:
> Sure, that works. But what if you are only using GoogleTest directly? You might not even know that you depend on Abseil.

I assume googletest will introduce a suitable abseil version automatically? It might not be the latest one, but it should work with googletest?

derekmauro:
> I assume googletest will introduce a suitable abseil version automatically? It might not be the latest one, but it should work with googletest?

What does it mean to "introduce a suitable abseil version automatically"?

meteorcloudy:
In googletest's MODULE.bazel file, depend on abseil-cpp via a `bazel_dep` statement.

derekmauro:
I think this misses the point I made above. Let me try an explicit example.

Abseil and GoogleTest are both at version 1.
Abseil introduces feature X in a dev commit.
GoogleTest uses feature X in a dev commit. It is able to do this because it uses the archive_override feature to support live at head philosophy.
GoogleTest introduces feature Y in a dev commit.
Abseil uses feature Y in a dev commit. Again, it is able to this because it uses the archive_override feature to support live at head philosophy.

Now it comes time to release Abseil and GoogleTest version 2.
If I release GoogleTest version 2 first, I can't refer to any released version of Abseil in MODULE.bazel, because none of them have feature X.
If I release Abseil version 2 first, I can't refer to any released version of GoogleTest in MODULE.bazel, because none of them have feature Y.

meteorcloudy:
That's for the example, I see the problem now.

I guess, we need to decouple the release processes and increase the frequency to address this issue.

Imagine if you release a new version every time you had a new commit to each of the libraries, we won't have this problem since this will fallback to the WORKSPACE use case (where you can point to any commit).

But that's the extreme case, we only need to have a new abseil version once feature X is introduced so that googletests can depend on a released version of abseil and vice versa.

Do you think this would be feasible? I'm not sure how expensive it is to make a new release for those two projects.

derekmauro:
Support has been added in https://github.com/google/googletest/commit/6a5938233b6519ba99ddb7c7314d45d3fa877969. We will publish support to BCR after the next release.

PR #4376, opened 2023-09-24
Deduce value type in container matchers

### LINKED ISSUE #4375: [Bug]: Container matchers fail to build on containers that do not expose value_type
### Describe the issue

Trying to use container matchers on containers that do not expose `value_type` as a member type results in a compiler error. This prevents from using the matchers on custom user containers that do not follow STL naming conventions for member types.

The value type could be deduced from the iterator type using `std::iterator_traits`, so the container should not need to expose `value_type` as a member.

Minimal reproducible example:
```cpp
// A minimal container that does not expose value_type
class ContainerWithoutValueType {
  using UnderlyingContainer = std::vector<int>;

 public:
  ContainerWithoutValueType(std::initializer_list<int> const args) : _v{args} {}

  UnderlyingContainer::const_iterator begin() const { return _v.begin(); }
  UnderlyingContainer::const_iterator end() const { return _v.end(); }
  UnderlyingContainer::size_type size() const { return _v.size(); }

  friend bool operator==(ContainerWithoutValueType const& lhs,
                         ContainerWithoutValueType const& rhs) {
    return lhs._v == rhs._v;
  }

 private:
  UnderlyingContainer _v;
};

TEST(PointwiseTest, DeducesValueType) {
  const ContainerWithoutValueType container{1, 2, 3};
  EXPECT_THAT(container, Pointwise(Eq(), container));
}
```

e40661d89b051e9ef4eb8a2420b74bf78b39ef41

Arch Linux (kernel 6.1.39-1-lts)

gcc version 13.1.1 20230714 (GCC) 

cmake version 3.27.1

### Additional context

_No response_

m-pilia:
Avoid directly using the member `value_type` of a container, and deduce the value type from the iterator instead (using `std::iterator_traits`). This allows to use container matchers on containers that do not expose their value type as a member type.

Resolves #4375

### DIFF
--- a/googlemock/include/gmock/gmock-matchers.h
+++ b/googlemock/include/gmock/gmock-matchers.h
@@ -2404,8 +2404,7 @@ class BeginEndDistanceIsMatcher {
 // elements in the containers (which don't properly matter to sets, but can
 // occur if the containers are vectors or lists, for example).
 //
-// Uses the container's const_iterator, value_type, operator ==,
-// begin(), and end().
+// Uses the container's const_iterator, operator ==, begin(), and end().
 template <typename Container>
 class ContainerEqMatcher {
  public:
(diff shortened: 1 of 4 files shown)


dinord:
Closing since we closed #4375, see explanation there.

PR #4391, opened 2023-10-12
fix: explicitly copy-convert to StringType in StartsWith/EndsWith matchers

muggenhor:
Instead of relying on automatic conversions, which won't happen if they're explicit! This fails with `boost::string_view` (GCC) and `std::string_view` (GCC and Clang both) for example.


muggenhor:
This change is not significant enough that I believe it's copyrightable. And I don't have the time to spend reading legal documents.

So do as you will with this change. E.g. go for your "own" fix when trying to use these matchers with a function taking `boost::string_view`/`std::string_view` parameter types on GCC.

Issue #4387, opened 2023-10-11
[FR]: Better exception printing for multi-line exceptions


vogelsgesang:
### Does the feature exist in the most recent commit?

no. I just double-checked the source code from [current main](https://github.com/google/googletest/blob/2dd1c131950043a8ad5ab0d2dda0e0970596586a/googletest/src/gtest.cc#L2570C20-L2570C45).

### Why do we need this feature?

This makes GTest easier to use, by making its messages slightly more readable. This is clearly nice-to-have. The patch is probably only 15 additional lines of code + test cases. We are happy to provide a patch for it.

### Describe the proposal.

For exceptions with multi-line `e.what()` descriptions, GTest currently prints

```
C++ exception with description "Error: 22000
Error while communicating with S3
DETAIL: aws-s3-bucket-name: mybucket
INTERNAL DETAIL: aws-s3-bucket-name: mybucket" thrown in the test body.
```

Instead, we would prefer 

```
C++ exception with description
> Error: 22000
> Error while communicating with S3
> DETAIL: aws-s3-bucket-name: mybucket
> INTERNAL DETAIL: aws-s3-bucket-name: mybucket
thrown in the test body.
```

We would propose to adjust `FormatCxxExceptionMessage` to detect if the `description` contains any newline characters. If it contains no newlines, keep the logic as is today (i.e. print the error message on the same line).

If the description contains newlines, print the exception with newlines in it replace every `\n` by `\n> ` before printing (of course, with special handling to correctly handle the potential presence/absence of a trailing `\n`).

### Is the feature specific to an operating system, compiler, or build system version?

no. This will be portable across all compilers

avrdan:
```
    auto strDesc = std::regex_replace(description, std::regex("\n"), "\n> ");
    auto trailingDesc = strDesc.substr(strDesc.size()-3);
    if (trailingDesc == "\n> ") {
      strDesc = strDesc.substr(0, strDesc.size()-3);
    }
    auto leadingDesc = strDesc.substr(0, 3);
    if (leadingDesc != "\n> ") {
      strDesc = "\n> " + strDesc;
    }
    message << "C++ exception with description \"" << strDesc << "\"";
```
This takes care of the following cases:

"Error: 22000\nError while communicating with S3\nDETAIL: aws-s3-bucket-name: mybucket\nINTERNAL DETAIL: aws-s3-bucket-name: mybucket\n"
[...]

PR #4393, opened 2023-10-14
Better exception printing for multi-line exceptions #4387

abhiota:
This makes GTest easier to use, by making its messages slightly more readable. This is clearly nice-to-have. The patch is probably only 15 additional lines of code + test cases. We are happy to provide a patch for it.

For exceptions with multi-line e.what() descriptions, GTest currently prints:
```
C++ exception with description "Error: 22000
Error while communicating with S3
DETAIL: aws-s3-bucket-name: mybucket
INTERNAL DETAIL: aws-s3-bucket-name: mybucket" thrown in the test body.
```

Modifications:
```
> Error: 22000
thrown in the test body.
```

The `FormatCxxExceptionMessage` function will detect if the description contains any newline characters. If it contains no newlines, keep the logic as is it was previously (i.e. print the error message on the same line). If the description contains newlines, print the exception with newlines in it and replace every `\n` by `\n>`  before printing (of course, with special handling to correctly handle the potential presence/absence of a trailing `\n`).

### DIFF
--- a/googletest/src/gtest.cc
+++ b/googletest/src/gtest.cc
@@ -2571,7 +2571,14 @@ static std::string FormatCxxExceptionMessage(const char* description,
                                              const char* location) {
   Message message;
   if (description != nullptr) {
-    message << "C++ exception with description \"" << description << "\"";
+    std::string desc(description);
+    if (desc.find('\n') == std::string::npos) {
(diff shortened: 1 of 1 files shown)


derekmauro:
https://github.com/google/googletest/issues/4387#issuecomment-1915528101

Issue #1930, opened 2018-10-24
gtest should support output stream redirect


ftrofin:
The default reporter (PrettyUnitTestResultPrinter) writes straight to stdout via printf calls. I would like to have an API so I can provide my own std::ostream, like boost Test library: https://www.boost.org/doc/libs/1_68_0/libs/test/doc/html/boost_test/test_output/logging_api/log_ct_output_stream_redirection.html

This is useful when you use gtest as part of a bigger framework and you need to redirect to output somewhere else (or convert it on the fly to a different encoding for example).

If for running the tests I can at least work around the problem by writing an event listener and writing into my own stream, for *listing* the tests, there is no workaround since `ListTestsMatchingFilter` is not accessible and the listeners are not invoked for the list tests action.

ftrofin:
Before we even talk about a PR let's have a conversation about how this should be approached. There are many ways in which this can be done but all of them imply lots of changes to the existing code. Here are two that come immediately to mind.
1. Replace all printf calls with std::cout << ... The advantage of doing this is that it's very easy to redirect cout (much easier than stdout).
2. Pass in the output stream as a parameter to RunAllTests(). Redirect all output to it.

I'm guessing I'm not the first one thinking about this and there might be already significant effort invested in the design of this feature. Let's discuss and agree on the best approach before spending lots of time to modify the code.

gennadiycivil:
Lack of participation signals lack of interest from the community. Closing

DannyVlasenko:
> Lack of participation signals lack of interest from the community. Closing

How many people from the community should vote for this issue to make you reopen it?

ftrofin:
So which approach should we take for this? See the two options I layed out above (or it can be a new one...)

milladgit:
This would be a great feature in the embedded world! In many cases, we need to redirect the textual output to a variable and then return that content back to our desktop system in the near future. I second this.

CraigHutchinson:
Any progress on this? I am running on Android and don't see any output from my application so I currently expect this is my issue as from what I understand printf() goes to dev/nul

gaspardpetit:
@CraigHutchinson for Android, I have this PR that's been waiting for almost a year: https://github.com/google/googletest/pull/3864 

For the printf case, I attempted a macro solution in 2018 which was closed because macros are evil: https://github.com/google/googletest/pull/1201 and I tried again with a `gtest_printf` solution last year but again, no interest to merge since: https://github.com/google/googletest/pull/3860

[...]

PR #4402, opened 2023-10-29
Add public interface for XmlUnitTestResultPrinter

LittleFox94:
Allowing custom gtest_main implementations to instantiate the XmlUnitTestResultPrinter with either a given file path or any other kind of std::ostream to write to.

This is useful for e.g. embedded cases where an XML report is still wanted, but not file system is available, by instantiating with a std::stringstream and delivering the data via any custom mean.

Related to #1930

This clearly needs some more polishing (e.g. docs in the header), but before I'll do the chores, I'd like some feedback :)

### DIFF
--- a/googletest/include/gtest/gtest.h
+++ b/googletest/include/gtest/gtest.h
@@ -1018,6 +1018,23 @@ class EmptyTestEventListener : public TestEventListener {
   void OnTestProgramEnd(const UnitTest& /*unit_test*/) override {}
 };
 
+class XmlUnitTestResultPrinter : public EmptyTestEventListener {
+ public:
+  XmlUnitTestResultPrinter(const char* output_file);
+
+  XmlUnitTestResultPrinter(std::ostream* output_stream);
+
+  XmlUnitTestResultPrinter(const XmlUnitTestResultPrinter&) = delete;
+  XmlUnitTestResultPrinter& operator=(const XmlUnitTestResultPrinter&) = delete;
+
+  void OnTestIterationEnd(const UnitTest& unit_test, int iteration) override;
+  void ListTestsMatchingFilter(const std::vector<TestSuite*>& test_suites);
+
+ private:
+  const std::string output_file_;
+  std::ostream* output_stream_;
+};
+
 // TestEventListeners lets users add listeners to track events in Google Test.
 class GTEST_API_ TestEventListeners {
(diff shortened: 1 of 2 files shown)


LittleFox94:
For some context, this is an example how that might be used:
https://praios.lf-net.org/littlefox/lf-os_amd64/-/blame/d499372dc4d35ddc468027955006f4af3edcebe8/src/runtime-tests/gtest_main.cxx#L19-21

(this is not merged to my main yet, so I hope that commit hash stays non-GC'd long enough to be useful ^^')

dinord:
Making this a public interface generally sounds good to me.

googletest/include/gtest/gtest.h
```diff
@@ -1018,6 +1018,27 @@ class EmptyTestEventListener : public TestEventListener {
...
+  XmlUnitTestResultPrinter(std::ostream* output_stream);
+
+  XmlUnitTestResultPrinter(const XmlUnitTestResultPrinter&) = delete;
+  XmlUnitTestResultPrinter& operator=(const XmlUnitTestResultPrinter&) = delete;
+
+  void OnTestIterationEnd(const UnitTest& unit_test, int iteration) override;
+  void ListTestsMatchingFilter(const std::vector<TestSuite*>& test_suites);
+
+ private:
+#if GTEST_HAS_FILE_SYSTEM
```
dinord, line 1036: Do you think we could have a separate interface for "output sink"? 

Perhaps we could even turn `XmlUnitTestResultPrinter` into an abstract class, and have a `FileXmlUnitTestResultPrinter` vs. a `StreamXmlUnitTestResultPrinter`?

LittleFox94: what do you think about having `XmlUnitTestResultPrinter` be the one that prints to a stream and `FileXmlUnitTestResultPrinter` inheriting from that, opening a file stream to give to the base?

I don't think the additional abstract base, from which both `Stream..` and `File..` inherit is useful here

LittleFox94:
> Making this a public interface generally sounds good to me.

thank you for that feedback - sorry for getting back here this late, will look into tidying this up, hopefully soon :)

byt3n33dl3 (commented):
Just want to see (XmlUnitTestResultPrinter)

PR #4403, opened 2023-10-29
[#4387] Better exception printing for multi-line exceptions.

avrdan:
Solution for multi-line exceptions:

- regex replace
- adds leading _\n_ if it does not exist, does not add another otherwise
- deals with trailing \n -> same output with or without a trailing _\n_, but note that two trailing \n\n will result in a line that starts with _>_
- changes default behavior to use a line with _>_ 

**Examples**

New line after each segment:

```
> Error: 22000
thrown in the test body.
```

Default behavior if no _\n_ chars detected:

```
> Error: 22000 Error while communicating with S3 DETAIL: aws-s3-bucket-name: mybucket INTERNAL DETAIL: aws-s3-bucket-name: mybucket
thrown in the test body.
```

For more info please see my comment on https://github.com/google/googletest/issues/4387

### DIFF
--- a/googletest/src/gtest.cc
+++ b/googletest/src/gtest.cc
@@ -5352,7 +5352,8 @@ void UnitTest::AddTestPartResult(TestPartResult::Type result_type,
       result);
 
   if (result_type != TestPartResult::kSuccess &&
-      result_type != TestPartResult::kSkip) {
+      result_type != TestPartResult::kSkip &&
+      result_type != TestPartResult::kNonFatalFailure) {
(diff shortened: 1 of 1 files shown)


avrdan:
This branch contains two commits, I made 2 separate pull requests for the 2 separate issues. Sorry for tracking against the wrong branch...

derekmauro:
The team discussed this idea and decided not to accept it. It is a valid style choice, but we would prefer to preserve exact error messages.

PR #4407, opened 2023-10-31
[#4387] Better exception printing for multi-line exceptions.

avrdan:
Solution for multi-line exceptions:

- regex replace
- adds leading \n if it does not exist, does not add another otherwise
- deals with trailing \n -> same output with or without a trailing \n, but note that two trailing \n\n will result in a line that starts with >
- changes default behavior to use a line with >

Examples

New line after each segment:

```
> Error: 22000
thrown in the test body.
```

Default behavior if no \n chars detected:
```

> Error: 22000 Error while communicating with S3 DETAIL: aws-s3-bucket-name: mybucket INTERNAL DETAIL: aws-s3-bucket-name: mybucket
```
thrown in the test body.

For more info please see my comment on https://github.com/google/googletest/issues/4387

### DIFF
--- a/googletest/src/gtest.cc
+++ b/googletest/src/gtest.cc
@@ -60,6 +60,7 @@
 #include <unordered_set>
 #include <utility>
 #include <vector>
+#include <regex>
 
 #include "gtest/gtest-assertion-result.h"
 #include "gtest/gtest-spi.h"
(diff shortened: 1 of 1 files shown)


avrdan:
This new pull request only tracks the commit related to #4387.

vogelsgesang:
Thanks for looking into this, @avrdan! Looks good to me (but I am not involved with GTest in any way, so I can't approve this Pull Request)

PR #4408, opened 2023-11-07
Cleanup include paths

tmiguelf:
Used local path instead of additional include directories to find needed gtest-internal-inl.h (improves compatibility with other build methods)


derekmauro:
The current #include structure works with all of our supported build systems. Supporting other systems introduces a maintenance  burden.

tmiguelf:
@derekmauro Please reconsider.
1. There is only 1 header file in the entire project that /src to be included in the additional includes directory.
2. It goes against best practices to use additional include directories for files that are side by side, and even more so if the header file is private.

PR #4413, opened 2023-11-13
Add sanitizers to GitHub Actions CI

bjosv:
Run all testcases with ASan, UBSan and TSan during Linux CI.
A few testcases which are not suitable to run with a sanitizer are exempted in these runs, like for ExitedWithCode and AllowLeak.

This PR is a continuation of PR #3648 to be able find issues already in CI.

Introduces the following defines to indicate when built with a sanitizer, which can be used to disable sanitizers on testcase level:
- `GTEST_HAS_ADDRESS_SANITIZER`
- `GTEST_HAS_UNDEFINED_BEHAVIOR_SANITIZER`
- `GTEST_HAS_THREAD_SANITIZER`

and the following bazel test tags to disable on test binary level:
`no_asan`, `no_ubsan`, `no_tsan`

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -1,0 +1,29 @@
+build:asan --strip=never
+build:asan --copt -fsanitize=address
+build:asan --copt -DGTEST_HAS_ADDRESS_SANITIZER
+build:asan --copt -O2
+build:asan --copt -g
+build:asan --copt -fno-omit-frame-pointer
+build:asan --copt -fno-sanitize-recover=all
+build:asan --linkopt -fsanitize=address
+build:asan --test_tag_filters=-no_asan
+
+build:ubsan --strip=never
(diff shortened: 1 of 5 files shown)


dinord:
We only ever accepted GitHub Actions as a best-effort community contribution, and do not maintain it ourselves. We use Google-internal tools (including sanitizers) on our end, and having e.g.  `defined(GTEST_HAS_ADDRESS_SANITIZER)` does not work with our internal tools.

We'd consider accepting sanitizers in GitHub Actions if you can figure out a way to make them work without adding new `#ifdef`s in code, either by fixing actual bugs in code or reconfiguring the sanitizers so they don't trigger on false positives.

bjosv:
Closing this since I cant find a way to config the sanitizer to handle ExitedWithCode and AllowLeak.

PR #4414, opened 2023-11-14
Cleaned up include statments

tmiguelf:
Used include directive properly.

### DIFF
--- a/googlemock/CMakeLists.txt
+++ b/googlemock/CMakeLists.txt
@@ -67,7 +67,6 @@ config_compiler_and_linker() # from ${gtest_dir}/cmake/internal_utils.cmake
 # Adds Google Mock's and Google Test's header directories to the search path.
 set(gmock_build_include_dirs
   "${gmock_SOURCE_DIR}/include"
-  "${gmock_SOURCE_DIR}"
   "${gtest_SOURCE_DIR}/include"
   # This directory is needed to build directly from Google Test sources.
   "${gtest_SOURCE_DIR}")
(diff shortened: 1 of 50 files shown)


derekmauro:
This breaks our code import procedure and internal build system. What problem are you trying to solve?

tmiguelf:
I'm trying to solve a maintainability problem.

Include directives using <> are meant to be found only trough additional include directories.
Include directives using "" are meant to find files relative to the current file location (although they can fallback to using additional include directories in case they are not found).
You should probably want to add a linter rule for that.

1. Source files in ./src were using include directives with "" to find files that are in ./include, this is incorrect as the header files can not be found using relative path, they can only be found using additional include directives.
2. Source files in ./src were using include directives with "src/*" to find files in source, this is triply incorrect for the following reasons:
a) File can not be found using relative path as there is no relative folder named "src", only by using additional include directives can this be found
b) These files can correctly be found using local path by not prepending "src"
c) This requires the project path to be added as an additional include directive, this is extremely problematic as it exposes obsoletely everything. It can also potential cause the problem where files are improperly included from public include files, which can lead to the situation where the library builds and passes all the tests but fails to build when included into other projects. Removing the project base path from the additional include directives (as I argue should the standard practice in any project) this problem can not occur.

This code should be able to compile just fine without breaking. If that breaks your build then that is a red flag indicating that there's something wrong with the build. If it's only breaking "policy", then I would strongly recommend revising the "policy" as the improper use of include directives can be project breaking.

PR #4422, opened 2023-11-24
docs: Document how to remove prefix for parametrized test

### LINKED ISSUE #4423: [FR]: Document how to remove prefix for parametrized test

Feature itself, yes. No documents.

https://chromium-review.googlesource.com/c/chromium/src/+/5040971/comment/c202d9b3_0ae8ff49/

In Chromium, lots of committers use "All" for instantiation name, even if multiple instances are not used.
I think this is because they don't know how to opt-out it. Documentation is required.

### Describe the proposal.

Document how to remove prefix for parametrized test

N/A

kenoss:
fix #4423


dinord:
As a Googler, would you mind submitting a change internally instead?

dinord:
Please send the change internally.

PR #4454, opened 2024-01-26
Better exception printing for multi-line exceptions

parth0809:
Fixes : #4387


derekmauro:
The team discussed this idea and decided not to accept it. It is a valid style choice, but we would prefer to preserve exact error messages.

PR #4461, opened 2024-02-08
Update quickstart-bazel.md to new version of GoogleTest

chrisbigelow:
previous url pointed to release that no longer builds


derekmauro:
We need the CLA.

PR #4478, opened 2024-02-26
Fix race condition when creating unique filename

luhenry:
Because the file is not opened with O_EXCL, there is no guarantee that multiple threads calling this function simultaneously across different processes will not generate the same filename. Instead make sure to use O_EXCL for the kernel to decide whether a filename already exists.


luhenry:
> Have you checked all the call sites to see if this is OK? I believe we even have a test (DirectoryCreationTest.CreateDirectoriesAndUniqueFilename) that should break here:

I've checked that all callers are fine with the file existing, and will simply open the path after the call to `FilePath::GenerateUniqueFileName`.

Would it be keep the code generating the unique file name, updating the doc, and updating the test? It would also closely match what [mktemp](https://www.man7.org/linux/man-pages/man3/mktemp.3.html) is doing.

luhenry:
> Sorry, I'm still confused. I haven't had a chance to test your code, so I may be missing something, but how is this line not failing with your change?

The test indeed does fail, and I failed to test before submitting (my bad) as I was expecting some of the test to run on CI and validate the change. However, despite the test failing, I've been carrying this patch in multiple projects without issue.

That's why I am asking whether you'd be more comfortable with adding a _new_ API similar to `mkstemp` rather than modifying the existing. We could then use that new API in places like `XmlUnitTestResultPrinter` or `JsonUnitTestResultPrinter` constructors. I'm happy to work on a quick proof of concept if that would help the discussion.

Issue #4493, opened 2024-03-18
[Bug]: A race condition exist between the Watcher thread and the main thread terminating.


cgyrling:
### Describe the issue

A race condition exist between the Watcher thread and main(). A case was found where the Watcher thread does not get execution time before the main function returns and calls atexit(). At that point the Watcher thread started runing tls_init() code while the main thread was shutting down. This resulted in rare crashes and deadlocks.

### Steps to reproduce the problem

As this is a race condition, it is tricky to reproduce. The easiest way is to do the following.
1) Add SCOPED_TRACE right before returning from GTest.
2) Add a thread_local instance of a class that is accessing a static shared_ptr. The shared pointer gets cleaned up on 'atexit' and if you access that via your constructor in the call from tls_init(), it will access freed memory.

### What version of GoogleTest are you using?

1.14.0

### What operating system and version are you using?

Windows 10

### What compiler and version are you using?

msc / Visual Studio 2022

### What build system are you using?

buck

justzh:
You need to see ＠Dｒ。Ｃａｌｖｉｎ　Ｌｉｎ

PR #4494, opened 2024-03-18
Fixed a race condition between the Watcher thread and the main thread during program exit.

cgyrling:
A race condition exist between the Watcher thread and main(). A case was found where the Watcher thread does not get execution time before the main function returns and calls atexit(). At that point the Watcher thread started runing tls_init() code while the main thread was shutting down. This resulted in rare crashes and deadlocks.

This fixes the following issue: (https://github.com/google/googletest/issues/4493)

### DIFF
--- a/googletest/src/gtest-port.cc
+++ b/googletest/src/gtest-port.cc
@@ -591,12 +591,34 @@ class ThreadLocalRegistryImpl {
   // StartWatcherThreadFor to WatcherThreadFunc.
   typedef std::pair<DWORD, HANDLE> ThreadIdAndHandle;
 
+  class WatcherThreadParams
+  {
+    public:
+    WatcherThreadParams(
+        DWORD threadId,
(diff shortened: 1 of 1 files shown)


derekmauro:
We can't get started on this without the CLA. Please sign the CLA and report back.

cgyrling:
It shows me as part of this Corporate Contract 
Google Corporate CLA | Facebook, Inc. | Apr 29, 2016 12:42 PDT.

I double-checked all the emails used and I believe I should be good to go. No individual CLA needed to be signed

derekmauro:
The CLA check is still failing. There is nothing I can do about that. Please figure out why.

If you would like to close this pull request I am happy to fix this myself (I would do it in a slightly different way anyway).

cgyrling:
Hi Derek. I'm fine with you fixing the issue. I'm having CLA issues on my end that is holding me up.

PR #4514, opened 2024-04-14
Improved safety with null pointer check in SetEnv class.

LeilaShcheglova:
Added a check for a null pointer name in the constructor and destructor of the SetEnv class to ensure safety during usage. Now the program will throw an exception if name is equal to nullptr or its length is zero, which helps prevent incorrect behavior.


derekmauro:
This is a test, not a library. No one would ever pass nullptr.

PR #4518, opened 2024-04-18
Minor fixes to gtest_build_tests=ON

puetzk:
Fix some failures identified in gtest_build_tests=ON when trying to build googletest with winegcc for a [winelib](https://wiki.winehq.org/Winelib) application (i.e. when using a POSIX libc/libstdc++, but with some win32 API available).

Context: I have additional patches in the branch https://github.com/puetzk/googletest/commits/wine/, but I don't think their maturity (or popularity) makes sense to upstream at this time. But these two commits are stand-alone fixes that seem generally applicable, so I've cherry-picked them.

### DIFF
--- a/googlemock/test/gmock_test.cc
+++ b/googlemock/test/gmock_test.cc
@@ -87,10 +87,10 @@ TEST(InitGoogleMockTest, ParsesSingleFlag) {
 
 TEST(InitGoogleMockTest, ParsesMultipleFlags) {
   int old_default_behavior = GMOCK_FLAG_GET(default_mock_behavior);
-  const wchar_t* argv[] = {L"foo.exe", L"--gmock_verbose=info",
-                           L"--gmock_default_mock_behavior=2", nullptr};
+  const char* argv[] = {"foo.exe", "--gmock_verbose=info",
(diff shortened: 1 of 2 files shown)


googlemock/test/gmock_test.cc
```diff
@@ -87,10 +87,10 @@ TEST(InitGoogleMockTest, ParsesSingleFlag) {
 
 TEST(InitGoogleMockTest, ParsesMultipleFlags) {
   int old_default_behavior = GMOCK_FLAG_GET(default_mock_behavior);
-  const wchar_t* argv[] = {L"foo.exe", L"--gmock_verbose=info",
```
KioFancy, line 90: [calc.zip](https://github.com/google/googletest/files/15074144/calc.zip)

googlemock/test/gmock_test.cc
```diff
@@ -87,10 +87,10 @@ TEST(InitGoogleMockTest, ParsesSingleFlag) {
...
-                           L"--gmock_default_mock_behavior=2", nullptr};
+  const char* argv[] = {"foo.exe", "--gmock_verbose=info",
```
KioFancy, line 90: 1234

PR #4538, opened 2024-05-14
ParseInt32: Check for str not being a null pointer before parsing string as an int32

tobbi:
Now, this is not too big of an issue since all the functions calling this already do this, but better be safe than sorry.


derekmauro:
Thanks, but we don't need this. This is an internal function with 4 callers.

tobbi:
@derekmauro Mind reviewing any of my other PRs just so I can have the good feeling of having contributed something to Google code?

PR #4552, opened 2024-05-31
Detect and report Rotten Green Tests

pogo59:
Rotten Green Tests are tests that have assertions that did not execute, even though they were contained in a test method that has a Pass result. This patch detects and reports such assertions contained in a TEST or TEST_F method, if the method result is Pass. Skipped, disabled, or otherwise not-run tests are not analyzed for rotten assertions.

By default, Rotten assertions do not count as a failure. Use `--gtest_treat_rotten_as_pass=0` to make rotten tests into failures. This is customizable.

Rotten assertions in helper functions are also reported if the overall result of the test program was Pass.

Currently, TEST_P, TYPED_TEST, and TYPED_TEST_P are treated as if they were helpers, for Rotten reporting purposes. This can lead to false positives for these kinds of Test methods, especially if they are skipped, disabled, or otherwise not run.

### DIFF
--- a/googlemock/test/gmock-internal-utils_test.cc
+++ b/googlemock/test/gmock-internal-utils_test.cc
@@ -275,15 +275,10 @@ TEST(LosslessArithmeticConvertibleTest, FloatingPointToFloatingPoint) {
 
   // Larger size => smaller size is not fine.
   EXPECT_FALSE((LosslessArithmeticConvertible<double, float>::value));
-  GTEST_INTENTIONAL_CONST_COND_PUSH_()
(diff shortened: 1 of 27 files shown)


pogo59:
This is a replacement for #4192, which could detect rotten assertions more reliably but would not work properly in Google's Bazel build environment.

pogo59:
Hi @derekmauro it looks like I'm unable to request reviewers, just wanted to make sure people were aware of this PR.

pogo59:
I've had no response, and my retirement is imminent. The PR is still here if someone else wants to take it over.

Issue #2790, opened 2020-04-01
Gtest compile error AIX


Marf97:
Hi,

I'm having trouble building newest version of gtest on AIX 6.1. I would like to build it with g++ not AIX xlc. After I downloaded release, opening it the procedure how I got to my error is:
```
cd googletest-release-1.10.0
mkdir build
cd build
cmake ..
make
```
I'm getting tons of errors like this:
```
googletest-release-1.10.0/googletest/include/gtest/gtest.h:2299:1: error: template with C linkage
 template <typename T1, typename T2>
 ^~~~~~~~
```
cmake output:
```
cmake ..
-- The C compiler identification is GNU 8.3.0
-- The CXX compiler identification is GNU 8.3.0
-- Check for working C compiler: /opt/freeware/bin/gcc
-- Check for working C compiler: /opt/freeware/bin/gcc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /opt/freeware/bin/c++
-- Check for working CXX compiler: /opt/freeware/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Found PythonInterp: /opt/freeware/bin/python (found version "2.7.17")
-- Looking for pthread.h
-- Looking for pthread.h - found
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Failed
-- Looking for pthread_create in pthreads
-- Looking for pthread_create in pthreads - found
-- Found Threads: TRUE
-- Configuring done
-- Generating done
-- Build files have been written to: MY PATH
```

Marf97:
Just to put some more context. That AIX machine is quite outdated and there is some possibility of missing libs or something else. I was trying to satisfy all requirements. If the problem is connected to `CMAKE_HAVE_LIBC_PTHREAD - Failed` please tell me what is needed because I couldn't find it in AIX toolbox or anywhere else.

asoffer:
Unfortunately AIX is an unsupported platform. This could be our bug, or a bug in g++'s compilation for that platform. If you determine what's wrong we'd welcome a PR. But we don't have these machines to test on and aren't familiar with them so we unfortunately cannot be of much help.

[...]

PR #4553, opened 2024-06-02
Make SYSTEM argument optional in target_include_directories

KOMMYHAP:
Hello!
I'd like to use cmake FetchContent to download googletest sources and clang-cl to compile it. But in this case gtest include directory was not added to gtest target. Seems like SYSTEM argument in target_include_directories does not pollute target properties correctly.
I wrap SYSTEM argument in cmake option gtest_enable_system_include to save backward compatibility.

### DIFF
--- a/googletest/CMakeLists.txt
+++ b/googletest/CMakeLists.txt
@@ -26,6 +26,8 @@ option(
   "Build gtest with internal symbols hidden in shared libraries."
   OFF)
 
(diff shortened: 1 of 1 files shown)


lilianmoraru:
Sounds like a CMake bug.
Did you try using the latest stable CMake?

KOMMYHAP:
Hm, you may be right. I'll try it later and come back

GregDomjan:
If similar was to be done for googlemock
This might also be used to address #2790 AIX build

PR #4554, opened 2024-06-04
new changes

Yoowwq:


googletest/cmake/internal_utils.cmake
```diff
@@ -154,7 +154,9 @@ macro(config_compiler_and_linker)
   set(cxx_no_rtti "${cxx_default} ${cxx_no_rtti_flags}")
 
   # For building the gtest libraries.
+  if(NOT DEFINED cxx_strict_flags)
```
lilianmoraru, line 157: `set(cxx_strict "${cxx_default} ${cxx_strict_flags}")`  is trying to add `cxx_strict_flags`, so this would break that.
This would always add an empty `cxx_strict_flags`.

PR #4563, opened 2024-06-24
Add support for all help flag variants mentioned in the documentation

codeinred:
The documentation specifies that '-h', '-?', and '/?' may be used, but the code failed to check for these flags. This commit fixes the behavior of the executable to match what is specified in the docs.

Excerpt from [advanced.md](docs/advanced.md#running-test-programs-advanced-options):

> To see a list of supported flags and their usage, please run your test program
> with the `--help` flag. You can also use `-h`, `-?`, or `/?` for short.

### DIFF
--- a/googletest/src/gtest.cc
+++ b/googletest/src/gtest.cc
@@ -6672,6 +6672,12 @@ static void LoadFlagsFromFile(const std::string& path) {
 }
 #endif  // GTEST_USE_OWN_FLAGFILE_FLAG_ && GTEST_HAS_FILE_SYSTEM
 
+// Checks if the given flag is a help flag.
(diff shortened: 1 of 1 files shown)


codeinred:
See issue #4564

derekmauro:
GoogleTest has (for now) optional dependency on Abseil, and the Abseil flags library is used when in this case. As far as I know, Abseil flags library only supports the `--help` flag. So in this case, I think we should make the documentation consistent with the code, not the other way around.

derekmauro:
I changed the documentation in https://github.com/google/googletest/commit/3ef16ef8b30f52a8f1ff677faa1730e76835aa40. Thanks for letting us know about this issue.

PR #4565, opened 2024-06-25
Print the error code when unable to create a temporary file on Windows

BenSluis:
Help root cause a sporadic error seen when running gtest applications


BenSluis:
Related-to: #4566

PR #4571, opened 2024-07-01
Fix Windows compilation when _CRT_DECLARE_NONSTDC_NAMES=0

### LINKED ISSUE #934: Doesn't compile with Clang/C2
GoogleTest doesn't compile with Visual Studio's new Clang/C2 giving lots of errors like:

```
error : use of undeclared identifier 'chdir'; did you mean '_chdir'?
```

See https://msdn.microsoft.com/en-us/library/ms235420.aspx

mon:
Closes #934 (which was closed due to inactivity, not due to it being fixed).

On Windows, using the POSIX names for some filesystem ops (chdir, fdopen, read, write, close) has been deprecated for some time (potentially as far back as [the early to mid 90s](https://devblogs.microsoft.com/oldnewthing/20200730-00/?p=104021)).

Of course, Microsoft has kept the old names around for backwards compat for over 25 years. But when the above #define is set, (or when clang-cl is used with a `/std:` specified, apparently), the names are left undefined, causing compilation failures.

Considering how long the underscored names have been available, I consider this change to have no compatibility concerns.


mon:
Apologies for the force push, I had a mistaken `#if` instead of `#ifdef` for one of the guards.

mon:
Got a "stale" email about this PR, still relevant!

PR #4594, opened 2024-08-13
Update primer.md

segsch:
For consistency and readability: the test below uses nullptr not NULL.


segsch:
For consistency and readability: the test below uses nullptr not NULL.

PR #4628, opened 2024-10-21
feat: add EXPECT_NO_COREFILE_DEATH

zhouronghua:
If we use expect_death, we really know it will coredump or raise signal, EXPECT_NO_COREFILE_DEATH is a helper for user to use expect_death but do not genreate core file.
EXPECT_NO_COREFILE_DEATH depends on ::testing::ExceptionTest, who will set rlimit to suppress core file generated.

printf in main is not good idea for all test.


zhouronghua:
ping

Issue #4289, opened 2023-06-19
[Bug]: CMake Quickstart: policy CMP0135 is not set.


Nickrader:
### Describe the issue

CMake Quickstart:
When building per instructions, after exectuting: cmake -S . -B build
 
"CMake Warning (dev) at /usr/share/cmake-3.24/Modules/FetchContent.cmake:1267 (message):
  The DOWNLOAD_EXTRACT_TIMESTAMP option was not given and policy CMP0135 is
  not set.  "

### Steps to reproduce the problem

https://github.com/Nickrader/GTestCMakeQuickstart

### What version of GoogleTest are you using?

Latest to fix error.

### What operating system and version are you using?

Lubuntu 22.10

### What compiler and version are you using?

gcc version 12.2.0 (Ubuntu 12.2.0-3ubuntu1) 

### What build system are you using?

cmake version 3.24.2

### Additional context

Unsure about implementation of fix.
Simple enough to use latest commit and GIT_REPOSITORY, GIT_TAG.

To use URL it seems setting CMP0135 to NEW would be solution, based on other similar problems/solutions.  Unsure where to set, as error didn't point to source code in Googletest.

CosmicPegasis:
## Steps to Reproduce
Using the given repo can't reproduce the problem.
```
cosmic@zack ~/c/c/p/GTestCMakeQuickstart (stable)> cmake -S . -B build
-- The C compiler identification is GNU 13.1.1
-- The CXX compiler identification is GNU 13.1.1
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Found Python: /usr/bin/python3.11 (found version "3.11.3") found components: Interpreter 
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
-- Found Threads: TRUE  
-- Configuring done (43.4s)
-- Generating done (0.0s)
-- Build files have been written to: /home/cosmic/coding/cpp/playground/GTestCMakeQuickstart/build
```
- cmake version: 3.26.4
- gcc version: 13.1.1
[...]

PR #4630, opened 2024-10-22
Address Issue #4289 by setting CMake policy CMP0135 to the new (recommended) behavior when CMake version supports

dodohand:
select new behavior for CMP0135 when supported by cmake so that: "CMake 3.24 and above prefers to set the timestamps of all extracted contents to the time of the extraction. This ensures that anything that depends on the extracted contents will be rebuilt whenever the URL changes."

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,6 +3,10 @@
 
 cmake_minimum_required(VERSION 3.13)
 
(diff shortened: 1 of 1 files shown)


dodohand:
Link to Issue #4289 :  https://github.com/google/googletest/issues/4289

PR #4662, opened 2024-11-15
Fix documentation bug.

asoffer:
Matchers require the signature of these three functions to be `const`.


asoffer:
I signed it.

dinord:
Thanks :-)

Issue #4665, opened 2024-11-20
[FR]: Streaming results format to stdout


davidmatson:
### Does the feature exist in the most recent commit?

No

### Why do we need this feature?

Currently, we have per-OS code for stream_result_to, and Windows support is not yet available.

### Describe the proposal.

Provide a switch to have stdout report progress in the same format used by stream_result_to rather than the human-readable format.

Perhaps:
--gtest-machine-results

### Is the feature specific to an operating system, compiler, or build system version?

No

RedMarcher:
Hello, I'm new to this particular project. Would you mind telling me which file the current implementation for stdout is located at? You mentioned that it was different for each OS, so is the feature implementation currently spread out among multiple different files?

RedMarcher:
Looking over your comment more closely, do you just want a feature where whatever is currently spit out of stream_result_to is also available to stdout?

davidmatson:
> Looking over your comment more closely, do you just want a feature where whatever is currently spit out of stream_result_to is also available to stdout?

yes (and also not having the pretty-print format written stdout)

RedMarcher:
> > Looking over your comment more closely, do you just want a feature where whatever is currently spit out of stream_result_to is also available to stdout?
> 
> yes (and also not having the pretty-print format written stdout)

So basically I should add a new flag using one of the names you proposed to activate this feature?

davidmatson:
> So basically I should add a new flag using one of the names you proposed to activate this feature?

This is a feature request to add a new flag that would control this behavior, yes.

davidmatson:
@derekmauro - thoughts on this proposal? We'd be interested in working on a draft implementation, if there's agreement that this would be a good idea.

RedMarcher:
> [@derekmauro](https://github.com/derekmauro) - thoughts on this proposal? We'd be interested in working on a draft implementation, if there's agreement that this would be a good idea.

I already have a sort of merge request. Unfortunately no one has looked at it yet. This project is a little bit quiet. If you could take a look and give me your thoughts, I'd appreciate it.

https://github.com/google/googletest/pull/4668

davidmatson:
@RedMarcher - I think we'll be best off if we can get someone from the GoogleTest team to consider the feature request before we get too deep into implementation details. Sometimes it can take a while for those folks to review, and I'm not sure of any way to accelerate that process.

RedMarcher:
[...]

PR #4668, opened 2024-11-21
New feature implementation: Streaming results format to stdout #4665

RedMarcher:
I haven't actually implemented the feature yet, for now I've just added the flags necessary for it.
I would like to ask for more clarifications about how result streaming is currently handled before adding in new feature code.


RedMarcher:
I noticed the use of lots of #if compiler commands. Is this applicable to the feature? If the flag is set to true, should I just have a function which prints results in the same format as from stream_result_to to the console using cout and stream operators?

RedMarcher:
After reviewing the rest of the code some more, I realize stream_result_to doesn't actually create a function, rather it adds a new listener to a preexisting list. I'm wondering if printing to the terminal should follow the same behavior, or if it should be more "direct" somehow.

PR #4680, opened 2024-12-12
Add CMake interface library `GTest::gtest_prod`

### LINKED ISSUE #4673: [FR]: Export a header-only library `GTest::gtest_prod` via CMake

The Bazel infrastructure already provides a dedicated `gtest_prod` library, which was introduced by commit 105579a6e43908bc88a289d309492eda8be896be.

Unfortunately, the same does not yet exist for CMake.

### User story
As a C++ application developer using CMake,
I want GoogleTest's CMake config to provide a dedicated header-only library for `gtest_prod.h`,
so that I can use the `FRIEND_TEST()` macro in my application code without having to link against the static `GTest::gtest` library.
Only my unit tests should link against `GTest::gtest`.

### Describe the proposal.

The implementation should look roughly like this:
* In `googletest/CMakeLists.txt`, add a library named `gtest_prod` of type `INTERFACE`.
* Include `gtest_prod` in the list of targets to be installed.
* Add a pkgconf template `googletest/cmake/gtest_prod.pc.in` that adds the correct include-path to the compiler flags.

This will cause a new target, `GTest::gtest_prod`, to be added to `GTestTargets.cmake`.

When a user application subsequently links against `GTest::gtest_prod` in its `CMakeLists.txt`, the correct include-path will be set. *The user application will not pull in the static gtest library*.

No; this is a platform-agnostic feature.
To the best of my knowledge, it does not pose any specific requirements on the CMake version.

mp-chet:
Production code that contains `#include <gtest/gtest_prod.h>` should link against this interface library: it sets the correct include path for the compiler but does not instruct the linker to pull in any GoogleTest library code.

Fixes #4673.

### DIFF
--- a/googletest/CMakeLists.txt
+++ b/googletest/CMakeLists.txt
@@ -116,10 +116,14 @@ include_directories(${gtest_build_include_dirs})
 #
 # Defines the gtest & gtest_main libraries. User tests should link
 # with one of them.
+# Production code that includes <gtest/gtest_prod.h> should link with the
+# interface library gtest_prod.
 
 # Google Test libraries. We build them using more strict warnings than what
 # are used for other targets, to ensure that gtest can be compiled by a user
 # aggressive about warnings.
+add_library_and_alias(gtest_prod INTERFACE)
+set_target_properties(gtest_prod PROPERTIES VERSION ${GOOGLETEST_VERSION})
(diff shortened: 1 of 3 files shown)


Hasan0971:
Good

mp-chet:
Rebased to the latest `main` branch.

mp-chet:
Rebased to the latest `main` branch.

PR #4685, opened 2024-12-18
Put the fake Fuchsia SDK in a module extension

tpudlik:
This allows users to override the fake SDK with a real one using https://bazel.build/rules/lib/globals/module#override_repo.

Without this change, it is impossible for a project that depends on googletest as a `bazel_dep` to build tests using the "real" Fuchsia SDK, because any references to `@fuchsia_sdk` within googletest `BUILD.bazel` files unconditionally resolve to the "fake" Fuchsia SDK. With this change, if you have the real Fuchsia SDK declared in your `MODULE.bazel`, you can add the following lines to coerce googletest to use the real Fuchsia SDK as well:

    fake_fuchsia_sdk_extension = use_extension("@com_google_googletest//:fake_fuchsia_sdk.bzl", "fuchsia_sdk")
    override_repo(fake_fuchsia_sdk_extension, "fuchsia_sdk")

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -39,29 +39,37 @@ module(
 # Only direct dependencies need to be listed below.
 # Please keep the versions in sync with the versions in the WORKSPACE file.
 
-bazel_dep(name = "abseil-cpp",
-          version = "20240116.2")
+bazel_dep(
+    name = "abseil-cpp",
+    version = "20240116.2",
+)
+bazel_dep(
+    name = "platforms",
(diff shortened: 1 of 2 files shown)


tpudlik:
Oops, sounds like I should submit this internally first. Will send it as cl/707973104.

tpudlik:
Merged as https://github.com/google/googletest/commit/e54519b09463cec3aea77a1739e02c97ca766da5.

PR #4687, opened 2024-12-23
Create Debian and RPM Packages on release

### LINKED ISSUE #4686: [FR]: Create Packages on every release (Debian, RPM, etc.)

No

This feature would make it easier for anyone to download the latest Debian/RPM package from GitHub rather than have to compile it.

Also, the current versions of GoogleTest, while functional are old (v1.11.0 is from June 2021)
Image

This would also make it easy for consistently having Debian/Ubuntu repositories updated if a package is generated each release.

### Describe the proposal.

I think the easiest solution would be to create a CPackConfig.txt to leverage your CMakeLists.txt file. Once this is done, create a GitHub workflow such that on every release, you create the packages and upload it for release. A workflow like the one listed below should do the trick, except you would be using CMake and `cpack -G DEB`, etc. for each package.

https://github.com/capstone-engine/capstone/blob/next/.github/workflows/build_release.yml

I have a working example on how to create Debian and RPM packages already using CPackConfig.txt here

https://github.com/capstone-engine/capstone/pull/2590

For now, start with Linux, but it would be nice to attempt other Operating Systems too.

AndrewQuijano:
Closes #4686

### DIFF
--- a/.github/workflows/build_release.yml
+++ b/.github/workflows/build_release.yml
@@ -1,0 +1,34 @@
+name: Build Source Release
+
+# Trigger whenever a release is created
+on:
+  release:
+    types:
+      - created
+
+jobs:
+  build:
+    name: build
+    runs-on: ubuntu-22.04
+    steps:
+    - name: clone the repository
+      uses: actions/checkout@v4
+
+    - name: Create the Debian and RPM package
+      run: |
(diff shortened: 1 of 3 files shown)


AndrewQuijano:
I just signed the CLA

PR #4722, opened 2025-02-07
fix: Take the C++ standard version into account when using `__has_cpp_attribute`

### LINKED ISSUE #4605: [Bug]: GTEST_INTERNAL_HAVE_CPP_ATTRIBUTE is not accounting for the standard in which attribute is introduced
### Describe the issue

The `GTEST_INTERNAL_HAVE_CPP_ATTRIBUTE` is not evaluating against the C++ standard in which an attribute is introduced. Currently the only usage is in context of `GTEST_INTERNAL_HAVE_CPP_ATTRIBUTE(maybe_unused)` which could lead in context of C++14 to usage `maybe_unused` which is only applicable from C++17 onwards. Usage of this could be leading to `-Wc++17-attribute-extensions` warning.

More robust approach would be therefore be:
```
#define GTEST_INTERNAL_HAVE_CPP_ATTRIBUTE(x) __cplusplus >= __has_cpp_attribute(x)
```

Compile googletest repository from root like:
```
$ CXXFLAGS="-Werror -Wall -Wextra -Wpedantic" cmake -DCMAKE_CXX_STANDARD=14 . -B /tmp/gtestbuildfolder/
$ cmake --build /tmp/gtestbuildfolder/
```

1.15.2

macOS Sonoma 14.6

Apple clang version 15.0.0 (clang-1500.0.40.1)

cmake version 3.30.2

### Additional context

_No response_

yuyawk:
Resolves #4605

As elaborated in the issue, `GTEST_INTERNAL_HAVE_CPP_ATTRIBUTE(maybe_unused)` failing to consider the C++ standard can cause `-Wc++17-attribute-extensions` warning in C++14.

This PR applies the patch equivalent to what the issue proposed, but uses `GTEST_INTERNAL_CPLUSPLUS_LANG` instead of `__cplusplus`.

### DIFF
--- a/googletest/include/gtest/internal/gtest-port.h
+++ b/googletest/include/gtest/internal/gtest-port.h
@@ -757,7 +757,7 @@ typedef struct _RTL_CRITICAL_SECTION GTEST_CRITICAL_SECTION;
 #if defined(__has_cpp_attribute)
 // NOTE: requiring __cplusplus above should not be necessary, but
 // works around https://bugs.llvm.org/show_bug.cgi?id=23435.
-#define GTEST_INTERNAL_HAVE_CPP_ATTRIBUTE(x) __has_cpp_attribute(x)
+#define GTEST_INTERNAL_HAVE_CPP_ATTRIBUTE(x) GTEST_INTERNAL_CPLUSPLUS_LANG >= __has_cpp_attribute(x)
 #else
 #define GTEST_INTERNAL_HAVE_CPP_ATTRIBUTE(x) 0
 #endif


thirtytwobits:
I ran into this issue and verified this fix on Ubuntu 24.04 using googletest v1.16.0 and clang 19.1.7

higher-performance:
I'm confused. If `__has_cpp_attribute(...) == 0` then wouldn't this condition evaluate to `true` instead of `false`?

I assume you need `__has_cpp_attribute(x) != 0 && ...` as part of the condition?

higher-performance:
I'm closing this because I don't think there is a way to make this behave the way you want at the library level. For example, `__has_cpp_attribute(nodiscard)` evaluates to `201907L` on C++17 and C++14, despite the fact that it is valid on C++17 and invalid on C++14.

PR #4723, opened 2025-02-13
Add support for streaming results on Windows (fixes #3935).

### LINKED ISSUE #3935: Support --gtest_stream_result_to on Windows

davidmatson:


davidmatson:
@derekmauro - would you be able to help reviewing this PR? thanks!

PR #4738, opened 2025-03-26
Added contribution.txt with an improvement idea

Aastha08a:
This pull request adds a contribution file with an idea to improve Google Test.


pgawro:
Out of curiosity, what was a reason for this change?

PR #4758, opened 2025-05-01
feat: Allow Property matcher with default arguments 

Richlee96112:
Used std::invoke in PropertyMatcher to support default arguments
All tests pass, including a new test for std::source_location.


marbub:
It seems Google is no longer merging pull requests.

Ahmedaxa (commented):
axa

DuPol:
I would really appreciate if this could be merged.

Issue #4803, opened 2025-07-18
`TYPED_TEST_SUITE_P` and `-Wexit-time-destructors`


cdesouza-chromium:
### Does the feature exist in the most recent commit?

No.

### Why do we need this feature?

I have been going about in Chromium, trying to make clang's `-Wexit-time-destructors` a default enabled warning, in order to reduce the risk with destruction time ordering issues. Upon further investigation, I noticed that some of the test that are coming back with warnings happen to trip the warning whenever they use `TYPED_TEST_SUITE_P`. This macro looks like this:

```cxx
// The variables defined in the type-parameterized test macros are
// static as typically these macros are used in a .h file that can be
// #included in multiple translation units linked together.
#define TYPED_TEST_SUITE_P(SuiteName)              \
  static ::testing::internal::TypedTestSuitePState \
  GTEST_TYPED_TEST_SUITE_P_STATE_(SuiteName)
```

And `::testing::internal::TypedTestSuitePState` looks like this:

```cxx
// State of the definition of a type-parameterized test suite.
class GTEST_API_ TypedTestSuitePState {
 public:
  TypedTestSuitePState() : registered_(false) {}

  // Adds the given test name to defined_test_names_ and return true
  // if the test suite hasn't been registered; otherwise aborts the
  // program.
  bool AddTestName(const char* file, int line, const char* case_name,
                   const char* test_name) {
    if (registered_) {
      fprintf(stderr,
              "%s Test %s must be defined before "
              "REGISTER_TYPED_TEST_SUITE_P(%s, ...).\n",
              FormatFileLocation(file, line).c_str(), test_name, case_name);
      fflush(stderr);
      posix::Abort();
    }
    registered_tests_.emplace(test_name, CodeLocation(file, line));
    return true;
  }

  bool TestExists(const std::string& test_name) const {
    return registered_tests_.count(test_name) > 0;
  }

  const CodeLocation& GetCodeLocation(const std::string& test_name) const {
    RegisteredTestsMap::const_iterator it = registered_tests_.find(test_name);
    GTEST_CHECK_(it != registered_tests_.end());
    return it->second;
  }

  // Verifies that registered_tests match the test names in
  // defined_test_names_; returns registered_tests if successful, or
  // aborts the program otherwise.
  const char* VerifyRegisteredTestNames(const char* test_suite_name,
                                        const char* file, int line,
                                        const char* registered_tests);

 private:
[...]

PR #4804, opened 2025-07-21
Avod exit-time destructors in macros when possible

cdesouza-chromium:
This change attemps to reduce the introduction of exit-time destructors by gtest/gmock whenever it is possible, to just skip the destructor during termination. This is useful for codebases that aim to comply with clang's `-Wexit-time-destructors`. However, this change does not attempt to make all of gtest/gmock complaint with this warning, but limits itself to what merely affects user code.

Bug: https://github.com/google/googletest/issues/4803


cdesouza-chromium:
cc @pizzud would you please able to help me get this PR moving for Chromium?

PR #4823, opened 2025-08-27
Use "test suite" rather than "test case" in error about mixing TEST and TEST_F

RedBeard0531:
I'm pretty sure that this was missed when changing terminology since moving to another test case doesn't seem to make sense in context.


RedBeard0531:
stalebot just pinged me. Given that there are no conflicts with the base, I'm pretty sure this is still an issue...

PR #4889, opened 2025-12-20
Create cmake-single-platform.yml

Yankuba-debug:
This branch has been merged


mkruskal-google:
Sorry for the delayed response, and I appreciate the contribution.  But we currently use an internal test platform for presubmits rather than GHA.  There's a very good chance we will be switching to GHA in the near future, but it will be a more holistic migration and there are strict performance requirements (e.g. we'll need sccache hooked up to our cloud build at the very least for cmake)

PR #4894, opened 2026-01-01
Resolve warning about undefined TARGET_OS_IPHONE

dpogue:
Check that `TARGET_OS_IPHONE` is defined before checking its value to determine iOS support. This causes a warning about undefined macros when building against old Mac OS X SDKs.


dpogue:
Adding a comment because apparently this PR is "stale" and will get closed due to lack of activity otherwise. Only activity needed here is someone to trigger the workflows and a code review.

PR #4914, opened 2026-01-27
windows: improve windows handling when creating tmp files

sluongng:
In certain Windows environment, user may encounter error such as

> source/common/gtest-port.cc(1057):: Condition success != 0 failed. Unable to create a temporary file in ?

Note on the ? which does not provide a lot of information on why the
test failed.

Turned out, this was because the ::GetTempPathA api failed, which went
unchecked and passed an invalid value down to ::GetTempFileNameA.

Improve the error handling logic by checking for zero value returned by
::GetTempPathA. When error happened, call ::GetLastError to print
a more descriptive message.

### DIFF
--- a/googletest/src/gtest-port.cc
+++ b/googletest/src/gtest-port.cc
@@ -29,6 +29,7 @@
 
 #include "gtest/internal/gtest-port.h"
 
+#include <errno.h>
 #include <limits.h>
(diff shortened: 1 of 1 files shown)


higher-performance:
Closing as this is a duplicate of #4578.

Issue #3960, opened 2022-07-26
Documentation inconsistent on test names and underscores


BMBurstein:
The docs on generating custom names for parameterized tests says that underscores are not allowed 

https://github.com/google/googletest/blob/b1e9b6323a17002b8359fef338c57bc0437caf57/docs/advanced.md?plain=1#L1184-L1186

but then in the example a few lines later, and underscore is used

https://github.com/google/googletest/blob/b1e9b6323a17002b8359fef338c57bc0437caf57/docs/advanced.md?plain=1#L1218-L1224

and in the documentation for the fourth parameter of `INSTANTIATE_TEST_SUITE_P `, it is explicitly allowed

https://github.com/google/googletest/blob/b1e9b6323a17002b8359fef338c57bc0437caf57/docs/reference/testing.md?plain=1#L113-L117

So which is it?

Abhimanyu164:
Bug@

PR #4915, opened 2026-01-28
Docs: fix underscore naming rules across documentation

AdityaSkhorne:
This PR fixes incorrect and inconsistent documentation regarding the use of
underscores in GoogleTest test suite and test names.

Changes included:
- Corrected the GoogleTest Primer description of underscore restrictions
- Clarified the FAQ explanation to match C++ reserved identifier rules
- Aligned the testing documentation with actual GoogleTest behavior

Underscores are allowed in identifiers except for C++ reserved patterns
(leading underscore followed by an uppercase letter, or double underscores)
and combinations that may cause internal name collisions.

These updates bring all related documentation sections into agreement with
each other, the C++ standard, and existing GoogleTest behavior.


higher-performance:
Closing as this is similar to #3960 which has an attached PR.

Issue #4912, opened 2026-01-26
[FR]: Make string matchers work with std::wstring_view


brandl-muc:
### Does the feature exist in the most recent commit?

AFAICT this feature is not contained in [85087857ad10bd407cd6ed2f52f7ea9752db621f](https://github.com/google/googletest/tree/85087857ad10bd407cd6ed2f52f7ea9752db621f) which is currently the most recent commit.

### Why do we need this feature?

Currently, `StrEq(std::string_view{})` compiles but `StrEq(std::wstring_view{})` does not. (note that the second call uses **w**string_view)
[Example on Compiler Explorer](https://godbolt.org/z/j5oP4o3Mn)

Strictly speaking, this is rather a nuisance than a must-have feature, but the inconsistency is annoying at the very least.

### Describe the proposal.

The simple fix would be to convert all wide string matchers to accept `std::wstring_view` instead of the current `const std::wstring& str`. This will support the aforementioned case and should also accept all cases that bound to `const std::wstring &`.

However, for narrow strings the implementation is more elaborate, allowing any single argument that can be used to construct a `std::string`:
[gtest-matchers.h:808](https://github.com/google/googletest/blob/85087857ad10bd407cd6ed2f52f7ea9752db621f/googletest/include/gtest/gtest-matchers.h#L808)
```
template <typename T, typename = typename std::enable_if<
                          std::is_constructible<std::string, T>::value>::type>
using StringLike = T;
```
[gmock-matchers.h:4783](https://github.com/google/googletest/blob/85087857ad10bd407cd6ed2f52f7ea9752db621f/googlemock/include/gmock/gmock-matchers.h#L4783)
```
template <typename T = std::string>
PolymorphicMatcher<internal::StrEqualityMatcher<std::string>> StrEq(
    const internal::StringLike<T>& str) {
  return MakePolymorphicMatcher(
      internal::StrEqualityMatcher<std::string>(std::string(str), true, true));
}
```
So probably, an anlogous implementation introducing a `WStringLike` type would be desired.

### Is the feature specific to an operating system, compiler, or build system version?

This is standard C++17, but obviously, `wstring` is way more common on Windows.

brandl-muc:
Sounds good and interesting.
But I'm just the bug reporter, not a maintainer. My opinion on the PR is probably of limited worth.

PR #4926, opened 2026-02-05
Extend support for StrEqualityMatcher class and factory functions for wide char strings

XAMeLeOH:
## Feature description

Extending support of wide characters for StrEqualityMatcher and factory functions using the class.

Feature request: #4912.

## Analysis and design

`StrEqualityMatcher` and the factory functions using it support `std::wstring`, it's not implemented the same way it is done for `std::string`. For instance `StrEq` object for `std::string` can be created out of `char *`, `std::string`, `std::string_view` thanks to using `internal::StringLike` template alias. Whereas the version for `std::wstring` only supports `const std::wstring&`.

Also there absolutely no support for char[8|16|32]_t whatsoever.

| Matcher / CharT | char | wchar_t | char16_t | char32_t | char8_t |
| -- | -- | -- | -- | -- | -- |
| StrEq         | ✅ | ✅* | ❌ | ❌ | ❌ |
| StrNe         | ✅ | ✅* | ❌ | ❌ | ❌ |
| StrCaseEq | ✅ | ✅* | ❌ | ❌ | ❌ |
| StrCaseNe | ✅ | ✅* | ❌ | ❌ | ❌ |
| HasSubstr | ✅ | ✅* | ❌ | ❌ | ❌ |
| StartsWith | ✅ | ✅* | ❌ | ❌ | ❌ |
| EndsWith   | ✅ | ✅* | ❌ | ❌ | ❌ |

\* - does not support creation from a `std::basic_string_view<>`

## Solution description

The solution proposed generalizes the `std::string` version of the class and functions for the whole variety of the character types.

Factory functions `HasSubstr`, `StartsWith`, `EndsWith` are also extended (atm lacking the tests).

| Matcher / CharT | char | wchar_t | char16_t | char32_t | char8_t |
| -- | -- | -- | -- | -- | -- |
| StrEq         | ✅ | ✅ | ✅ | ✅ | ✅ |
| StrNe         | ✅ | ✅ | ✅ | ✅ | ✅ |
| StrCaseEq | ✅ | ✅ | ❌* | ❌* | ❌* |
| StrCaseNe | ✅ | ✅ | ❌* | ❌* | ❌* |
| HasSubstr | ✅ | ✅ | ✅ | ✅ | ✅ |
| StartsWith | ✅ | ✅ | ✅ | ✅ | ✅ |
| EndsWith   | ✅ | ✅ | ✅ | ✅ | ✅ |

\* - Case insensitive comparison is not provided by the standard library, thus is not implemented here.

## Areas affected

Public interface - factory functions (e.g. `StrEq`) are backward compatible.

Extended `StringLike` which is an internal type has been extended to include wide characters.

`StrEqualityMatcher` class (also an internal entity) has been changed the following way:

* The c'tor argument `case_sensitive` was removed in favour of the template parameter. This was done in order to avoid calling case insensitive comparison functions for the character types that don't support such a comparison (char[8|16|32]_t).

### DIFF
--- a/googlemock/include/gmock/gmock-matchers.h
+++ b/googlemock/include/gmock/gmock-matchers.h
@@ -905,16 +905,34 @@ bool CaseInsensitiveStringEquals(const StringType& s1, const StringType& s2) {
   return CaseInsensitiveStringEquals(s1.substr(i1 + 1), s2.substr(i2 + 1));
 }
 
+// Case insensitive comparison support based on the CharT
+
+template<typename CharT>
+struct ci_matcher_supported : std::false_type {};
+
+template<>
+struct ci_matcher_supported<char> : std::true_type {};
+
+template<>
+struct ci_matcher_supported<wchar_t> : std::true_type {};
+
 // String matchers.
 
 // Implements equality-based string matchers like StrEq, StrCaseNe, and etc.
-template <typename StringType>
+// StrCaseEq and StrCaseNe are only supported for char and wchar_t strings.
+template <typename StringType, typename CaseSensitive = std::true_type,
+          typename std::enable_if_t<
+              std::disjunction_v<
+                  CaseSensitive,
+                  ci_matcher_supported<typename StringType::value_type>
+              >,
+              int> = 0
+          >
 class [[nodiscard]] StrEqualityMatcher {
  public:
-  StrEqualityMatcher(StringType str, bool expect_eq, bool case_sensitive)
+  StrEqualityMatcher(StringType str, bool expect_eq)
       : string_(std::move(str)),
-        expect_eq_(expect_eq),
-        case_sensitive_(case_sensitive) {}
+        expect_eq_(expect_eq) {}
 
 #if GTEST_INTERNAL_HAS_STRING_VIEW
   bool MatchAndExplain(const internal::StringView& s,
@@ -947,8 +965,7 @@ class [[nodiscard]] StrEqualityMatcher {
   bool MatchAndExplain(const MatcheeStringType& s,
                        MatchResultListener* /* listener */) const {
     const StringType s2(s);
-    const bool eq = case_sensitive_ ? s2 == string_
-                                    : CaseInsensitiveStringEquals(s2, string_);
+    const bool eq = MatchStrings(s2, ci_matcher_supported<typename StringType::value_type>{});
     return expect_eq_ == eq;
   }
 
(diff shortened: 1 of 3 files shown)


googlemock/include/gmock/gmock-matchers.h
```diff
@@ -961,18 +978,28 @@ class [[nodiscard]] StrEqualityMatcher {
...
+  template <typename S>
+  bool MatchStrings(const S& s2, std::false_type) const {
+    return s2 == string_;
+  }
+
   void DescribeToHelper(bool expect_eq, ::std::ostream* os) const {
     *os << (expect_eq ? "is " : "isn't ");
     *os << "equal to ";
-    if (!case_sensitive_) {
+    if (!CaseSensitive::value) {
```
brandl-muc: Could now be constexpr-if if I'm not mistaken.
(a C++17 feature, which [should not be a problem](https://github.com/google/googletest/releases/tag/v1.17.0))

googletest/include/gtest/gtest-matchers.h
```diff
@@ -815,8 +815,64 @@ class [[nodiscard]] ImplicitCastEqMatcher {
...
+template <typename T>
+struct char_type_traits : char_type_traits_impl<std::decay_t<remove_cvref_t<T>>> {};
+
+template<typename T>
+using char_type_traits_t = typename char_type_traits<T>::type;
+
+template <typename T,
+          typename = std::enable_if_t<
+                  std::is_constructible_v<std::basic_string<char_type_traits_t<T>>, T>>>
 using StringLike = T;
```
brandl-muc: If I'm reading it correctly, then you changed the meaning of `StringLike`.
Before, it would be defined for any type from which you could construct a `std::string`.
After, it is only defined for c strings, basic_strings and basic_string_views of any character type. This seems to betray the intention of `StringLike`, you basically could have used `std::basic_string_view` directly as it, too, supports all three types.

To phrase it differently, suppose my codebase had the following type with a conversion function:
```
class MyString {
  // implementation omitted

  operator std::string() const
};
```
Without having tested this, the compiler in my head says that I can do `StrEq(MyString{ "foo" }, "foo")` without your change, but I can't anymore with your change. It's thus a breaking change.

You probably need to do something along these lines:
```
template <typename T, typename CharT, typename = typename std::enable_if_t<
                          std::is_constructible_v<std::basic_string<CharT>, T>>>
using StringLike = T;
```
And then use it like so:
```
template <typename T = std::string,
          typename CharT = internal::char_type_traits_t<T>>
PolymorphicMatcher<internal::StrEqualityMatcher<std::basic_string<CharT>>> StrEq(
    const internal::StringLike<T, CharT>& str) {
  return MakePolymorphicMatcher(
      internal::StrEqualityMatcher<std::basic_string<CharT>>(std::basic_string<CharT>(str), true));
}
```

But as I'm not a maintainer and since there are no tests for such a class like `MyString` this can or cannot be OK, I simply can't say.

XAMeLeOH: I did try to make it work with everything that can be used to construct a string, but I missed that case. Thanks for bringing it up! I added a few tests to verify that scenario and changed the logic the compiler deducts the character type from a StringLike.

higher-performance:
I think 488c47cc68353ac94fda363e501c327e31261906 resolved this for `wchar_t`? Given the title and the description are solely asking for wide-char strings, and newer character types seem to be just a bonus, I'm closing the bug. (I don't believe we have reason to prioritize other character types at the moment; there doesn't seem to be much demand for it either.)

Issue #4935, opened 2026-02-28
[Bug]: -Wdeprecated-declarations warning from clang23+gcc12


sendaoYan:
### Describe the issue

Below is the compiler warnings log snippet:

```
In file included from /home/yansendao/git/googletest/googletest/src/gtest-all.cc:38:
In file included from /home/yansendao/git/googletest/googletest/include/gtest/gtest.h:55:
In file included from /usr/lib/gcc/x86_64-linux-gnu/12/../../../../include/c++/12/memory:66:
/usr/lib/gcc/x86_64-linux-gnu/12/../../../../include/c++/12/bits/stl_tempbuf.h:263:8: warning: 'get_temporary_buffer<testing::TestInfo *>' is deprecated [-Wdeprecated-declarations]
  263 |                 std::get_temporary_buffer<value_type>(_M_original_len));
      |                      ^
/usr/lib/gcc/x86_64-linux-gnu/12/../../../../include/c++/12/bits/stl_algo.h:4996:15: note: in instantiation of member function 'std::_Temporary_buffer<__gnu_cxx::__normal_iterator<testing::TestInfo **, std::vector<testing::TestInfo
      *>>, testing::TestInfo *>::_Temporary_buffer' requested here
 4996 |       _TmpBuf __buf(__first, (__last - __first + 1) / 2);
      |               ^
/usr/lib/gcc/x86_64-linux-gnu/12/../../../../include/c++/12/bits/stl_algo.h:5070:23: note: in instantiation of function template specialization 'std::__stable_sort<__gnu_cxx::__normal_iterator<testing::TestInfo **,
      std::vector<testing::TestInfo *>>, __gnu_cxx::__ops::_Iter_comp_iter<(lambda at /home/yansendao/git/googletest/googletest/src/gtest.cc:3057:20)>>' requested here
 5070 |       _GLIBCXX_STD_A::__stable_sort(__first, __last,
      |                       ^
/home/yansendao/git/googletest/googletest/src/gtest.cc:3056:8: note: in instantiation of function template specialization 'std::stable_sort<__gnu_cxx::__normal_iterator<testing::TestInfo **, std::vector<testing::TestInfo *>>,
      (lambda at /home/yansendao/git/googletest/googletest/src/gtest.cc:3057:20)>' requested here
 3056 |   std::stable_sort(test_info_list_.begin(), test_info_list_.end(),
      |        ^
[...]

PR #4936, opened 2026-02-28
Fix -Wdeprecated-declarations warning from clang

sendaoYan:
Hi all,

This PR disable the `-Wdeprecated-declarations` compiler warning for clang explicitly. The issue detail shows in #4935 

Why not add "-Wno-deprecated-declarations" to makefile to disable the compiler waning? Because OpenJDK will include the whole gtest source and build the gtest with their makefile, disable the compiler warning in source will avoid these project do this disable work again.

### DIFF
--- a/googletest/include/gtest/gtest.h
+++ b/googletest/include/gtest/gtest.h
@@ -52,7 +52,16 @@
 #include <cstddef>
 #include <cstdint>
 #include <limits>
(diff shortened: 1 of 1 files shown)


higher-performance:
Closing per https://github.com/google/googletest/issues/4935#issuecomment-5038020952.

PR #4947, opened 2026-03-28
Minimize heavy STL headers in GoogleTest and GoogleMock public headers

### LINKED ISSUE #4943: [FR]: Improve compilation time by minimizing inclusions of `<ostream>`, `<string>`, and `<vector>`

No.

GTest is the main compilation time bottleneck in our codebase.

### Describe the proposal.

Our codebase has around ~700 unit test translation units that include `gtest/gtest.h`. We are suffering from slow compilation and CI times.

Profiling via `ClangBuildAnalyzer` shows the following heavy includes:

```
802664 ms: external/toolchains_llvm++llvm+llvm_toolchain_llvm/include/c++/v1/ostream (included 5982 times, avg 134 ms), included via:
  717x: gtest.h
  ...

696285 ms: external/toolchains_llvm++llvm+llvm_toolchain_llvm/include/c++/v1/string (included 7443 times, avg 93 ms), included via:
  252x: gtest.h 
  ...

591285 ms: external/toolchains_llvm++llvm+llvm_toolchain_llvm/include/c++/v1/vector (included 6843 times, avg 86 ms), included via:
  486x: gtest.h
  ...
```

Would it be possible to internally minimize the use of these headers? For example, `<ostream>` could be replaced with `<iosfwd>` in a few headers, and potentially isolated to `.cpp` files.

`<string>` could be conditionally replaced with `<string_view>` or even C-style strings, and `<vector>` could be conditionally be replaced with bespoke containers.

Is this something the GTest team is willing to consider?

Thanks.-

No.

unseen2004:
Description:
  This PR addresses Issue #4943 (https://github.com/google/googletest/issues/4943) by minimizing the inclusion of heavy standard library headers like
  <ostream> and <sstream> in public GoogleTest and GoogleMock headers. This should significantly improve compilation times for projects with many test
  translation units.

  Key Changes:
   * Introduced internal::StreamTo Helpers: Added a series of StreamTo overloads in namespace internal to handle the streaming of common types (built-ins,
     strings, etc.) without requiring the full definition of std::ostream.
   * Header Optimization: Swapped <ostream> and <sstream> for <iosfwd> in the following public headers:
       * gtest/gtest.h
       * gtest/gtest-printers.h
       * gtest/gtest-matchers.h
       * gmock/gmock-matchers.h
       * gmock/gmock-spec-builders.h
       * (and several internal headers)
   * Refactored StringMatchResultListener: Moved the std::stringstream member of StringMatchResultListener into the implementation file
     (gmock-matchers.cc) using a std::unique_ptr<std::ostream>, hiding the heavy dependency from the public header.
   * Code Migration: Moved several non-template PrintTo and StreamTo implementations from headers to .cc files.
   * Enhanced API: Made MatchResultListener::stream() const to facilitate better access in const contexts.

  Verification:
   * A full build of the entire project (googletest and googlemock) was completed successfully.
   * All existing tests were built and linked correctly with the new optimizations.

  Fixes #4943

### DIFF
--- a/googlemock/include/gmock/gmock-cardinalities.h
+++ b/googlemock/include/gmock/gmock-cardinalities.h
@@ -42,7 +42,7 @@
 #include <limits.h>
 
 #include <memory>
-#include <ostream>  // NOLINT
+#include <iosfwd>  // NOLINT
 
 #include "gmock/internal/gmock-port.h"
 #include "gtest/gtest.h"
--- a/googlemock/include/gmock/gmock-matchers.h
+++ b/googlemock/include/gmock/gmock-matchers.h
@@ -264,8 +264,7 @@
 #include <iterator>
 #include <limits>
 #include <memory>
-#include <ostream>  // NOLINT
-#include <sstream>
+#include <iosfwd>
 #include <string>
 #include <tuple>
 #include <type_traits>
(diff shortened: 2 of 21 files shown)


higher-performance:
Closing per https://github.com/google/googletest/issues/4943#issuecomment-5037606414.

PR #4951, opened 2026-04-04
CMake: Replace CMAKE_COMPILER_IS_GNUCXX with CMAKE_CXX_COMPILER_ID

petk:
CMAKE_COMPILER_IS_GNUCXX is obsolete variable and can be replaced with CMAKE_CXX_COMPILER_ID (also available since early CMake versions).

In the past CMake versions also LCC and QCC compilers had this variable set to boolean true but these aren't relevant here.


ameerabuhadwan94-tech (commented):
Startup

PR #4959, opened 2026-04-18
Fix: InitGoogleTest now consumes and removes the --help flag

Root-Aamir:
### Description
Currently, when `--help` is passed to a GoogleTest binary, `InitGoogleTest` sets the internal help flag but does not remove the flag from `argv`. This prevents downstream argument parsers from knowing that the flag has already been handled.

### Changes
- Updated `ParseGoogleTestFlagsOnlyImpl` in `gtest.cc` to set `remove_flag = true` when `--help` or other unrecognized GTest flags are encountered.
- This ensures `argc` is decremented and the help flag is removed from the arguments list after initialization.

### Verification
Reproduced the issue with a manual test runner. 
- **Before fix:** `argc` remained 2 after `InitGoogleTest` when passing `--help`.
- **After fix:** `argc` correctly becomes 1.


Root-Aamir:
I have signed the CLA.

Root-Aamir:
@googlebot I signed it.

PR #4988, opened 2026-06-01
Fix Python 3 debug output writes

### LINKED ISSUE #3943: googletest-output-test.py not fully Python3 ready
**Describe the bug**

googletest-output-test.py has a feature where you can set the
environment variable DEBUG_GTEST_OUTPUT_TEST to get dumps
of the output it saw and the output it expected to see.  However,
setting this caused the script to crash with a Python error along
the lines of "a bytes-like object is required."

`env DEBUG_GTEST_OUTPUT_TEST=1 ctest -R googletest-output-test -V`
The -V lets you see the crash; the script will create a file
`_googletest-output-test_normalized_actual.txt` but it is empty.

yes

Ubuntu 18.04

Python 3.8.0

cmake 3.17.3

**Additional context**

[stackoverflow](https://stackoverflow.com/questions/33054527/typeerror-a-bytes-like-object-is-required-not-str-when-writing-to-a-file-in) suggests that adding `.encode` to the strings being written
will solve the problem, and indeed that worked for me.

KirtiRamchandani:
Fixes #3943.

## Problem

`googletest-output-test.py` opens the `DEBUG_GTEST_OUTPUT_TEST` debug dump files in binary mode, then writes Python strings to them. Under Python 3, enabling the debug environment variable can crash with `TypeError: a bytes-like object is required`.

## Root cause

The debug output path writes normalized text data, but the files were opened with `wb`.

## Solution

Add a small helper that writes the debug dumps as UTF-8 text with newline conversion disabled, then use it for both normalized debug files. While validating the script with Python 3, I also fixed an invalid escape sequence warning in the Windows color-normalization regex without changing the intended pattern.

## Tests

- `python -m py_compile googletest/test/googletest-output-test.py`
- `git diff --check -- googletest/test/googletest-output-test.py`

### DIFF
--- a/googletest/test/googletest-output-test.py
+++ b/googletest/test/googletest-output-test.py
@@ -160,7 +160,7 @@ def NormalizeToCurrentPlatform(test_output):
 
   if IS_WINDOWS:
     # Removes the color information that is not present on Windows.
-    test_output = re.sub('\x1b\\[(0;3\d)?m', '', test_output)
+    test_output = re.sub(r'\x1b\[(0;3\d)?m', '', test_output)
     # Changes failure message headers into the Windows format.
     test_output = re.sub(r': Failure\n', r': error: ', test_output)
     # Changes file(line_number) to file:line_number.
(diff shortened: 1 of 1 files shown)


KirtiRamchandani:
Closing this to avoid duplicating #4986, which already covers the same issue.

PR #4993, opened 2026-06-04
Allow splitting MOCK_METHOD declaration and definition to improve build times

### LINKED ISSUE #3434: Splitting MOCK_METHOD declaration and definition
I'm working with classes that mocks a lot of methods and each MOCK_METHODxx results in method beining defined in each translation unit. Cookbook mentions defining constructor and destructor for mocks in separate .cpp file so it can be reused, it seems that it makes sense to split declaration and definition of each MOCK_METHOD in order to save some compilation time.

There is closed PR implementing this: https://github.com/google/googletest/pull/1592. Would you be interested in pull request with similar functionality?

nautical-exact:
Creates new macros `DECLARE_MOCK_METHOD(<args>)` and `DEFINE_MOCK_METHOD(MockClassName, <args>)`  (where `<args>` represents the same arguments that `MOCK_METHOD` takes). The intent is that (for users concerned about compilation speed) `DECLARE_MOCK_METHOD` is used inside the class body in place of `MOCK_METHOD` and simply _declares_ the mock method and `gmock_` method, and then `DEFINE_MOCK_METHOD` is used in a `.cpp` file and actually _defines_ those methods. Instantiating template functions like `FunctionMocker<>::InvokeWith` is a major source of build slowdowns, so moving those function calls into a `.cpp` file can significantly speed up compilation when the same mock is used in multiple test files.

Closes #3434 

--- 

### Appendix A: Build Speed Testing

<details>
  <summary>Build speed testing</summary>

Using a Python script, I generated several mocks each with several random mock methods, and then created several test files which expect a call on each mock method of each mock class. The script creates two versions of each mock: One using the standard "inline" mock definitions (`MOCK_METHOD`), and the other using the "split" mock definition (`DECLARE_MOCK_METHOD` and `DEFINE_MOCK_METHOD`) created in this PR.

<details>
  <summary>generate.py</summary>

```python

#!/usr/bin/env python3

import random
from pathlib import Path
import shutil

ROOT_DIR=Path(__file__).resolve().parent
OUTPUT_DIR = ROOT_DIR / "out-generated"

INT_TYPES=[
    "char", "signed char", "unsigned char",
    "short", "unsigned short",
    "int", "unsigned int",
    "long", "unsigned long",
    "long long", "unsigned long long"
]

def helper_gen_mock_header(mock_method_macro: str, n: int, mock: list[list[str]]) -> str:
    mock_method_defs = "\n".join(
        f"    {mock_method_macro}({spec[0]}, f_{i}, ({', '.join(spec[1:])}));"
        for i, spec in enumerate(mock)
    )
    
    return """
#pragma once
#include "gmock/gmock-function-mocker.h"
struct MOCK_%(n)s {
    MOCK_%(n)s();
    ~MOCK_%(n)s();
%(mock_method_defs)s
};
""" % {"n": n, "mock_method_defs": mock_method_defs}

def gen_inline_mock_header(n: int, mock: list[list[str]]) -> str:
    return helper_gen_mock_header("MOCK_METHOD", n, mock)

def gen_split_mock_header(n: int, mock: list[list[str]]) -> str:
    return helper_gen_mock_header("DECLARE_MOCK_METHOD", n, mock)

def gen_inline_mock_source(n: int) -> str:
    return """
#include "mock_%(n)s.h"

MOCK_%(n)s::MOCK_%(n)s() = default;
MOCK_%(n)s::~MOCK_%(n)s() = default;
""" % {"n": n}

def gen_split_mock_source(n: int, mock) -> str:
    mock_method_defs = "\n".join(
        f"DEFINE_MOCK_METHOD(MOCK_{n}, {spec[0]}, f_{i}, ({', '.join(spec[1:])}));"
        for i, spec in reversed(list(enumerate(mock)))
    )
    return "\n".join((gen_inline_mock_source(n), mock_method_defs))

def gen_test_source(n: int, mocks: list[list[list[str]]]) -> str:
    segments = []
    mock_includes = "\n".join(f'#include "mock_{i}.h"' for i in range(len(mocks)))
    mock_vars = "\n    ".join(f"MOCK_{i} mock_{i};" for i in range(len(mocks)))
    
    expect_calls_list = []
    actual_calls_list = []
    for i_mock, mock in enumerate(mocks):
        for i_method, method in enumerate(mock):
            expect_calls_list.append(
                f"EXPECT_CALL(mock_{i_mock}, f_{i_method}({', '.join(f'A<{t}>()' for t in method[1:])})).WillOnce(Return(0));"
            )
            actual_calls_list.append(
                f"EXPECT_EQ(0, mock_{i_mock}.f_{i_method}({', '.join('0' for _ in range(len(method) - 1))}));"
            )
    expect_calls = "\n    ".join(expect_calls_list)
    actual_calls = "\n    ".join(actual_calls_list)
    
    return """
#include "gtest/gtest.h"
%(mock_includes)s

using namespace testing;

TEST(Suite_%(n)s, Test_%(n)s)
{
    %(mock_vars)s
    %(expect_calls)s
    %(actual_calls)s
}
""" % {
        "n": n,
        "mock_includes": mock_includes,
        "mock_vars": mock_vars,
        "expect_calls": expect_calls,
        "actual_calls": actual_calls
    }


if __name__ == "__main__":
    NUM_TESTS = 8
    NUM_MOCKS = 4
    METHODS_PER_MOCK = 4
    PARAMS_PER_METHOD = 6
    mocks = [
        [random.choices(INT_TYPES, k=PARAMS_PER_METHOD) for _ in range(METHODS_PER_MOCK)]
        for _ in range(NUM_MOCKS)
    ]
    
    shutil.rmtree(OUTPUT_DIR, ignore_errors=True)
    OUTPUT_DIR.mkdir(parents=True)

    INLINE_OUTPUT_DIR = OUTPUT_DIR / "inline"
    SPLIT_OUTPUT_DIR = OUTPUT_DIR / "split"
    INLINE_OUTPUT_DIR.mkdir()
    SPLIT_OUTPUT_DIR.mkdir()

    for i, mock in enumerate(mocks):
        (INLINE_OUTPUT_DIR / f"mock_{i}.h").write_text(gen_inline_mock_header(i, mock))
        (INLINE_OUTPUT_DIR / f"mock_{i}.cpp").write_text(gen_inline_mock_source(i))
        (SPLIT_OUTPUT_DIR / f"mock_{i}.h").write_text(gen_split_mock_header(i, mock))
        (SPLIT_OUTPUT_DIR / f"mock_{i}.cpp").write_text(gen_split_mock_source(i, mock))
        
    for i in range(NUM_TESTS):
        text = gen_test_source(i, mocks)
        (INLINE_OUTPUT_DIR / f"test_{i}.cpp").write_text(text)
        (SPLIT_OUTPUT_DIR / f"test_{i}.cpp").write_text(text)
````
</details>

<details>
<summary>compile.sh</summary>

```bash
#!/usr/bin/env bash

set -euo pipefail

ROOT_DIR="$(dirname -- "$(realpath -- "${BASH_SOURCE[0]}")")"

COMPILER="${CXX:-g++}"

SOURCE_DIR="$(realpath "$1")"
BUILD_DIR="${SOURCE_DIR}/build"

FLAGS=(
    "-I${SOURCE_DIR}"
    "-I${ROOT_DIR}/googletest/googletest/include"
    "-I${ROOT_DIR}/googletest/googlemock/include"
    "-L${ROOT_DIR}/googletest/build/lib"
    '-std=c++17'
)

declare -A SOURCE_TO_OBJECT=()

while IFS= read -r -d $'\0' S; do
    O="$S"
    O="$BUILD_DIR${O#"${SOURCE_DIR}"}"
    O="${O%.cpp}.o"
    SOURCE_TO_OBJECT["$S"]="$O"
done < <(find "$SOURCE_DIR" -name '*.cpp' -print0)

rm -rf "$BUILD_DIR"
mkdir -p "$BUILD_DIR"

time {
    paste <(printf "%s\n" "${SOURCE_TO_OBJECT[@]}") <(printf "%s\n" "${!SOURCE_TO_OBJECT[@]}") | \
        xargs ${PARALLEL+"-P${PARALLEL}"} -L1 "$COMPILER" "${FLAGS[@]}" -c -o

    "$COMPILER" "${FLAGS[@]}" "${SOURCE_TO_OBJECT[@]}" -lgmock -lgtest_main -lgtest -o "$BUILD_DIR/test.bin"
}
```
</details>

The exact results depend on the number of mock classes and test files generated, and how many mock methods are in each mock class, but I typically saw speedups of about 25% for sequential builds:

```bash
$ ./generate.py

$ ./compile.sh out-generated/inline/

real    0m41.871s
user    0m38.528s
sys     0m3.342s

$ ./compile.sh out-generated/split/

real    0m30.587s
user    0m27.481s
sys     0m3.102s
```

EDIT:

Running with various numbers of tests and mocks gives us a bit more data:

#### Total compilation time (seconds) using split mocks (blue) or inline mocks (red) vs. number of mocks and tests:
Figure_1

#### Percent change in total compilation time vs. number of mocks and tests:
Screenshot from 2026-06-07 13-13-35

Unfortunately there's not much of a trendline, I think I would have to run this quite a few times to filter out random noise. But broadly we're seeing around 15-25% time savings, and the time savings are better the more mocks and tests we have.

</details>


nautical-exact:
This PR is ready for review.

Not sure exactly what the process is in this repo for merging external PRs or who the active maintainers are. @derekmauro, the last external PR mentioned you, could you take a look at this?

CDAlexanderW:
+1 for getting this please. We have build performance problems for a very large code base and think this could help a lot!

CDAlexanderW:
@nautical-exact I had a short look on your changes and from what I tried, it looks like someone should not or cant use the override keyword in the DEFINE_ macro, since it simply tries to append it to the function definition which does not work and does not make sense. Someone can simply not put the override keyword there. But be aware that you put that in the documentation text sample which in my perspective is wrong because it does not compile. Please correct me if I am wrong.
I would prefer that when having the override keyword thee it simply gets ignored in the DEFINE_ macro, as someone could the share the DECLARE_ and DEFINE_ by an include file where we redefine the macro DECLARE_ in header and DEFINE_ in cpp. Although I am not sure if this is a really good idea.
Besides, some parameter types in my code seem to be more strictly checked for scope in the DEFINE_ macro (which may make sense) in contrast to normal MOCK_METHOD macro, so I simply added it.
Hopefully we can also find a maintainer that reviews this and approves for merge.
The speed improvement for large code bases is significant to support this!

nautical-exact:
@CDAlexanderW Thank you so much for taking a look!

You're indeed correct, `override` and `final` don't work, because they're only valid on method declarations and not definitions. I think you're right that it's best to have the DEFINE macros simply ignore them; it makes migration/copy-pasting from the declaration to the definition easier. I've made that fix.

To resolve the scoping issues, I've switched to using trailing return type for the DEFINE macros (and also moved some static asserts into the function body). The cause of the issue is that if we have a type like `struct S { using X = int; X f(X); };`, then an out-of-class function definition `X S::f(X) { /* ... */ }` is invalid (since the `X` typedef isn't in scope when writing the return type, it would need to be `S::X S::f(X) { /* ... */ }`). However a definition using a trailing return type `auto S::f(X) -> X { /* ... */ }` _is_ valid, so we can just use that (because everything after the `S::` is already within the scope of the class).

I've added test cases to ensure that both of the above work. Let me know if you find any other issues!

CDAlexanderW:
@nautical-exact thx for the quick reply and code change. I tested your changed code and can confirm override can stay there without error and in my test case I also didn't need to change anything regarding scope.

CDAlexanderW:
Now the question is what needs to be done so this gets one step further?
@derekmauro I can imagine you are very busy, can you please help out, or give us at least a hint? This would really be a game changer for us in gtest.

notable-equivalent:
+1 on this, build times have been a big issue for us so anything that could help would be much appreciated. 

I apologize if this is rude, but since it's not clear who's maintaining this repo, I'm just going to @ everyone at Google who merged anything into this repo recently, in the hopes that someone knows what the process is for merging external PRs. @mkruskal-google @jacobsa @dsharletg @tweenk @Mizux

CDAlexanderW:
@fuzzard Sorry to disturb you, but we are lost here. Do you know who is maintaining this repo? Or who we have to contact? Any help in any direction would be helpful.

Mizux:
my bet on @derekmauro IIRC tech lead of abseil-cpp and googletest ?

spacible:
@derekmauro sorry to ping you. But we are stuck here without any help. No one reacts or gives even the slightest information. 

Can you tell us who is responsible?
Who we should contact?
Or also if there are no public pull request even accepted.

I think that much more people would appreciate this changes (and wait for it), than it might seem here. Just think about how much build time is wasted worldwide in large projects that could get improved.

CDAlexanderW:
@mkruskal-google Can you help here by providing at least minimal info, one sentence. So we know how things are. Even if it is the info you don't know anything or can't help with who we should contact. thank you!

spacible:
@nautical-exact can you have a look why the CI builds failed (I can’t access the logs) and if it comes from your changes provide a fix?

thx to Mike Kruskal for bringing things forward. Appreciate it very much.

nautical-exact:
I actually also can't access the logs, the "details" links seem to point to `fusion2/ci/kokoro/prod%3Agoogletest%2Fgithub%2Fpresubmit%2Flinux/activity/7aef57e4-3a9c-46d4-9dcf-43c9c35c134c` which seems to be an internal Google link of some kind (at the very least, it doesn't look like a valid public URL to me).

I see some presibmit scripts in the `ci` folder, I should have some time to try running those locally later

googlemock/include/gmock/gmock-function-mocker.h
```diff
@@ -208,6 +309,70 @@ using internal::FunctionMocker;
...
+      GMOCK_PP_REPEAT(GMOCK_INTERNAL_MATCHER_PARAMETER, _Signature, _N))       \
+      GMOCK_PP_IF(_Constness, const, ) _RefSpec;                               \
+  ::testing::MockSpec<GMOCK_PP_REMOVE_PARENS(_Signature)> gmock_##_MethodName( \
+      const ::testing::internal::WithoutMatchers&,                             \
+      GMOCK_PP_IF(_Constness, const, )::testing::internal::Function<           \
+          GMOCK_PP_REMOVE_PARENS(_Signature)>*) const _RefSpec _NoexceptSpec;  \
+  ::testing::FunctionMocker<GMOCK_PP_REMOVE_PARENS(_Signature)>&               \
+      gmock_get_mocker_##_MethodName(                                          \
+          ::testing::internal::TypeTag<GMOCK_PP_REMOVE_PARENS(                 \
+              _Signature) GMOCK_PP_IF(_Constness, const, ) _RefSpec>) const {  \
```
nautical-exact: Found the issue. This line is causing a warning because `_Signature` is often already wrapped in something like `identity_t<>`, causing this line to expand to types like `::testing::internal::TypeTag<::testing::internal::identity_t<int(int)> const>`, which is a "function type" which is const, which is not valid (or rather, it's valid but it's the same as just `::testing::internal::TypeTag<int(int)>`, the const gets dropped, which is what the compiler warns us about). What we _want_ is `::testing::internal::TypeTag<int(int) const>`, which is a "const-qualified function type" (that is, it's a function type which specifies that it would be called on a const object). 

Since the signature could be a typedef or wrapped in `identity_t` (and therefore we can't just blindly add `const` after it with the macro), and it's difficult to apply arbitrary cv-ref qualifiers to a function type using templates (because we don't want `T const`, instead we need to match `T` as `Ret(Args...)` and then we want `Ret(Args...) const`, I'm just going to add a second template argument to the TypeTag helper which is a normal type like an `int` with all the qualifiers added to that.

nautical-exact:
Alright everything should be working now. I confirmed that the linux-presubmit.sh script passes locally

mkruskal-google:
Apologies for the delay here and for confusing things by running tests with invisible logs (kudos on debugging that!).

While I completely understand the desire to improve build times by avoiding mock recompilation, I don't think we can accept this PR today. To be transparent, our team doesn't currently have the bandwidth to keep up with the rate of PRs and issues to Googletest, along with the internal maintenance. Because of that, we have to be extremely strict about taking on any new API surface or ongoing maintenance.

A couple of alternatives to consider:

- C++/Clang modules, which could have larger wins on your build times overall (and also avoid recompilation of mocks)
- Maintain these internally, as a custom extension of Googletest. They depend on internals and may not be extremely stable, but if the savings are valuable enough it might be worth it.

CDAlexanderW:
@mkruskal-google Big thx for the response!

Who do we need to contact or get approval from, so this can be merged then?
What is needed so this PR gets an approval?

Regarding your suggestions (and thanks for taking time to even think about a solution):
I don't think modules will work without additional gtest/gmock PRs, which brings us back to start. Can you give us any example where this has been used already in a large codebase? The new defines are quite easy to use (also because they are optional), even for larger code base.
There are rules which prevent many from modifying anything (or using on top) beside the official release/branch/commit. So this is also no solution.

@nautical-exact thx for all the good work!

mkruskal-google:
You found the right people, we're just a small team and not currently able to merge this PR.

- Modules - it's not that we won't accept *any* PRs to Googletest, it's just that the maintenance costs of this one are too high.  A PR that fixes issues to make Googletest compatible with modules is a different story, and would be much easier to approve (assuming they're small, which I would guess they would be).  https://clang.llvm.org/docs/Modules.html has some more info on C++ and clang modules

- Internal maintenance - While I understand you might have strict internal rules about using an unreleased Googletest, these macros don't actually need to be in *this* project.  You could define these macros in a separate repo or in your own project as custom extensions build *on top of* Googletest.

PR #5002, opened 2026-06-21
Initialize `match` variable in `MatchAndExplain`'s loop

poyea:
This prevents the warning of `variable 'match' is not initialized [cppcoreguidelines-init-variables]`.

Testing: The warning is gone with `clang-tidy`.

```
googlemock/include/gmock/gmock-matchers.h:3617:12: warning: variable 'match' is not initialized [cppcoreguidelines-init-variables]
 3617 |       bool match;  // Does the current element match the current matcher?
      |            ^    
      |                  = false
```


mkruskal-google:
Thanks for the contribution!  But tbh this seems like just a bad warning?  MSAN will catch *uses* of uninitialized variables, but I don't really see the harm in an uninitialized variable that is provably initialized before it's used.  Even if we accepted this PR, I don't think we'd want to enable this warning in our CI, and future changes could easily regress it

PR #5012, opened 2026-07-07
Add default case to death test AssumeRole() switch

r-barnes:
Meta has `-Wswitch-default` enabled for large parts of its repository. This warning requires a `default` case to cover non-exhaustive enumeration handling in `switch` statements.

The `switch (gtest_dt->AssumeRole())` statement in `gtest-death-test-internal.h` does not currently have a `default` case, which triggers this warning. This diff provides one, logging a fatal error for any unexpected `AssumeRole()` return value.

I expect this is better than alternatives such as error suppression at the build or file level because it will be cross-compatible with every compiler.

### DIFF
--- a/googletest/include/gtest/internal/gtest-death-test-internal.h
+++ b/googletest/include/gtest/internal/gtest-death-test-internal.h
@@ -244,6 +244,8 @@ GTEST_API_ bool ExitedUnsuccessfully(int exit_status);
           gtest_dt->Abort(::testing::internal::DeathTest::TEST_DID_NOT_DIE);   \
           break;                                                               \
         }                                                                      \
+        default:                                                               \
+          GTEST_LOG_(FATAL) << "Unexpected AssumeRole() return value.";        \
       }                                                                        \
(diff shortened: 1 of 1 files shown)


mkruskal-google:
I'm a little confused why we're tripping the warning, since this *is* an exhaustive switch.  This also means we won't get a warning if we ever add a new enum value

mkruskal-google:
Ah ok to clarify, in Google we generally use clang's default behavior of `-Wswitch`.  *That* warning only triggers when you have no default on non-exhaustive switches.  For exhaustive switches it allows a missing default.  This has the benefit that it makes exhaustive switches start warning if a new enum value is added.

`-Wswitch-default` unconditionally requires a default case.  This could be safer for switches on variables you don't control (e.g. if they're user input and could fall outside of known values).  But it won't alert you when a new value is added.

I think in this case disabling the warning is correct.  You generally shouldn't be applying strict warnings over your dependencies, and this code is all purely internal to googletest

PR #5020, opened 2026-07-09
fix: detect pthread-enabled Emscripten in GTEST_HAS_PTHREAD

### LINKED ISSUE #5015: [Bug]: Inconsistent `GTEST_HAS_PTHREAD` causes ODR violation that crashes any `EXPECT_CALL`
### Describe the issue

When GoogleTest is built for Emscripten with pthreads (`-pthread`), the gtest/gmock libraries end up compiled with `GTEST_HAS_PTHREAD=1`, but consumer translation units are compiled with `GTEST_HAS_PTHREAD=0`. The two disagree, and the result is an ODR violation that makes essentially any gmock test crash at startup.

- gtest's CMake (`internal_utils.cmake`) runs `find_package(Threads)`, which succeeds under Emscripten's `-pthread`, so it appends `-DGTEST_HAS_PTHREAD=1` to the gtest/gmock targets' compile flags.
- gtest's header (`gtest-port.h`) auto-detects `GTEST_HAS_PTHREAD=0` on Emscripten, because Emscripten is not in the list of pthread-capable platforms.
- That macro is private to gtest's own targets, so it is **not** propagated to consumers. Consumer TUs therefore fall back to the header's auto-detection and get `0`.

`GTEST_HAS_PTHREAD` selects between two different definitions of `testing::internal::ThreadLocal<T>` (a pthread-backed one vs. a single-threaded one) that have different member layouts. At runtime the `FunctionMocker<F>` class uses this. Building gtest/gmock with `-fsanitize=alignment` results in the following error:

```
gmock-spec-builders.h:1585:26: runtime error: member call on misaligned address
0x000000000002 for type 'testing::Sequence', which requires 8 byte alignment
```

`CMakeLists.txt`:

```cmake
cmake_minimum_required(VERSION 3.16)
project(repro CXX)

set(THREADS_PREFER_PTHREAD_FLAG ON)
find_package(Threads REQUIRED)

include(FetchContent)
FetchContent_Declare(googletest
  GIT_REPOSITORY https://github.com/google/googletest.git
  GIT_TAG        main)
FetchContent_MakeAvailable(googletest)

add_executable(repro test.cpp)
target_link_libraries(repro GTest::gmock GTest::gmock_main Threads::Threads)
```

`test.cpp`:

```cpp
#include <gmock/gmock.h>
#include <gtest/gtest.h>
#include <string_view>
using ::testing::StrictMock;
namespace {
class M { public: MOCK_METHOD(void, call, (std::string_view), (const)); };
}
TEST(Repro, ExpectCall) {
  StrictMock<M> m;
  EXPECT_CALL(m, call(std::string_view("a")));
  m.call("a");
}
```

Build and run (Emscripten + Node):

```sh
emcmake cmake -B build -G Ninja \
  -DCMAKE_CXX_FLAGS="-pthread" \
  -DCMAKE_EXE_LINKER_FLAGS="-pthread -sALLOW_MEMORY_GROWTH=1"
cmake --build build
node build/repro.js
```

`git rev-parse HEAD` = `7917641ff965959afae189afb5f052524395525c` (2025-09-11).

- Target platform: Emscripten (emsdk 6.0.2), run under Node.js v25.
- Build host: Ubuntu 24.04 (Linux).

The bug is specific to the Emscripten platform; the host OS is not relevant.

Emscripten `em++` 6.0.2 (`em++ -v`):

```
emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 6.0.2 (7a2d97d627ff4945eae28847ce0387ac52b92c09)
clang version 23.0.0git (https://github.com/llvm/llvm-project 787619a4072e0eb7887357d5d284e86c17548aed)
Target: wasm32-unknown-emscripten
```

CMake 3.28.3 (`cmake --version`), using `FetchContent` to add GoogleTest (the standard integration). The problem also reproduces with direct `em++` compile/link commands and no build system at all.

## Potential fix
In `gtest-port.h`:

```c
#if (defined(GTEST_OS_LINUX) || defined(GTEST_OS_MAC) || ... ||
     defined(GTEST_OS_AIX) || defined(GTEST_OS_ZOS) ||
     (defined(__EMSCRIPTEN__) && defined(__EMSCRIPTEN_PTHREADS__)))   // <-- add this
#define GTEST_HAS_PTHREAD 1
#else
#define GTEST_HAS_PTHREAD 0
#endif
```

`__EMSCRIPTEN_PTHREADS__` is defined by Emscripten if and only if the TU is compiled with `-pthread`.

## Workaround for users
```cmake
set(gtest_disable_pthreads ON CACHE BOOL "" FORCE)
# or alternatively
add_compile_definitions(GTEST_HAS_PTHREAD=1)

# both have to come before
FetchContent_MakeAvailable(googletest)
```

devsarthak2005:
## Summary

Fix inconsistent `GTEST_HAS_PTHREAD` detection on Emscripten when building with `-pthread`.

GoogleTest's CMake configuration correctly enables `GTEST_HAS_PTHREAD=1` for the library when pthreads are available. However, `gtest-port.h` does not recognize pthread-enabled Emscripten during header auto-detection, causing consumer translation units to use `GTEST_HAS_PTHREAD=0`.

This mismatch results in different implementations of internal synchronization types such as `ThreadLocal<T>`, leading to an ODR violation and runtime crashes in GMock (for example, `EXPECT_CALL`).

## Changes

* Update `gtest-port.h` to treat Emscripten as pthread-capable only when `__EMSCRIPTEN_PTHREADS__` is defined.
* Keep behavior unchanged for all other platforms.

## Why this fix?

`__EMSCRIPTEN_PTHREADS__` is defined only when Emscripten is compiled with `-pthread`, making it the correct indicator for enabling pthread support. This keeps header-based detection consistent with the build configuration while avoiding any impact on non-pthread Emscripten builds.

Fixes #5015.

### DIFF
--- a/googletest/include/gtest/internal/gtest-port.h
+++ b/googletest/include/gtest/internal/gtest-port.h
@@ -593,7 +593,8 @@ typedef struct _RTL_CRITICAL_SECTION GTEST_CRITICAL_SECTION;
      defined(GTEST_OS_DRAGONFLY) || defined(GTEST_OS_GNU_KFREEBSD) || \
      defined(GTEST_OS_OPENBSD) || defined(GTEST_OS_HAIKU) ||          \
      defined(GTEST_OS_GNU_HURD) || defined(GTEST_OS_SOLARIS) ||       \
-     defined(GTEST_OS_AIX) || defined(GTEST_OS_ZOS))
+     defined(GTEST_OS_AIX) || defined(GTEST_OS_ZOS) ||                \
+     (defined(GTEST_OS_EMSCRIPTEN) && defined(__EMSCRIPTEN_PTHREADS__)))
 #define GTEST_HAS_PTHREAD 1
 #else
 #define GTEST_HAS_PTHREAD 0


mkruskal-google:
Closing as dupe of #5019

PR #5036, opened 2026-07-21
Prefer -pthread flag on CMake

### LINKED ISSUE #2482: -pthread should probably be used instead of -lpthread

Tachi107:
In the past, GoogleTest could not set this flag unconditionally as it would break CUDA on CMake versions older than 3.13. As the project now requires at least CMake 3.16, it's safe to enable and finally prefer -pthread to -lpthread.

This does not make any difference on systems with glibc 2.34 as newer, as libpthread is part of libc now.

Properly fixes #2482

Cc: @higher-performance

### DIFF
--- a/googletest/cmake/internal_utils.cmake
+++ b/googletest/cmake/internal_utils.cmake
@@ -63,6 +63,7 @@ macro(config_compiler_and_linker)
   unset(GTEST_HAS_PTHREAD)
   if (NOT gtest_disable_pthreads AND NOT MINGW)
     # Defines CMAKE_USE_PTHREADS_INIT and CMAKE_THREAD_LIBS_INIT.
+    set(THREADS_PREFER_PTHREAD_FLAG TRUE)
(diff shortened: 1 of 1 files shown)


higher-performance:
Thanks for sending, I'll give it a shot.

Tachi107:
Thanks for merging!

PR #5044, opened 2026-07-25
fix: avoid double invocation of side-effecting matchers in PredicateFormatterFromMatcher

### LINKED ISSUE #4073: EXPECT_THAT with Throws invokes callable twice when exception is not thrown
**Describe the bug**

`EXPECT_THAT` macro with the `Throws` expectation invokes the callable twice when it does not throw an exception. This does not happen when the exception is thrown.

This causes failures for stateful lambdas or lambdas with side effects.

Repro code:
```cpp
#include <exception>
#include <iostream>
#include <gmock/gmock.h>

int main()
{
    int state = 1;
    EXPECT_THAT(([&] { std::cout << "I am invoked " << state++ << '\n'; }),
        testing::Throws<std::exception>());
}
```

Outputs:
```
I am invoked 1
I am invoked 2
...
```

Also see https://godbolt.org/z/WE44hn5Kc.

Yes.

Does not matter. The issue happens on both Windows and Linux.

MSVC 17.4.1, clang trunk.

K-S-Manikandan:
EXPECT_THAT with Throws<E>() previously invoked the callable twice on failure: once in Matches() and again in MatchPrintAndExplain() when formatting the diagnostic message. This caused issues with stateful or side-effecting callables.

Fix: call MatchAndExplain() once with an interested listener, capture the result and explanation, then manually format the failure message using UniversalPrint() + type info + listener explanation.

BREAKING CHANGE: PredicateFormatterFromMatcher no longer detects flaky matchers that change their result based on IsInterested(). This only affects pathological matchers; real matchers return consistent results regardless of listener interest.

Fixes #4073

### DIFF
--- a/googlemock/include/gmock/gmock-matchers.h
+++ b/googlemock/include/gmock/gmock-matchers.h
@@ -1719,9 +1719,13 @@ class [[nodiscard]] PredicateFormatterFromMatcher {
     // potentially unsafe downcasting of the matcher argument.
     const Matcher<const T&> matcher = SafeMatcherCast<const T&>(matcher_);
 
-    // The expected path here is that the matcher should match (i.e. that most
-    // tests pass) so optimize for this case.
-    if (matcher.Matches(x)) {
+    // Capture the explanation during the initial match attempt so that
+    // side-effecting matchers (like Throws) are not invoked a second time
+    // when formatting the failure message.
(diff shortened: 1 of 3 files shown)


K-S-Manikandan:
Superseded by #5030 (same fix, earlier PR)

PR #5057, opened 2026-08-11
Fix -Wunused-template warnings

victorvianna:
Clang 23 turns on this warning under -Wall.
https://github.com/llvm/llvm-project/blob/release/23.x/clang/docs/ReleaseNotes.md?plain=1#L602

### DIFF
--- a/googlemock/include/gmock/gmock-matchers.h
+++ b/googlemock/include/gmock/gmock-matchers.h
@@ -6028,8 +6028,8 @@ PolymorphicMatcher<internal::ExceptionMatcherImpl<Err>> ThrowsMessage(
     };                                                                         \
   };                                                                           \
   template <GMOCK_INTERNAL_MATCHER_TEMPLATE_PARAMS(args)>                      \
-  inline full_name<GMOCK_INTERNAL_MATCHER_TYPE_PARAMS(args)> name(             \
-      GMOCK_INTERNAL_MATCHER_FUNCTION_ARGS(args)) {                            \
+  [[maybe_unused]] inline full_name<GMOCK_INTERNAL_MATCHER_TYPE_PARAMS(args)>  \
+  name(GMOCK_INTERNAL_MATCHER_FUNCTION_ARGS(args)) {                           \
     return full_name<GMOCK_INTERNAL_MATCHER_TYPE_PARAMS(args)>(                \
         GMOCK_INTERNAL_MATCHER_ARGS_USAGE(args));                              \
   }                                                                            \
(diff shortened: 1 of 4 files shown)


googlemock/include/gmock/gmock-matchers.h
```diff
@@ -6028,8 +6028,8 @@ PolymorphicMatcher<internal::ExceptionMatcherImpl<Err>> ThrowsMessage(
     };                                                                         \
   };                                                                           \
   template <GMOCK_INTERNAL_MATCHER_TEMPLATE_PARAMS(args)>                      \
-  inline full_name<GMOCK_INTERNAL_MATCHER_TYPE_PARAMS(args)> name(             \
-      GMOCK_INTERNAL_MATCHER_FUNCTION_ARGS(args)) {                            \
+  [[maybe_unused]] inline full_name<GMOCK_INTERNAL_MATCHER_TYPE_PARAMS(args)>  \
```
mkruskal-google, line 6031: Where does this one show up?  This doesn't seem right.  If someone defines a matcher and it's unused, I would *expect* this warning to trigger

victorvianna: googlemock/test/gmock-matchers-comparisons_test.cc:2214-2215 — Construct1Arg and Construct2Args. TEST(MatcherConstruct, ExplicitVsImplicit) checks the generated class templates (Construct1ArgMatcherP<int>, Construct2ArgsMatcherP2<int, double>) as types only, so the factory templates are never instantiated.

Worth flagging on scope, since it bears on your point:

- This only fires for matchers with internal linkage — anonymous namespace or static. A matcher defined in a shared header never warns either way, so the suppression is narrower than it looks.
- Per the PR description, Clang 23 promotes -Wunused-template into -Wall, so this stops being an opt-in -Weverything complaint and starts hitting everyone who defines a matcher in an anonymous namespace in a test file — which is the common case.

On whether it should warn: the thing that pushed me toward suppressing is that a user can't silence it themselves. The declaration lives inside the macro expansion, so there's nowhere to attach [[maybe_unused]]; they'd have to wrap the matcher definition in pragmas. That's also presumably why non-parameterized MATCHER already suppresses -Wunused-function / -Wunused-member-function on its factory at gmock-matchers.h:5947-5950 — same situation, and -Wunused-template is just the template-valued version of it. So this reads to me as restoring consistency with a call that was already made, rather than making a new one.

That said, it's your call. If you'd rather keep unused matchers flagged in user code, I'll drop the gmock-matchers.h change and instantiate the factories in the test instead (e.g. via `(void)Construct1Arg(1);`). Let me know which you prefer.

PR #5069, opened 2026-08-19
Don't repeat the failure summary in the XML report body

### LINKED ISSUE #4785: [Bug]: gtest_output xml copies failure message into both <failure> element body and attribute 'message'
### Describe the issue

Google test seems to fill the junit-style XML document produced with with the failure from ASSERT or EXPECT statements twice. This simply reduces the quality of the failure report.

I'm working with some very large matrices that get dumped on failure and I lost several minutes of time before realizing that my reporting system was showing me the same data twice.

write a simple test to induce a failure

```cpp
TEST(test_suite, test_function)
{
  EXPECT_EQ(1, 2) << "I figure 1 is close enough to 2 they should be the same";
}
```

run this test suite with `--gtest_output=xml`

expected test_detail.xml:

_note: what follows is a manually mangled version of an existing failing test I have_

```xml
<?xml version="1.0" encoding="UTF-8"?>
<testsuites tests="3" failures="1" disabled="0" errors="0" time="0.978" timestamp="2025-06-26T14:44:15.422" name="AllTests">
  <testsuite name="test_api_optimal_lh" tests="2" failures="1" disabled="0" skipped="0" errors="0" time="0.978" timestamp="2025-06-26T14:44:15.422">
    <testcase name="test_function" file="/home/groostav/sscce/demo.cpp" line="14" status="run" result="completed" time="0.004" timestamp="2025-06-26T14:44:15.422" classname="test_suite">
      <failure message="expectation violation" type=""><![CDATA[/home/groostav/sscce/demo.cpp:52
  Actual: 1
Expected: 2
I figure 1 is close enough to 2 they should be the same	
]]></failure>
    </testcase>
  </testsuite>
</testsuites>
```

(note: a very terse summary in `message`)

actual:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<testsuites tests="3" failures="1" disabled="0" errors="0" time="0.978" timestamp="2025-06-26T14:44:15.422" name="AllTests">
  <testsuite name="test_api_optimal_lh" tests="2" failures="1" disabled="0" skipped="0" errors="0" time="0.978" timestamp="2025-06-26T14:44:15.422">
    <testcase name="test_function" file="/home/groostav/sscce/demo.cpp" line="14" status="run" result="completed" time="0.004" timestamp="2025-06-26T14:44:15.422" classname="test_suite">
      <failure message="/home/groostav/sscce/demo.cpp:52&#x0A; Actual: 1&#x0A;Expected: 2&#x0A;I figure 1 is close enough to 2 they should be the same&#x0A;" type=""><![CDATA[/home/groostav/sscce/demo.cpp:52
  Actual: 1
Expected: 2
I figure 1 is close enough to 2 they should be the same	
]]></failure>
    </testcase>
  </testsuite>
</testsuites>
```

note the duplication of the comparison failure and error message; it is repeated twice, once it is encoded for XML attribute string, the other time it is dumped in the body of an element as a CDATA block.

1.14.0

pulled via cmake's FetchContent: https://github.com/google/googletest/archive/refs/tags/v1.14.0.zip

Ubuntu 24.04 (linux 6.6.8)

```
uname -a
Linux Batou 6.6.87.2-microsoft-standard-WSL2 #1 SMP PREEMPT_DYNAMIC Thu Jun  5 18:30:46 UTC 2025 x86_64 x86_64 x86_64 GNU/Linux
groostav@Batou:~/Metamodelling/cmake-build-debug-ninja-vcpkg-wsl/bin$ lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 24.04.2 LTS
Release:        24.04
Codename:       noble
```

gcc 13.3.0

```
gcc --version
gcc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
```

cmake 3.28.3
ninja 1.11.1

### Additional context

I use junit all the time but i don't actually have the runner installed, if need be i can find out what junit does exactly but IIRC the `message` will just be the java exception's message and the failure body will be the exception printed with stack trace.

nithin42:
Fixes #4785.

`XmlUnitTestResultPrinter::OutputXmlTestResult` writes the summary of a test part result into the `message` attribute of the `<failure>` (or `<skipped>`) element and the full message into the CDATA body of the same element:

```cpp
const std::string summary = location + "\n" + part.summary();
*stream << "      <failure message=\"" << EscapeXmlAttribute(summary) << "\" type=\"\">";
const std::string detail = location + "\n" + part.message();
OutputXmlCDataSection(stream, RemoveInvalidXmlCharacters(detail).c_str());
```

`summary()` is `message()` with the stack trace cut off, so the two are only different when a stack trace was recorded. In a build without stack trace support they are identical, and for a skipped result they are always identical because `ShouldEmitStackTraceForResultType` never asks for a trace on `kSkip`. The report then carries the same text twice, once escaped as an attribute and once as CDATA.

Before, running a test that does `EXPECT_EQ(1, 2)`:

```xml
<failure message="test.cc:62&#x0A;Expected equality of these values:&#x0A;  1&#x0A;  2&#x0A;" type=""><![CDATA[test.cc:62
Expected equality of these values:
  1
  2
]]></failure>
```

After:

```xml
<failure message="test.cc:62&#x0A;Expected equality of these values:&#x0A;  1&#x0A;  2&#x0A;" type="" />
```

The CDATA body is now written only when it holds more than the attribute already does, which in practice means when there is a stack trace to add. The `message` attribute is left alone, still holding the summary, which is the part JUnit consumers read and the part that the docs already show as the only content of a `<failure>` element. Nothing is dropped from the report: whatever used to be in the body but is no longer written was a byte for byte copy of the attribute. The JSON printer already reports the message just once.

Testing:

* `gtest_xml_output_unittest.py` gets a new `testFailureDetailIsNotACopyOfTheMessage` case that walks every `<failure>` and `<skipped>` element in the report and asserts its detail is not a copy of its `message` attribute. It fails on the current code and passes with the change.
* The expected documents in that file were updated. `<skipped>` elements lose their body unconditionally since they never carry a stack trace; `<failure>` bodies are dropped only in the no stack trace configuration, which is what the new `RemoveRepeatedDetails` helper does, so the expectations still describe both configurations.
* The example in `docs/advanced.md` was changed to match the output.
* Ran the CMake test suite (`cmake -Dgtest_build_tests=ON -Dgmock_build_tests=ON`, then `ctest`) on Linux with GCC.

### DIFF
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -2173,8 +2173,8 @@ could generate this report:
 <testsuites tests="3" failures="1" errors="0" time="0.035" timestamp="2011-10-31T18:52:42" name="AllTests">
   <testsuite name="MathTest" tests="2" failures="1" errors="0" time="0.015">
     <testcase name="Addition" file="test.cpp" line="1" status="run" time="0.007" classname="">
-      <failure message="Value of: add(1, 1)&#x0A;  Actual: 3&#x0A;Expected: 2" type="">...</failure>
-      <failure message="Value of: add(1, -1)&#x0A;  Actual: 1&#x0A;Expected: 0" type="">...</failure>
+      <failure message="Value of: add(1, 1)&#x0A;  Actual: 3&#x0A;Expected: 2" type=""/>
+      <failure message="Value of: add(1, -1)&#x0A;  Actual: 1&#x0A;Expected: 0" type=""/>
     </testcase>
     <testcase name="Subtraction" file="test.cpp" line="2" status="run" time="0.005" classname="">
     </testcase>
--- a/googletest/src/gtest.cc
+++ b/googletest/src/gtest.cc
@@ -4372,11 +4372,20 @@ void XmlUnitTestResultPrinter::OutputXmlTestResult(::std::ostream* stream,
           internal::FormatCompilerIndependentFileLocation(part.file_name(),
                                                           part.line_number());
       const std::string summary = location + "\n" + part.summary();
-      *stream << "      <failure message=\"" << EscapeXmlAttribute(summary)
-              << "\" type=\"\">";
       const std::string detail = location + "\n" + part.message();
-      OutputXmlCDataSection(stream, RemoveInvalidXmlCharacters(detail).c_str());
-      *stream << "</failure>\n";
+      *stream << "      <failure message=\"" << EscapeXmlAttribute(summary)
+              << "\" type=\"\"";
+      // The "message" attribute already carries the summary, so only write a
+      // CDATA body when the full message adds something to it (a stack trace,
+      // for instance). Otherwise the exact same text is reported twice.
+      if (detail == summary) {
+        *stream << " />\n";
+      } else {
+        *stream << ">";
+        OutputXmlCDataSection(stream,
+                              RemoveInvalidXmlCharacters(detail).c_str());
+        *stream << "</failure>\n";
+      }
     } else if (part.skipped()) {
       if (++skips == 1 && failures == 0) {
         *stream << ">\n";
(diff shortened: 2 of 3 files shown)


nithin42:
@googlebot I signed it!

PR #5071, opened 2026-08-20
Register googlemock Python tests with CTest

### LINKED ISSUE #5070: [Bug]: GoogleMock Python tests are never registered with CTest (regression of #4124)
### Describe the issue

The two GoogleMock Python tests, `gmock_leak_test` and `gmock_output_test`, are not registered with CTest in any CMake configuration, even when a Python 3 interpreter is present. Their C++ helper binaries are compiled, so the build spends time producing `gmock_leak_test_` and `gmock_output_test_`, but nothing ever runs them.

This is a regression of #4124. That issue was fixed in February 2023 by fd36851c8d383a2b711905eb488692da1152a3d8, which stored the interpreter variables as `CACHE INTERNAL` so that they were visible outside the googletest directory scope. In June 2023, 812f35b26b88ab1607a806b87539175591bba8c2 replaced that code with `find_package(Python3)` and dropped the cache variables, and the tests silently stopped being registered again.

Both tests pass when invoked by hand against a normal CMake build, so this is purely a registration problem and not a problem with the tests themselves.

Configure with both test options enabled and count the registered tests:

```
cmake -S googletest -B build -Dgtest_build_tests=ON -Dgmock_build_tests=ON
ctest --test-dir build -N | tail -1
ctest --test-dir build -N | grep -E "gmock_leak_test|gmock_output_test"
```

Output:

```
Total Tests: 63
```

The `grep` produces no output and exits with status 1: neither test is present.

Both tests pass when run manually against the same build tree:

```
$ PYTHONPATH=. python3 googlemock/test/gmock_output_test.py --build_dir=build/googlemock
Ran 1 test in 0.014s

OK

$ PYTHONPATH=. python3 googlemock/test/gmock_leak_test.py --build_dir=build/googlemock
Ran 5 tests in 0.013s

OK
```

For comparison, disabling Python entirely drops 18 tests, all of them googletest's:

```
cmake -S googletest -B build-nopython -Dgtest_build_tests=ON -Dgmock_build_tests=ON \
  -DCMAKE_DISABLE_FIND_PACKAGE_Python3=ON
ctest --test-dir build-nopython -N | tail -1
```

```
Total Tests: 45
```

63 minus 45 is 18, and every one of those 18 belongs to googletest. GoogleMock contributes zero Python tests in either case.

### Root cause

`find_package(Python3 COMPONENTS Interpreter QUIET)` is called from `googletest/cmake/internal_utils.cmake`, which is included by `googletest/CMakeLists.txt`. When googlemock is built, it adds googletest as a subdirectory, so googletest's directory scope is a child of googlemock's. CMake functions are global, so `py_test()` is callable from `googlemock/CMakeLists.txt`, but the variable `Python3_Interpreter_FOUND` and the imported target `Python3::Interpreter` are both directory-scoped and do not propagate upward. `py_test()` therefore takes its early return every time it is called from googlemock.

A related consequence of the guard added in 8760db15: configuring with only `-Dgmock_build_tests=ON`, which is one of the commands suggested in `CONTRIBUTING.md`, skips the `find_package` call altogether.

Because `find_package` is `QUIET` and `py_test()` returns silently, there is no diagnostic in the configure output. A contributor who runs the documented build and sees CTest report success has no way to notice that two tests were never registered.

### Suggested fix

Calling `find_package(Python3 COMPONENTS Interpreter QUIET)` inside `if (gmock_build_tests)` in `googlemock/CMakeLists.txt`, before the Python tests section, restores registration. Caching the variable as fd36851c did is no longer sufficient on its own, because `py_test()` now uses the imported `Python3::Interpreter` target, which is also directory-scoped.

A more conservative alternative, if adding the two tests to CI is not desirable right now, is to leave registration as it is and emit a `message(STATUS ...)` when the interpreter is not found, so that the silent skip at least becomes visible.

I am happy to send a pull request for either approach.

### Scope

Not included, and can be filed separately if useful:

- `CONTRIBUTING.md` still documents `FindPythonInterp` behaviour: it mentions the error message `Could NOT find PythonInterp (missing: PYTHON_EXECUTABLE)` and suggests `-DPYTHON_EXECUTABLE=path/to/python`. With `FindPython3` that variable is ignored; CMake reports `Manually-specified variables were not used by the project: PYTHON_EXECUTABLE`. The hint variable is now `Python3_EXECUTABLE`.
- `ci/windows-presubmit.bat` passes `-DPYTHON_EXECUTABLE:FILEPATH=...`, which is likewise no longer read.

### Known limitation

Registering these tests puts them back into the official CI, which runs `-Dgtest_build_tests=ON -Dgmock_build_tests=ON` followed by `ctest` on both Linux and MSVC. `gmock_output_test` compares output against a golden file and has historically been sensitive to MSVC output differences: 0a3b403f fixed it for MSVC on the same day fd36851c enabled these tests, and 097f64e9 reverted that fix a week later. I have only been able to verify Linux, so the MSVC behaviour of the golden comparison is unknown to me and would need checking before merging.

0daf775f8c9324e7d42582a09de1240805f25a54 is the commit I reproduced on; `git rev-parse --short HEAD` reports 0daf775f.

Ubuntu 22.04.3 LTS, running under WSL2 on Windows.

g++ 11.4.0 (Ubuntu 11.4.0-1ubuntu1~22.04.3).

cmake version 3.22.1. Python 3.10.12 is present on `PATH`.

### Additional context

The build currently compiles `gmock_leak_test_` and `gmock_output_test_` and then never runs them, so the cost of these tests is already being paid.

YdiBik:
Closes #5070

## What changed

`googlemock/CMakeLists.txt` now calls `find_package(Python3 COMPONENTS Interpreter QUIET)` inside `if (gmock_build_tests)`, before the Python tests section. One statement plus a comment; no other file is touched.

## Why

`py_test()` is defined in `googletest/cmake/internal_utils.cmake` and is global, but the `Python3_Interpreter_FOUND` variable and the imported `Python3::Interpreter` target it depends on are both directory-scoped. googlemock adds googletest as a subdirectory, so those live in a child scope and are not visible where `py_test()` is called from. Every call from googlemock took the early return, and `gmock_leak_test` and `gmock_output_test` were never registered with CTest.

This is a regression of #4124. fd36851c8d383a2b711905eb488692da1152a3d8 fixed it in February 2023 by marking the interpreter variables `CACHE INTERNAL`, which made them global. 812f35b26b88ab1607a806b87539175591bba8c2 later replaced that code with `find_package(Python3)` and dropped the cache variables, restoring the original problem. Re-adding `CACHE INTERNAL` alone would not be enough today, because `py_test()` now runs the interpreter through the imported target rather than through a path variable, and imported targets are directory-scoped as well. Performing the search in googlemock's own scope creates both the variable and the target where they are needed.

Because `find_package` is `QUIET` and `py_test()` returns silently, there was no diagnostic for this in the configure output.

## Verification

All commands run on Ubuntu 22.04.3 (WSL2), g++ 11.4.0, cmake 3.22.1, Python 3.10.12, against upstream/main at 0daf775f8c9324e7d42582a09de1240805f25a54.

Before the change:

```
$ cmake -S . -B build -Dgtest_build_tests=ON -Dgmock_build_tests=ON
$ ctest --test-dir build -N | tail -1
Total Tests: 63
$ ctest --test-dir build -N | grep -E "gmock_leak_test|gmock_output_test"
$ make -j8 && ctest
100% tests passed, 0 tests failed out of 63
```

After the change:

```
$ ctest --test-dir build -N | tail -1
Total Tests: 65
$ ctest --test-dir build -N | grep -E "gmock_leak_test|gmock_output_test"
  Test #19: gmock_leak_test
  Test #20: gmock_output_test
$ make -j8 && ctest
100% tests passed, 0 tests failed out of 65
```

The two tests are executed, not merely registered, and both pass.

The `-Dgmock_build_tests=ON` invocation suggested in `CONTRIBUTING.md` is also fixed by this: it went from 18 registered tests to 20.

No formatting tools were run: the change is confined to a CMake file, and `.clang-format` does not apply to it.

## Known limitation

This puts `gmock_leak_test` and `gmock_output_test` back into the CMake CI configurations, which run `-Dgtest_build_tests=ON -Dgmock_build_tests=ON` followed by `ctest` on both Linux and MSVC. `gmock_output_test` compares program output against a golden file and has been sensitive to MSVC output differences in the past: 0a3b403f fixed it for MSVC on the same day fd36851c first enabled these tests, and 097f64e9 reverted that fix a week later. I have only been able to verify Linux and cannot check the MSVC behaviour of the golden comparison.

If re-enabling the tests in CI is not wanted at this point, I am happy to replace this with a `message(STATUS ...)` diagnostic instead, so that the silent skip at least becomes visible without changing which tests run. Please let me know which you prefer.

## Environment

- Ubuntu 22.04.3 LTS under WSL2
- g++ (Ubuntu 11.4.0-1ubuntu1~22.04.3) 11.4.0
- cmake version 3.22.1
- Python 3.10.12

### DIFF
--- a/googlemock/CMakeLists.txt
+++ b/googlemock/CMakeLists.txt
@@ -202,6 +202,11 @@ if (gmock_build_tests)
   ############################################################
   # Python tests.
 
+  # Python is found in googletest's directory scope, which does not propagate
+  # to this parent scope, and the imported Python3::Interpreter target used by
+  # py_test() is directory-scoped as well, so the search is repeated here.
+  find_package(Python3 COMPONENTS Interpreter QUIET)
+
   cxx_executable(gmock_leak_test_ test gmock_main)
   py_test(gmock_leak_test)
 


AdamerGitHub:
@googlebot I signed it!

AdamerGitHub:
@googlebot I fixed it.

